PR #21, opened 2019-12-20
Implement loadable external checks

jeromepin:
This allows users to pass directories paths on the command line to load user-defined checks.

As I am not a seasoned Python developer, there is two things to take care of :

* Should I commit Pipfile and Pipfile.lock given the changes
* I may need help regarding tests


schosterbarak:
Great feature! I'll try to add docs in the upcoming days and will merge.

Thank you for the contribution! Much appreciated

Issue #50, opened 2020-01-10
Checkov Docker Image


jdhornsby:
**Is your feature request related to a problem? Please describe.**
We are using Gitlab CI. It would be easier/faster to integrate checkov if we had a docker image of it.

**Describe the solution you'd like**
Create a docker image and publish it to docker hub.

**Describe alternatives you've considered**
We are just getting started with checkov and just install it manually at build time. This works ok. We could build a docker image ourselves, but it'd be nice if it was official.

**Additional context**
Forgive me if this does not mesh with your vision for the tool or an image already exists. I couldn't find one docker hub.

schosterbarak:
Hey @jdhornsby , thank you for interest. Do you mind sharing/contributing the dockerfile you have? It would save us some time. I'll add the publishing to dockerhub later on next week if the dockerfile is contributed :)

jdhornsby:
Sure, I'll work on a PR this weekend for you. Thanks!

schosterbarak:
@jdhornsby you should be able to open a PR from a fork

schosterbarak:
Thanks! @jdhornsby docker is now available at: https://hub.docker.com/r/bridgecrew/checkov
Have a great week!

jdhornsby:
Awesome. Thank you! I will update my CI pipeline this week.

schosterbarak:
@jdhornsby if you'd like to share integration docs to circleci under docs/integrations/gitlab (just like the jenkins) that would be  super awesome and will help the community adapt your approach :)

jdhornsby:
Sure thing. I'll work on that and send you a PR with the docs! Thanks for your work on this awesome tool.

PR #52, opened 2020-01-11
chore(docker): adding a docker file

jdhornsby:
The resulting image can be used by mounting the terraform files at a location and directing checkov to scan that location:
$ docker container run -v /host/path/tf:/tf <image> -d /tf

The entry point can also be overridden to bash for other uses, such as CI commands:
$ docker container run -i -v /host/path/tf:/tf --entrypoint=/bin/bash <image>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM python
+
+RUN pip install checkov
+
+ENTRYPOINT ["checkov"]


nimrodkor:
Related to https://github.com/bridgecrewio/checkov/issues/50

schosterbarak:
Thank you @jdhornsby , much appreciated!
when would the entrypoint require override? can you give an example?

jdhornsby:
Sure. When I use the image in GitLab CI, I'll override the entry point so I can run scripts with the results of the scan.

schosterbarak:
Thanks! @jdhornsby docker is now available at: https://hub.docker.com/r/bridgecrew/checkov
Have a great week!

PR #96, opened 2020-02-07
Fail CKV_AWS_8 if block_device is not specified

m-wynn:

It is easy to create an `aws_instance` without a specified block device.  Unfortunately, the default block device is not encrypted.

This PR fails CKV_AWS_8 if no block_device is specified.


schosterbarak:
Thank you @m-wynn !

PR #105, opened 2020-02-24
[Add] enum for convention type checks

TechIsCool:
This is intented to be used for Stylistic and Convention based checks.

For example underscores vs dashes in terraform entity names.
This might be opinionated but I am currently using custom checks. So this won't impact master


schosterbarak:
Hey @TechIsCool  can you add a markdown under `docs/` directory with an example of. a convention check so other's might know ?

TechIsCool:
@schosterbarak I added a little note. If you want me to be more specific I can be and Add notes on what each classification is.

schosterbarak:
@TechIsCool Thank you!

PR #108, opened 2020-02-28
Pipenv Windows complaint version

JamesWoolfenden:

This version is pipenv safe version for Windows, the cmd replaces the hardcoded bat which could fail in some possible set-ups. 
This version will work wherever Python is (and would work with other versions of Python) and works with Pre-commit and the virtual envs it can create.   Hope that makes sense?


schosterbarak:
Thank you @JamesWoolfenden I don't own a Windows workstation.
 Can you please confirm it works on Windows?

JamesWoolfenden:
Yeah It works on WIndows, both at the cli and via pipenv and via a commit hook.

PR #109, opened 2020-03-01
unifty boolean checks

schosterbarak:
unify boolean checks

### DIFF
--- a/checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
+++ b/checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
@@ -1,0 +1,25 @@
+from abc import abstractmethod
+
+import dpath
+
+from checkov.terraform.models.enums import CheckResult
+
+from checkov.terraform.checks.resource.base_check import BaseResourceCheck
+
+expected_value = True
+
+
+class BaseResourceBooleanValueCheck(BaseResourceCheck):
+    def __init__(self, name, id, categories, supported_resources):
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        inspected_key = self.get_inspected_key()
+        if dpath.util.search(conf, inspected_key) != {}:
+            if dpath.util.get(conf, inspected_key)[0] == expected_value:
+                return CheckResult.PASSED
+        return CheckResult.FAILED
+
+    @abstractmethod
+    def get_inspected_key(self):
+        raise NotImplemented()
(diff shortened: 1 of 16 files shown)


checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
```diff
@@ -0,0 +1,25 @@
...
+expected_value = True
+
+
+class BaseResourceBooleanValueCheck(BaseResourceCheck):
+    def __init__(self, name, id, categories, supported_resources):
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        inspected_key = self.get_inspected_key()
+        if dpath.util.search(conf, inspected_key) != {}:
```
tronxd: can just write 
`if dpath.util.search(conf, inspected_key):`

checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
```diff
@@ -0,0 +1,25 @@
+from abc import abstractmethod
+
+import dpath
+
+from checkov.terraform.models.enums import CheckResult
+
+from checkov.terraform.checks.resource.base_check import BaseResourceCheck
+
+expected_value = True
```
tronxd: What if in some point we will need to check for `False` or maybe just generalize the class into `BaseResourceValueCheck`? which will check a certain attribute match a certain value, not just `True`
I think this should be refactored as a function of kind:
```
class BaseResourceValueCheck(...):
@abstractmethod
    def get_inspected_key(self):
        raise NotImplemented()
@abstractmethod
    def get_inspected_value(self):
        raise NotImplemented()     
```
and then implement respectively in a specific check.

checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
schosterbarak: doesn't it mean "not null"?
not "empty dict" is different. no?

checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
schosterbarak: Like 👍  added

checkov/terraform/checks/resource/BaseResourceBooleanValueCheck.py
tronxd: No difference in that case

Issue #59, opened 2020-01-12
Add new check: File system should be encrypted at rest 


schosterbarak:
(encrypted = true) https://www.terraform.io/docs/providers/aws/r/efs_file_system.html

schosterbarak:
Solved: https://github.com/bridgecrewio/checkov/pull/112

PR #112, opened 2020-03-06
Adding check for EFS volume encryption

JamesWoolfenden:

Ok i'm giving extending Checkov a go.  I made a TF module for EFS, nothing too amazing but I noticed that it didn't check for it being encrypted- https://github.com/JamesWoolfenden/terraform-aws-efs.
So i added 2 new rules in Checkov to find this.
I've added a test, I pretty much duplicated one of the checks you have already.
I have also updated it to have tests too. The tests pass for me if there is a kms_key_id specified and a its set to encrypted. I don't this fixes issue 59 completely. I don't think the checks expand variables? be happy to be wrong?


tests/terraform/runner/resources/example/example.tf
```diff
@@ -467,4 +467,20 @@ resource "aws_s3_bucket" "bridgecrew_cws_bucket" {
...
+
+resource "aws_efs_file_system" "sharedstore" {
+  creation_token                  = "my-product"
+
+  lifecycle_policy {
+    transition_to_ia = "AFTER_30_DAYS"
+  }
+
+    kms_key_id                      = "aws/efs"
+    encrypted                       = True
```
nimrodkor: @JamesWoolfenden Should be `true`, I believe

checkov/terraform/checks/resource/aws/EFSEncryptionEnabled.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        """
+            Looks for encryption configuration at EFS:
+            https://www.terraform.io/docs/providers/aws/r/efs_file_system.html
+        :param conf: efs configuration
+        :return: <CheckResult>
+        """
+        if "encrypted" in conf.keys():
```
nimrodkor: @JamesWoolfenden The fact that the key `encrypted` exists isn't enough as `encrypted` can be set to false

schosterbarak:
@JamesWoolfenden Checkov do expand variables defined at the same file or directory. see:
https://www.checkov.io/2.Concepts/Evaluations.html

checkov/terraform/checks/resource/aws/EFSEncryptionEnabled.py
JamesWoolfenden: i agree. The problem i was having was that my efs was a module and the values got no expansion at all.

tests/terraform/runner/resources/example/example.tf
JamesWoolfenden: Yep thanks for finding it. Updated

checkov/terraform/checks/resource/aws/EFSEncryptionEnabled.py
JamesWoolfenden: Ive change the checks, one for the encrypted value and another so that kms_key_id is set. I split out the tests to separate files to cover this .

JamesWoolfenden:
@nimrodkor Im not sure why the PR is failing its checks, the coverage report and tests all work locally.

schosterbarak:
@JamesWoolfenden can you please refresh from latest master and see if it is fixed?

nimrodkor:
Hi @JamesWoolfenden!

Regarding the tests - see @schosterbarak 's comments.
Could you please explain why checking if `kms_key_id` exists (`EFSEncryption.py`) is required?
What I mean is that EFS can have encryption set with `encrypted` set to true and no `kms_key_id`. If that's the case - isn't `EFSEncryptionEnabled.py` enough?
BTW - your check `EFSEncryptionEnabled.py` definitely resolves https://github.com/bridgecrewio/checkov/issues/59! Thanks!

JamesWoolfenden:
I've made the changes you requested. I'm learning hopefully...

PR #124, opened 2020-03-13
add --soft-fail mode

frenchi:
Hello Bridgecrew 👋 

I'd like to be able to run checkov, but not break the build. This PR adds the `--soft-fail` or `-s` mode which suppresses the return code of 1, but otherwise leaves the report the same.

Screen Shot 2020-03-13 at 4 55 05 PM


schosterbarak:
Thank you! Much appreciated!

PR #136, opened 2020-03-20
Adding DNSSEC checks for Cloud DNS

gherynos:
I added a couple of new checks to ensure that DNSSEC is enabled in Cloud DNS and that the RSASHA1 algorithm is not used for the keys.

https://www.terraform.io/docs/providers/google/r/dns_managed_zone.html#state
https://www.terraform.io/docs/providers/google/r/dns_managed_zone.html#algorithm


schosterbarak:
Thanks!

PR #207, opened 2020-03-30
Add tests for validation of check IDs

### LINKED ISSUE #206: We have colliding check IDs
**Describe the bug**
There are 11 checks with the ID `CKV_AWS_23`, for example.

Need to add tests to make sure check IDs don't collide

nimrodkor:

Resolves #206 by making sure we actually don't

### DIFF
--- a/tests/terraform/runner/test_runner.py
+++ b/tests/terraform/runner/test_runner.py
@@ -61,6 +61,40 @@ class TestRunnerValid(unittest.TestCase):
         self.assertEqual(summary['failed'], 0)
         self.assertEqual(summary["parsing_errors"], 0)
 
+    def test_check_ids_dont_collide(self):
+        runner = Runner()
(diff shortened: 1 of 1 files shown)


tests/terraform/runner/test_runner.py
```diff
@@ -61,6 +61,40 @@ def test_runner_specific_file(self):
         self.assertEqual(summary['failed'], 0)
         self.assertEqual(summary["parsing_errors"], 0)
 
+    def test_check_ids_dont_collide(self):
```
schosterbarak: i thought we are testing this in tests/terraform/test_scanner_registry.py
it failed builds several times. what is the diff?

nimrodkor: @schosterbarak I wasn't aware of that test. nice. 
However, that test checked only one registry - this checks across all registries...

PR #209, opened 2020-03-30
Support directory as a repetitive parameter

### LINKED ISSUE #115: checkov directory switch not repeatable
**Describe the bug**
When running checkov -h, the directory switch claims it is repeatable but when you repeat it, only the last directory listed is scanned

```
  -d DIRECTORY, --directory DIRECTORY
                        Terraform root directory (can not be used together with --file). Can be repeated
```

**To Reproduce**
1. Create two directories (directory1, directory2) containing valid terraform files
2. Run cli command  `checkov -d ./directory1 -d ./directory2`
3. Only directory2 is scanned

**Expected behavior**
Both directory1 and directory2 are scanned

 - Checkov Version: 1.0.180 

**Additional context**
I would actually prefer a recursive switch, using the -d directive multiple times is a workaround

nimrodkor:

Related to #115 , still no support for recursive

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -22,7 +22,7 @@ def run():
     parser = argparse.ArgumentParser(description='Infrastructure as code static analysis')
     parser.add_argument('-v', '--version',
                         help='Checkov version', action='store_true')
-    parser.add_argument('-d', '--directory',
+    parser.add_argument('-d', '--directory', action='append',
                         help='IaC root directory (can not be used together with --file). Can be repeated')
     parser.add_argument('-f', '--file', action='append',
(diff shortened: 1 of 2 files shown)


checkov/main.py
```diff
@@ -40,9 +40,9 @@ def run():
     if args.list:
         print_checks()
         return
-    else:
```
schosterbarak: why remove the else?

nimrodkor: @schosterbarak Because in the case of the `if`, there is a `return`. So the else does nothing

schosterbarak: missed that 👍

PR #216, opened 2020-03-30
add check : x-ray tracing is enabled for Lambda; closes #191

### LINKED ISSUE #191: Add new check: Lambda function should have X-Ray tracing enabled
tracing_config: https://www.terraform.io/docs/providers/aws/r/lambda_function.html

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-tracingconfig.html

JamesWoolfenden:

Adds a check for enabling x-ray on a lambda, so resolves #191 
Checks for existence of the bock tracing_config, I think the value of mode - PassThrough or Active isn't relevant?
Wasn't sure what Category this went in? Used CheckCategories.GENERAL_SECURITY  as there's nothing like CheckCategories.DEBUGGING.


checkov/terraform/checks/resource/aws/LambdaXrayEnabled.py
```diff
@@ -0,0 +1,18 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+class LambdaXrayEnabled(BaseResourceCheck):
+    def __init__(self):
+        name = "X-ray tracing is enabled for Lambda"
+        id = "CKV_AWS_50"
+        supported_resources = ['aws_lambda_function']
+        categories = [CheckCategories.GENERAL_SECURITY]
```
schosterbarak: maybe should be under the logging category? WDYT?

JamesWoolfenden: It's your call. It didnt feel like it fitted anywhere and there's a bunch of xray tickets, which is more interactive than logging. ill change it to logging?

schosterbarak:
Thanks!

PR #221, opened 2020-04-05
Local context parser

tronxd:
support of local variable evaluations, major refactor of context parsers and record class


checkov/terraform/context_parsers/base_parser.py
```diff
@@ -68,43 +73,41 @@ def _compute_definition_end_line(self, start_line_num):
...
-            if not self.context.get(entity_type):
-                self.context[entity_type] = {}
-            if not self.context.get(entity_type).get(entity_name):
-                self.context[entity_type][entity_name] = {}
+        for i, entity_block in enumerate(definition_blocks):
+            entity_context_path = self.get_entity_context_path(entity_block)
             for line_num, line in parsed_file_lines:
                 line_tokens = [x.replace('"', "") for x in line.split()]
-                if all(x in line_tokens for x in [self.get_block_type(), entity_type, entity_name]):
+                if self._is_block_signature(line_tokens, entity_context_path):
```
schosterbarak, line 106: is that returning a boolean value? name hints it is when it does not.

I know it's not part of the original scope, but can you please add class comments and methods comments to the common classes you have modified?

checkov/terraform/context_parsers/parsers/locals_context_parser.py
```diff
@@ -0,0 +1,29 @@
...
+    def _collect_local_values(self, local_block):
+        for local_name, local_value in local_block.items():
+            local_value = local_value[0]
+            if type(local_value) in (int, float, bool, str):
+                dpath.new(self.context, ['assignments', local_name], local_value)
+
+    def get_block_type(self):
+        return self.definition_type
+
+    def get_entity_context_path(self, entity_block):
```
schosterbarak, line 19: what is this method used for?

checkov/terraform/context_parsers/parsers/resource_context_parser.py
```diff
@@ -3,11 +3,18 @@
 
 class ResourceContextParser(BaseContextParser):
     def __init__(self):
-        definition_type = 'resource'
+        definition_type = "resource"
         super().__init__(definition_type=definition_type)
 
-    def get_block_type(self):
-        return self.definition_type
+    def _is_block_signature(self, line_tokens, entity_context_path):
```
schosterbarak, line 9: "_" prefix is for private methods. but you call this method outside of this class - maybe that should be changed?

schosterbarak:
Overall very good. let's talk about some minor comments today :)

checkov/terraform/context_parsers/base_parser.py
tronxd: @schosterbarak It's actually returns a boolean value
docs are on the way anyway :)

checkov/terraform/context_parsers/parsers/locals_context_parser.py
tronxd: @schosterbarak given an entity (may be a `resource`, `data`, `provider` etc.), return the entity's path in the context parser object. It is an overridden method in each context parser that implements the base context parser.

checkov/terraform/context_parsers/parsers/resource_context_parser.py
tronxd: @schosterbarak Thanks, it's a code leftover. the implementation is in `base_context_parser` and usage is only class-internal.

checkov/terraform/context_parsers/parsers/locals_context_parser.py
schosterbarak: great. maybe we should add that in a pydoc on that method? :)

Issue #218, opened 2020-03-31
Create a brew package for checkov


nimrodkor:
Creating a `brew` tap + formula will enable faster and easier adoption to this tool

PR #226, opened 2020-04-12
Adding brew package for homebrew (OSX)

lfacchinelli:

This PR adds the ability to install checkov using homebrew on OSX (as requested in #218)
Once merged into master, OSX users should execute:

> $ brew tap bridgecrewio/checkov https://github.com/bridgecrewio/checkov
> $ brew update
> $ brew install checkov

On each release, the formula should be updated to reflect the latest release

### DIFF
--- a/HomebrewFormula/checkov.rb
+++ b/HomebrewFormula/checkov.rb
@@ -1,11 +1,16 @@
 class Checkov < Formula
-  desc "Prevent cloud misconfigurations during build time - Terraform & Cloudformation static analysis"
+  include Language::Python::Virtualenv
+  desc "Prevent cloud misconfigurations during build time. Terraform & CFN static analysis"
   homepage "https://www.checkov.io/"
   url "https://github.com/bridgecrewio/checkov/archive/1.0.239.tar.gz"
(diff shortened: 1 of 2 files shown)


schosterbarak:
@lfacchinelli  thank you! 
I guess that means we should add "brew update" stage on each release. correct? - if yes, i can do that. just verifying since i'm not very familiar with brew

lfacchinelli:
Sorry guys, by mistake pressed the “merge” button in github. Can someone revert it back?

lfacchinelli:
@schosterbarak On each release the user should only run:
> $ brew upgrade checkov 

Brew update only refresh the internal database with references about what formula is available on each tap

Homebrew works pretty much like YUM/APT

schosterbarak:
thank you @lfacchinelli  do you know if there is a way to automate the sync between formula and pipenv? 
that way we won't have to maintain package twice

lfacchinelli:
@schosterbarak Based on what's available in the official documentation, no. 
Maybe a solutions is to add a script that release a new version to pypi (using pipenv) and updates lines 5 and 6 of the homebrew Formula

PR #235, opened 2020-04-27
Bridgecrew CLI integration

tronxd:
Integrate checkov with bridgecrew's platform by a set of API calls, using a bridgecrew issued API key


schosterbarak:
can you please add markdown under docs/integrations too?

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -0,0 +1,131 @@
...
+        persist_checks_results(reduced_scan_reports, self.s3_client, self.bucket, self.repo_path)
+
+    def commit_repository(self):
+        """
+        Finalize the repository's scanning in bridgecrew's platform.
+        """
+        request = None
+        try:
+            request = http.request("PUT", INTEGRATIONS_API_URL,
+                                   body=json.dumps({"path": self.repo_path, "branch": "master"}),
```
nimrodkor, line 110: @tronxd Please add a todo to get the actual branch name!

checkov/main.py
```diff
@@ -35,22 +36,35 @@ def run(banner = checkov_banner):
                         help='Report output format')
     parser.add_argument('-s', '--soft-fail',
                         help='Runs checks but suppresses error code', action='store_true')
+    parser.add_argument('--bc-api-key', help='Bridgecrew API key')
+    parser.add_argument('--repo-id',
+                        help='Identity string of the repository, with form <repo_owner>/<repo_name>')
```
nimrodkor, line 41: @tronxd Maybe add a branch parameter, which has the default value "master"?

tronxd: @schosterbarak WDYT?

schosterbarak: I agree. Branch should be a parameter. The SaaS should show results only for the default branch (usually master). 
maybe that's a value you can populate from circleci.

tronxd: @schosterbarak adding the branch as an arg to checkov, defaults to master.

PR #239, opened 2020-04-29
handle dynamic sse_block when its a string instead of dict

arielkru:

### DIFF
--- a/tests/terraform/runner/resources/example/example.tf
+++ b/tests/terraform/runner/resources/example/example.tf
@@ -475,6 +475,28 @@ resource "aws_s3_bucket" "bridgecrew_cws_bucket" {
   }
 }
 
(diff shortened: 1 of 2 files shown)


tests/terraform/runner/resources/example/example.tf
```diff
@@ -475,6 +475,28 @@ resource "aws_s3_bucket" "bridgecrew_cws_bucket" {
   }
 }
 
+resource "aws_s3_bucket" "dynamic ssee block as string" {
```
nimrodkor, line 478: @arielkru This is not sse block as string, it is a full dict. Does this pass or fail?
Please add another case to show the other case (fail / pass)

arielkru:
all bugfix commits are left in ont PR

PR #242, opened 2020-04-30
Bugfix/s3 versioning string

arielkru:

### DIFF
--- a/checkov/terraform/checks/resource/aws/S3Encryption.py
+++ b/checkov/terraform/checks/resource/aws/S3Encryption.py
@@ -19,12 +19,14 @@ class S3Encryption(BaseResourceCheck):
         """
         if 'server_side_encryption_configuration' in conf.keys():
             sse_block = conf['server_side_encryption_configuration']
-            if 'rule' in sse_block[0].keys():
-                rule_block = sse_block[0]['rule']
-                if 'apply_server_side_encryption_by_default' in rule_block[0].keys():
-                    encryption_block = rule_block[0]['apply_server_side_encryption_by_default']
-                    if  'sse_algorithm' in encryption_block[0].keys():
-                        return CheckResult.PASSED
+            if type(sse_block[0]) is dict:
+                # all non-dict sse_blocks will return as FAILED
+                if 'rule' in sse_block[0].keys():
+                    rule_block = sse_block[0]['rule']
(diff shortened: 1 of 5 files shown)


checkov/terraform/checks/resource/aws/S3Encryption.py
```diff
@@ -19,12 +19,14 @@ def scan_resource_conf(self, conf):
         """
         if 'server_side_encryption_configuration' in conf.keys():
             sse_block = conf['server_side_encryption_configuration']
-            if 'rule' in sse_block[0].keys():
-                rule_block = sse_block[0]['rule']
-                if 'apply_server_side_encryption_by_default' in rule_block[0].keys():
-                    encryption_block = rule_block[0]['apply_server_side_encryption_by_default']
-                    if  'sse_algorithm' in encryption_block[0].keys():
-                        return CheckResult.PASSED
+            if type(sse_block[0]) is dict:
```
nimrodkor, line 22: @arielkru I believe that if it isn't a `dict`, we won't be able to assess it - so we should let it pass, not fail. WDYT?

checkov/terraform/checks/resource/aws/S3MFADelete.py
```diff
@@ -17,10 +17,11 @@ def scan_resource_conf(self, conf):
         if 'versioning' in conf.keys():
             versioning_block = conf['versioning']
             for block in versioning_block:
-                if 'mfa_delete' in block.keys():
-                    if block['mfa_delete']:
+                if type(block) is dict:
```
nimrodkor, line 20: @arielkru Here too - if it is not a dict, we won't be able to assess it so I think it should pass

tests/terraform/runner/resources/example/example.tf
```diff
@@ -344,6 +344,43 @@ resource "aws_cloudfront_distribution" "s3_distribution" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+dynamic "ordered_cache_behavior" {
+    for_each = var.ordered_cache
```
nimrodkor, line 348: @arielkru What was throwing me off is the indent here - missing one tab. Otherwise - @schosterbarak this is an example of the string block because `var.ordered_cache` is null, so this is parsed by checkov as a string and not as a dict. Even though we have the `content` block.

PR #249, opened 2020-05-03
handle environment value as int instead of string.

arielkru:
the tf example for an error (delete_after):

  environment {
    variables = {
      es_endpoint  = var.es_endpoint
      index        = 'all'
      delete_after = 15
      index_format = var.index_format
      sns_arn      = var.sns_arn
    }
  }

### DIFF
--- a/checkov/terraform/checks/resource/aws/LambdaEnvironmentCredentials.py
+++ b/checkov/terraform/checks/resource/aws/LambdaEnvironmentCredentials.py
@@ -18,14 +18,14 @@ class LambdaEnvironmentCredentials(BaseResourceCheck):
         if 'environment' in conf.keys():
             if isinstance(conf['environment'][0], dict):
                 if 'variables' in conf['environment'][0]:
-                    if type (conf['environment'][0]['variables']) is not list:
+                    if not isinstance(conf['environment'][0]['variables'], list):
                         conf['environment'][0]['variables'] = [conf['environment'][0]['variables']]
-                    if type(conf['environment'][0]['variables'][0]) is dict:
+                    if isinstance(conf['environment'][0]['variables'][0], dict):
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/aws/LambdaEnvironmentCredentials.py
```diff
@@ -16,17 +16,19 @@ def __init__(self):
...
-                    for values in list(conf['environment'][0]['variables'][0].values()):
-                        if type(values) is not list:
-                            values = [values]
-                        for value in values:
-                            if re.match(access_key_pattern, value) or re.match(secret_key_pattern, value):
-                                return CheckResult.FAILED
+            if isinstance(conf['environment'][0], dict):
+                if 'variables' in conf['environment'][0]:
+                    if type (conf['environment'][0]['variables']) is not list:
+                        conf['environment'][0]['variables'] = [conf['environment'][0]['variables']]
```
nimrodkor, line 22: @arielkru It seems like we're doing this quite a lot. Maybe it is time to extract to a method?

checkov/terraform/checks/resource/aws/LambdaEnvironmentCredentials.py
```diff
@@ -16,17 +16,19 @@ def __init__(self):
...
+                    if type(conf['environment'][0]['variables'][0]) is dict:
+                        # variables can be a string, which in this case it points to a variable
+                        for values in list(conf['environment'][0]['variables'][0].values()):
+                            if type(values) is not list:
```
tronxd, line 26: Prefer using `isinstance(var,type)` instead of `type is ..`

arielkru: np, I'll go over it

PR #250, opened 2020-05-04
3 changes: add type (forcers) for list and int + bugfix for passwords + bugfix for cluster config

arielkru:


checkov/terraform/checks/resource/aws/ElasticsearchNodeToNodeEncryption.py
```diff
@@ -18,21 +18,21 @@ def scan_resource_conf(self, conf):
...
-            if isinstance(instance_count,int):
-                if instance_count > 1:
-                    if "node_to_node_encryption" in conf.keys():
-                        if conf["node_to_node_encryption"][0]["enabled"][0]:
-                            return CheckResult.PASSED
+            if isinstance(conf["cluster_config"][0], dict):
+                instance_count = conf["cluster_config"][0]["instance_count"][0]
+                if isinstance(instance_count, int):
+                    if instance_count > 1:
+                        if "node_to_node_encryption" in conf.keys():
```
nimrodkor, line 25: @arielkru Please use the `and` operator to simplify this `if-else` case

checkov/terraform/checks/resource/aws/PasswordPolicyExpiration.py
```diff
@@ -19,7 +20,7 @@ def scan_resource_conf(self, conf):
         """
         key = 'max_password_age'
         if key in conf.keys():
-            if conf[key][0] >= 90:
+            if force_int(conf[key][0]) >= 90:
```
nimrodkor, line 23: @arielkru What happens if the value isn't a number, but a string that points to a variable?

checkov/terraform/checks/resource/aws/PasswordPolicyLength.py
```diff
@@ -19,7 +20,7 @@ def scan_resource_conf(self, conf):
         """
         key = 'minimum_password_length'
         if key in conf.keys():
-            if conf[key][0] >= 14:
+            if force_int(conf[key][0]) >= 14:
```
nimrodkor, line 23: @arielkru Same here

checkov/terraform/checks/resource/aws/PasswordPolicyReuse.py
```diff
@@ -19,7 +20,7 @@ def scan_resource_conf(self, conf):
         """
         key = 'password_reuse_prevention'
         if key in conf.keys():
-            if conf[key][0] >= 24:
+            if force_int(conf[key][0]) >= 24:
```
nimrodkor, line 23: @arielkru Same here

checkov/common/util/type_forcers.py
```diff
@@ -0,0 +1,10 @@
+def force_list (var):
+    if not isinstance(var, list):
+        var = [var]
+    return var
+
+
+def force_int (var):
+    if not isinstance(var, int):
+        var = int(var)
```
nimrodkor, line 9: @arielkru I would suggest having a try-catch here, and in case of an exception return `None`

checkov/terraform/checks/resource/aws/ElasticsearchNodeToNodeEncryption.py
arielkru: I added only the first row, but no problem i'll edit the rest

Issue #264, opened 2020-05-13
False negative for CKV_AWS_24 (unrestricted SSH ingress)


acdha:
**Describe the bug**

The CKV_AWS_24 rule appears to have stopped working at some point.

**To Reproduce**
Run checkov — e.g. `docker run -it --rm -v $(pwd):/code bridgecrew/checkov -d /code` — using this Terraform file:

<details>

```hcl
resource "aws_security_group" "inline_rules" {
  description = "SG with inline rules"
  ingress = [
    {
      cidr_blocks      = ["0.0.0.0/0"]
      description      = "Wide Open"
      from_port        = 0
      ipv6_cidr_blocks = []
      prefix_list_ids  = []
      security_groups  = []
      protocol         = "-1"
      self             = false
      to_port          = 65535
    }
  ]
}

resource "aws_security_group" "separate_rules" {
  description = "SG with separate rules"
}

resource "aws_security_group_rule" "bad" {
  description       = "Allow inbound SSH from the world!"
  type              = "ingress"
  security_group_id = aws_security_group.separate_rules.id
  protocol          = "tcp"
  from_port         = 22
  to_port           = 22
  cidr_blocks       = ["0.0.0.0/0"]
}

resource "aws_security_group_rule" "worse" {
  description       = "Allow everything!"
  type              = "ingress"
  security_group_id = aws_security_group.separate_rules.id
  protocol          = "-1"
  from_port         = 0
  to_port           = 65535
  cidr_blocks       = ["0.0.0.0/0"]
}
```

</details>

**Expected behavior**

The security groups should be flagged

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS using Docker.app
 - Checkov Version: version: 1.0.290 

```
Using default tag: latest
latest: Pulling from bridgecrew/checkov
Digest: sha256:8ad4123715f6c75df75810aecea0b861e02dbe37dee823f4d2514147d7dd64d7
Status: Image is up to date for bridgecrew/checkov:latest
docker.io/bridgecrew/checkov:latest
```

**Additional context**

<details>

```
docker run -it --rm -v (pwd):/code bridgecrew/checkov -d /code

       _               _              
   ___| |__   ___  ___| | _______   __
  / __| '_ \ / _ \/ __| |/ / _ \ \ / /
 | (__| | | |  __/ (__|   < (_) \ V / 
  \___|_| |_|\___|\___|_|\_\___/ \_/  
                                      
version: 1.0.290 

[...]

PR #265, opened 2020-05-13
Fix AWS Security Group ingress check for SSH

acdha:
This fixes several of the problems in #264 by updating the expected checks for the input rule's data structure. It does not add a check for inline ingress rules on the `aws_security_group` itself.


acdha:
Has there been anything thought to not hard-coding the test data? This bug would have been caught quickly if the test suite parsed a Terraform file.

nimrodkor:
Yes, there has @acdha . However, please note that the tests fail both for the hardcoded confs and for the terraform file, for example `example.tf`, is failing to parse

nimrodkor:
Handled through #270

PR #275, opened 2020-05-18
Base resource value checks enhancements

tronxd:
- Support of hierarchy on `BaseResourceValueCheck` checks, in order to support non-root attribute checks
- Support of multiple expected values
- Support of a default expected value, according to best practices
- Identify attributes which are variable dependant, and mark these resources as `PASSED` under the check
- Incorporation of `CKV_ANY` value,  in the case the check looks for any value for the inspected key
- Conversion of `S3Encryption`, `LaunchConfigurationEBSEncryption` and `SQSQueueEncryption` to implement the  `BaseResourceValueCheck` class

### DIFF
--- a/checkov/common/models/consts.py
+++ b/checkov/common/models/consts.py
@@ -1,1 +1,2 @@
 SUPPORTED_FILE_EXTENSIONS = [".tf", ".yml", ".yaml", ".json", ".template"]
+ANY_VALUE = ["CKV_ANY"]
--- a/checkov/common/models/consts.py
+++ b/checkov/common/models/consts.py
@@ -1,2 +1,2 @@
 SUPPORTED_FILE_EXTENSIONS = [".tf", ".yml", ".yaml", ".json", ".template"]
-ANY_VALUE = ["CKV_ANY"]
+ANY_VALUE = "CKV_ANY"
--- a/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
+++ b/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
@@ -1,14 +1,17 @@
 from checkov.common.models.enums import CheckResult, CheckCategories
-from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
-class LaunchConfigurationEBSEncryption(BaseResourceCheck):
+class LaunchConfigurationEBSEncryption(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure all data stored in the Launch configuration EBS is securely encrypted "
         id = "CKV_AWS_8"
-        supported_resources = ['aws_launch_configuration','aws_instance']
+        supported_resources = ['aws_launch_configuration', 'aws_instance']
         categories = [CheckCategories.ENCRYPTION]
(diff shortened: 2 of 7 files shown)


checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
```diff
@@ -1,15 +1,18 @@
...
         id = "CKV_AWS_8"
-        supported_resources = ['aws_launch_configuration','aws_instance']
+        supported_resources = ['aws_launch_configuration', 'aws_instance']
         categories = [CheckCategories.ENCRYPTION]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
+    def get_inspected_key(self):
+        return "*_block_device/[0]/encrypted"
+
     def scan_resource_conf(self, conf):
```
nimrodkor, line 16: @tronxd Is this method still relevant?

checkov/terraform/checks/resource/base_resource_value_check.py
```diff
@@ -1,29 +1,92 @@
...
+
+    def get_expected_values(self):
+        """
+        Override the method with the list of acceptable values if the check has more than one possible expected value, given
+        the inspected key
+        :return: List of expected values, defaults to a list of the expected value
+        """
+        return [self.get_expected_value()]
 
     def get_expected_value(self):
```
nimrodkor, line 87: @tronxd Maybe we should rename to `get_default_passing_value`?

checkov/common/models/consts.py
```diff
@@ -1 +1,2 @@
-SUPPORTED_FILE_EXTENSIONS = [".tf", ".yml", ".yaml", ".json", ".template"]
\ No newline at end of file
+SUPPORTED_FILE_EXTENSIONS = [".tf", ".yml", ".yaml", ".json", ".template"]
+ANY_VALUE = ["CKV_ANY"]
```
nimrodkor, line 2: @tronxd Would expect this to be a string

checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
schosterbarak: backword compatability of external checks maybe?

checkov/terraform/checks/resource/base_resource_value_check.py
schosterbarak: please do not rename method names - for backward compatibility with eternal checks

checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
tronxd: @nimrodkor @schosterbarak 
This check is a special case where the "generic" `scan_resource_conf` of `BaseResourceValueCheck` won't satisfy the check:
The check looks for 2 expected keys that share a common postfix (`root_block_device` and `ebs_block_device`, but doesn't accept a key with the same postfix  `ephemeral_block_device`
That's why the implementation of `scan_resource_conf` still remains.

checkov/terraform/checks/resource/base_resource_value_check.py
tronxd: @nimrodkor didn't change because of the reason above

PR #278, opened 2020-05-20
Handle digests in image names for image pull checks

mikeurbanski1:
The image checks fail if the image has a digest (e.g., `repo/image:tag@sha256:12345`), because splitting on `:` returns three values.

This fixes that by removing the digest if it's present.


schosterbarak:
Thank you!

PR #283, opened 2020-05-21
Docker image name parsing

tronxd:
Parse docker image name to to extract the image name and tag via a best practice regex


checkov/common/models/consts.py
```diff
@@ -1,2 +1,3 @@
 SUPPORTED_FILE_EXTENSIONS = [".tf", ".yml", ".yaml", ".json", ".template"]
 ANY_VALUE = "CKV_ANY"
+DOCKER_IMAGE_REGEX = r'(?:[^\s\/]+/)?([^\s:]+)(?::)?([^\s]*)'
```
nimrodkor, line 3: @tronxd To be simpler, can be `:?` instead of `(?::)?`

tronxd: Absolutely!

PR #286, opened 2020-05-22
Add ECS cluster container insights check

### LINKED ISSUE #167: Add new check: ECS ensure container insights is enabled
https://www.terraform.io/docs/providers/aws/r/ecs_cluster.html
ClusterSettings: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-cluster.html#cfn-ecs-cluster-clustersettings

	https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-cluster-clustersetting.html

twmartin:
Closes #167

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/ECSClusterContainerInsights.py
+++ b/checkov/cloudformation/checks/resource/aws/ECSClusterContainerInsights.py
@@ -1,0 +1,27 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class ECSClusterContainerInsights(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure container insights are enabled on ECS cluster"
+        id = "CKV_AWS_65"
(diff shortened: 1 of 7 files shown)


nimrodkor:
Thanks for the contribution @twmartin, both the CF and the TF checks!

However, I think the `conf` you provided in the terraform tests are not the same structure as the actual result of hcl2 parsing - which means your terraform check is lacking. From my tests, the conf would look like this:
`{'name': ['test-cluster'], 'setting': [{'name': ['containerInsights'], 'value': ['enabled']}]}`

I suggest you check [this test out](https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/aws/test_SecurityGroupUnrestrictedIngress22.py) - it is in a different format, which ensures correctness of the conf passsed to the test

twmartin:
@nimrodkor - good call! I think starting the test from the HCL makes a lot of sense. Do things look better after 85cf049?

schosterbarak:
looks great. Thanks @twmartin

PR #288, opened 2020-05-22
Add cfn-lint to Github Actions PR workflow

twmartin:
It may be beneficial to hit the Cloudformation unit test examples with some `cfn-lint` (https://github.com/aws-cloudformation/cfn-python-lint) as a guardrail against using bunk cfn templates as test inputs.

Yay or nay?


twmartin:
I went down a bit of a rabbit hole on the Cloudformation security group rule description check. I feel like the check was doing more than was necessary in checking for descriptions on the security group resources themselves since those properties are required in the Cloudformation spec. Therefore, I removed that logic leaving only SG rule descriptions to be checked and pared down the unit tests accordingly.

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-4.json
```diff
@@ -1,40 +0,0 @@
-{
```
nimrodkor, line 1: @twmartin Why delete this test?

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-3.json
```diff
@@ -0,0 +1,14 @@
+{
```
nimrodkor, line 1: @twmartin Why convert it to json?

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED.yaml
```diff
@@ -1,21 +1,10 @@
-Description: Security Group Example
```
nimrodkor, line 1: @twmartin This is a good test for variable rendering, why remove the parameters block?

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-PASSED-2.yaml
```diff
@@ -1,17 +1,10 @@
-AWSTemplateFormatVersion: 2010-09-09
 Resources:
-  TargetSG:
```
nimrodkor, line 3: @twmartin Why remove the reference to this resource?

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-3.json
twmartin: I'm totally cool with converting this to yaml instead. I had seen that a json test existed in the samples previous to my changes, so I figured perhaps there was some significance to including a json cfn template in a check's unit test examples and therefore I should retain that. However, if there isn't much value in the inclusion of a json example here, I think it would be nice for all the test examples for a particular check to be of the same format.

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-4.json
twmartin: The new `SecurityGroupRuleDescription-FAILED-3.json` now plays the role of the "json-egress-rule-with-no-description" test.

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-3.json
twmartin: The role of "yaml-ingress-rule-with-no-description" test is now played by `SecurityGroupRuleDescription-FAILED-2.yaml`

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED-4.json
twmartin: I believe the 3 failed/passed test examples remaining in this PR provide sufficient coverage for the `SecurityGroupRuleDescription` Cloudformation check:
- SG resource w/ inline rule - `SecurityGroupRuleDescription-(FAILED|PASSED).yaml`
- Ingress rule resource - `SecurityGroupRuleDescription-(FAILED|PASSED)-2.yaml`
- Egress rule resource - `SecurityGroupRuleDescription-(FAILED|PASSED)-3.yaml`

For that reason, I deleted the examples above number 3.

twmartin:
> I agree with your code changes, however some of the changes you made to the tests files led them to becoming their most degenerate form - which is not what we tried to achieve. Do they fail cfn-lint?

My sentiment is that the best unit tests are those that are as minimal as reasonably possible yet still provide adequate coverage for the item being tested. I believe this leads to more reliable, maintainable, and performant unit tests. If that sentiment is not shared, I can certainly add back the example template contents that I had removed.

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-FAILED.yaml
twmartin: I feel as though it is beneficial to keep unit tests stripped as minimally as reasonably possible while still providing adequate coverage to the test case. Perhaps variable rendering test coverage should live in its own explicit test and not passively in the `SecurityGroupRuleDescription` unit test. With the absence of the parameters block from this file, is test coverage for variable rendering completely lost, or is it still retained by way of other example cfn templates/tests?

tests/cloudformation/checks/resource/aws/example_SecurityGroupRuleDescription/SecurityGroupRuleDescription-PASSED-2.yaml
twmartin: It seemed like another opportunity to strip the unit test down to the bare bones. I can certainly add things back, if my sentiment on the value of minimal unit tests is not shared.

nimrodkor:
Hi @twmartin !
I completely agree with your sentiment about unit tests. However, we do have some variable rendering  - for example extracting the `Default` value out of parameters ([Example](https://github.com/bridgecrewio/checkov/blob/c33d4ffdcc744fb43f5de8c3e4fba0df89d06e8c/checkov/cloudformation/runner.py#L66)).
The simplifications you made to the relevant tests "resolve" the rendering, hence leaving parts of the code that used to be tested, untested.
JSON vs yaml is not important and I thank you for removing the unnecessary checks & tests. However, for the cases that are still tested, I would prefer leaving them as is.

How does that sound?

twmartin:
@nimrodkor - alrighty. Are things looking better now, or should all of the examples include a parameter with associated `!Ref`?

PR #303, opened 2020-05-31
gcp new checks/virtual machines

arielkru:

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeBootDiskEncryption.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeBootDiskEncryption.py
@@ -1,0 +1,21 @@
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class GoogleComputeBootDiskEncryption(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure VM disks for critical VMs are encrypted with CustomerSupplied Encryption Keys (CSEK)"
+        id = "CKV_GCP_38"
+        supported_resources = ['google_compute_instance']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if 'boot_disk' in conf.keys():
+            if 'disk_encryption_key_raw' in conf['boot_disk'][0] or 'kms_key_self_link' in conf['boot_disk'][0]:
+                return CheckResult.PASSED
+        return CheckResult.FAILED
+
+
+
+check = GoogleComputeBootDiskEncryption()
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccountFullAccess.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccountFullAccess.py
@@ -22,16 +22,16 @@ class GoogleComputeDefaultServiceAccountFullAccess(BaseResourceCheck):
         :param conf: google_compute_instance configuration
         :return: <CheckResult>
         """
+        if conf['name'][0].startswith('gke-'):
+            return CheckResult.PASSED
         if 'service_account' in conf.keys():
             if 'email' in conf['service_account'][0]:
                 if re.match(DEFAULT_SERVICE_ACCOUNT, conf['service_account'][0]['email'][0]):
-                    if FULL_ACCESS_API in conf['service_account'][0]['scopes']:
+                    if FULL_ACCESS_API in conf['service_account'][0]['scopes'][0]:
                         return CheckResult.FAILED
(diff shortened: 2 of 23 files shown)


checkov/terraform/checks/resource/gcp/GoogleComputeBootDiskEncryption.py
```diff
@@ -0,0 +1,21 @@
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class GoogleComputeBootDiskEncryption(BaseResourceCheck):
```
nimrodkor, line 5: @arielkru Can this be refactored into `BaseResourceValueCheck`?

checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccountFullAccess.py
```diff
@@ -0,0 +1,37 @@
...
+        :param conf: google_compute_instance configuration
+        :return: <CheckResult>
+        """
+        if conf['name'][0].startswith('gke-'):
+            return CheckResult.PASSED
+        if 'service_account' in conf.keys():
+            if 'email' in conf['service_account'][0]:
+                if re.match(DEFAULT_SERVICE_ACCOUNT, conf['service_account'][0]['email'][0]):
+                    if FULL_ACCESS_API in conf['service_account'][0]['scopes'][0]:
+                        return CheckResult.FAILED
```
nimrodkor, line 31: @arielkru is `scopes` a list ([]) or a list of lists? ([[]]). If it is a list, need to skip the `[0]` as `in` is an operator for lists and strings

checkov/terraform/checks/resource/gcp/GoogleComputeDiskEncryption.py
```diff
@@ -0,0 +1,20 @@
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class GoogleComputeDiskEncryption(BaseResourceCheck):
```
nimrodkor, line 5: @arielkru Can this be refactored into `BaseResourceValueCheck`?

checkov/terraform/checks/resource/gcp/GoogleComputeIPForward.py
```diff
@@ -0,0 +1,23 @@
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class GoogleComputeIPForward(BaseResourceCheck):
```
nimrodkor, line 5: @arielkru Let's talk about how this can inherit from `BaseResourceValueCheck`

checkov/terraform/checks/resource/gcp/GoogleComputeShieldedVM.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        Shielded VM can only be used with boot images with shielded VM support.
+        See the complete list of supporting images here: https://cloud.google.com/compute/docs/images#shielded-images
+        :param conf:
+        :return: <checkResult>
+        """
+
+        if 'shielded_instance_config' in conf.keys():
+            if 'enable_vtpm' in conf['shielded_instance_config'][0] and \
+                    not conf['shielded_instance_config'][0].get('enable_vtpm')[0]:
```
nimrodkor, line 23: @arielkru `something.get('enable_vtpm')[0]` will fail if something doesn't have any value for `enable_vtpm`. please have the get return `[]` by default

checkov/terraform/checks/resource/gcp/GoogleComputeBootDiskEncryption.py
tronxd: @nimrodkor The scan logic doesn't match the use case, it checks for 2 existing keys

checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccountFullAccess.py
arielkru: It's a list of list

checkov/terraform/checks/resource/gcp/GoogleComputeDiskEncryption.py
arielkru: the scan is based on `len(conf['disk_encryption_key'][0]) > 0`, which is not an option in `BaseResourceValueCheck`  (if I use ANY_VALUE it would pass with an empty list, which is not the same check)

checkov/terraform/checks/resource/gcp/GoogleComputeShieldedVM.py
arielkru: It must have a value for it because the `if` first checks if the key `enable_vtpm` exists. actually there's no need to use the get function because I first check if the key exists, so I removed it.

PR #322, opened 2020-06-04
Do not fail on unsupported, complex variable evaluations

nimrodkor:

### DIFF
--- a/checkov/terraform/evaluation/evaluation_methods/const_variable_evaluation.py
+++ b/checkov/terraform/evaluation/evaluation_methods/const_variable_evaluation.py
@@ -1,7 +1,10 @@
-from checkov.terraform.evaluation.base_variable_evaluation import BaseVariableEvaluation
+import logging
 import os
-import dpath
 import re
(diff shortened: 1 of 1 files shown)


checkov/terraform/evaluation/evaluation_methods/const_variable_evaluation.py
```diff
@@ -43,6 +46,11 @@ def _assign_definition_value(self, definition_type, var_name, var_value, var_ass
                 definition_path = assignment_obj.get('definition_path')
                 entry_expression = assignment_obj.get('definition_expression')
                 definition_name = assignment_obj.get('definition_name')
```
tronxd, line 48: Isn't `local.ip_ranges.ipv4Prefixes[*].prefix` a string?

nimrodkor: @tronxd It is evaluated as a tree
image
Unlike regular variables:
image

PR #339, opened 2020-06-10
Fix logic in IAM Password Policy Expiration check

acdha:
A Terraform configuration like this fails with Checkov 1.0.366:

```hcl
resource "aws_iam_account_password_policy" "default" {
  allow_users_to_change_password = true
  minimum_password_length        = 16
  require_lowercase_characters   = true
  require_numbers                = true
  require_uppercase_characters   = true
  require_symbols                = true
  password_reuse_prevention      = 24
  max_password_age               = 60
}
```

I had not noticed because my configuration previously used 90 days following CIS but there's a separate issue with the configuration not matching the “within 90 days or less” description (see below).

The logic in this check was previously inverted and would only pass configurations which were outside of the expected range:

```
(Pdb) conf[key][0] = None
(Pdb) p not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90)
True
(Pdb) conf[key][0] = 0
(Pdb) p not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90)
True
(Pdb) conf[key][0] = 180
(Pdb) p not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90)
True
(Pdb) conf[key][0] = 90
(Pdb) p not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90)
True
(Pdb) conf[key][0] = 89
(Pdb) p not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90)
False
```

This commit also changes the check to be less than or equal to 90 days to match the AWS Config rule which is used by the AWS Security Hub CIS control:

https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-cis-controls.html#securityhub-cis-controls-1.11

### DIFF
--- a/checkov/terraform/checks/resource/aws/PasswordPolicyExpiration.py
+++ b/checkov/terraform/checks/resource/aws/PasswordPolicyExpiration.py
@@ -20,7 +20,8 @@ class PasswordPolicyExpiration(BaseResourceCheck):
         """
         key = 'max_password_age'
         if key in conf.keys():
-            if not (force_int(conf[key][0]) and force_int(conf[key][0]) < 90):
+            max_age = force_int(conf[key][0])
+            if max_age and max_age <= 90:
                 return CheckResult.PASSED
         return CheckResult.FAILED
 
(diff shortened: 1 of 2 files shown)


nimrodkor:
Thanks @acdha

PR #356, opened 2020-06-17
ignore common private directories

schosterbarak:
Ignore node_modules , .terraform and .serverless directories when parsing files

### DIFF
--- a/checkov/cloudformation/runner.py
+++ b/checkov/cloudformation/runner.py
@@ -6,11 +6,11 @@ from functools import reduce
 
 from checkov.cloudformation.checks.resource.registry import resource_registry
 from checkov.cloudformation.parser import parse
+from checkov.cloudformation.parser.node import dict_node
 from checkov.common.output.record import Record
 from checkov.common.output.report import Report
-from checkov.common.runners.base_runner import BaseRunner, ignored_directories
+from checkov.common.runners.base_runner import BaseRunner, filter_ignored_directories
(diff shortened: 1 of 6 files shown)


checkov/common/runners/base_runner.py
```diff
@@ -1,7 +1,12 @@
+import os
+
 from abc import ABC, abstractmethod
 
 from checkov.runner_filter import RunnerFilter
 
+IGNORED_DIRECTORIES_ENV = os.getenv('IGNORED_DIRECTORIES',"node_modules,.terraform,.serverless")
```
tronxd, line 7: I would add this env variable to docs, explaining it's a comma separated string of ignored directories.

checkov/cloudformation/runner.py
```diff
@@ -35,6 +35,7 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
 
         if root_folder:
             for root, d_names, f_names in os.walk(root_folder):
+                [d_names.remove(d) for d in list(d_names) if d in ignored_directories]
```
tronxd, line 38: I would refactor the filtering logic into checkov.common or to base runner class

checkov/common/runners/base_runner.py
prabhu: `CK_IGNORED_DIRECTORIES` something with a prefix

prabhu:
lgtm

PR #358, opened 2020-06-17
Created base_resource_negative_value_check.py

rotemavni:
- Modified checks in gcp and aws to implement BaseResourceNegativeValueCheck instead of BaseResourceCheck
- Overridden the abstract methods when needed
- Added one more test case in test_GoogleComputeInstanceOSLogin.py

### DIFF
--- a/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
+++ b/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
@@ -13,5 +13,8 @@ class RDSPubliclyAccessible(BaseResourceNegativeValueCheck):
     def get_inspected_key(self):
         return 'publicly_accessible'
 
+    def get_forbidden_values(self):
+        return []
+
 
 check = RDSPubliclyAccessible()
--- a/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
+++ b/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
@@ -1,4 +1,4 @@
-from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.common.models.enums import CheckCategories
 from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
 
 
@@ -6,22 +6,9 @@ class RDSPubliclyAccessible(BaseResourceNegativeValueCheck):
     def __init__(self):
         name = "Ensure all data stored in the RDS bucket is not public accessible"
         id = "CKV_AWS_17"
-        supported_resources = ['aws_db_instance','aws_rds_cluster_instance']
+        supported_resources = ['aws_db_instance', 'aws_rds_cluster_instance']
         categories = [CheckCategories.NETWORKING]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
-
-    def scan_resource_conf(self, conf):
-        """
-            Looks for publicly_accessible configuration at aws_db_instance:
-            https://www.terraform.io/docs/providers/aws/d/db_instance.html
-        :param conf: publicly_accessible configuration
-        :return: <CheckResult>
-        """
-        if 'publicly_accessible' in conf.keys():
-            key = conf['publicly_accessible'][0]
-            if key:
-                return CheckResult.FAILED
-        return CheckResult.PASSED
 
     def get_inspected_key(self):
         return 'publicly_accessible'
--- a/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
+++ b/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
@@ -1,3 +1,4 @@
+from checkov.common.models.consts import ANY_VALUE
 from checkov.common.models.enums import CheckCategories
 from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
 
@@ -14,7 +15,7 @@ class RDSPubliclyAccessible(BaseResourceNegativeValueCheck):
         return 'publicly_accessible'
 
     def get_forbidden_values(self):
-        return []
+        return [ANY_VALUE]
 
 
 check = RDSPubliclyAccessible()
--- a/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
+++ b/checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
@@ -1,8 +1,8 @@
 from checkov.common.models.enums import CheckResult, CheckCategories
-from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
 
 
-class RDSPubliclyAccessible(BaseResourceCheck):
+class RDSPubliclyAccessible(BaseResourceNegativeValueCheck):
     def __init__(self):
         name = "Ensure all data stored in the RDS bucket is not public accessible"
         id = "CKV_AWS_17"
@@ -22,6 +22,9 @@ class RDSPubliclyAccessible(BaseResourceCheck):
             if key:
                 return CheckResult.FAILED
         return CheckResult.PASSED
+
+    def get_inspected_key(self):
+        return 'publicly_accessible'
 
 
 check = RDSPubliclyAccessible()
--- a/checkov/terraform/checks/resource/base_resource_negative_value_check.py
+++ b/checkov/terraform/checks/resource/base_resource_negative_value_check.py
@@ -21,7 +21,7 @@ class BaseResourceNegativeValueCheck(BaseResourceCheck):
                     return CheckResult.PASSED
 
         inspected_key = self.get_inspected_key()
-        bad_values = self.get_bad_values()
+        bad_values = self.get_forbidden_values()
         if dpath.search(conf, inspected_key) != {}:
             if len(bad_values) == 0:
                 return CheckResult.FAILED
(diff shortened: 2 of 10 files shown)


checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        return None
+
+    def get_excluded_condition(self):
+        """
+        Returns the excluded value for the excluded key, which states the vulnerable value is valid for it
+        """
+        def check_condition(value):
+            return self.get_excluded_key() is not None
+        return check_condition
```
nimrodkor, line 37: @rotemavni Why wrap the method in a method?

checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,38 @@
+from abc import abstractmethod
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class BaseResourceNegativeValueCheck(BaseResourceCheck):
+    def __init__(self, name, id, categories, supported_resources):
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
```
nimrodkor, line 9: @rotemavni This method should be implemented here, for all the other checks

checkov/terraform/checks/resource/base_resource_negative_value_check.py
rotemavni: @nimrodkor Because this way you can provide different conditions for excluding a value. For example in `GoogleComputeIPForward` you would want to exclude the fix if `value.startswith('gke-')`. In other cases you might want to have a different condition such has `endwith` etc.

rotemavni: @nimrodkor issue fixed

checkov/terraform/checks/resource/base_resource_negative_value_check.py
rotemavni: @nimrodkor implemented the method

checkov/terraform/checks/resource/gcp/GoogleComputeSerialPorts.py
```diff
@@ -12,10 +12,16 @@ def __init__(self):
 
     def scan_resource_conf(self, conf):
```
tronxd: `BaseResourceNegativeValueCheck` defines how resource is scanned via `scan_resource_conf`
That means that its implementing checks shouldn't override it

checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        :return: JSONPath syntax path of the checked attribute
+        """
+        raise NotImplementedError()
+
+    def get_bad_values(self):
+        """
+        Returns the vulnerable value for the inspected, governed by provider best practices
+        """
+        return []
```
tronxd, line 48: If there's no default "bad" value, I would raise a `NotImplememtedError` exception here, so there will be a guarantee that the scanning would check some values against forbidden values

checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,38 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_inspected_key(self):
+        """
+        :return: JSONPath syntax path of the checked attribute
+        """
+        raise NotImplementedError()
+
+    def get_bad_values(self):
```
tronxd, line 44: I would rename from `bad` to `forbidden`

checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,62 @@
...
+                value = dpath.get(conf, excluded_key)
+                if isinstance(value, list) and len(value) == 1:
+                    value = value[0]
+                if self.check_excluded_condition(value):
+                    return CheckResult.PASSED
+
+        inspected_key = self.get_inspected_key()
+        bad_values = self.get_bad_values()
+        if dpath.search(conf, inspected_key) != {}:
+            if len(bad_values) == 0:
```
tronxd, line 26: In what scenario would this condition would be true? If the check implements the negative-values check, it should have at least one forbidden value of the inspected key

rotemavni: @tronxd There are two checks that forbid the key from existing at all, regardless of its value:
`RDSPubliclyAccessible (CKV_AWS_17)` https://github.com/bridgecrewio/checkov/pull/358/files#diff-1925ff1f8b443c99418c8f5e5efee48d
`GoogleComputeExternalIP (CKV_GCP_40)` https://github.com/bridgecrewio/checkov/pull/358/files#diff-8a9c56fa1e4a7208e776d2ae7716ece3

checkov/terraform/checks/resource/base_resource_negative_value_check.py
rotemavni: Done.

checkov/terraform/checks/resource/base_resource_negative_value_check.py
rotemavni: Done.

checkov/terraform/checks/resource/base_resource_negative_value_check.py
tronxd: @rotemavni If the attribute shouldn't exist at all, wouldn't it suffice to check the upper if? (`dpath.search(conf, inspected_key) != {}`)

rotemavni: @tronxd I don't think it will suffice because when you check `if value in bad_values`, and `bad_values` is empty, the condition will be `False` and the check will pass instead of fail. 
I can merge these conditions into one if : `if value in bad_values or len(bad_values) == 0: return CheckResult.FAILED`

tronxd: @rotemavni agree. just rename the vars `bad_values` to `forbidden_values` :)

checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -0,0 +1,60 @@
...
+                if self.check_excluded_condition(value):
+                    return CheckResult.PASSED
+
+        inspected_key = self.get_inspected_key()
+        bad_values = self.get_forbidden_values()
+        if dpath.search(conf, inspected_key) != {}:
+            value = dpath.get(conf, inspected_key)
+            if isinstance(value, list) and len(value) == 1:
+                value = value[0]
+            if value in bad_values or len(bad_values) == 0:
```
nimrodkor, line 29: @rotemavni Please use Checkov's `ANY_VALUE` instead of the empty list

checkov/terraform/checks/resource/aws/RDSPubliclyAccessible.py
```diff
@@ -1,27 +1,20 @@
...
-        if 'publicly_accessible' in conf.keys():
-            key = conf['publicly_accessible'][0]
-            if key:
-                return CheckResult.FAILED
-        return CheckResult.PASSED
+    def get_inspected_key(self):
+        return 'publicly_accessible'
+
+    def get_forbidden_values(self):
+        return []
```
nimrodkor, line 17: @rotemavni Should be `[ANY_VALUE]`

checkov/terraform/checks/resource/gcp/GoogleComputeExternalIP.py
```diff
@@ -1,18 +1,20 @@
...
 
-    def scan_resource_conf(self, conf):
-        if 'access_config' in conf.keys():
-            return CheckResult.FAILED
-        return CheckResult.PASSED
+    def get_inspected_key(self):
+        return 'access_config'
+
+    def get_forbidden_values(self):
+        return []
```
nimrodkor, line 17: @rotemavni Should be `[ANY_VALUE]`

checkov/terraform/checks/resource/gcp/GoogleComputeSerialPorts.py
```diff
@@ -1,21 +1,20 @@
...
-        if 'metadata' in conf.keys():
-            if 'serial-port-enable'in conf['metadata'][0]:
-                if conf['metadata'][0]['serial-port-enable'] not in ['0', False]:
-                    return CheckResult.FAILED
-        return CheckResult.PASSED
+    def get_inspected_key(self):
+        return 'metadata/[0]/serial-port-enable'
+
+    def get_forbidden_values(self):
+        return ['1', True]
```
nimrodkor, line 17: @tronxd Is this required? Or does chekov parse `1` and `"true"` as `true`?

tronxd: @nimrodkor yes, as a backward compatibility feature for Terraform v0.11 syntax
refer the function `evaluate_string_booleans` at `checkov/terraform/runner.py`

schosterbarak: @tronxd maybe we should maintain that list in a common enum?

checkov/terraform/checks/resource/base_resource_negative_value_check.py
rotemavni: Done

checkov/terraform/checks/resource/gcp/GoogleComputeSerialPorts.py
tronxd: @schosterbarak @rotemavni scanning occurs after `1` would be evaluated into `True`, so the forbidden values should be `[True]` only

PR #361, opened 2020-06-18
Use feature detection instead of version detection

cclauss:
Python porting best practice [___use feature detection instead of version detection___](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).


schosterbarak:
Thank you @cclaus

KevinHock:
@cclauss is everywhere I look 😮

Issue #363, opened 2020-06-18
False Negative CKV_AWS_8


horuscoding:
**Describe the bug**
The check CKV_AWS_8 will incorrectly pass if there are multiple *_block_device blocks configured, and the first has encryption enabled while the others do not.

**To Reproduce**
1. Generate file ckv_aws_8_test.tf in current directory
```
resource "aws_instance" "test" {
  ami                  = var.ami_id
  instance_type        = var.instance_type
  key_name             = var.key_name

  root_block_device {
    volume_type = "gp2"
    volume_size = var.root_volume_size
    encrypted = true
  }


  ebs_block_device {
    volume_type = "gp2"
    volume_size = var.ebs_volume_size
    device_name = "/dev/xvdb"
    encrypted = false
  }
}
```
2. Run cli command
```
checkov -f ckv_aws_8_test.tf
```
3. CKV_AWS_8 check passes

**Expected behavior**
The check should fail if there exists any *block_device configuration with encryption disabled.

**Screenshots**
Check passing incorrectly:
![image](https://user-images.githubusercontent.com/17894112/85059254-54c77500-b171-11ea-86f7-df370d3cc21f.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu: 18.04.1 LTS
 - Checkov Version: 1.0.382

**Additional context**
If the order of the two block_device blocks are switched like in this example code:
```
resource "aws_instance" "test" {
  ami                  = var.ami_id
  instance_type        = var.instance_type
  key_name             = var.key_name

  ebs_block_device {
    volume_type = "gp2"
    volume_size = var.ebs_volume_size
    device_name = "/dev/xvdb"
    encrypted = false
  }

 root_block_device {
    volume_type = "gp2"
    volume_size = var.root_volume_size
    encrypted = true
  }
}
```

The result is a fail, as expected
![image](https://user-images.githubusercontent.com/17894112/85059610-e931d780-b171-11ea-95fb-31817d36eb87.png)

arielkru:
Hi @horuscoding, 
We just updated checkov with a fix to your issue, it should now return the check's result after checking for encryption in all of the *_block_device blocks.
Thanks!

PR #392, opened 2020-06-29
CKV_AWS_8 - bug fix - check for encryption in all of the block_device…

arielkru:
…s before defining result

fixes: https://github.com/bridgecrewio/checkov/issues/363

### DIFF
--- a/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
+++ b/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
@@ -20,6 +20,7 @@ class LaunchConfigurationEBSEncryption(BaseResourceValueCheck):
         :param conf: aws_launch_configuration configuration
         :return: <CheckResult>
         """
+        all_blocks_encrypted = False
         for key in conf.keys():
             if "block_device" in key and "ephemeral" not in key:
                 if "encrypted" in conf[key][0]:
--- a/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
+++ b/checkov/terraform/checks/resource/aws/LaunchConfigurationEBSEncryption.py
@@ -26,9 +26,11 @@ class LaunchConfigurationEBSEncryption(BaseResourceValueCheck):
                     if conf[key][0]["encrypted"] == [False]:
                         return CheckResult.FAILED
                     else:
-                        return CheckResult.PASSED
+                        all_blocks_encrypted = True
                 else:
                     return CheckResult.FAILED
+        if all_blocks_encrypted:
+            return CheckResult.PASSED
         return CheckResult.FAILED
 
 
--- a/tests/terraform/checks/resource/aws/test_LaunchConfigurationEBSEncryption.py
+++ b/tests/terraform/checks/resource/aws/test_LaunchConfigurationEBSEncryption.py
@@ -1,5 +1,7 @@
 import unittest
 
+import hcl2
+
 from checkov.terraform.checks.resource.aws.LaunchConfigurationEBSEncryption import check
 from checkov.common.models.enums import CheckResult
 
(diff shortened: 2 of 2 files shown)


arielkru:
Solves #363

acdha:
This PR is causing false-positives on anything with a non-hardcoded value. I have a template which launches instances using a dynamic block (in this case, to force encrypted=true on the volumes declared in that AMI) and that causes this check to fail because it's getting the HCL expression rather than the resolved value:

```
(Pdb) p conf[key]
['${[for bdm in data.aws_ami.cis_amazon_linux2.block_device_mappings : bdm if bdm.device_name != data.aws_ami.cis_amazon_linux2.root_device_name]}']
(Pdb) pprint(conf)
{'ami': ['${data.aws_ami.cis_amazon_linux2.id}'],
 'associate_public_ip_address': [False],
 'count': [1],
 'dynamic': [{'ebs_block_device': {'content': [{'delete_on_termination': [True],
                                                'device_name': ['${ebs_block_device.value["device_name"]}'],
                                                'encrypted': [True],
                                                'iops': ['${ebs_block_device.value["ebs"]["iops"]}'],
                                                'kms_key_id': ['${var.kms_key_id}'],
                                                'snapshot_id': ['${ebs_block_device.value["ebs"]["snapshot_id"]}'],
                                                'volume_size': ['${lookup(local.volume_size_overrides,,ebs_block_device.value["device_name"],,ebs_block_device.value["ebs"]["volume_size"],)}'],
                                                'volume_type': ['${ebs_block_device.value["ebs"]["volume_type"]}']}],
                                   'for_each': ['${[for bdm in '
                                                'data.aws_ami.cis_amazon_linux2.block_device_mappings '
                                                ': bdm if bdm.device_name != '
                                                'data.aws_ami.cis_amazon_linux2.root_device_name]}']}}],
 'ebs_block_device': ['${[for bdm in '
                      'data.aws_ami.cis_amazon_linux2.block_device_mappings : '
                      'bdm if bdm.device_name != '
                      'data.aws_ami.cis_amazon_linux2.root_device_name]}'],
 'iam_instance_profile': ['${var.iam_instance_profile}'],
 'instance_type': ['m5n.large'],
 'metadata_options': [{'http_endpoint': ['enabled'],
                       'http_put_response_hop_limit': [1],
                       'http_tokens': ['required']}],
 'root_block_device': [{'encrypted': [True],
                        'kms_key_id': ['${var.kms_key_id}'],
                        'volume_size': [16]}],
 'subnet_id': ['${var.subnet_ids[count.index % length(var.subnet_ids)]}'],
 'tags': ["${merge(var.tags,{'Name': '${var.name_prefix}-${count.index + "
          "1}'})}"],
 'user_data_base64': ['${base64encode(var.user_data)}'],
 'volume_tags': ["${merge(var.tags,{'Name': '${var.name_prefix}-${count.index "
                 "+ 1}'})}"],
 'vpc_security_group_ids': ['${var.vpc_security_group_ids}']}
```

acdha:
The logic for this appears to have a latent bug, as well: it has a path which sets `all_blocks_encrypted = True` but that value can never be cleared. It looks like that's currently safe because the loop exits early with a failure the first time an unencrypted block is encountered but if that changed in the future you'd have a false-negative if there was ever an encrypted device followed by an unencrypted one.

I would personally favor a simpler structure:

```python
        for key in conf.keys():
            if "block_device" in key and "ephemeral" not in key:
                if conf[key][0].get("encrypted") != [True]:
                    return CheckResult.FAILED

        return CheckResult.PASSED
```

This makes the failure to handle dynamic expressions obvious because `"encrypted" in conf[key][0]` silently fails because that's a valid string-contains check on Python rather than the expected key-in-dictionary check the code assumes but `conf[key][0].get("encrypted")` will get an exception.

arielkru:
Hi @acdha,
Thanks for the input, you're right the check didn't handle properly dynamic blocks.
I created a new PR based on your suggestion: #397, which should solve the problem.

PR #423, opened 2020-07-06
Handle assignment bug, unnamed variables, non-existent params

nimrodkor:

### DIFF
--- a/checkov/arm/context_parser.py
+++ b/checkov/arm/context_parser.py
@@ -43,10 +43,9 @@ class ContextParser(object):
         for key_entry in keys_w_params:
             param = re.sub("\[parameters\('|'\)\]", "", self._get_from_dict(dict(self.arm_template),
                                                                             key_entry[:-1])[key_entry[-1]])
-            self._set_in_dict(dict(self.arm_template), key_entry, parameter_defaults[param])
-            logging.debug(
-                "Replacing parameter {} in file {} with default value: {}".format(param, self.arm_file,
-                                                                                  parameter_defaults[param]))
+            if param in parameter_defaults:
+                logging.debug(f"Replacing parameter {param} in file {self.arm_file} with default value: {parameter_defaults[param]}")
+                self._set_in_dict(dict(self.arm_template), key_entry, parameter_defaults[param])
 
         for key_entry in keys_w_vars:
             param = re.sub("\[variables\('|'\)\]", "", self._get_from_dict(dict(self.arm_template),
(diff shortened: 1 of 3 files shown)


schosterbarak:
@nimrodkor can you elaborate?

checkov/arm/context_parser.py
```diff
@@ -43,10 +43,9 @@ def evaluate_default_parameters(self):
         for key_entry in keys_w_params:
             param = re.sub("\[parameters\('|'\)\]", "", self._get_from_dict(dict(self.arm_template),
                                                                             key_entry[:-1])[key_entry[-1]])
-            self._set_in_dict(dict(self.arm_template), key_entry, parameter_defaults[param])
-            logging.debug(
-                "Replacing parameter {} in file {} with default value: {}".format(param, self.arm_file,
-                                                                                  parameter_defaults[param]))
+            if param in parameter_defaults:
```
nimrodkor, line 46: @schosterbarak Handled case of parameter having `null` as the default value, which means it does not have a default value

checkov/cloudformation/runner.py
```diff
@@ -40,7 +40,10 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
 
             for file in files_list:
                 relative_file_path = f'/{os.path.relpath(file, os.path.commonprefix((root_folder, file)))}'
-                (definitions[relative_file_path], definitions_raw[relative_file_path]) = parse(file)
+                try:
+                    (definitions[relative_file_path], definitions_raw[relative_file_path]) = parse(file)
+                except TypeError:
+                    logging.info(f'CloudFormation skipping {file} as it is not a valid CF template')
```
nimrodkor, line 46: @schosterbarak Handles cases where parsing of CF file failed. there are many `yaml` and `json` files, and some although have blocks similar to CF templates, are not CF templates. we are using the AWS [cfn_yaml](checkov/cloudformation/parser/cfn_yaml.py) parser. So if it fails - it's safe to assume this is not a valid CFN template

checkov/terraform/evaluation/evaluation_methods/const_variable_evaluation.py
```diff
@@ -101,7 +101,9 @@ def _assign_definitions(self, assignment_files, folder):
         for var_file, variable_assignments in assignment_files.items():
             relative_var_file = f'/{os.path.relpath(var_file, self.root_folder)}'
             for definition_type in variable_assignments.keys():
-                for var_name, var_value in variable_assignments[definition_type]['assignments'].items():
+                for var_name, var_value in variable_assignments[definition_type].get('assignments', {}).items():
+                    if var_name == '':
+                        continue
```
nimrodkor, line 106: @schosterbarak Checkov was run agains a repository with a bad commit, which included the following statement in the variables block:
```
variable "" {
}
```
This variable does not have a name, and therefore cannot be referenced. It will now be skipped

Issue #146, opened 2020-03-25
Add new check: CloudFront Distribution should have WAF enabled 


jonjozwiak:
AWS WAF gives you control over how traffic reaches your applications by enabling you to create security rules that block common attack patterns, such as SQL injection or cross-site scripting, and rules that filter out specific traffic patterns you define. 

web_acl_id: https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html
WebAclId: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-distributionconfig.html

njgibbon:
https://github.com/bridgecrewio/checkov/pull/736 

This should close this issue.

PR #431, opened 2020-07-10
Add check to test is WAF is attached to Cloudfront distribution #146

JamesWoolfenden:
This is an attempt to resolve "Add new check: CloudFront Distribution should have WAF enabled #146" for Terraform. Im trying to use you new HCL2 format so hopefully this conforms?
 I tested and build and installed the package and it works with my module https://github.com/JamesWoolfenden/terraform-aws-cloudfront-s3.


checkov/terraform/checks/resource/aws/WAFEnabled.py
```diff
@@ -0,0 +1,21 @@
...
+from checkov.common.models.consts import ANY_VALUE
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class WAFEnabled(BaseResourceValueCheck):
+    def __init__(self):
+        name = "CloudFront Distribution should have WAF enabled"
+        id = "CKV_AWS_68"
+        supported_resources = ['aws_cloudfront_distribution']
+        categories = [CheckCategories.ENCRYPTION]
```
schosterbarak, line 11: I think WAF deserves a new category. WDYT on CheckCategories.APPLICATION_SECURITY ?

checkov/terraform/checks/resource/__init__.py
```diff
@@ -1,3 +1,4 @@
 from checkov.terraform.checks.resource.aws import *
 from checkov.terraform.checks.resource.gcp import *
 from checkov.terraform.checks.resource.azure import *
+from checkov.terraform.checks.resource.github import *
```
schosterbarak, line 4: sweet!

JamesWoolfenden: i got a little carried away here, looks like this PR also has the new provider and test for github

checkov/terraform/checks/resource/aws/WAFEnabled.py
JamesWoolfenden: Cool, added.

checkov/terraform/checks/resource/__init__.py
schosterbarak: np. looks awesome. would you like to merge those together? 
If you feel like the github is ready, we can merge those.

JamesWoolfenden: this was supposed to be a separate PR but its all the code to add support for the github provider and its first check. Hope thats ok or i can split to another PR. Got a bit keen.

PR #435, opened 2020-07-11
Handle GCP firewall ports as int as well

nimrodkor:


schosterbarak:
See failure in bandit

PR #456, opened 2020-07-19
Policies sharing / Enable loading external checks from git

schosterbarak:
ability to load external checks from github repo


schosterbarak:
> This looks brilliant. Use of go-getter for specific sub-dirs is a really smart addition to a great feature request!

I only used go-getter as a source of inspiration. haven't used it directly since it's in go. 
I do hope checkov will achieve the same capability as it grows.

Thanks for the feedback!

Issue #171, opened 2020-03-25
Add new check: Ensure Amazon MQ Brokers are not publicly exposed 


jonjozwiak:
publicly_accessible: https://www.terraform.io/docs/providers/aws/r/mq_broker.html#publicly_accessible
PubliclyAccessible: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-amazonmq-broker.html

PR #467, opened 2020-07-23
171: add check for MQ Broker not publicly exposed

mattfenwick:

----

Hi 👋 excited to put up my first checkov PR, hope I followed [all the guidelines](https://www.checkov.io/1.Introduction/Policies.html)!

This address issue https://github.com/bridgecrewio/checkov/issues/171.

### Details I'm not sure about

 - [x] did I use the right `ckv_aws` id?
 - [x] did I use the right category?
 - [x] is it right to treat "not present" as "passing" in the check, or should that be "unknown"?

### Notes

How I got the next id:  `git grep -h -i -P -o ckv_aws_\(\\d+\) | sort | uniq`

Test plan: `pipenv run python -m coverage run -m pytest tests/terraform/checks/resource/aws`

### DIFF
--- a/checkov/terraform/checks/resource/aws/MQBrokerNotPubliclyExposed.py
+++ b/checkov/terraform/checks/resource/aws/MQBrokerNotPubliclyExposed.py
@@ -1,0 +1,27 @@
+from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class MQBrokerNotPubliclyExposed(BaseResourceNegativeValueCheck):
+    def __init__(self):
+        name = "Ensure MQ Broker is not publicly exposed"
+        id = "CKV_AWS_69"
+        supported_resources = ['aws_mq_broker']
+        categories = [CheckCategories.NETWORKING]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self):
+        """
+        Looks for public accessibility:
+            https://www.terraform.io/docs/providers/aws/r/mq_broker.html#publicly_accessible
+            https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-amazonmq-broker.html
+        :param conf: aws_launch_configuration configuration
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/aws/MQBrokerNotPubliclyExposed.py
```diff
@@ -0,0 +1,24 @@
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
+from checkov.common.models.enums import CheckResult, CheckCategories
+
+
+class MQBrokerNotPubliclyExposed(BaseResourceCheck):
```
nimrodkor, line 5: @mattfenwick I seems this class can inherit from `BaseResourceNegativeValueCheck` - just specify the path to the bad key and it will do the rest

tests/terraform/checks/resource/aws/test_MQBrokerNotPubliclyExposed.py
```diff
@@ -0,0 +1,61 @@
+import unittest
+
+from checkov.terraform.checks.resource.aws.MQBrokerNotPubliclyExposed import check
+from checkov.common.models.enums import CheckResult
+
+
+class TestMQBrokerNotPubliclyExposed(unittest.TestCase):
+
+    def test_failure_mqbroker_logging(self):
+        resource_conf = {
```
nimrodkor, line 10: @mattfenwick To ensure the conf is configured correctly, try to write the matching HCL block and then parse it using hcl2. [Example for this pattern](https://github.com/bridgecrewio/checkov/blob/ddb9b8eca0774796ba44d29f0460fe4576b69792/tests/terraform/checks/resource/aws/test_IAMRoleAllowsPublicAssume.py#L11)

checkov/terraform/checks/resource/aws/MQBrokerNotPubliclyExposed.py
mattfenwick: 👍

tests/terraform/checks/resource/aws/test_MQBrokerNotPubliclyExposed.py
mattfenwick: 👍

PR #471, opened 2020-07-24
Unpin dpath, update to 2.0.1

robeden:


schosterbarak:
Hey @robeden thank you!
What is the scenario that led you into the need for a newer version of dpath?

robeden:
My main tool is an AWS Lambda which has checkov as a pip dependency. I then fork it from the same env. As a result, the dpath used by checkov is the same one I’m required to use.
There are other ways I could do that, so it’s not a huge deal if you don’t want to update for some reason, but I figured it would be good to keep up.

robeden:
Oh, and dpath 2 allows defaults on `get`, so you don’t have to raise a KeyError if an element isn’t found. That was the main functionality I wanted.

schosterbarak:
Hi @robeden this currently breaks internal integrations on our private repos. we'll have to wait for merging this one.

robeden:
@schosterbarak - Understood, thanks for letting me know. Hopefully that tech debt can get paid down at some point.

robeden:
FYI: https://github.com/akesterson/dpath-python/issues/136

schosterbarak:
@tronxd @nimrodkor is it ok to coordinate this for this week releases?

nimrodkor (commented):
Hey @robeden !

Sorry for getting to this so late. Just remove the update to the brew formula and we're good :)

robeden:
@nimrodkor - This was on hold because you guys had downstream conflicts. Has that been resolved so you're ready to move forward? If so, this has grown pretty stale so I can resync it.

Issue #452, opened 2020-07-17
False positive: CKV_AWS_19: "Ensure all data stored in the S3 bucket is securely encrypted at rest"


ismailyenigul:
**Describe the bug**

checkov fails on encryption enabled bucket

**To Reproduce**
```
check: CKV_AWS_19: "Ensure all data stored in the S3 bucket is securely encrypted at rest"
	FAILED for resource: aws_s3_bucket.my
	Guide: https://docs.bridgecrew.io/docs/s3_14-data-encrypted-at-rest
	File: /my/golden/s3.tf:1-18

		1  | resource "aws_s3_bucket" "my" {
		2  |   bucket = "my-${lower(var.environment)}-my-${lower(var.my)}"
		3  |   acl    = "private"
		4  |   # Delete S3 Bucket even it has some objects in it
		5  |   force_destroy = "true"
		6  | 
		7  |   server_side_encryption_configuration {
		8  |     rule {
		9  |       apply_server_side_encryption_by_default { sse_algorithm = "AES256" }
		10 |     }
		11 |   }
		12 | 
		13 |   versioning {
		14 |     enabled = false
		15 |   }
		16 | 
		17 | 
		18 | }
```
 
Version:
``` 
checkov -v
1.0.447
```

nimrodkor:
Hi @ismailyenigul !

I think the problem is not a bad check. When scanning this block:
```terraform
resource "aws_s3_bucket" "my" {
  bucket = "my-${lower(var.environment)}-my-${lower(var.my)}"
  acl    = "private"
  # Delete S3 Bucket even it has some objects in it
  force_destroy = "true"

  server_side_encryption_configuration {
    rule {
      apply_server_side_encryption_by_default { 
        sse_algorithm = "AES256" 
      }
    }
  }

  versioning {
    enabled = false
  }
}
```

The difference is that the `sse_algorithm` is in it's own line - and the result is `"PASSED"` for me.
We know of this issue and plan to handle it soon. Thanks for reporting!

ismailyenigul:
Cool thanks @nimrodkor

ismailyenigul:
Another check about image pull policy.
I think this check should failed if the `latest` tag used.
If we  are not using `latest` tag, it should not be an issue no?


```
Check: CKV_K8S_15: "Image Pull Policy should be Always"
	FAILED for resource: Deployment.calico-typha-horizontal-autoscaler.kube-system (container 0)
[...]

Issue #355, opened 2020-06-17
Checkov crashes when parsing kube_config field


holgiosalos:
**Describe the bug**
Parsing fails when processing a resource or provider related to a Kubernetes cluster and the following notation is used in some field `kube_config.0.property` notation.

**To Reproduce**
1. Provider defined in a similar way than in this example:
```
provider "helm" {
  version = "1.1.1"
  kubernetes {
    host                   = module.aks_cluster.kube_config.0.host
    client_certificate     = base64decode(module.aks_cluster.kube_config.0.client_certificate)
    client_key             = base64decode(module.aks_cluster.kube_config.0.client_key)
    cluster_ca_certificate = base64decode(module.aks_cluster.kube_config.0.cluster_ca_certificate)
    load_config_file       = false
  }
}

provider "kubernetes" {
  version                = "1.10.0"
  host                   = module.aks_cluster.kube_config.0.host
  client_certificate     = base64decode(module.aks_cluster.kube_config.0.client_certificate)
  client_key             = base64decode(module.aks_cluster.kube_config.0.client_key)
  cluster_ca_certificate = base64decode(module.aks_cluster.kube_config.0.cluster_ca_certificate)
}
```
2. Module defined as
```
module "new_relic" {
  source                            = "s3::https://s3.amazonaws.com/my-artifacts/new-relic-k8s-0.2.5.zip"
  kubernetes_host                   = module.aks_cluster.kube_config.0.host
  kubernetes_client_certificate     = base64decode(module.aks_cluster.kube_config.0.client_certificate)
  kubernetes_client_key             = base64decode(module.aks_cluster.kube_config.0.client_key)
  kubernetes_cluster_ca_certificate = base64decode(module.aks_cluster.kube_config.0.cluster_ca_certificate)
  cluster_name                      = module.naming_conventions.aks_name
  new_relic_license                 = data.vault_generic_secret.new_relic_license.data["license"]
  cluster_ca_bundle_b64             = module.aks_cluster.kube_config.0.cluster_ca_certificate
  module_depends_on                 = [null_resource.delay_aks_deployments]
}
```
3. Run cli command : checkov -d folder
```
DEBUG:checkov.terraform.parser:failed while parsing file /home/holgiosalos/myterraform/myazuremodule/1-main.tf
Traceback (most recent call last):
  File "/home/holgiosalos/.local/lib/python3.8/site-packages/lark/parsers/lalr_parser.py", line 62, in get_action
    return states[state][token.type]
[...]

PR #477, opened 2020-08-02
fix hcl2 parser 

schosterbarak:
fix 2 scenarios where hcl2 parser crashes by repackaging it with fixed lark config.
This solves:
https://github.com/bridgecrewio/checkov/issues/355
https://github.com/bridgecrewio/checkov/issues/452


robeden:
Where is the code for bc-python-hcl2? I assume it's a Bridgecrew fork, but couldn't find the source.

schosterbarak:
@robeden it's here: https://github.com/bridgecrewio/python-hcl2
We have two open PRs to the original one. They are open for a while now... I decided to build another package for now.

Issue #480, opened 2020-08-06
Enable checks for terraform module call


angatha:
I wan't to wirte a costume check for the source of a terraform module call.

```hcl
module "some_module {
    source = "something to check"
}
```

I didn't found a way how this check could be implemented in the current version.

An implementation could be really similar to the current implementation of terraform resources.
Using `module` as the type, adapted from the usage in terraform, could be used to implement a check registry.

I already implemented this way and will add it in the related pull request.

I don't see a general check for module calls, but having no support prevents costume checks that make sense in a special case like mine.

schosterbarak:
Hi @angatha thank you for submitting this. can you share examples of the use case? when would a user should do such check?

angatha:
Hi @schosterbarak, I use another git repository as the source and refer to specific versions of the module via tags. To keep my code up to date, I wan't a check that knows about the versions available and will fail if a newer version is available.

I think that is really specific and therefore I didn't suggest the check.

PR #481, opened 2020-08-06
Enable checks for terraform module calls

angatha:
This is my solution to #480.
I tested it locally using the check for which i requested the feature.


schosterbarak:
Looks great!
Do you mind adding a UT with a dummy check to continuously validates the loading is working?

angatha:
Thanks a lot too for the fast acceptance.

PR #486, opened 2020-08-07
Use defaultdict

angatha:
`collections.defaultdict` provides an elegant way to initialise the value for a key on the fly.


schosterbarak:
Thanks

Issue #45, opened 2020-01-07
Allow checks to run globally across all resources


m-wynn:
**Is your feature request related to a problem? Please describe.**
I would like to define global checks, e.g. a Naming Convention check, that can apply globally to all resources, without me having to write out all the supported resources

**Describe the solution you'd like**
I would like to write something like:

```python
from checkov.terraform.checks.resource.base_check import BaseResourceCheck

class NamingCheck(BaseResourceCheck):
    def __init__(self):
        name = "Ensure naming convention is correct"
        id = "MY_GLOBAL_001"
        supported_resources = ['*']
        # ...
```
The registry's get_checks could also look at the `*` resource, de-duplicating as necessary (maybe it could be a set?)

**Describe alternatives you've considered**
Get a list of all terraform resources and put that in the list

**Additional context**
It could also be neat to do, for example, `supported_resources = ["aws_*"]`, or more advanced matching, but that is not necessary for my purposes right now.

"Resource names must start with a letter or underscore, and may contain only letters, digits, underscores, and dashes" ([source](https://www.terraform.io/docs/configuration/resources.html#resource-syntax)) so anything else should be fair game for matching

schosterbarak:
Thank you @m-wynn . Do you see additional use-cases for this feature outside of naming convention?
I wonder if there is a built-in best practice check that we should add to utilize this capability.

m-wynn:
I'll try to think of a few.  Some that we might look at are:
* Avoid hard-coding resource names ARNs
* Add tags on resources
* jsonencode HCL objects intead of writing heredocs with embedded json
* check strings for common misspellings

JamesWoolfenden:
the tags is a good one, as far as AWS json policies shouldn't they all be data sources like data aws_iam_policy_document now?

m-wynn:
They *should* but they syntactically don't have to be (see https://www.terraform.io/docs/providers/aws/r/iam_policy.html )

m-wynn:
There's also other endpoints like https://www.terraform.io/docs/providers/vault/r/generic_endpoint.html which, while non-AWS, could still use jsonencode

bobdoah:
I have a use case to check that all resources _that can be tagged_ are tagged with specific tags. I'm not sure if that could be defined using just wildcards, but it would probably be a start.

kmroz0991:
[...]

PR #487, opened 2020-08-08
Implement wildcard for resource selector

angatha:
As requested in #45 this implements patterns as resource selectors. The example code from the issue:

```python
from checkov.terraform.checks.resource.base_check import BaseResourceCheck

class NamingCheck(BaseResourceCheck):
    def __init__(self):
        name = "Ensure naming convention is correct"
        id = "MY_GLOBAL_001"
        supported_resources = ['*']
        # ...
```

will register the test for every terraform resource. This happens, because the test is registered in this registry. This gives the limitation of `"*"` meaning not everything but in the case of terraform every data or every resource, ...

Maybe a generic base for terraform in combination with class decorators or mixins to define for which parts (data, resource, etc.) the check should apply.


angatha:
This implementation makes it (nearly) impossible to exclude some resources like: 

> I wan't the check for all `aws_*` but not on `aws_se_bucket`

A similar issue rises for the use cases of validation names as @m-wynn stated in #45. For `aws_s3_bucket`, this would be bucket and not name as it is usually. A clean implementation should use the resource type.

One way to implement this would be the suggested generic base class, but maybe this is a generell enhancement for all checks. You are already able to specify multiple resources for a check. On the other hand, this may be a motivation to write complex checks for multiple resources that should be done in multiple small checks.

The changes for the generic solution wold change the signature of `scan_entity_conf` in `checkov/common/checks/base_check.py#BaseCheck` to:

```python

    @abstractmethod
    def scan_entity_conf(self, conf, entity_type):
        raise NotImplementedError()
```

angatha:
I will postpone support for registering into multiple registries. I don't see a use case where such wide range is required.

schosterbarak:
Hi @angatha  Thank you for this. it looks like a big change. i think changing BaseResourceCheck and similar classes might break custom policies that are imported dynamically ( i haven't checked)  - if it does it might cause issues to some of the checkov users. 
I'll try to review it sometime in the upcoming week or so. Sorry for the late reply.

angatha:
I added some code that keep old code working that extends `checkov.common.checks.base_check.BaseCheck` directly.

If the subclass doesn't expect the `entity_type` parameter, it is wrapped into another method that expects it but discard it. The code is written in such a way, that such changes are easily done in future (if checks need more context.)

angatha:
I thought about removing the Typed base classes and applying the same method that is used to patch `scan_entity_conf` to them.
To make it more easily usable with an annotation that enables patching and adds a function to the method object that is by itself an annotation, linke the property annotation does, to register the wrapping functions:

```python
class BaseResourceCheck(BaseCheck):

    @fixable
    @abstractmethod
    def scan_resource_conf(self, conf, entity_type):
        raise NotImplementedError()

    @scan_resource_conf.wrapper_for("self", "conf")
    def _scan_resource_conf_self_conf(wrapped):
        def wrapper(self, conf, entity_type):
            return wrapped(self, conf)

        return wrapper
```

This will allow the implementation of `scan_resource_conf` to be:

```python
    ...
```

angatha:
I added `checkov.common.multi_signature`. This should provide compatibility in every place, not just from the check base clases.

tests/terraform/checks/resource/aws/test_IAMAdminPolicyDocument.py
```diff
@@ -13,7 +13,7 @@ def test_success(self):
                          "Action": [\n        "ec2:Describe*"\n      ],\n      \
                          "Effect": "Allow",\n     \
                           "Resource": "*"\n    }\n  ]\n}']}
-        scan_result = check.scan_entity_conf(conf=resource_conf)
+        scan_result = check.scan_entity_conf(conf=resource_conf, entity_type='aws_iam_policy')
```
schosterbarak, line 16: why is entity_type addition required?

angatha: As mentioned in #45 the wildcard feature could be used for a naming convention. In the case of aws with terraform this is usually name but e.g. for S3 buckets this is bucket. `entity_type` is a way to catch such edge cases. Checking if `conf` contains the value is possible but I can't guarantee that this not lead to false positives in some cases. Giving a ground truth about the type gives such guarantee.

Do you have another solution?

nimrodkor (commented):
@angatha Thanks for this! Looks pretty cool.
I agree that the `conf` isn't enough and we need the resource type. However, I don't see any reason why the `scan_entity_conf` needs to know the resource type. That would allow us to avoid the major change.
So - why does `scan_entity_conf` need to know the resource type?

angatha:
@nimrodkor how would you pass the type to the check? `scan_entity_conf` is the function that is called to perform the check.

An other way, which I definitely advise against, is a field of the check instance itself. But this introduces a state to each check.

nimrodkor:
@angatha My question is a step beforehand - why does the check need to know the type of the conf?

angatha:
> @angatha Thanks for this! Looks pretty cool.
> I agree that the `conf` isn't enough and we need the resource type. However, I don't see any reason why the `scan_entity_conf` needs to know the resource type. That would allow us to avoid the major change.
> So - why does `scan_entity_conf` need to know the resource type?

You agreed that the resource type is needed. I thought you agreed, that resource type is needed at the place where all the existing checks analyse `conf`.

One example why `resource_type` is required at this place, I wrote here: https://github.com/bridgecrewio/checkov/pull/487#discussion_r471928217

Or did I misunderstand you?

nimrodkor:
Hey @angatha !
It will be much easier to speak on slack, mind joining our workspace?
https://codified-security.herokuapp.com/
We'll sort it out quickly, I'm sure!

checkov/terraform/checks/provider/base_check.py
```diff
@@ -11,9 +12,19 @@ def __init__(self, name, id, categories, supported_provider):
...
         raise NotImplementedError()
 
-    def scan_entity_conf(self, conf):
-        return self.scan_provider_conf(conf)
+    @classmethod
+    @scan_provider_conf.add_signature(args=["self", "conf"])
+    def _scan_provider_conf_self_conf(cls, wrapped):
+        def wrapper(self, conf, entity_type=None):
+            # keep default argument for entity_type so old code, that doesn't set it, will work.
+            return wrapped(self, conf)
```
nimrodkor, line 28: @angatha Wouldn't the `entity_type` be always `provider` here?

checkov/terraform/checks/module/base_module_check.py
```diff
@@ -11,9 +12,19 @@ def __init__(self, name, id, categories, supported_resources):
...
+    def scan_module_conf(self, conf, entity_type):
         raise NotImplementedError()
 
-    def scan_entity_conf(self, conf):
-        return self.scan_resource_conf(conf)
+    @classmethod
+    @scan_module_conf.add_signature(args=["self", "conf"])
+    def _scan_module_conf_self_conf(cls, wrapped):
+        def wrapper(self, conf, entity_type=None):
+            # keep default argument for entity_type so old code, that doesn't set it, will work.
```
nimrodkor, line 27: @angatha Wouldn't the entity_type be always `module` here?

tests/terraform/checks/resource/aws/test_IAMAdminPolicyDocument.py
nimrodkor: @angatha If I understand correctly, this change is not required. Right?
If that's so - let's revert changes to the old tests (`test_IAMAdminPolicyDocument.py` & `test_IAMStarActionPolicyDocument.py`) please

nimrodkor (commented):
@angatha Thanks!
Looks pretty cool indeed. Just some minor comments before we merge this.

💪

checkov/terraform/checks/provider/base_check.py
angatha: Yes

checkov/terraform/checks/module/base_module_check.py
angatha: Yes

tests/terraform/checks/resource/aws/test_IAMAdminPolicyDocument.py
angatha: @schosterbarak sorry, when I first answered to your question I hade something else in mind.

@nimrodkor I would say it is required. Let me explain, why entity_type on `checkov.common.checks.base_check.BaseCheck.scan_entity_conf(...)` does not have a default value. This then implies, why the test, that called this directly needs to be adjusted.

When writing a check, that uses `entity_type`, you don't want to check if it is not None, because how should a check, that requires this information, handle it? It is nothing that depends on the files checkov is scanning but entirely on the implementation of checkov. And this implementation can guarantee that when running a check by calling `scan_entity_conf` from `checkov.common.checks.base_check.BaseCheck.run(...)` is not None (we can anotate the type as `str` and not `typing.Optional[str]`).

At this point, the check implementation expects a `str` and therefore the check code should pass the correct `entity_type`. (Even if the check doesn't use it.)

checkov/terraform/checks/provider/base_check.py
angatha: @nimrodkor In this case (and the one below) the `entity_type` add no value and it could be removed.

tests/terraform/checks/resource/aws/test_IAMAdminPolicyDocument.py
angatha: When patching the function, I set a default value. This (unintentionally) hides other places, where this would occur. But in this cases, the test calls the function on the check directly e.g. `tests.terraform.checks.resource.aws.test_IAMPolicyAttachedToGroupOrRoles`. There you can argument, that the function doesn't define it on the other hand, the new signature specifies it.

nimrodkor (approved):
🙏

checkov/terraform/checks/provider/base_check.py
angatha: @nimrodkor this was not correct. In the case of provider, this is the name of the provider. `tests.terraform.runner.TestRunnerValid.test_runner_two_checks_only` will fail, if I would have added an assert statement.

PR #489, opened 2020-08-08
Fix NSGRulePortAccessRestricted crashing on a terraform dynamic block

### LINKED ISSUE #488: Checkov crashes when evaluating a Terraform dynamic block in NSGRulePortAccessRestricted.py
**Describe the bug**
When checking azure_security_group_rule, azurerm_network_security_rule or azurerm_network_security_group Terraform resource types, NSGRulePortAccessRestricted.py throws a "TypeError: string indices must be integers" error whenever there's a dynamic block.

**To Reproduce**
1. Create a resource in terraform, containing a dynamic security rule -
```
resource "azurerm_network_security_group" "snet_nsgs" {
  count               = "${length(local.subnets)}"
  name                = "${local.root}-snet-${lookup(local.subnets[count.index], "name")}-nsg"
  location            = "${azurerm_resource_group.net_rg.location}"
  resource_group_name = "${azurerm_resource_group.net_rg.name}"
  tags                = "${local.tags}"


  dynamic "security_rule" {
    for_each = [for s in local.subnets[count.index].nsg_rules : {
      name                       = s.name
      priority                   = s.priority
      direction                  = s.direction
      access                     = s.access
      protocol                   = s.protocol
      source_port_range          = s.source_port_range
      destination_port_range     = s.destination_port_range
      source_address_prefix      = s.source_address_prefix
      destination_address_prefix = s.destination_address_prefix
      description                = s.description
    }]
    content {
      name                       = security_rule.value.name
      priority                   = security_rule.value.priority
      direction                  = security_rule.value.direction
      access                     = security_rule.value.access
      protocol                   = security_rule.value.protocol
      source_port_range          = security_rule.value.source_port_range
      destination_port_range     = security_rule.value.destination_port_range
      source_address_prefix      = security_rule.value.source_address_prefix
      destination_address_prefix = security_rule.value.destination_address_prefix
      description                = security_rule.value.description
    }
  }
}
```
2. Run checkov
3. Error!

**Expected behavior**
As checkov cannot evaluate the dynamic block, I expect the check to be skipped without throwing an error.

 - OS: Ubuntu
 - Checkov Version 1.0.479

davenicoll:
Fixes #488

### DIFF
--- a/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
+++ b/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
@@ -27,7 +27,7 @@ class NSGRulePortAccessRestricted(BaseResourceCheck):
 
     def scan_resource_conf(self, conf):
         if any("dynamic" in s for s in conf):
-            return CheckResult.SKIPPED
+            return CheckResult.UNKNOWN
             
         rule_confs = [conf]
         if 'security_rule' in conf:
--- a/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
+++ b/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
@@ -26,6 +26,9 @@ class NSGRulePortAccessRestricted(BaseResourceCheck):
         return False
 
     def scan_resource_conf(self, conf):
+        if any("dynamic" in s for s in conf):
+            return CheckResult.PASSED
+            
         rule_confs = [conf]
         if 'security_rule' in conf:
             rule_confs = conf['security_rule']
--- a/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
+++ b/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
@@ -26,7 +26,7 @@ class NSGRulePortAccessRestricted(BaseResourceCheck):
         return False
 
     def scan_resource_conf(self, conf):
-        if any("dynamic" in s for s in conf):
+        if "dynamic" in conf:
             return CheckResult.UNKNOWN
             
         rule_confs = [conf]
--- a/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
+++ b/checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
@@ -27,7 +27,7 @@ class NSGRulePortAccessRestricted(BaseResourceCheck):
 
     def scan_resource_conf(self, conf):
         if any("dynamic" in s for s in conf):
-            return CheckResult.PASSED
+            return CheckResult.SKIPPED
             
         rule_confs = [conf]
         if 'security_rule' in conf:


checkov/terraform/checks/resource/azure/NSGRulePortAccessRestricted.py
```diff
@@ -26,6 +26,9 @@ def is_port_in_range(self, conf):
         return False
 
     def scan_resource_conf(self, conf):
+        if any("dynamic" in s for s in conf):
```
angatha, line 29: This test is really wide. Why not just `if "dynamic" in conf`?

Currently dynamic can only be used for `security_rule` so this will not skip other cases. What do you think about an additional check that dynamic is used for `security_rule`?

nimrodkor: Hey @davenicoll !

Thanks for the contribution. However, this change will mean if the resource has **any** dynamic blocks in it, it will return as `UNKNOWN`. IMO it would be better to check the `rule_conf` - verify it is a dict before checking fields inside `rule_conf`. WDYT?

davenicoll:
@angatha @nimrodkor both sensible suggestions, thanks - I'll work on this later today.

davenicoll:
I've reduced the scope as requested.

I also started to parse the dynamic block, before realising it's pointless -

```
            if 'access' in rule_conf and rule_conf['access'][0] == "Allow":
                if 'direction' in rule_conf and rule_conf['direction'][0] == "Inbound":
                    if 'protocol' in rule_conf and rule_conf['protocol'][0] == 'TCP':
                        if 'destination_port_range' in rule_conf and self.is_port_in_range(rule_conf):
                            if 'source_address_prefix' in rule_conf and rule_conf['source_address_prefix'][0] in INTERNET_ADDRESSES:
```

If a variable has been used in `access`, `direction`, `protocol` or `port`, the test will fail. As checkov is unable to parse the `for` loop within the dynamic, I think the best option is to return UNKNOWN if a dynamic is found. What do you think?

angatha:
> I've reduced the scope as requested.
> 
> I also started to parse the dynamic block, before realising it's pointless -
> 
> ```
>             if 'access' in rule_conf and rule_conf['access'][0] == "Allow":
>                 if 'direction' in rule_conf and rule_conf['direction'][0] == "Inbound":
>                     if 'protocol' in rule_conf and rule_conf['protocol'][0] == 'TCP':
>                         if 'destination_port_range' in rule_conf and self.is_port_in_range(rule_conf):
>                             if 'source_address_prefix' in rule_conf and rule_conf['source_address_prefix'][0] in INTERNET_ADDRESSES:
> ```
> 
> If a variable has been used in `access`, `direction`, `protocol` or `port`, the test will fail. As checkov is unable to parse the `for` loop within the dynamic, I think the best option is to return UNKNOWN if a dynamic is found.
> 
> Checking within dynamic blocks, without being able to determine the values of attributes correctly seems like it's giving a false sense of security.
> 
> What do you think?

Yes, I think returning UNKNOWN is the better choice.

davenicoll:
### ###

schosterbarak (approved):

schosterbarak:
Thanks!

PR #493, opened 2020-08-13
Add any principal check

mikeurbanski1:

Adds checks for S3 policies and IAM assume role policies for statements allowing any principal.

### DIFF
--- a/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
+++ b/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
@@ -13,27 +13,28 @@ class S3AllowsAnyPrincipal(BaseResourceCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf):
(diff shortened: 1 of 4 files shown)


checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
```diff
@@ -0,0 +1,40 @@
...
+        id = "CKV_AWS_70"
+        supported_resources = ['aws_s3_bucket', 'aws_s3_bucket_policy']
+        categories = [CheckCategories.IAM]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if isinstance(conf['policy'][0], str):
+            try:
+                policy_block = json.loads(conf['policy'][0])
+                if 'Statement' in policy_block.keys():
```
angatha, line 19: `if 'Statement' in policy_block:` is enough. But this is still correct.

I just ask to gather some opinions about this case. Calling keys is not necessary but in some way more expressive. On the other hand asking if something is in a dict is also really clear.

PR #495, opened 2020-08-14
sort checks in the same framework by number

mikeurbanski1:

Adds sorting when listing checks, so, e.g., `CKV_ABC_2` appears before `CKV_ABC_10`

### DIFF
--- a/checkov/common/util/docs_generator.py
+++ b/checkov/common/util/docs_generator.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from functools import cmp_to_key
 
 from tabulate import tabulate
 
@@ -9,6 +10,25 @@ from checkov.cloudformation.checks.resource.registry import cfn_registry as cfn_
 from checkov.kubernetes.registry import registry as k8_registry
 from checkov.serverless.registry import sls_registry
 from checkov.arm.registry import arm_registry
+
+
+def check_compare(c1, c2):
+    c1: str = c1[0]
+    c2: str = c2[0]
+
+    c1_index = c1.rindex('_')
+    c2_index = c2.rindex('_')
+
+    c1_prefix = c1[0:c1_index]
+    c1_number = int(c1[c1_index + 1:])
+
+    c2_prefix = c2[0:c2_index]
+    c2_number = int(c2[c2_index + 1:])
+
+    if c1_prefix == c2_prefix:
+        return c1_number - c2_number
+    else:
+        return -1 if c1 < c2 else 1
 
 
 def print_checks(framework="all"):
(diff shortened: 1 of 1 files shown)


checkov/common/util/docs_generator.py
```diff
@@ -11,6 +12,25 @@
...
 
+def check_compare(c1, c2):
+    c1: str = c1[0]
+    c2: str = c2[0]
+
+    c1_index = c1.rindex('_')
+    c2_index = c2.rindex('_')
+
+    c1_prefix = c1[0:c1_index]
+    c1_number = int(c1[c1_index + 1:])
```
angatha, line 23: There is no general naming convention so there does not have to be a number at the end.

Maybe you can use the iac (last column) and the second column to get the desired result.
Sorting by iac and then by name will group them by iac. The second column can be used to sort the different parts of terraform.

mikeurbanski1: Hm, well, maybe we should be enforcing a check numbering convention then.

The problem I'm really trying to solve is that it's a little annoying to find how many checks there are for a given framework, because the sorting is totally arbitrary today (it just sorts alphabetically, so it's not by framework/type OR by number - "10" comes before "2"). Since I'm adding checks regularly, I really just want to be able to run `checkov --list` and see what the next check number should be. Sorting by framework and by resource type doesn't solve that.

I'd be open to a recommendation to actually implement real sorting like you describe. It doesn't solve my problem, but it's a huge improvement over how the output looks today. And it's probably the most useful for a normal end user.

angatha: Yes, this will not solve your problem. Maybe adding a separate script for exactly that task is a viable solution.

On the other hand imho the --list option needs some polishing. It is not really useful to generate costume documentation, because it ignores the other arguments like excluded checks or costume checks.

Having that said, actually your code will work, because all tests follow the naming convention of checkov.

Maybe we can do the following:

1. slitting the id into text and numbers so we have a list `"CKV_AWS_01"` is parsed as `["CKV_AWS_", 1]` and `"MY_01_CHECK_1"` as `["MY_", 1, "_CHECK_", 1]`
2. comparing element by element using sorting for strings or numbers

Open points: How to handle leading zeros?

angatha:
I implemented a more generic version in my version of the checkov repository: https://github.com/bridgecrewio/checkov/compare/fix-list-output-order...angatha:fix-list-output-order

checkov/common/util/docs_generator.py
mikeurbanski1: I like this. It makes sense to check every "element" of the check name and apply the same logic.

As far as leading zeroes, I think I am comfortable ignoring them. If somebody really wants to create checks named  `_1` and `_01` and `_001` and then complains that they get sorted arbitrarily, we can fix it then.

angatha: @mikeurbanski1 i created a pull (#500) request for my implementation into this branch.

angatha (approved):

PR #500, opened 2020-08-17
Create a comparison key based on the check id

angatha:
The id is split into text and numbers. They are compared from left to
right. It also handles leading zeros, which are placed after te number
with one leading zero fewer. (1 -> 01 -> 001 ...)


robeden:
Good tests. 👍

angatha:
Thank you very much.

schosterbarak:
Thanks!

PR #507, opened 2020-08-19
Revert "add generic no-tags check for any resource type"

schosterbarak:
Reverts bridgecrewio/checkov#505


sherryxcheng:
May I ask why the tag check is removed? Are tags checked by checkov?

schosterbarak:
@sherryxcheng it just created too much failures for too many people. maybe this is better added as a custom check

PR #509, opened 2020-08-20
fix skipped repo-id format check

mikeurbanski1:
Fixes the validation that makes sure the repo-id has a slash. It got skipped before, causing the creation of invalid "integrations" in BC.


schosterbarak (approved):

schosterbarak:
good catch

angatha (approved):

PR #522, opened 2020-08-28
have pre-commit install/run checkov

oasys:
Pre-commit will install checkkov in its own environment, without
   first requiring the user to have already installed checkov globally.


oasys:
Seeking feedback.  This is valuable to me, as I can encourage checkov use by my team without requiring them to install yet another tool.

If you think it is important to keep the `language: script` hook in place, I can modify the PR to add a new hook (named something like `checkov-python`?) to give users the option.

holyender:
hey does this mean that we can have pre-commit hooks that accept args? 
im new to pre-commit so please correct me. it looks like currently if you use checkov pre-commit hooks like
```
-   repo: https://github.com/bridgecrewio/checkov.git
    rev: 1.0.513
    hooks:
    -   id: checkov
        files: \.tf$
```
It doesnt seem like it will accept any args like most pre-commit hooks. so if you added args: ['--soft-fail'] for example it would just return
```dirname: illegal option -- -
usage: dirname path
``` 
when trying to commit.  i would like to have it so that all the optional arguments can be part of the pre-commit hook. does your pr solve this of if you know how i can solve this I can work on that.
thanks!

oasys:
Yes @holyender, it will allow arguments.  I just tried with a `.pre-commit-config.yaml` like

```yaml
repos:
- repo: https://github.com/oasys/checkov
  rev: 999ab97
  hooks:
  - id: checkov
    verbose: true
    args: [ '--soft-fail' ]
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v3.2.0
  hooks:
  - id: trailing-whitespace
```

on a repo with failing checkov tests.   Checkov output was sent to STDOUT, and `pre-commit` proceeded to the next test.  Presumably this addresses your use-case of having an "informational" feedback from checkov without failing a commit hook.

holyender:
yes this solves my issue completely. thank you. hope this gets merged soon. im also in the process of seeing if i should recommend my team this tool

schosterbarak:
Very nice. Do you think we should change something in the docs: https://github.com/bridgecrewio/checkov/blob/master/docs/4.Integrations/pre-commit.md ?

oasys:
I took a look at the docs when I originally authored the PR, and didn't see any changes that need to be made.  As they are written, they still apply with this change.  If there's something specific you'd like changed/added, I'd be happy to do it.

schosterbarak:
Let's add just the topic in this discussion "how to add custom params. Might be beneficial to others too.

oasys:
Sounds good.  I rebased from current master and updated docs with a couple of example use cases.

Issue #506, opened 2020-08-19
Serverless: files specified with `-f` may not be processed


robeden:
**Describe the bug**
Specifying a target serverless file which isn't in the immediate working directory will prevent it from being processed. For example:
```
--framework serverless -f /Users/robeden/git/third-party/checkov/tests/serverless/checks/aws/example_AdminPolicyDocument/AdminPolicyDocument-FAILED-func_level/serverless.yml
```
... doesn't work. This is due to an insufficient check [here](https://github.com/bridgecrewio/checkov/blob/dcaa5980104262cd9326f3bb9e0c0bc79b53f358/checkov/serverless/runner.py#L33), which doesn't use the file base name.

**To Reproduce**
Run checkov targeting the test file as shown above (note: your absolute path will be different)

**Expected behavior**
There should check results, but none will be reported (since no files were processed)

robeden:
I have a fix for this in some serverless changes I'm currently working on.

Issue #501, opened 2020-08-18
Serverless variable resolution


robeden:
The serverless parser doesn’t currently handle [variable resolution](https://www.serverless.com/framework/docs/providers/aws/guide/variables/). Support for some of the basics (self references & external files, probably) should be supported.

robeden:
Nearly complete support for "self" and "file" references here:
https://github.com/robeden/checkov/compare/master...robeden:serverless-parsing

It currently doesn't support [custom variable syntax](https://www.serverless.com/framework/docs/providers/aws/guide/variables#using-custom-variable-syntax) or ["overwriting variables"](https://www.serverless.com/framework/docs/providers/aws/guide/variables#overwriting-variables).

robeden:
Basic support is now in.

PR #529, opened 2020-08-31
Serverless variable parsing, -f option fix

robeden:
Variable parsing handles "self", "file" and "env" sources. Handles deep references, overrides, quoting and custom variable patterns.

Note that there are lots of tests. To see what is supported, that's probably the best place to start. [Serverless variable guide](https://www.serverless.com/framework/docs/providers/aws/guide/variables/) for reference.

Fixes: #506, #501


checkov/serverless/runner.py
```diff
@@ -30,18 +30,23 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
                     parse_result = parse(file)
                     if parse_result:
                         (definitions[file], definitions_raw[file]) = parse_result
 
         if root_folder:
             for root, d_names, f_names in os.walk(root_folder):
+                # Don't walk in to "node_modules" directories under the root folder. If –for some reason–
+                # scanning one of these is desired, it can be directly specified.
+                if "node_modules" in d_names:
+                    d_names.remove("node_modules")
```
schosterbarak, line 43: would it work for internal dirs too?

schosterbarak (approved):
You rock! just one small question (i haven't tested it).

checkov/serverless/runner.py
robeden: Do you mean nested? For example: “<root_folder>/node_modules” as well as “<root_folder>/something/node_modules”? If so, yes. It’s called at each directory with the sub-dir names that will traversed if left alone.

PR #538, opened 2020-09-04
refactor secrets regexes to be more expandable

mikeurbanski1:
This PR reworks the secrets regexes (currently only used for EC2 / Lambda secrets) to be more extensible for use in other checks (e.g., Azure or GCP).

It adds an extended list of regexes taken from git-secrets, and provides a mechanism for this list to be expanded without any changes to checks themselves.

It also adds some categories to the regexes, so that if desired, a particular check can only use a certain list (e.g., AWS only).

The existing EC2 and Lambda checks will continue to only check AWS values, so this should not change any existing check results.


schosterbarak (approved):
Sweet!

Issue #152, opened 2020-03-25
Add new check: Enforce IMDSv2 in EC2 instance and Launch Configuration


jonjozwiak:
Blog Post Here: https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/

CloudFormation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html
	- Doesn’t appear possible for individual EC2 instance… 

Terraform: It does not appear configurable, but suggest further research.  
https://www.terraform.io/docs/providers/aws/r/launch_template.html
https://www.terraform.io/docs/providers/aws/r/instance.html

KevinHock:
This would be a great feature to have!

For Terraform, I think we would do
```python
if (
	metadata_options["http_tokens"] == "required"
	or
	metadata_options["http_endpoint"] == "disabled"
):
    # return and don't alert...
# alert 
```
from https://www.terraform.io/docs/providers/aws/r/instance.html#metadata-options

KevinHock:
This is kinda done as far as Terraform goes, re: my commit above ref'ing this issue

What should we do for cloudformation?

rollwagen:
FYI - as far Cloudformation goes, it seems this is still not possible on an EC2 instance level; still marked as 'open' on the roadmap. For details see <https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/655>

nimrodkor:
I believe this has been implemented for [cloudformation](https://github.com/bridgecrewio/checkov/blob/master/checkov/cloudformation/checks/resource/aws/IMDSv1Disabled.py) and [terraform](https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/IMDSv1Disabled.py). If someone disagrees - please reopen!

rollwagen:
@nimrodkor - thanks for the update and looking into this. 
[...]

PR #540, opened 2020-09-05
:tada: Add IMDSv1 Disabled check

KevinHock:

Solves #152, for Terraform only.

### DIFF
--- a/checkov/terraform/checks/resource/aws/IMDSv1Disabled.py
+++ b/checkov/terraform/checks/resource/aws/IMDSv1Disabled.py
@@ -1,0 +1,51 @@
+from checkov.common.models.enums import (
+    CheckCategories,
+    CheckResult,
+)
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class IMDSv1Disabled(BaseResourceCheck):
+
+    def __init__(self):
+        name = "Ensure Instance Metadata Service Version 1 is not enabled"
+        id = "CKV_AWS_79"
+        categories = [CheckCategories.GENERAL_SECURITY]
+        supported_resources = ['aws_instance', 'aws_launch_template']
+        super().__init__(
+            name=name,
+            id=id,
+            categories=categories,
+            supported_resources=supported_resources,
+        )
+
+    def scan_resource_conf(self, conf):
+        """
+        Looks for if the metadata service is disabled or requires session tokens:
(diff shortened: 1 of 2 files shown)


tests/terraform/checks/resource/aws/test_IMDSv1Disabled.py
```diff
@@ -0,0 +1,63 @@
+import unittest
```
KevinHock, line 1: We may want to migrate to `pytest`, more modern.

checkov/terraform/checks/resource/aws/IMDSv1Disabled.py
```diff
@@ -0,0 +1,51 @@
...
+        metadata_options = conf['metadata_options'][0]
+        return (
+            CheckResult.PASSED if
+            (
+                metadata_options.get("http_tokens") == ["required"]
+                or
+                metadata_options.get("http_endpoint") == ["disabled"]
+            )
+            else CheckResult.FAILED
+        )
```
angatha, line 48: Thank you for your contribution.

In my opinion a good old if statement would be more readable.

```python
        if (
                metadata_options.get("http_tokens") == ["required"]
                or metadata_options.get("http_endpoint") == ["disabled"]
        ):
            return CheckResult.PASSED
        else:
            return CheckResult.FAILED
```

angatha (approved):

tests/terraform/checks/resource/aws/test_IMDSv1Disabled.py
schosterbarak: The build is using pytest runner, so it might work ootb

checkov/terraform/checks/resource/aws/IMDSv1Disabled.py
schosterbarak: I agree, but i think it's fine this way too

schosterbarak:
Thanks @kevinhock

schosterbarak (approved):

checkov/terraform/checks/resource/aws/IMDSv1Disabled.py
KevinHock: I think I agree, sometimes I like to try something different, and see how I feel about it over time. e.g. I tried to center a readme once https://github.com/mandatoryprogrammer/TrustTrees#--trusttrees, but I'm still on the fence about it 🤔

PR #542, opened 2020-09-05
Make supported_resources deprecated

angatha:

There is only one way to use `supported_resources` and that is setting it to `['module']`. To reduce confusion when using the base class, it is now deprecated and set automatically.


nimrodkor (approved):

schosterbarak:
Thank you @angatha

PR #545, opened 2020-09-05
Add MSK encryption check

### LINKED ISSUE #137: Add new check: MSK - Check encryption at rest and in transit
Kafka traffic should be encrypted at rest and also in transit between clients as well as nodes in the cluster

* At rest should specify a KMS key
* In transit should be set to TLS with in cluster set to true

CloudFormation and Terraform documentation as follows:  
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-msk-cluster.html
https://www.terraform.io/docs/providers/aws/r/msk_cluster.html

DrFaust92:
Closes #137


schosterbarak:
Thank you @DrFaust92

PR #546, opened 2020-09-05
Add Athena Workgroup config check

### LINKED ISSUE #150: Add new check: Athena Workgroup should enforce configuration to prevent client disabling encryption
Verify enforce Work Group Configuration is true has ‘EnforceWorkGroupConfiguration’ True

EnforceWorkGroupConfiguration: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-athena-workgroup.html
enforce_workgroup_configuration: https://www.terraform.io/docs/providers/aws/r/athena_workgroup.html

DrFaust92:
Closes #150

### DIFF
--- a/checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
+++ b/checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
@@ -1,4 +1,4 @@
-from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.common.models.enums import CheckCategories
 from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
--- a/checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
+++ b/checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
@@ -1,8 +1,9 @@
 from checkov.common.models.enums import CheckResult, CheckCategories
-from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
-class AthenaWorkgroupConfiguration(BaseResourceCheck):
+class AthenaWorkgroupConfiguration(BaseResourceValueCheck):
+
     def __init__(self):
         name = "Ensure Athena Workgroup should enforce configuration to prevent client disabling encryption"
         id = "CKV_AWS_82"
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
```diff
@@ -0,0 +1,23 @@
...
+        name = "Ensure Athena Workgroup should enforce configuration to prevent client disabling encryption"
+        id = "CKV_AWS_82"
+        supported_resources = ['aws_athena_workgroup']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if 'configuration' in conf.keys():
+            if 'enforce_workgroup_configuration' in conf['configuration'][0]:
+                if conf['configuration'][0]['enforce_workgroup_configuration'] is True:
```
nimrodkor, line 16: @DrFaust92 Can we inherit from `BaseResourceValueCheck` for this check to save the duplication of `scan_resource_conf`?

DrFaust92:
Thanks @nimrodkor fixed and tested

DrFaust92:
@nimrodkor, Having issues with this as the test that doesn't have the block fails (default behaviour is no block and enabled).
Should i revert? thats why i originally did this the other way.

nimrodkor:
@DrFaust92 Great point. Just added support for overriding that case - #551 . Merge from master and it should solve your issue 😄

DrFaust92:
Excellent, this will simplify other checks as well!

checkov/terraform/checks/resource/aws/AthenaWorkgroupConfiguration.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class AthenaWorkgroupConfiguration(BaseResourceValueCheck):
+
+    def __init__(self):
+        name = "Ensure Athena Workgroup should enforce configuration to prevent client disabling encryption"
+        id = "CKV_AWS_82"
+        supported_resources = ['aws_athena_workgroup']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
```
nimrodkor, line 12: @DrFaust92 Missing `missing_block_result=CheckResult.PASSED`

DrFaust92:
@nimrodkor, seems that the change you made was for the `arm` variant of the `BaseResourceValueCheck`. can you do the same for terraform?

nimrodkor (approved):

DrFaust92:
Now the explicit failure test is failing that the config is valid  (when it isn't). i'll take a closer look.

DrFaust92:
It seems to be treating `configuration/[0]/enforce_workgroup_configuration` as missing because of the `False` value and returning the default passing result.

nimrodkor:
@DrFaust92 That was a bug in my feature, resolved in #554

DrFaust92:
Thanks! Ready!

PR #547, opened 2020-09-05
enable upstream runner

schosterbarak:

### DIFF
--- a/checkov/common/runners/runner_registry.py
+++ b/checkov/common/runners/runner_registry.py
@@ -2,6 +2,8 @@ import json
 import logging
 from abc import abstractmethod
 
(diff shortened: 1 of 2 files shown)


checkov/common/runners/runner_registry.py
```diff
@@ -29,11 +31,7 @@ def run(self, root_folder=None, external_checks_dir=None, files=None, guidelines
         return self.scan_reports
 
     def print_reports(self, scan_reports, args):
-        if args.output not in (
-            "github_failed_only",
-            "json",
-            "junitxml",
-        ):
+        if args.output not in OUTPUT_CHOICES:
```
angatha, line 34: @schosterbarak why is the banner no longer displayed for `'cli'`?

And when looking at the arg parser, there is no way, this could ever by `True`.

PR #557, opened 2020-09-07
11 new GCP Cloud SQL checks

GiladMark:

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GoogleCloudMySqlLocalInfileOff.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleCloudMySqlLocalInfileOff.py
@@ -23,16 +23,11 @@ class GoogleCloudMySqlLocalInfileOff(BaseResourceCheck):
             if 'MYSQL' in key:
                 if 'settings' in conf.keys():
                     for attribute in conf['settings'][0]:
-                        if attribute=='database_flags':
+                        if attribute == 'database_flags':
                             for flag in conf['settings'][0]['database_flags']:
-                                if (flag['name'][0]=='local_infile') and (flag['value'][0]=='on'):
+                                if (flag['name'][0] == 'local_infile') and (flag['value'][0] == 'on'):
                                     return CheckResult.FAILED
         return CheckResult.PASSED
-''' 
-    def get_inspected_key(self):
-        return 'settings/[0]/database_flags/[0]/local_infile'
 
-    def get_expected_value(self):
-        return "off"
-'''
+
 check = GoogleCloudMySqlLocalInfileOff()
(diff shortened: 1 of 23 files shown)


checkov/terraform/checks/resource/gcp/GoogleCloudMySqlLocalInfileOff.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+            Looks for google_sql_database_instance which enables local_infile:
+            :param
+            conf: google_sql_database_instance
+            configuration
+            :return: < CheckResult >
+        """
+        if 'database_version' in conf.keys():
+            key = conf['database_version'][0]
+            if key=='MYSQL_5_6' or key=='MYSQL_5_7' or key=='MYSQL_8_0':
```
arielkru, line 23: In all of the examples like the three lines above, it'd be better to create a few lists of approved / restricted values for the different sql dbs, and then just compare on value to the imported list. It would be easier to read, and whenever the list should be updated we can update it only in one place.

e.g:
mysql_version = ['MYSQL_5_6', 'MYSQL_5_7', 'MYSQL_8_0']
**
import mysql_version
**
if 'database_version' in conf.keys() and conf['database_version'][0] in mysql_versions:

checkov/terraform/checks/resource/gcp/GoogleCloudPostgreSqlLogMinMessage.py
```diff
@@ -0,0 +1,35 @@
...
+                    key = conf['database_version'][0]
+                    if key == 'POSTGRES_9_6' or key == 'POSTGRES_10' or key == 'POSTGRES_11' or key == 'POSTGRES_12':
+                        if 'settings' in conf.keys():
+                            for attribute in conf['settings'][0]:
+                                if attribute == 'database_flags':
+                                    for flag in conf['settings'][0]['database_flags']:
+                                        if (flag['name'][0] == 'log_min_messages'):
+                                            key_logmin=flag['value'][0]
+                                            if (key_logmin != 'fatal' and key_logmin != 'panic' and key_logmin != 'log' and key_logmin != 'error' and key_logmin != 'warning' and key_logmin != 'notice'
+                                                    and key_logmin != 'info' and key_logmin != 'debug1' and key_logmin != 'debug2' and key_logmin != 'debug3' and key_logmin != 'debug4' and key_logmin != 'debug5'):
```
arielkru, line 30: same here, you can create a list of excluded values and then check if key_logmin is not in list, easier to read and edit

tests/terraform/checks/resource/gcp/test_GoogleCloudMySqlLocalInfileOff.py
```diff
@@ -0,0 +1,173 @@
...
+
+from checkov.terraform.checks.resource.gcp.GoogleCloudMySqlLocalInfileOff import check
+from checkov.common.models.enums import CheckResult
+
+
+class TestCloudMySqlLocalInfileOff(unittest.TestCase):
+
+    def test_failure(self):
+        hcl_res = hcl2.loads("""
+                resource "google_sql_database_instance" "tfer--gilad-002D-mysql81" {
```
arielkru, line 13: better remove all of the 'gilad' instances and replace it with something more generic, since it's a public code

checkov/terraform/checks/resource/gcp/GoogleCloudMySqlLocalInfileOff.py
arielkru: Better, why not just check if starts with MYSQL, without the version? the list is updating, so we don't even need to track it

nimrodkor (approved):

Issue #478, opened 2020-08-03
Allow serverless checks against any data (particularly provider, layers, resources blocks)


robeden:
**Is your feature request related to a problem? Please describe.**
There is currently no want to write a custom serverless check for things like required tags. The reason is the "entity" is currently just the function data. So, here are some things that can't be inspected:

- Tags in the `provider` block (e.g., check for tags required by policy)
- `layers` (e.g., check for layers required by policy)
- `resources` (probably needs to run through CloudFormation scans eventually... but can keep it simple for now)

**Describe the solution you'd like**
I think the simplest thing for the moment would be to include all the data from `serverless.yml` parsing in the "entity" data set. This allows maximum flexibility... at least for the moment.

**Describe alternatives you've considered**
Similar to the terraform framework, there probably could be separate `Base*Check` classes to ease pulling out the right data. To be honest, I'm not sure this is necessary though since the format of Serverless is so much simpler than Terraform.
I think the simplest solution is to call the whole data map from the `serverless.yml` file parse the entity. Convenience functions can probably more easily fill the need in this case. ... at least until we do some experimentation and come up with something better.

robeden:
🤔 Although now that I think about it... the "alternative" approach of having similar `Base*Check` classes that reduce to the desired data set would be easy enough and would probably be nice for consistency. Maybe that is the way to go.

PR #567, opened 2020-09-11
Additional check types for Serverless

robeden:
This adds a variety of check types for Serverless corresponding to additional block types:

* custom (BaseCustomCheck)
* layer (BaseLayerCheck)
* package (BasePackageCheck)
* plugin (BasePluginCheck)
* provider (BaseProviderCheck)

Each of these is passed the appropriate data from the corresponding block, if present. In addition a "complete" (BaseCompleteCheck) type has been added which is provided the entire Serverless configuration to allow for additional blocks which may not be caught elsewhere or to assess the overall document.

To facilitate the additional classes, the structure has been changed to mirror the Terraform layout with a check type under the serverless package: `checkov.serverless.checks.<type>...`. The "function" type has been moved into this structure and I *think* I've maintained backwards compatibility by keeping `base_function_check.py` in the `checkov.serverless` package with an import to the new script in `...<function>`. However, this was not easily tested, so I can't completely vouch.

Finally, "enrichment" was extended to copy `stackTags` from the provider block to `tags` in function blocks.

For examples of all check types, see [`tests/serverless/checks/test_check_types.py`](tests/serverless/checks/test_check_types.py)

One additional note: line numbers and content was not straight-forward in two scenarios:

1. For "plugins", line numbers and content could not be pulled out because apparently that doesn't work with lists.
2. For "complete", line numbers are okay (whole file range), but reporting the whole content of the file doesn't seem right.

To deal with these scenarios, the line numbers of the whole file are reported, but no file content is given to the report. This is sub-optimal but for case # 1 the issue is related to the `CfnContextParser` and outside the scope of this change and for # 2, there really isn't a good answer unless the API changes to allow checks to provide back more detail about where the exact problem is located.

Fixes: #478


schosterbarak (approved):
Very nice. This will probably enable easier support of multi-cloud SLS. 
@tronxd @nimrodkor @metahertz can you take a look too?

@robeden do you know of OOTB best practices (policies) should be added to checkov after this cool update?

robeden:
@schosterbarak - We’re evaluating that now. I think most are going to be handled by CloudFormation checks of resources. Most of the raw-Serverless config  seems to be reasonable defaults. Checks we’re writing are more company-specific policy. But if I find other cases, I’ll see about including them.

nimrodkor (commented):
@robeden Super cool!
I have a question regarding provider confs vs function confs - a lot of the times, the function confs override the ones set in the provider level. In case I'm building a check for a function conf, I'd expect it to check the `default` value from the provider as well as the conf on the function. Is this implemented? I did not see it...

robeden:
@nimrodkor - I'm not entirely sure if it's completely doing all that it needs there. "Enrichment" of functions happens here:
https://github.com/robeden/checkov/blob/sls-check-types-478/checkov/serverless/runner.py#L131
Implementation here:
https://github.com/robeden/checkov/blob/sls-check-types-478/checkov/serverless/parsers/context_parser.py#L39

Most of that pre-existed my changes and copies down the environment and IAM blocks. I added copying of `stackTags` to `tags`.

My guess would be that you're correct and there's more stuff that should be included there, but I'm not sure offhand what it would be. Do you know of any specific items that are missed or is there something else that needs to be handled differently?

Note that there's a test case for that functionality here:
https://github.com/robeden/checkov/blob/sls-check-types-478/tests/serverless/checks/test_check_types.py#L118

schosterbarak:
Thanks @robeden , @nimrodkor and I just synced. Great contribution.

PR #598, opened 2020-10-07
fix: Minor typo in GCP test

adamjdeacon:


nimrodkor (approved):
@adamjdeacon Good catch!
Our docs are auto-generated so the change to `resource-scans.md` will be overritten. It doesn't matter, just in the future - no need to fix the generated docs 👍

PR #604, opened 2020-10-09
Decrease docker image size and improve cache

lcostea:

Followed some best practices for dockerfiles and improved the size (around 17MB on my machine) and it also should be better in terms of layers and caching them.


metahertz:
Thanks for the contribution @lcostea :)

PR #627, opened 2020-10-22
Prevent S3 lockouts, tests for S3:* DENY resource policies.

metahertz:
Suggestion based on this thread: https://twitter.com/0xdabbad00/status/1319065946770661376

Check that a resource policy is not causing a "Deny" to all principals, locking all non-root users out of the bucket.
Requirements needing the root user credentials to be available to undo a fix should be limited/removed at buildtime where possible. 

Requesting review from @nimrodkor and validation of usecase & test examples by https://twitter.com/0xdabbad00 

### DIFF
--- a/checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
+++ b/checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
@@ -27,13 +27,9 @@ class S3ProtectAgainstPolicyLockout(BaseResourceCheck):
                             if 'NotAction' in statement.keys():
                                 continue
 
-                            if 'Effect' not in statement.keys():
-                                # Defaults to allow if doesn't exist.
+                            if statement.get('Effect') != 'Deny':
                                 continue
-
-                            if statement['Effect'] == 'Allow':
-                                continue
-
+                                
                             principal = statement['Principal']
 
                             if principal == '*':
--- a/checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
+++ b/checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
@@ -1,5 +1,6 @@
 from checkov.common.models.enums import CheckResult, CheckCategories
 from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.common.util.type_forcers import force_list
 import json
 
 
@@ -7,7 +8,7 @@ class S3ProtectAgainstPolicyLockout(BaseResourceCheck):
 
     def __init__(self):
         name = "Ensure S3 bucket policy does not lockout all but root user. (Prevent lockouts needing root account fixes)"
-        id = "CKV_AWS_89"
+        id = "CKV_AWS_93"
         supported_resources = ['aws_s3_bucket', 'aws_s3_bucket_policy']
         categories = [CheckCategories.IAM]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
@@ -18,7 +19,18 @@ class S3ProtectAgainstPolicyLockout(BaseResourceCheck):
                 try:
                     policy_block = json.loads(conf['policy'][0])
                     if 'Statement' in policy_block.keys():
-                        for statement in policy_block['Statement']:
+                        for statement in force_list(policy_block['Statement']):
+                            if 'Condition' in statement.keys():
+                                # https://github.com/bridgecrewio/checkov/pull/627#issuecomment-714681751
+                                continue
+
+                            if 'NotAction' in statement.keys():
+                                continue
+
+                            if 'Effect' not in statement.keys():
+                                # Defaults to allow if doesn't exist.
+                                continue
+
                             if statement['Effect'] == 'Allow':
                                 continue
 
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,48 @@
...
+        supported_resources = ['aws_s3_bucket', 'aws_s3_bucket_policy']
+        categories = [CheckCategories.IAM]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if 'policy' in conf.keys():
+            if isinstance(conf['policy'][0], str):
+                try:
+                    policy_block = json.loads(conf['policy'][0])
+                    if 'Statement' in policy_block.keys():
```
nimrodkor, line 20: @metahertz `Statement` might not always be a list, need to use `force_list()` from our force utils

checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,48 @@
...
+                        for statement in policy_block['Statement']:
+                            if statement['Effect'] == 'Allow':
```
nimrodkor, line 22: @metahertz `Effect` might not exist - which means it is an `Allow` as well. Please check for explicit `Deny`

checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,48 @@
...
+
+                            if principal == '*':
+                                return CheckResult.FAILED
+                            elif 'AWS' in statement['Principal']:
+                                # Can be a string or an array of strings
+                                aws = statement['Principal']['AWS']
+                                if (type(aws) == str and aws == '*') or (type(aws) == list and '*' in aws):
+                                    return CheckResult.FAILED
+
+                            action = statement['Action']
```
nimrodkor, line 35: @metahertz Policy might have `NotAction` instead of `Action` (they are exclusive). I suggest we skip statements with `NotAction` blocks

checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,48 @@
...
+
+                            if action == '*':
```
nimrodkor, line 37: @metahertz statement['Action'] can also be a list or a string. Again, please use the `force_list()` method

0xdabbad00:
I'm not too familiar with the codebase of checkov, but the general idea looks good, but something should be added to deal with conditions.  Looking at https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html there are some sample policies that I think this will alert on, but should not be.  For example:

```
{
  "Version": "2012-10-17",
  "Id": "S3PolicyId1",
  "Statement": [
    {
      "Sid": "IPAllow",
      "Effect": "Deny",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": [
	 "arn:aws:s3:::awsexamplebucket1",
         "arn:aws:s3:::awsexamplebucket1/*"
      ],
      "Condition": {
	 "NotIpAddress": {"aws:SourceIp": "54.240.143.0/24"}
      }
    }
  ]
}
```

That policy is a reasonable policy to apply, but I believe your code would alert on it.  That policy could lock someone out if they don't set the IP correctly, but that is a harder problem to solve that is outside the scope of what I think would be reasonable for someone to expect checkov to detect.

Another example there is:
```
{
    "Version": "2012-10-17",
    "Id": "123",
    "Statement": [
      {
        "Sid": "",
        "Effect": "Deny",
        "Principal": "*",
        "Action": "s3:*",
        "Resource": "arn:aws:s3:::awsexamplebucket1/taxdocuments/*",
        "Condition": { "Null": { "aws:MultiFactorAuthAge": true }}
      }
    ]
 }
```


A simple fix for now may be to just check whether or not a `condition` exists in the statement, and if so, don't alert.

checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,48 @@
...
+                                return CheckResult.FAILED
+                            elif 's3' in statement['Action']:
+                                # Can be a string or an array of strings
+                                s3 = statement['Action']['s3']
+                                if (type(s3) == str and s3 == '*') or (type(s3) == list and '*' in s3):
+                                    return CheckResult.FAILED
+                except: # nosec
```
nimrodkor, line 44: @metahertz Missing a `Condition` check step. If there is a Condition, let it pass IMO

metahertz:
Thanks for the feedback @nimrodkor Added some more tests for those cases.
Also, we may want to look through some of the other checks not using force_list for for `statement in policy_block['Statement']`

checkov/terraform/checks/resource/aws/S3ProtectAgainstPolicyLockout.py
```diff
@@ -0,0 +1,60 @@
...
+                                continue
+
+                            if 'NotAction' in statement.keys():
+                                continue
+
+                            if 'Effect' not in statement.keys():
+                                # Defaults to allow if doesn't exist.
+                                continue
+
+                            if statement['Effect'] == 'Allow':
```
nimrodkor, line 34: @metahertz Can be merged into `if statement.get('Effect') == 'Deny':`

nimrodkor (approved):
GJ!
Would love it if you could open an issue for those other checks!

Issue #165, opened 2020-03-25
Add new check: Ensure DocumentDB Encryption in transit is NOT disabled


jonjozwiak:
Encryption in transit is enabled by default for new clusters unless you explicitly create a parameter group to disable it.  

https://www.terraform.io/docs/providers/aws/r/docdb_cluster_parameter_group.html#parameter
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbclusterparametergroup.html

njgibbon:
Ok cool, I will look at the CloudFormation implementation of this check as well :)

PR #633, opened 2020-10-23
check - test - Document DB Encryption in transit not disabled - Terraform

njgibbon:
Hello, first time adding a check to this project but i've tried to follow the helper documentation and i've found it very useful thus far!

This change is the terraform check implementation and unit tests for ensuring DocumentDB does not have in-flight encryption disabled.

Issue: https://github.com/bridgecrewio/checkov/issues/165

As mentioned in the issue the DocDB default is to have tls enabled, so in this check we just need to check that an explicit disable is not found for the configuration.

Provider doc for the configuration: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/docdb_cluster_parameter_group

The type was a bit confusing for me so I verified through the provider code to ensure it was a Type Set and I used the hcl2 library to ensure that the test data was input correctly as I saw from some of the other tests.
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_docdb_cluster_parameter_group.go
https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset

EDIT: type list and type set

Note that the 'parameter' option is optional but if it is used the 'name' and 'value' are then required.

Please let me know if anything needs adjusting :) Great project.

### DIFF
--- a/checkov/terraform/checks/resource/aws/DocDBTLS.py
+++ b/checkov/terraform/checks/resource/aws/DocDBTLS.py
@@ -1,0 +1,23 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
+
+
+class DocDBTLS(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure DocDB TLS is not disabled"
+        id = "CKV_AWS_90"
+        supported_resources = ['aws_docdb_cluster_parameter_group']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if 'parameter' in conf:
+            for elem in conf["parameter"]:
+                if elem["name"][0] == "tls" and elem["value"][0] == "disabled":
+                    return CheckResult.FAILED
+        else:
+            return CheckResult.PASSED
+        return CheckResult.PASSED
+
+
+check = DocDBTLS()
--- a/tests/terraform/checks/resource/aws/test_DocDBTLS.py
+++ b/tests/terraform/checks/resource/aws/test_DocDBTLS.py
@@ -1,0 +1,63 @@
+import unittest
+
+import hcl2
+
+from checkov.common.models.enums import CheckResult
+from checkov.terraform.checks.resource.aws.DocDBTLS import check
+
+
+class TestDocDBTLS(unittest.TestCase):
+
+    def test_failure(self):
+        hcl_res = hcl2.loads("""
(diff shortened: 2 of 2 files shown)


checkov/terraform/checks/resource/aws/DocDBTLS.py
```diff
@@ -0,0 +1,23 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
+
+
+class DocDBTLS(BaseResourceCheck):
```
schosterbarak, line 5: i think inheriting from `BaseResourceValueCheck`
like:
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/SQSQueueEncryption.py
might be a better fit

njgibbon: Hello,

I did want to use `BaseResourceValueCheck` and through this I found it may be best to use `BaseResourceNegativeValueCheck`. Because we want to fail this check if and only if the TLS config exists and is set to 'disabled'.

However, when you look at the specific data that gets used to implement the configuration it doesn't appear to match up.

In this case, the configuration is exposed through a list of KVPs which all look like this:
```
{
    name: "name-value",
    value: "value-value"
}
```
And so you can never find the tls configuration by name as it isn't exposed as a key anywhere. So the way to find if it exists is to iterate over all objects and check: 

```
if and only if:
    conf.parameter[X].name[0] is "tls" 
then check if:
    conf.parameter[X].value[0] is "disabled"
then:
    fail
# hcl2 input is parsing name to have single entry list though it is just a string value (TypeString in tf) though this doesn't effect the check
```

I don't think the Negative Check class provides the structure to check for this type of thing, and since this is a fairly uncommon way for terraform to expose configuration so I don't know if you would want to make an abstraction to handle in a slicker manner.
For example - DyanmoDB attributes are exposed in the same way (with slightly different key names): https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_dynamodb_table.go - but it is still mostly uncommon I think.

So, I think my implementation (or something pretty close) is actually the best for this check at this time given all the above.

Happy to try to refactor if i've missed something obvious!

nimrodkor: I agree, this check requires 2 attributes:
`if elem["name"][0] == "tls" and elem["value"][0] == "disabled":`

nimrodkor (approved):

checkov/terraform/checks/resource/aws/DocDBTLS.py
schosterbarak: thanks @nimrodkor for the help!

Issue #130, opened 2020-03-16
Add new check: DocDB snapshot encryption


schosterbarak:
Encryption of data at rest.
See: https://www.terraform.io/docs/providers/aws/r/docdb_cluster_snapshot.html

DrFaust92:
DocDb Snapshot does have an argument to define encryption, this is done in the parent cluster resource.

njgibbon:
Hey all. 

As @DrFaust has mentioned snapshot encryption for DocDB is controlled by enabling Encryption at Rest for DocDB.

And there is already a terraform check for this: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/DocDBEncryption.py

I believe that all that is needed here it to implement the DocDBEncryption for cloud formation and then this issue could be closed.

> You encrypt data at rest in your Amazon DocumentDB cluster by specifying the storage encryption option when you create your cluster. Storage encryption is enabled cluster-wide and is applied to all instances, including the primary instance and any replicas. It is also applied to your cluster’s storage volume, data, indexes, logs, automated backups, and snapshots. [0]

[0] AWS DocDB Enc at rest doc: https://docs.aws.amazon.com/documentdb/latest/developerguide/encryption-at-rest.html

PR #644, opened 2020-10-28
added check and tests for docdb encryption for cloudformation

njgibbon:
Hello,

First CF check I have added so I've tried to replicate based on some other similar checks.

To close issue https://github.com/bridgecrewio/checkov/issues/130 (see comments there).

This check corresponds with this terraform check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/DocDBEncryption.py and so I have replicated the type and description.

Notice that there are not many fields in the provided positive and negative tests data but this is because there are no required fields -  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbcluster.html

2 more fields were added to test data based on CF lint failure from GH actions feedback.

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/DocDBEncryption.py
+++ b/checkov/cloudformation/checks/resource/aws/DocDBEncryption.py
@@ -1,0 +1,16 @@
+from checkov.common.models.enums import CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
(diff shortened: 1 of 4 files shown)


schosterbarak (approved):
sweet

schosterbarak:
thanks @njgibbon

njgibbon:
No probs @schosterbarak i'm having fun :) I think this stuff is so valuable so i'm happy to help. Starting on the easier issues ! :)

PR #647, opened 2020-11-01
Add multiple variables to definitions context

rotemavni:


tronxd (approved):

robeden:
Is there a particular problem/case this is fixing? I'm in the process of reworking all this code and want to make sure I don't miss a scenario.

rotemavni:
@robeden 
There was a bug with creating the variable context. Before this fix the variable context used to contain only one variable for each file, because they were overwriting each other.
Adding the entity_name to the entity_context_path solved the problem.

robeden:
Got it, thanks

PR #648, opened 2020-11-03
Exclude NLBs from CKV_AWS_2 HTTPS check

### LINKED ISSUE #424: Cloudformation NLB Listener Rules are incorrectly detected as ALB rules and fail the HTTPS expectation
**Describe the bug**
NLB listeners are evaluated as ALB listeners, and fail because they don't use HTTPS

**To Reproduce**
1. Create a Cloudformation template with NLB listeners (e.g. TCP)
2. Run default tests
3. Tests will fail with `"AWS::ElasticLoadBalancingV2::Listener.<listenername>\" failed in check \"Ensure ALB protocol is HTTPS\"

**Expected behavior**
NLB rules should pass. 

**Screenshots**

 - OS: Ubuntu 20.04
 - Checkov Version: 1.0.423

**Additional context**
This resource:
```
    "ConsulDnsListener": {                                          
      "Type": "AWS::ElasticLoadBalancingV2::Listener",              
      "Properties": {                                               
        "DefaultActions": [                                         
          {                                                         
            "TargetGroupArn": {                                     
              "Ref": "ConsulDnsTargetGroup"                         
            },                                                      
            "Type": "forward"                                       
          }                                                         
        ],                                                          
        "LoadBalancerArn": {                                        
          "Ref": "LoadBalancerArn"                                  
        },                                                          
        "Port": 53,                                                 
        "Protocol": "TCP_UDP"                                       
      }                                                             
    },     
```
Produces this error:
```
{
  "type": "failure",
  "message": "Resource \"AWS::ElasticLoadBalancingV2::Listener.ConsulDnsListener\" failed in check \"Ensure ALB protocol is HTTPS\""
}```

### PR DESCRIPTION
**Pull request by jdheyburn**

Closes #424 

CloudFormation config may need to be verified too - I haven't worked with it before (I'm more of a TF dude).

I wanted to add protocols to exclude from the check, as opposed to only checking for `HTTP` so that we don't miss any new protocols supported by the ALB. If new protocols supported by ALB are added then we would miss them in this check.

### DIFF
--- a/checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
+++ b/checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
@@ -23,6 +23,10 @@ class ALBListenerHTTPS(BaseResourceCheck):
                 conf[key] == ["HTTPS"]
                 or
                 conf[key] == ["TLS"]
+                or
+                conf[key] == ["TCP"]
+                or
+                conf[key] == ["UDP"]
             ):
                 return CheckResult.PASSED
             elif conf[key] == ["HTTP"]:
--- a/checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
+++ b/checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
@@ -19,17 +19,7 @@ class ALBListenerHTTPS(BaseResourceCheck):
         """
         key = 'protocol'
         if key in conf.keys():
-            if (
-                conf[key] == ["HTTPS"]
-                or
-                conf[key] == ["TLS"]
-                or
-                conf[key] == ["TCP"]
-                or
-                conf[key] == ["UDP"]
-                or
-                conf[key] == ["TCP_UDP"]
-            ):
+            if conf[key] in (["HTTPS"], ["TLS"], ["TCP"], ["UDP"], ["TCP_UDP"]):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py [hunk 1]
```diff
@@ -23,6 +23,12 @@ def scan_resource_conf(self, conf):
                 conf[key] == ["HTTPS"]
                 or
                 conf[key] == ["TLS"]
+                or
+                conf[key] == ["TCP"]
+                or
+                conf[key] == ["UDP"]
+                or
+                conf[key] == ["TCP_UDP"]
```
nimrodkor, line 31: @jdheyburn Might be more readable to replace this prolonged condition as it is now to:
`conf[key] in ("HTTPS", "TLS", "TCP"...)`

nimrodkor (approved):
Thanks for the contribution @jdheyburn !

PTAL at the small comment below

checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
jdheyburn: Fair suggestion! I've made the change now 👍

schosterbarak:
super! 
thanks @jdheyburn

PR #649, opened 2020-11-03
Support suppressions by Bridgecrew IDs as well as Checkov IDs

nimrodkor:
1. Turn BcPlatformIntegration into a singelton imported from anywhere
2. Ingest the new `idMapping` field to allow suppression by BC IDs


tronxd (approved):

schosterbarak:
@nimrodkor should we change the printed ids in case of api token exists?

PR #650, opened 2020-11-03
Update CFN S3Encryption check to use BaseResourceValueCheck

mikeurbanski1:
The update to the check itself is straightforward.

I also had to update BaseResourceValueCheck to handle list indexes in parsing so that it works more like the Terraform BaseResourceValueCheck, despite the different path parsing logic.

All CFN tests pass.


schosterbarak:
Super!

checkov/cloudformation/checks/resource/base_resource_value_check.py
```diff
@@ -46,6 +46,13 @@ def scan_resource_conf(self, conf):
         matches = ContextParser.search_deep_keys(path_elements[-1], conf, [])
         if len(matches) > 0:
             for match in matches:
+                # CFN files are parsed differently from terraform, which causes the path search above to behave differently.
+                # The tesult is path parts with integer indexes, instead of strings like '[0]'. This logic replaces
```
schosterbarak, line 50: can you give an example?

mikeurbanski1: The key difference is that the `conf` in a Terraform check is a `dict`, whereas the `conf` in CFN checks is a `dict_node`. A couple levels down the stack in the `dpath.search(conf, inspected_key)` call in the TF BaseResourceValueCheck, there's a statement `type(obj)()`, where `obj == conf`. `dict` can be instantiated this way; `dict_node` cannot.

I think that's why this method of `BaseResourceValueCheck` is written differently for the two frameworks. (I tried just copying the code from the TF method to the CFN method and it bombs out.)

So the way the `ContextParser` works in this method, the `match` var on line 48 is: `['Properties', 'BucketEncryption', 'ServerSideEncryptionConfiguration', 0, 'ServerSideEncryptionByDefault', 'SSEAlgorithm', 'AES256']` - notice the `0` as an `int`. So this is the easiest way to make `path_elements = inspected_key.split('/')` match the `match` value while allowing the `inspected_key` to use the same syntax as it does for Terraform.

PR #654, opened 2020-11-04
DocDB TLS not disabled check for Cloudformation

njgibbon:
Check to match up with the terraform one already completed to close https://github.com/bridgecrewio/checkov/issues/165

Doc DB Param Group in Cloudformation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbclusterparametergroup.html
Doc DB Param Group 'Parameters' Doc: https://docs.aws.amazon.com/documentdb/latest/developerguide/cluster_parameter_groups-list_of_parameters.html

The default value is enabled and it is not disabled unless there is an explicit 'disabled' value for 'tls'. So only fail in this scenario and pass in all others.

**License**


schosterbarak (approved):
Thank you @njgibbon !

PR #655, opened 2020-11-06
Add case insensitivity to Azure SG rule values

mikeurbanski1:
This PR adds some flexibility in parsing components of security group rules for Azure. Values like "internet" / "Internet", "allow" / "Allow" all seem to work when creating a resource via Terraform.

Also checks for "*" in the protocol.

(Note: we may need to update a lot of checks for case insensitivity, which Azure seems to be flexible about)


justmeforyou (commented):
Good

PR #659, opened 2020-11-09
docs: update brew install/upgrade instruction

chenrui333:

relates to https://github.com/Homebrew/homebrew-core/pull/64401

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,13 @@ pip3 install checkov
 or using homebrew (MacOS only)
 
 ```sh
-brew tap bridgecrewio/checkov https://github.com/bridgecrewio/checkov
-brew update
 brew install checkov
+```
+
+or
+
+```sh
+brew upgrade checkov
 ```
 
 ### Configure an input folder


nimrodkor:
Hi @chenrui333 !

thanks for the contribution, it does seem very very cool. However, the question that's bothering me is how can we update the formula in homebrew? Checkov is still growing, and every release updates the homebrew formula. Implementing this change would mean we will need to update the remote formula every version, right?

chenrui333:
> Hi @chenrui333 !
> 
> thanks for the contribution, it does seem very very cool. However, the question that's bothering me is how can we update the formula in homebrew? Checkov is still growing, and every release updates the homebrew formula. Implementing this change would mean we will need to update the remote formula every version, right?

We (homebrew maintainers as well as contributors) will take care of it, we are pretty active keeping it up to date.

If we run into any issues, we will report into each of the repo.

chenrui333:
There is also another way that you can have it part of the release process. Once I merge my PR, I will share some that insights as well (which is what I have used in tflint)

schosterbarak:
hi @chenrui333  i think this is awesome. I'll merge it first thing tomorrow. 
one thing to notice is that we have a pipeline that auto updates the formula on every build of master (release): https://github.com/bridgecrewio/checkov/blob/master/.github/workflows/build.yml#L166. should we trigger a similar one on each master build on the brew repo?

chenrui333:
very much similar to that, https://github.com/terraform-linters/tflint/blob/master/.github/workflows/homebrew.yml

you need a repo token to PR from bridgecrewio/homebrew-core fork

nimrodkor (approved):

schosterbarak:
Sweet!

PR #660, opened 2020-11-09
DocDB Export to CloudWatch Logs - Cloudformation Implementation

njgibbon:
Check to implement the corresponding check for cloudformation for DocDB logs export to CloudWatch as is already done in terraform. You can see that the logic is the same and that even some of the implementation is identical.

Cloudformation API Doc for cluster to show test data is correct: https://docs.amazonaws.cn/en_us/AWSCloudFormation/latest/UserGuide/aws-resource-docdb-dbcluster.html

Terraform check for this:
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/DocDBLogging.py

**License**


schosterbarak (approved):
nice!

PR #663, opened 2020-11-10
workflows: release to homebrew

chenrui333:

Follow up of #659, a repo token and homebrew-core would be needed.

cc @schosterbarak @nimrodkor

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -14,4 +14,4 @@ jobs:
           # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
           formula-name: checkov
         env:
-          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
+          COMMITTER_TOKEN: ${{ secrets.PAT_TOKEN }}
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -1,0 +1,17 @@
+name: homebrew
+
+on:
+  push:
+    tags: 'v*'
+
+jobs:
+  homebrew:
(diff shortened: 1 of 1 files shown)


.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,17 @@
...
+  homebrew:
+    name: Bump Homebrew formula
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v1
+        with:
+          # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
+          formula-name: checkov
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
```
nimrodkor, line 17: Thanks @chenrui333 !

We do have a token already in our secrets, can we reuse it?
It's in use in: https://github.com/bridgecrewio/checkov/blob/52c7a73f9788a88a7f90408309c5fb6fa827f4d5/.github/workflows/build.yml#L176
```suggestion
          COMMITTER_TOKEN: ${{ secrets.PAT_TOKEN }}
```

chenrui333: Nice!! that is what I want.

nimrodkor (approved):

.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,17 @@
+name: homebrew
+
+on:
+  push:
+    tags: 'v*'
```
chenrui333, line 5: this needs to be `*` instead of `v*`

chenrui333: I just noticed the tags do not have `v`

PR #666, opened 2020-11-10
Add v to version tag

nimrodkor:

This will cause the workflow from #663 to run - the tag should be `v<VERSION>`

cc @chenrui333


chenrui333:
Thanks!

schosterbarak:
@nimrodkor please don't change release pattern. keep to the existing one.

PR #667, opened 2020-11-10
Replace local formula update with homebrew update

nimrodkor:

Merges flow from #663 to the existing release flow on every push to master

cc @chenrui333


schosterbarak (approved):

chenrui333 (commented):
by nature the build and release are separate, otherwise it would end up being throttled on the homebrew-core side.

chenrui333:
I think the github tag triggering should be good enough, no? Release on every master build seems little bit overwhelming.

schosterbarak:
@chenrui333 the following configuration ended up with the following failure. any ideas what we did wrong?
image

schosterbarak:
@chenrui333 ?

PR #669, opened 2020-11-11
Only check cluster for DocDB encryption on Terraform

rajivshah3:

Currently, checkov checks both the DocDB cluster and its instances for `storage_encrypted = true`. However, `storage_encrypted` is only an attribute of `aws_docdb_cluster`, not `aws_docdb_cluster_instance`.


schosterbarak (approved):
thank you @rajivshah3

Issue #386, opened 2020-06-28
Support for .tfvars


ymohamme:
We use .tfvars files to assign values to variables. Currently it looks like there is no support for files with a .tfvars extension. This is a great tool and adding this support would greatly benefit The terraform community.

schosterbarak:
Thank you @ymohamme for reporting this one. We are working on it. We do see it as part of a larger task that will enable deep-assignment of variables. We have estimated in a few weeks of work and @rotemavni is working hard on this one.

robeden:
@schosterbarak - I have an implementation (outside checkout) that I can contribute as a starting point if @rotemavni would like something to reference.

schosterbarak:
Thanks, @robeden  that will definitely help. let me connect I'll open a joined channel on this topic on our slack workspace :)

robeden:
I think we could handle "simple" variable parsing similar to what was done with serverless variables in #529. I've handled .tfvars, module and local references with a similar approach. There's some data that trickier (like existing resource lookups), but for basic variables it's pretty straight-forward.

jdheyburn:
This looks useful - any further progress on this one?

robeden:
I’m actively working on it right now. Hopefully we’ll have something to show soon.

schosterbarak:
@ymohamme try the latest version by @robeden and @rotemavni  :)

geertn:
I'm building my Storage Account names like this
```
name                = "${local.saname_prefix}diagnostics${module.tf-var-project.random_id}
```

With https://github.com/bridgecrewio/checkov/pull/429 merged I now get a Check failure on the SA name:

```
Check: CKV_AZURE_43: "Ensure the Storage Account naming rules"
	FAILED for resource: azurerm_storage_account.diagnostics
	File: /az_diag_sa.tf:8-22

		8  | resource "azurerm_storage_account" "diagnostics" {
		9  |   #checkov:skip=CKV_AZURE_35:Public access is allowed
		10 |   name                = "${local.saname_prefix}diagnostics${module.tf-var-project.random_id}"

````

robeden:
@geertn - Would you mind opening a new bug with your report? It will help raise it for visibility. Thanks!

Issue #651, opened 2020-11-04
Does not evaluate Terraform null variable default values


jdheyburn:
**Describe the bug**
I noticed this for CKV_AWS_86 "Ensure Cloudfront distribution has Access Logging enabled", unsure if it exists elsewhere for other checks. 

Checkov does not seem to evaluate Terraform `null` default values for variables in modules. Null in Terraform is the equivalent of having the attribute being set to that value as absent. 

I believe this is causing false negatives to appear in reports (passing when they should fail).

**To Reproduce**

main.tf
```hcl
resource "aws_cloudfront_distribution" "cf_dis" {
  # ...

  logging_config {
    include_cookies = var.logging_include_cookies
    bucket          = var.logging_bucket_id
    prefix          = var.logging_bucket_prefix
  }
 # ...
}
```

variables.tf
```hcl
variable "logging_include_cookies" {
  type        = bool
  description = "Whether to enable cookies in access logging"
  default     = null
}

variable "logging_bucket_id" {
  type        = string
  description = "The bucket ID where to store access logs"
  default     = null
}

variable "logging_bucket_prefix" {
  type        = string
  description = "The prefix where to store access logs"
  default     = null
}
```

The current settings means that no logging config is actually set, when when running:
```
$ checkov -d . -c CKV_AWS_86
```

It passes:

```
   ___| |__   ___  ___| | _______   __
  / __| '_ \ / _ \/ __| |/ / _ \ \ / /
 | (__| | | |  __/ (__|   < (_) \ V / 
  \___|_| |_|\___|\___|_|\_\___/ \_/  
                                      
By bridgecrew.io | version: 1.0.612 

terraform scan results:

Passed checks: 1, Failed checks: 0, Skipped checks: 0

Check: CKV_AWS_86: "Ensure Cloudfront distribution has Access Logging enabled"
        PASSED for resource: aws_cloudfront_distribution.cf_dis
```

**Expected behavior**
I would expect Checkov to evaluate the `null` value and fail the check.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Checkov Version [e.g. 22]: 1.0.612

**Additional context**
Additionally, even if the default value was set to a value, the calling module could overwrite the attribute with `null` - causing another false negative. I am not sure if it is related to this issue I am describing though.

On the other hand, the null check seems to work fine for S3 versioning for the below, where setting it to null causes a failed check.

main.tf
```hcl
[...]

Issue #622, opened 2020-10-21
Can't evaluate terraform list(string) default variable value


njgibbon:
**Describe the bug**
Check in question: CKV_AWS_37: "Ensure Amazon EKS control plane logging enabled for all log types" - https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/EKSControlPlaneLogging.py

When running checkov against the terraform 'aws_eks_cluster' resource and the list of log types is inline then the check passes.

When running checkov against the terraform 'aws_eks_cluster' resource and the list of log types is passed in as a variable then the check does not pass.

I suspect the issue is more wide-spanning than just this check though. 

**To Reproduce**
When running checkov against the following configuration (inline) the check will pass:
```
resource "aws_eks_cluster" "example" {
  name     = "example"
  enabled_cluster_log_types =  ["api", "audit", "authenticator", "controllerManager", "scheduler"]
}
```
When running checkov against the following configuration the check will fail:
```
resource "aws_eks_cluster" "example" {
  name     = "example"
  enabled_cluster_log_types = var.log_types_enabled
}

variable "log_types_enabled" {
 type = list(string)
 default = ["api", "audit", "authenticator", "controllerManager", "scheduler"]
}
```

**Expected behavior**
It is expected that the list variable default value will be evaluated and replaced and this is what will be analysed.

https://github.com/bridgecrewio/checkov/blob/master/docs/2.Concepts/Evaluations.md

**Screenshots**
None.

**Desktop (please complete the following information):**
 - OS: Found this error consistent on Windows, Mac and Linux. In all cases downloaded via pip.
 - Checkov Version: version: 1.0.590

njgibbon:
Hey - I just re-tested this with the new version and it has worked asexpected! With the varibales being correctly evaluated and the check passing.

This issue can now be closed. Great work @robeden

Issue #494, opened 2020-08-14
S3 logging check isn't working in some cases


mlushpenko:
**Describe the bug**
For one of our S3 buckets, `Ensure all data stored in the S3 bucket have versioning enabled` check passes when it should fail.

**To Reproduce**

Run checkov on the following resource :
```
resource "aws_s3_bucket" "k8s-logs" {
  #checkov:skip=CKV_AWS_52:we can't enable mfa delete via terraform
  bucket = var.logs-bucket-name
  acl    = "private"

  logging {
    target_bucket = var.kops-state-bucket-name
    target_prefix = "s3-access-logs/"
  }

  lifecycle_rule {

    id      = "logs"
    enabled = "true"

    transition {
      days          = 60
      storage_class = "GLACIER"
    }

    expiration {
      days = "365"
    }
  }

  server_side_encryption_configuration {
    rule {
      apply_server_side_encryption_by_default {
        sse_algorithm = "AES256"
      }
    }
  }
}
```

And the results is:

```
checkov -d test --quiet

       _               _              
   ___| |__   ___  ___| | _______   __
  / __| '_ \ / _ \/ __| |/ / _ \ \ / /
 | (__| | | |  __/ (__|   < (_) \ V / 
  \___|_| |_|\___|\___|_|\_\___/ \_/  
                                      
by bridgecrew.io | version: 1.0.484 

terraform scan results:

Passed checks: 6, Failed checks: 0, Skipped checks: 1
```

**Expected behavior**
Fail as versioning wasn't enabled. It only works when I disable versioning explicitly:

```
Passed checks: 5, Failed checks: 1, Skipped checks: 1

Check: CKV_AWS_21: "Ensure all data stored in the S3 bucket have versioning enabled"
        FAILED for resource: aws_s3_bucket.k8s-logs
        File: /s3.tf:1-37
        Guide: https://docs.bridgecrew.io/docs/s3_16-enable-versioning

                1  | resource "aws_s3_bucket" "k8s-logs" {
                2  |   #checkov:skip=CKV_AWS_52:we can't enable mfa delete via terraform
                3  |   bucket = var.logs-bucket-name
                4  |   acl    = "private"
                5  | 
                6  |   versioning {
                7  |     enabled = false
                8  |   }
```

**Desktop (please complete the following information):**
 - OS: MacOS
 - Checkov Version 1.0.484 

**Additional context**
[...]

PR #671, opened 2020-11-12
Terraform parser & variable evaluation rework

robeden:
⚠️ This is a big change, please review carefully. Some development and testing is still ongoing, so it is currently a draft PR. All feedback is welcome! ⚠️ 

This PR is an overhaul of the parsing and evaluation logic for Terraform. It brings number of new abilities and improvements:

- `.tfvars` support (`terraform.tfvars` and `.auto.tfvars`)
- JSON-format Terraform support (`.tf.json` and `.tfvars.json`)
- Improved variable handling
- Function support: `tomap`, `tolist`, `toset`, `tostring`, `tobool`, `tonumber` currently. More in development.
- module loading framework with local module loading currently (_mostly_) implemented
- Improved testability of TF parsing scenarios
- Environment variable support

**Implementation notes & details**

I set out to improve the existing parser and evaluation code and kept hitting walls in terms of how and when evaluation needed to be done, so this ended up being a rewrite of those pieces of code into `parser2`. (That name doesn't need to stick, but it was easiest to keep around for comparison and testing for the time being.)

Like my serverless parser implementation, this is looping parser/evaluator which keeps processing until the structure is stabilized (or it hits a limit). This might have performance implications and should be checked against large data sets.

There are many test scenarios under `tests/terraform/evaluation/resources/parser_scenarios` and the corresponding `test_parser_scenarios.py`.

One of the biggest struggles I had was keeping track of what data needed to go where in the various dicts floating around, particularly around evaluations. To aid that, I created a dataclass in `checkov/common/variables/context.py` which `parser2` emits and the `runner` expands for the report. This will feel different for anyone familiar with the previous approach.

I used typing in new pieces I wrote to help with documentation and keeping things straight. I haven't used any tool to formally enforce, but simply for documentation and IDE convenience. I know some python devs don't like typing, so it can be pulled if we don't want it cluttering things.

**Remaining work**

- ~Module loading basically works, but references to outputs from modules is not yet implemented.~
- ~How module data is stored in the structure map is a little weird. Currently everything is designed to organized under the file it's defined in, but this gets pretty inconvenient with loaded modules (for resolving `module.` references). I haven't settled on what we'd want here and opinions are welcome. See the `maze_of_variables` test case.~ <br>Resolved module data is now stored within the calling module block. See comment [here](https://github.com/bridgecrewio/checkov/pull/671/commits/27c62beb3f90c136398c0f160bf36aa09613414e).
- ~Test cases were light previously for cases that I know must be relevant. More work should be done to find special cases from the old parser/evaluators to ensure they're captured in test cases.~ <br>I've dug fairly deep on this and think I've caught everything.
- ~Some test cases aren't thoroughly checking their variable evaluation. Need to add `eval.json` files to their scenarios.~ <br>Much better now.
- ~There are probably a few other functions that would be good.~ <br>Can be handled in future work.
- ~Context info (code & line numbers) is not properly associated with loaded module data by the runner.~ <br>Fixed.

Fixes: #386, #622, #651, #446
Related to: #314, #494 


checkov/terraform/parser2.py
```diff
@@ -0,0 +1,527 @@
...
+                    return orig_variable     # no change
+    # TODO ROB - format() support
+    # elif orig_variable.startswith("format(") and orig_variable.endswith(")"):
+    #     format_tokens = orig_variable[7:-1].split(",")
+    #     return format_tokens[0].format([_to_native_value(t) for t in format_tokens[1:]])
+
+    return orig_variable        # fall back to no change
+
+
+def _load_or_die_quietly(file: os.PathLike, parsing_errors: Dict) -> Optional[Mapping]:
```
nimrodkor, line 488: @robeden Loved the name 🤣

checkov/terraform/runner.py
```diff
@@ -87,21 +94,23 @@ def check_tf_definition(self, report, root_folder, runner_filter, collect_skip_c
             definitions_context = {}
             for definition in self.tf_definitions.items():
                 definitions_context = parser_registry.enrich_definitions_context(definition, collect_skip_comments)
-            variable_evaluator = ConstVariableEvaluation(root_folder, self.tf_definitions, definitions_context)
-            variable_evaluator.evaluate_variables()
-            self.tf_definitions, self.definitions_context = variable_evaluator.tf_definitions, variable_evaluator.definitions_context
```
nimrodkor, line 92: @robeden The `evaluate_variables()` method also turned all hardocded "true" and "false" in resources to True / False. Is that taken care of in your new parser?

checkov/terraform/runner.py
```diff
@@ -115,8 +124,22 @@ def run_block(self, entities, definition_context, full_file_path, report, scanne
...
+                    #     'region': {
+                    #         'var_file': '/tf/example.tf',
+                    #         'value': 'us-east-1',
+                    #         'definitions': [{
+                    #             'definition_name': 'region',
+                    #             'definition_expression': '${var.region}',
+                    #             'definition_path': 'resource/0/aws_s3_bucket/foo-bucket/region/0'
+                    #         }]
+                    #     }
+                    # }
```
nimrodkor, line 139: @robeden What's this? Is it `self.evaluations_context[full_file_path]`?

nimrodkor (commented):
@robeden Great job! Loved the documentation and extensive testing.

checkov/terraform/parser2.py
```diff
@@ -0,0 +1,527 @@
...
+                        dir_filter=lambda d: self._check_process_dir(d))
+
+    @staticmethod
+    def parse_file(file: str, parsing_errors: Dict[str, Exception] = None) -> Optional[Dict]:
+        if not file.endswith(".tf") and not file.endswith(".tf.json"):
+            return None
+        return _load_or_die_quietly(Path(file), parsing_errors)
+
+
+def parse_directory(directory: str, include_sub_dirs: bool, out_definitions: Dict,
```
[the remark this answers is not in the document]
robeden: Yes, it should be. Will change.

checkov/terraform/parser2.py
```diff
@@ -0,0 +1,527 @@
...
+        except OSError:
+            # Skip files that can't be accessed
+            continue
+
+        # Variable files
+        # See: https://www.terraform.io/docs/configuration/variables.html#variable-definitions-tfvars-files
+        if file.name == "terraform.tfvars.json":
+            json_tfvars = file
+            continue
+        elif file.name == "terraform.tfvars":
```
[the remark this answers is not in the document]
robeden: Well... maybe. I [followed the spec](https://www.terraform.io/docs/configuration/variables.html#variable-definitions-tfvars-files) for which files are automatically loaded and the order they're loaded in. Of course, the problem that comes up is that you can also specify files via the `-var-file` flag. The problem is, generally in my experience, the reason to that would be that you have alternate sets of variables (test, etc), so just blindly loading them all will simply produce conflicts that we don't know how to resolve or whether they should have been loaded in the first place. I couldn't think of a simple solution to that problem.

checkov/terraform/parser2.py
robeden: Thanks. 🙂

checkov/terraform/runner.py
robeden: Yes, sorry... that's some old guidance I left for myself. That structure really took me a while to wrap my head around, so I left it in there to remind me. I'm good to pull it now.

checkov/terraform/runner.py
robeden: I think `true` and `false` are handled, but let me make sure there's a test case to be sure. It also converted `1` and `0`, which I'm less excited about converting.

I have a plan for getting back to TF 0.11 support and that location might be a better place to special-case that if we want, since I believe it's behavior specific to <0.12?

robeden: Looks like it's not currently handled. I threw in the easy implementation, but that broke some more sophisticated logic in the `tomap_function` parser scenario (specifically the MIXED_BOOL) test case which is ensuring the datatype in maps is consistent. I'll have to think on how to prevent that. (If it matters. For checkov's purposes, some portions of the spec probably aren't important.)

checkov/terraform/parser2.py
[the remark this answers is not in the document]
nimrodkor: Agreed

checkov/terraform/runner.py
nimrodkor: I agree that `false` and `true` strings are more important to handle. Can you handle them?

robeden: Yep, will do.

robeden: What do you think about converting “0” and “1”? How big a deal are they?

checkov/terraform/parser2.py
```diff
@@ -0,0 +1,522 @@
...
+    #          variable default values.
+    #          Variable files are also flagged for later processing.
+    var_values: Dict[str, Any] = {}
+    hcl_tfvars: Optional[os.DirEntry] = None
+    json_tfvars: Optional[os.DirEntry] = None
+    auto_vars_files: Optional[List[os.DirEntry]] = None      # lazy creation
+    for file in os.scandir(directory):
+        # Ignore directories and hidden files
+        try:
+            if not file.is_file() or file.name.startswith("."):
```
rotemavni, line 119: @robeden In this iteration form we don't iterate inner folders. Is this the desired behaviour?

robeden: It's a little non-obvious because the walk happens in `_parse_directory` (line 84).

I did it this way because Terraform doesn't include subdirectories automatically, so `_internal_dir_load` takes that approach and `_parse_directory` deals with Checkov's recursive loading behavior.

checkov/terraform/runner.py
robeden: `true` and `false` are now handled. I haven't done anything with `1` and `0` for now.

robeden:
@schosterbarak - trufflehog is finding the fake key from your branch, I think? Not sure what to do about that.

checkov/terraform/runner.py
nimrodkor: 0 and 1 are old conventions, and prone to many false positives as they are valid values by themselves. I'd leave it as is for now

nimrodkor (approved):
LGTM. Want to make this PR a non-draft one?

schosterbarak:
@robeden thanks. fix on master

checkov/terraform/parser2.py
rotemavni: @robeden Thanks for the response.
I'm adding some support in loading modules from terraform registry, and I encountered a dilemma: After loading a module, it will only be iterated using `_internal_dir_load` and won't parse any subdirectory. 
What do you think should be done in this context? perhaps replacing the call to `_internal_dir_load` from to `_parse_directory`?

robeden: I think it should only hit the main directory and then branch off from there based on additional module references. But it’s possible I’m wrong in my understanding. Do you have an example or test case you can point me at?

I should be on slack in an hour or so if you want to talk in real-time there.

robeden:
@nimrodkor - I have a couple module things I’d like to finish up today (see remaining work in description). But I suppose those could be done in a separate PR if you want this in quickly.

Has anyone had a way to verify the performance isn’t horrible? I don’t have a big enough dataset from before to assess.

checkov/terraform/parser2.py
robeden: FYI, @rotemavni and I spoke offline and I think we've convinced ourselves that the current behavior is correct. What we didn't know is that it looks like you can reference deeper into a module with syntax like this ([reference](https://github.com/terraform-aws-modules/terraform-aws-security-group)):
```
module "web_server_sg" {
  source = "terraform-aws-modules/security-group/aws//modules/http-80"
```
So, we'll look for that double slash as the possible entry point.

robeden:
I’ve improved module handling so their data is properly checked and references into modules now work. The location of module data has changed to aide references and to keep the location where it was loaded from. However, context info is not properly associated yet. 

I’m still struggling to wrap my brain around how some of that code wants to work, so someone more familiar with it might be able to quickly spot what should happen. For now, I made it fail gracefully if it can’t find the context.

I still expect this code to have sharp edges and be a bit brittle, but I’ll pull this out if “draft” mode so it’s mergeable when you’re comfortable with it.

robeden:
After talking with @rotemavni , we're going to change the way the resolved module data is stored in the TF structure. So, I've moved this back to draft status while we work that out.

robeden:
All working again.

robeden:
Here's output showing a problem found deep down in a module (from [this test case](https://github.com/robeden/checkov/tree/terraform-variable-parsing/tests/terraform/parser/resources/parser_scenarios/module_matryoshka):
```
terraform scan results:

Passed checks: 3, Failed checks: 1, Skipped checks: 0

Check: CKV_AWS_20: "S3 Bucket has an ACL defined which allows public READ access."
	PASSED for resource: aws_s3_bucket.example0
	File: /buckets.tf:5-7

Check: CKV_AWS_20: "S3 Bucket has an ACL defined which allows public READ access."
	PASSED for resource: aws_s3_bucket.example2
	File: /bucket1/bucket2/bucket.tf:5-7

Check: CKV_AWS_20: "S3 Bucket has an ACL defined which allows public READ access."
	PASSED for resource: aws_s3_bucket.example1
	File: /bucket1/bucket.tf:5-7

Check: CKV_AWS_20: "S3 Bucket has an ACL defined which allows public READ access."
	FAILED for resource: aws_s3_bucket.example3
	File: /bucket1/bucket2/bucket3/bucket.tf:1-4

		1 | resource "aws_s3_bucket" "example3" {
		2 |   bucket = "bucket3"
		3 |   acl    = "public-read"
		4 | }
```

checkov/terraform/parser2.py
```diff
@@ -0,0 +1,695 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+_VAR_PATTERN = re.compile(r'\${([^{}]+?)}')
+_SIMPLE_TYPES = frozenset(["string", "number", "bool"])
+
+_RESOURCE_REF_PATTERN = re.compile(r'[\d\w]+(\.[\d\w]+)+')
+
+
+class Parser2:
```
rotemavni, line 26: @robeden  
I think it's safe to rename this class to `Parser` now (and the file name as well).

rotemavni (approved):
Looks great! GJ!

checkov/terraform/parser2.py
robeden: Done

PR #674, opened 2020-11-17
Check - Cloudformation - API Gateway Access Logging

njgibbon:
Hello,

I have implemented the Cloudformation version of this check following the same logic as with the terraform one that already exists: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/APIGatewayAccessLogging.py

The following references have been used to ensure the v1 and v2 test data is correct:
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-stage.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-stage-accesslogsetting.html
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigatewayv2-stage-accesslogsettings.html

See PR comment for why this has been split in to 2 checks instead of 1 as it is with Terraform. 

**License**


njgibbon:
Oops. Note the cflint output
`E3002 Invalid Property Resources/MyStage/Properties/AccessLogSettings`
V1 has AccessLogSetting and V2 has AccessLogSetting**s**. Annoying.
As a consequence I think this should be split in to 2 simple checks. 1 for V1 and 1 for V2.

njgibbon:
Now I see there is a conflict because I have used the same ID across 2 checks in the same category. I'm going offline now.

Would it be better for the API V1 and V2 to have different IDs?

Or should I implement a ResourceCheck where I check if either AccessSetting or AccessSettings exist?

nimrodkor:
Hi @njgibbon !
Thanks for the contribution. I think 2 separate checks with two different IDs is a good fit here, since the APIs are broken between the versions.

njgibbon:
Ok @nimrodkor that's what i've done. 

In Terraform they have made sure that both V1 and V2 have the same config API so that check still works correctly for both resources. Now you have the option of splitting the terraform check to CKV_AWS_76 and CKV_AWS_95 if you want symmetry - but you don't need to.

schosterbarak:
Thank you @njgibbon

PR #676, opened 2020-11-19
CloudFormation Check - KMS key policy wildcard principal

avishayil:
Hi everyone,
I implemented this check as I believe this should be a security best practice - not to allow wildcard principal on KMS Key policies.

This check tackles both cases with `"Principal": "*"` and `"Principal": {"AWS": "*"}` which should be considered dangerous, and tests are handling both implementations.

More information can be found on the AWS documentation here:
https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
@@ -4,7 +4,7 @@ from checkov.cloudformation.checks.resource.base_resource_value_check import Bas
 
 class KMSKeyWildCardPrincipal(BaseResourceValueCheck):
     def __init__(self):
-        name = "Ensure rotation for customer created CMKs is enabled"
+        name = "Ensure KMS key policy does not contain wildcard (*) principal"
         id = "CKV_AWS_28"
         supported_resources = ['AWS::KMS::Key']
         categories = [CheckCategories.ENCRYPTION]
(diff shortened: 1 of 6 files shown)


checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self):
+        return 'Properties/KeyPolicy/Statement/Principal'
+
+    def scan_resource_conf(self, conf):
+        if conf.get('Properties'):
+            if conf['Properties'].get('KeyPolicy'):
+                policy_block = conf['Properties']['KeyPolicy']
+                principals_list = get_recursively(policy_block, 'Principal')
```
schosterbarak, line 48: @avishayil  looks great! 
do you think the recursive function can be replaced with `dpath`? - was not sure since i haven't drilled down a lot on it.

avishayil: Hi, I thought about using `dpath` initially, but I couldn't get the `conf` object to work with it.
For example:
```python
for x in dpath.util.search(str(policy_block), "Statement", yielded=True):
    print(x)
```

ends with `dpath` error: `E       TypeError: __init__() missing 3 required positional arguments: 'x', 'start_mark', and 'end_mark'`

That's why I chose the recursive function to resolve the principals.

schosterbarak: Makes sense. Thank you!

PR #677, opened 2020-11-19
Remote Modules Loading

rotemavni:
Added support for loading external modules from terraform registry, github, bitbucket and general git links.

**Some background regarding loading from terraform registry**:
In order to load from the registry, the source and version should be retrieved.
The source will have a prefix of `<namespace>/<name>/<provider>` and may contain a suffix `//<inner_dir>/...`
If the suffix exists we will load the files in these sub-directories.
The version may be specified as a version constraint (such as "~> v3.0") or not at all. The algorithm evaluates which version out of the existing available versions is the best match for the provided version. If the version doesn’t exist, we will load the latest version.
In order to load the module an API call is made to the registry, and the response contains a header with the git url to download.

**The code includes the following changes:**
1. Refactored `get_git.py` to support cloning with git refs versions and avoid cloning into ‘clone’ and ‘result’ directories.
2. Updated `ModuleLoaderRegistry`:
- determine whether to download the external modules and to which folder.
- loop through the loaders as long as there is `next_url` in the ModuleContent, which indicates that the loading process is not yet complete.
3. Updated `RegistryLoader` with the process of retrieving the git url for the module download and checking if the source contains inner directories.
4. Added to `main` three new args that will be added to the `RunnerFilter`: 
- `download_external_modules` - default is False
-`external_modules_download_path` - defaults to `".external_modules"`
- `evaluate_variables` defaults to True
5. Refactored terraform parser by transforming some of the functions into class methods to avoid passing many parameters and use class members instead.
6. Fixed an unrelated bug I found during testing in `ALBListenerHTTPS` terraform check.


tests/terraform/parser/resources/parser_scenarios/tobool_function/expected.json
```diff
@@ -6,7 +6,7 @@
         "STRING_T": [true],
         "RAW_F": [false],
         "STRING_F": [false],
-        "JUNK": ["${tobool(\"invalid\")}"]
+        "JUNK": ["\"invalid\""]
```
robeden, line 9: This is certainly a valid option, but I don’t know if it’s the best thing to convert to. In terraform, this would generate an error. My approach has generally otherwise been to avoid changing anything is a value couldn’t be resolved. So, personally, I have a slight preference to leave the statement unresolved.

checkov/terraform/module_loading/loader.py
```diff
@@ -13,25 +15,31 @@
...
-    the module files
- 3) the loader tried to load the module content but and error occurred, in which case an exception
-    is raised.
-        :param current_dir: Directory containing the reference to the module.
-        :param source: the raw source string from the module's `source` attribute (e.g.,
-                       "hashicorp/consul/aws" or "git::https://example.com/vpc.git?ref=v1.2.0")
-        :param source_version: contains content from the module's `version` attribute, if provided
-
-        :return: A ModuleContent object which may or may not being loaded.
-        """
```
robeden, line 36: Is dropping this documentation intentional?

checkov/terraform/module_loading/content.py
```diff
@@ -3,8 +3,9 @@
 
 
 class ModuleContent(object):
-    def __init__(self, dir: Optional[Union[tempfile.TemporaryDirectory, str]]) -> None:
-        self.dir = dir
+    def __init__(self, dir: Optional[Union[tempfile.TemporaryDirectory, str]], next_url=None, exception=None) -> None:
+        self.dir = dir.replace('//', '/') if dir else dir
```
nimrodkor, line 7: @rotemavni `if dir else dir`? In which case is dir None? If that's the case - let the `else` case be None to be clearer please

checkov/terraform/module_loading/loaders/versions_parser.py
```diff
@@ -0,0 +1,65 @@
...
+    raw_version = raw_version.replace(' ', '')
+    raw_version_constraints = raw_version.split(',')
+    version_constraints = []
+    for constraint in raw_version_constraints:
+        constraint_parts = re.search(VERSION_REGEX, constraint).groupdict()
+        version_constraints.append(VersionConstraint(constraint_parts))
+    return version_constraints
+
+
+def order_versions_by_size(versions_strings):
```
nimrodkor, line 50: @rotemavni Not sure what `by_size` means, and also it would be great if the method name would be explicit about the DESC ordering :)

checkov/terraform/parser.py
```diff
@@ -23,10 +25,38 @@
...
     def __init__(self):
         self._parsed_directories = set()
 
+    def _init(self, directory: str, out_definitions: Optional[Dict],
+              out_evaluations_context: Dict[str, Dict[str, EvaluationContext]],
+              out_parsing_errors: Dict[str, Exception],
+              env_vars: Mapping[str, str],
+              download_external_modules: bool,
+              external_modules_download_path: str, evaluate_variables):
+        self._parsed_directories.clear()
```
nimrodkor, line 43: @rotemavni Why does `init()` do a `clear()`?
If you need it in `parse_directory` - do it there please

nimrodkor (approved):
LGTM, small comments.
GJ @rotemavni !

ismailyenigul:
Hi
It seems this feature is not working well with module source with `ref=`

```
module "label" {
  source  = "git::https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.19.2"
  context = module.this.context
}

 checkov --download-external-modules  true -d .
ERROR:checkov.terraform.module_loading.loader:failed to get git::https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.19.2 because of Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.19.2 /Users/ismail/dev/terraform-aws-vpc-flow-logs-s3-bucket/.external_modules/git::https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.19.2
  stderr: 'Cloning into '/Users/ismail/dev/terraform-aws-vpc-flow-logs-s3-bucket/.external_modules/git::https://github.com/cloudposse/terraform-null-label.git?ref=tags/0.19.2'...
warning: redirecting to https://github.com/cloudposse/terraform-null-label?ref=tags/0.19.2/
fatal: https://github.com/cloudposse/terraform-null-label?ref=tags/0.19.2/info/refs not valid: is this a git repository?
'

```

You can get the full version of the file from https://github.com/cloudposse/terraform-aws-vpc-flow-logs-s3-bucket/blob/f38956c669a371d1540d3c9d23383d58801579ba/main.tf

PR #682, opened 2020-11-21
Avoid parser crashes on count.index

nimrodkor:

FYI @robeden @rotemavni


schosterbarak:
Thanks. Can you please add UT with the scenario?

PR #694, opened 2020-11-27
Check - CF - API GW Auth - correspond with TF

njgibbon:
Hello,

Continuing to bring CF checks inline with TF. Followed the logic of the tf already implemented.

CF Doc(s) for Test Data:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html
https://docs.aws.amazon.com/apigateway/api-reference/resource/method/

Terraform check:
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/APIGatewayAuthorization.py

**License**


schosterbarak:
Thanks @njgibbon

Issue #518, opened 2020-08-26
Junitxml output returns multiple files 


secustor:
**Describe the bug**
If multiple frameworks are triggered, checkov will print two separate junit.xml files to the output. This prevents Gitlab to parse the artifact as the XML is malformed. 

**To Reproduce**
1. Prepare a folder containing files of two frameworks
2. Run:
   a. `- checkov -d . --repo-id $CI_PROJECT_NAMESPACE/$CI_PROJECT_NAME --branch $CI_COMMIT_REF_NAME -s`
   b  `- checkov -d . -o junitxml > checkov.xml`
3. This results in following file: 

``` xml
<?xml version="1.0" ?>
<testsuites failures="3" errors="0" tests="5" disabled="0" time="0.0">
	<testsuite name="Ensure RBAC is enabled on AKS clusters" disabled="0" failures="0" errors="0" skipped="0" time="0" tests="1" package="checkov.terraform.checks.resource.azure.AKSRbacEnabled">
		<testcase name="terraform Ensure RBAC is enabled on AKS clusters azurerm_kubernetes_cluster.monitoring" classname="checkov.terraform.checks.resource.azure.AKSRbacEnabled" file="/aks.tf"/>
	</testsuite>
	<testsuite name="Ensure Kube Dashboard is disabled" disabled="0" failures="0" errors="0" skipped="0" time="0" tests="1" package="checkov.terraform.checks.resource.azure.AKSDashboardDisabled">
		<testcase name="terraform Ensure Kube Dashboard is disabled azurerm_kubernetes_cluster.monitoring" classname="checkov.terraform.checks.resource.azure.AKSDashboardDisabled" file="/aks.tf"/>
	</testsuite>
	<testsuite name="Ensure AKS has an API Server Authorized IP Ranges enabled" disabled="0" failures="1" errors="0" skipped="0" time="0" tests="1" package="checkov.terraform.checks.resource.azure.AKSApiServerAuthorizedIpRanges">
		<testcase name="terraform Ensure AKS has an API Server Authorized IP Ranges enabled azurerm_kubernetes_cluster.monitoring" classname="checkov.terraform.checks.resource.azure.AKSApiServerAuthorizedIpRanges" file="/aks.tf">
			<failure type="failure" message="Resource &quot;azurerm_kubernetes_cluster.monitoring&quot; failed in check &quot;Ensure AKS has an API Server Authorized IP Ranges enabled&quot;"/>
		</testcase>
	</testsuite>
[...]

PR #705, opened 2020-12-01
Combine junit reports

mikeurbanski1:
This PR fixes JUnit XML output when there are results for multiple reports (e.g., Terraform and CFN). Previously, it would print multiple distinct outputs, resulting in an invalid XML file.


schosterbarak:
fix https://github.com/bridgecrewio/checkov/issues/518

PR #716, opened 2020-12-04
Add PoC Parliament integration check

KevinHock:
So @schosterbarak, I wanted to check your thoughts on this, because I dynamically modified the `.name` attribute, 🤔 , and I wasn't sure if this broke expectations for other users/you.

If you approve of my methods, then I'll add a test for this PR.

### DIFF
--- a/checkov/terraform/checks/resource/aws/ParliamentProblem.py
+++ b/checkov/terraform/checks/resource/aws/ParliamentProblem.py
@@ -1,0 +1,85 @@
+import json
+from pathlib import Path
+
+import parliament
+from checkov.common.models.enums import (
+    CheckCategories,
+    CheckResult,
+)
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from parliament.cli import is_finding_filtered
+
+
+class ParliamentProblem(BaseResourceCheck):
+
+    def __init__(self):
+        name = "duo-labs/parliament issue(s) found"
+        self.base_name = "duo-labs/parliament issue(s) found"
+        id = "CKV_AWS_96"
+        categories = [CheckCategories.IAM]
+        supported_resources = ['aws_iam_policy']
+        super().__init__(
+            name=name,
+            id=id,
+            categories=categories,
+            supported_resources=supported_resources,
+        )
+
+    def scan_resource_conf(self, conf):
+        """
+        TODO
+
+        :param conf: dict of supported resource configuration
+        :return: <CheckResult>
+        """
+        # We change the name only if we find issues
+        self.name = self.base_name
+
+        policy_string = conf['policy'][0].strip()
+
+        try:
+            json.loads(policy_string)
+        except ValueError:
+            return CheckResult.PASSED
+
+        enhanced_filtered_findings = [
+            enhanced_finding
+            for enhanced_finding in map(
+                parliament.enhance_finding,
+                parliament.analyze_policy_string(
+                    policy_string,
+                    include_community_auditors=True,
+                    config=parliament.config,
+                ).findings,
+            )
+            if not is_finding_filtered(
+                enhanced_finding,
+                minimum_severity="MEDIUM",
+            )
+        ]
+        if enhanced_filtered_findings:
+            self.name = f"{self.name}\n" + "-\n".join(
+                map(
+                    str,
+                    enhanced_filtered_findings,
+                )
+            )
+            return CheckResult.FAILED
+        return CheckResult.PASSED
+
+
+
+def add_config_for_community_auditors():
+    community_auditors_override_file = (
+        Path(parliament.config_path).parent
+        / "community_auditors"
+        / "config_override.yaml"
+    )
+    # This adds to parliament.config
+    parliament.override_config(community_auditors_override_file)
+
+
+add_config_for_community_auditors()
+
+
+check = ParliamentProblem()


KevinHock:
Sample output:

```
Passed checks: 36, Failed checks: 2, Skipped checks: 0



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.example
	File: /aws_iam_policy_document_parse_error.tf:30-34



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.lambda_logging
	File: /aws_iam_policy_lambda.tf:1-21



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.example
	File: /aws_iam_policy_document_empty.tf:6-10



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.example
	File: /aws_iam_policy_document_bad_resource_formats.tf:48-52



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.policy
	File: /aws_iam_policy.tf:1-20



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.example
	File: /aws_iam_policy_single_line.tf:1-4





Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.policy
	File: /aws_iam_policy_bad_action.tf:1-21



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found"
	PASSED for resource: aws_iam_policy.example
	File: /aws_iam_policy_document.tf:47-51



Check: CKV_AWS_96: "duo-labs/parliament issue(s) found
CREDENTIALS_EXPOSURE -  - {'actions': ['redshift:getclustercredentials', 'ecr:getauthorizationtoken'], 'filepath': None}-
PERMISSIONS_MANAGEMENT_ACTIONS -  - {'actions': ['redshift:getclustercredentials'], 'filepath': None}"
	FAILED for resource: aws_iam_policy.policy
	File: /my_file.tf:1-21

		1  | resource "aws_iam_policy" "policy" {
		2  |   name        = "test_policy"
		3  |   path        = "/"
		4  |   description = "My test policy"
		5  | 
		6  |   policy = <<EOF
		7  | {
		8  |   "Version": "2012-10-17",
		9  |   "Statement": [
		10 |     {
		11 |       "Effect": "Allow",
		12 |       "Action": [
		13 |         "redshift:getclustercredentials",
		14 |         "ecr:getauthorizationtoken"
		15 |       ],
		16 |       "Resource": "*"
		17 |     }
		18 |   ]
		19 | }
		20 | EOF
		21 | }


Check: CKV_AWS_96: "duo-labs/parliament issue(s) found
RESOURCE_POLICY_PRIVILEGE_ESCALATION - Possible resource policy privilege escalation on * due to s3:PutObject not being allowed, but does allow s3:PutBucketAcl - {'filepath': None}-
RESOURCE_POLICY_PRIVILEGE_ESCALATION - Possible resource policy privilege escalation on * due to s3:GetObject not being allowed, but does allow s3:PutBucketAcl - {'filepath': None}-
RESOURCE_POLICY_PRIVILEGE_ESCALATION - Possible resource policy privilege escalation on * due to s3:DeleteObject not being allowed, but does allow s3:PutBucketAcl - {'filepath': None}-
PERMISSIONS_MANAGEMENT_ACTIONS -  - {'actions': ['ram:createresourceshare', 'lambda:addpermission', 's3:putbucketacl'], 'filepath': None}"
	FAILED for resource: aws_iam_policy.policy
	File: /test_permissions_management.tf:1-22

		1  | resource "aws_iam_policy" "policy" {
		2  |   name        = "test_permissions_management"
		3  |   path        = "/"
		4  |   description = "My test policy"
		5  | 
		6  |   policy = <<EOF
		7  | {
		8  |   "Version": "2012-10-17",
		9  |   "Statement": [
		10 |     {
		11 |       "Effect": "Allow",
		12 |       "Action": [
		13 |         "lambda:addpermission",
		14 |         "s3:putbucketacl",
		15 |         "ram:CreateResourceShare"
		16 |       ],
		17 |       "Resource": "*"
		18 |     }
		19 |   ]
		20 | }
		21 | EOF
		22 | }
```

kinda dirty, and my severity `"MEDIUM"` filter is opinionated, and the difference between high-sev findings by parliament and medium-sev findings are blurred.


It does work though, it seems, and so I might be lazy and add custom checks for my own uses that work for more resource types.

I couldn't think of a better way to do this without a major architectural change to checkov, maybe you have a better idea.

KevinHock:
_gonna ignore the import errors, this is mainly a PoC just to get feedback on_

robeden:
I think dynamically modifying the `name` on the check would probably work once, but could do strange things on subsequent runs. The problem is there is only one instance of each check, so when the name is changed on the first check, the name will still be changed on the second check. This means the "passed" report for the second run would have the name from the first.

This could be worked around to probably work at the moment, but the behavior is always going to be brittle.

I've wanted to change the name reported based on the result (to provide more detail) often, so this is something that would be nice to have an answer for.
What about something like this:
```
@dataclass
class CheckResultAndName:
    result: CheckResult
    new_name: str
```

Then BaseCheck.run (etc.) would handle either CheckResult or CheckResultAndName? (Type: `Union[CheckResult, CheckResultAndName]`)

KevinHock:
Sounds like a great idea @robeden!

I tested the current PR, and the order of the checks didn't matter for the output, restoring the name to the `base_name` at the start of the check function ended up working fine.

I agree it is dirty and brittle, so I'll make this PR use the `dataclass`.

KevinHock:
_will update this PR on Monday, had a couple other things pop up today_

Have a great weekend ;)

robeden:
@KevinHock - if you want any help with that, let me know. 

I was thinking today and “ExtendedCheckResult” might be a better class name. And probably something like “name_override” for the attribute.

schosterbarak:
Thanks @KevinHock 
@nimrodkor let's sync on that. i wonder if it will break something upstream

KevinHock:
@schosterbarak Lemme know what the results of the sync are, in full transparency I would like to get this PR done by the end of the week, for reasons :D 😉

schosterbarak:
@KevinHock do you think it's possible to create a separate checkov check (with separate name) for any parliament finding instead of one class?

KevinHock:
Do you mean dynamically create classes depending on how many Parliament checkers there are?

I am not sure, but for some velocity concerns, I think I might full-speed ahead with keeping this as a custom check and maybe revisit this PR later 🤔

KevinHock:
Hmm, what ID would these have then @schosterbarak? Each check needs to have an e.g. `CKV_AWS_96` that is unique 🤔 

Creating classes ain't bad because all the findings have the right attributes, but not sure how to handle ID  🤷🏼

KevinHock:
The checks in Parliament are spread all through out the Parliament codebase, in `add_finding` calls.
 
We need to scan a resource with Parliament -> get the list of findings -> then create classes based on the attributes of the findings.

The `scan a resource` code in this PR happens in 1 Parliament check.

I am not sure where we could do "Let's run all the regular checkov checks, on this resource. Now let's run parliament on it too." 🤔

robeden:
Maybe it would be best to have another field for additional data in the report. That could be used to report the details of the parliament problem. 

Another use case where this makes sense: we have a check that a tag is present and contains one of a set of values (identifying the product to which infrastructure belongs). It would be nice to specify the invalid value when there’s a typo.

schosterbarak:
@robeden i agree, we will need to modify a little the reporting object to make that happened without breaking anything. 
@KevinHock we  (@nimrodkor) want to run parliament on a few of our databases and see if it's not too noisy. it might take us time.  
for now, i suggest running this as a custom policy: https://www.checkov.io/2.Concepts/Custom%20Policies.html

KevinHock:
sgtm, definitely understand wanting to test the noise, plus you'll have to redo the process whenever you bump parliament, in the future.

Gonna close for now, and we can re-open once you circle back with the analysis. 

I'm gonna hit my specific goal with a custom check of the current PR

Thanks everyone!! 😉

PR #717, opened 2020-12-04
handle off-by-one errors in CFN line range calculation

mikeurbanski1:
This PR changes the file line range to include the resource name in the output for CFN yaml files.

Before:
```
		11 |     Type: 'AWS::S3::Bucket'
		12 |     Properties:
		13 |       BucketName: 'hello'
```

After:
```
		10 |   MyBucket:
		11 |     Type: 'AWS::S3::Bucket'
		12 |     Properties:
		13 |       BucketName: 'hello'
```

It also handles the case when there is not a newline at the end of a file, which caused the last line of a resource to be truncated.

It *also* handles blank lines between resources.

Finally, it adds tests for these cases.

### DIFF
--- a/checkov/cloudformation/context_parser.py
+++ b/checkov/cloudformation/context_parser.py
@@ -59,11 +59,32 @@ class ContextParser(object):
             start_line = min(find_lines_result_list)
             end_line = max(list(self.find_lines(cf_resource, ENDLINE)))
 
-            entity_lines_range = [start_line, end_line - 1]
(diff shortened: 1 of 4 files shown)


schosterbarak:
@tronxd you would probably want to take a look at this one

mikeurbanski1:
Also see https://github.com/bridgecrewio/platform/pull/2363

tronxd (approved):
GJ!

PR #720, opened 2020-12-07
MSK checks not pulling in nested values correctly

nomeelnoj:
We have had to customize these checks for our environment so I figured I would send them back upstream.

The objects for MSK are nested one level further than is currently being read in the checks.  Here are some examples of the objects as checkov sees them:

cloudwatch_logging: 
```
{'cloudwatch_logs': [{'enabled': [True], 'log_group': ['${aws_cloudwatch_log_group.msk[0].name}']}]}
```

encryption:
```
{'encryption_at_rest_kms_key_arn': ['${var.kms_key_id == "" ? module.kms.arn : var.kms_key_id}'], 'encryption_in_transit': [{'client_broker': ['TLS'], 'in_cluster': [True]}]}
```

As you can see from the data objects above, we need an additional `[0]` in order for the values to be compared properly


schosterbarak:
Thank you @nomeelnoj

Issue #724, opened 2020-12-08
setup.py very out of sync with Pipfile & pinning deps


robeden:
**Describe the bug**
`setup.py` pins all its dependencies against [general guidance](https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files):

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

Also, it was pinning old versions of many things such as boto3. I hit this when moving a tool to poetry, which had trouble resolving dependencies. When I finally managed to get it to do so, here are the upgrades that happened when I *removed* checkov as a dependency:
<img width="338" alt="Screen Shot 2020-12-08 at 2 27 43 PM" src="https://user-images.githubusercontent.com/1891321/101538032-21b4b000-3962-11eb-8f9d-352fb7655318.png">

**Expected behavior**
I believe `setup.py` should be more imprecisely defined (as Pipfile is) to allow consuming tools to allow library updates.

schosterbarak:
@robeden good catch. do you know of a way to automate the sync between the two?

robeden:
@schosterbarak - I looked a good bit and I think we could do it with a bit more work than I wanted to put in right this instant. [dephell](https://github.com/dephell/dephell) does conversion and would probably work great if we had a `pyproject.toml` file. I think poetry can manage it a bit better too, though I'm not 100% sure. I think a Github action step similar to the homebrew action would definitely be nice though.

schosterbarak:
@robeden thank you!

PR #725, opened 2020-12-08
Bring setup.py in line with Pipfile

robeden:
- Use un-pinned version specification in setup.py
- Remove some unused deps from setup.py
- Better version specification for a few deps (like `dpath`)

Fixes: #724


Pipfile.lock
```diff
@@ -405,27 +397,13 @@
             "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
             "version": "==2.10"
         },
-        "importlib-metadata": {
```
nimrodkor, line 408: @robeden It seems dropping this caused our UT to fail, but actual scanning (`integration-tests` job) completed successfully. Can you add it to the dev-packages in our Pipfile?

robeden: Thanks, I reverted. The lock didn't need to be updated.

nimrodkor (approved):

Issue #714, opened 2020-12-03
CKV_GIT_1 fails when github_repository visibility is "private"


jasonhuling:
**Describe the bug**
CKV_GIT_1 fails even though the `visibility` is set to "private".

**To Reproduce**
1. Define a github_repository with visibility = "private"
```
resource "github_repository" "this" {
  name = "my_repo"
  visibility = "private"
}
```
2. Run checkov

**Expected behavior**
This check should pass.

**Screenshots**
```
Check: CKV_GIT_1: "Ensure Repository is Private"
        FAILED for resource: github_repository.this
        File: /example/main.tf:21-24

                21 | resource "github_repository" "this" {
                22 |   name       = "my_repo"
                23 |   visibility = "private"
                24 | }
```

**Desktop (please complete the following information):**
 - OS: macOS 10.15.7
 - Checkov Version 1.0.645

**Additional Context**
It seems as though the `private` flag is depreciated which is why I am using `visibility`. If you set `private = true` you will get the following warning:
```
Warning: "private": [DEPRECATED] use visibility instead

  on main.tf line 21, in resource "github_repository" "this":
  21: resource "github_repository" "this" {
```

njgibbon:
Hello :) 

Thanks for the report!

I think I can take a look at this.

According to the Provider docs[0] currently 'private' is still supported though it does seem to be deprecated as you've evidenced. So, we just need to adjust the check to also check for 'visibility' is 'private' or 'internal' which would also be acceptable.

I should be able to fix this in the next few days.

[0]: https://www.terraform.io/docs/providers/github/r/repository.html

schosterbarak:
@jasonhuling thank you for reporting this. 
Thanks to @njgibbon  this is now resolved

PR #728, opened 2020-12-09
Fix - Check CKV_GIT_1 - Terraform - Github

njgibbon:
Hello again :-)

Check re-implementation to resolve issue: https://github.com/bridgecrewio/checkov/issues/714

Some new version of the github provider has added 'visibility' as a configuration option which overrides 'private'. 'private' is still a valid configuration but will be fully deprecated at some point in the future.

Provider docs:
https://www.terraform.io/docs/providers/github/r/repository.html

**License**


schosterbarak:
Thanks @njgibbon

PR #729, opened 2020-12-13
Fix - Issue with Terraform base resource value check class

njgibbon:
Hello (Sorry for the verbosity on this one but I think it's needed):

I was looking through issues and found this: https://github.com/bridgecrewio/checkov/issues/494

I re-tested expecting some false positive or maybe some minor issue with the check implementation.

I confirmed that the bug report was valid but the check and unit tests both looked fine.

Terraform S3 Versioning CKV_AWS_21 Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/S3Versioning.py

Tests:
https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/aws/test_S3Versioning.py

**Investigation**
With some functional testing I was able to narrow down the issue and focus in on one oddity. Along with the given scenario in the issue the following scenarios also cause the S3 Version check to pass.
```
resource "aws_s3_bucket" "test" {
  enabled = true
}
```

```
resource "aws_s3_bucket" "test" {
  foo {
    enabled = true
  }
}
```
```
resource "aws_s3_bucket" "test" {
  foo {
    bar {
      enabled = true
    }
  }
}
```
Where the following configuration is the only thing that should cause this to pass:
```
resource "aws_s3_bucket" "test" {
  versioning {
    enabled = true
  }
}
```

**Root Cause**
I narrowed down through some debugging and mutation that the issue is occurring in the terraform `base_resource_value_check` class, in `scan_resource_conf` where `if dpath.search(conf, inspected_key) != {}:` condition isn't met and so it goes to `# Look for the configuration in a bottom-up fashion`.

Here there is a semantic failure where if a partial match is made on the key we are searching for on the last child field i.e `enabled` for `versioning.enabled` would then match as positive in a check even if it only finds `enabled` or `foo.enabled` or `foo.bar.enabled`.

I have put in what seems to be a simple fix and tested it but here is where I _may_ be needing some help with a better implementation implementation because I have just waded into an area of the code base where I don't have a great understanding. Seems right though :-)

I have added several unit tests which may look odd under normal circumstances but are there to fully catch and ensure this issue isn't occurring.

**Action**
Because of the nature of this issue it isn't just related to S3 Versioning but can be causing false positives and false negatives all over the place just depending on the luck of the other configuration in the block where fields share a partial name.

I think this should be prioritised whenever possible when people are in. Either to take this change, or to work on a different change to achieve the same intention.

This will need some eyes. I have recorded all I can to provide all the context possible.

@schosterbarak  @nimrodkor 

**Ripple Effect**
See the unit test history on this PR. You can see that my new tests are failing and then with the next commit they will be solved but this then causes some other tests to fail so this needs to be investigated and the implementation adjusted.

From looking, these new failures appear to be things that were covered up by the previous faulty logic.

You can see from the file changes that all of these check changes were **obvious problems being hidden by the same root cause**.

**License**


nimrodkor (approved):
Thanks @njgibbon ! Looks solid to me.
However, would love another review on this from @tronxd before merging as this is a core piece

tronxd (approved):
LGTM, thanks @njgibbon for the contribution!

Issue #586, opened 2020-09-28
CKV_AWS_16 Encryption at rest Incorrect for Aurora 


jonjozwiak:
**Describe the bug**
When checking a CloudFormation template, CKV_AWS_16 flags encryption at rest not being enabled against an aurora instance.  However, according to docs for aurora this is set at DB cluster level:
 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-storageencrypted

This check needs to be updated to exclude the check for aurora or updated to also check the AWS::RDS::DBCluster for the appropriate StorageEncrypted setting.  

**To Reproduce**
Here is an example resource: 
		138 |     Type: AWS::RDS::DBInstance
		139 |     Properties:
		140 |       AllowMajorVersionUpgrade: false
		141 |       AutoMinorVersionUpgrade: true
		142 |       CopyTagsToSnapshot: true
		143 |       DBClusterIdentifier:
		144 |         Ref: MyDatabaseCluster
		145 |       DBInstanceIdentifier: MyInstance0
		146 |       DBInstanceClass: db.r5.large
		147 |       DBSubnetGroupName:
		148 |         Ref: GreenDataSubnetGroup
		149 |       Engine: aurora-mysql
		150 |       Tags:
		151 |       - Key: Name
		152 |         Value: MyInstance0
		153 |     DeletionPolicy: Delete

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - Checkov Version: 1.0.540

**Additional context**
Add any other context about the problem here (e.g. code snippets).

schosterbarak:
Thanks @jonjozwiak ! What's up 🙂.
Is it something you'd like to contribute?

nickgibbon-nationwide:
I think that this issue along with https://github.com/bridgecrewio/checkov/issues/139 can now be closed.

schosterbarak:
Thanks @nickgibbon-nationwide

njgibbon:
I think that this issue along with #139 can now be closed.

EDIT: diff account aha

PR #731, opened 2020-12-14
Fix - Check - CF - CKV_AWS_16 - Aurora Encryption

njgibbon:
Hello :-)

Issue: https://github.com/bridgecrewio/checkov/issues/586

Adjusting the implementation of CKV_AWS_16 in CloudFormation because Aurora instances should not fail because their encryption is configured differently. 

CF Doc Evidence: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-storageencrypted

This change is to stop the current erroneous failure.

**License**


nimrodkor (commented):
@njgibbon Thanks for the contribution!

That seems good, but I wouldn't skip the `BaseResourceValueCheck` so quickly. What do you think about this solution:
Keep the inheritance, and override the `scan_resource_conf` method. In the overriding method, implement the check for aurora as you implemented, and otherwise call the `super`'s `scan_resource_conf`
?

njgibbon:
Hey. Just stabilising the stuff I put in first.

Hmm, so that would be like overriding the method and then conditionally deciding not to use the overridden method in the method its self aha. I think I get it but is that really easier to understand? And in that case couldn't the same argument be made in almost all instances where `BaseResourceValueCheck` get's used?

I can absolutely have a go at it. It would be like a new pattern in (at least) this part of the codebase.

What do ya think?

nimrodkor:
@njgibbon Right. There's an example for it - check out `GoogleCloudDNSSECEnabled`
Yes, that argument can be made for most of the `BaseResourceValueCheck`s that have an exception! 👍

njgibbon:
@nimrodkor  - Cool i'll do that then.

njgibbon:
@nimrodkor Looks pretty clean and I'll similarily use `UNKNOWN` because it's better semantically because just because, yeah, technically it isn't a pass, it's just not a fail.

njgibbon:
@nimrodkor Alright, changed things around :-)

nimrodkor (approved):
GJ!

Issue #688, opened 2020-11-24
Harmonize Python versions


Porkepix:
**Describe the bug**
The Dockerfile is using python 3.8 for six months (https://github.com/bridgecrewio/checkov/commit/5d2d2e5f23ed4fdf3fa0df256563dcb89112d465) but everywhere else in the repository, python seems fixed to 3.7 version.

I started to change it to 3.8, but first I'm not sure you want to do it, and what does it imply for you. But anyway 3.8 seems to be working great looking at how it's used with the docker images for months. Also, 3.9 is out for now 4/5 months. What do you think about those versions?

robeden:
Good catch on the Dockerfile. Technically we only need 3.7 right now, though I did wish I had some functionality from 3.8 a bit ago.

metahertz:
@robeden do you see any downsides bumping everything in unison to 3.9? 
Most of our install methods (pipenv, docker, etc) are pretty self-contained anyway.

Thinking out loud incase i've missed something!

robeden:
We run checkov inside an AWS lambda, which doesn't currently support 3.9. So, I would hold back on that for the moment.

robeden:
One slightly related thing: locking from python 3.9 doesn't work right now due to this error when running tests: `ModuleNotFoundError: No module named 'importlib_metadata'. Not sure if updating to python 3.8 would fix it or not, but it would probably at least be a step in the right direction.

Porkepix:
PR https://github.com/bridgecrewio/checkov/pull/745#issuecomment-833226410 should resolve this issue? right?

njgibbon:
Hey @Porkepix  et al. Thanks for raising the conversation on this. I think this can close now with ref to the comments i've made here to summarise our state on this topic - https://github.com/bridgecrewio/checkov/pull/745#issuecomment-869174272

gruebel:
Closing with the reference to @njgibbon comment.

PR #745, opened 2020-12-22
chore(ci): use python 3.9

chenrui333:

The formulae is buildable with [python 3.9](https://github.com/Homebrew/homebrew-core/pull/67440), hence update the CI to pointing to the latest version.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -37,4 +37,4 @@ semantic_version = "*"
 packaging = "*"
 
 [requires]
-python_version = "3.7"
+python_version = ">=3.7"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: build
 
-on: 
+on:
   push:
     branches:
     - master
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
       - uses: dschep/install-pipenv-action@v1
       - name: Build & install checkov package
         run: |
@@ -56,10 +56,10 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@master
-      - name: Set up Python 3.7
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.9
       - uses: dschep/install-pipenv-action@v1
       - name: Install dependencies
         run: |
@@ -114,10 +114,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - name: Set up Python 3.7
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.9
       - uses: dschep/install-pipenv-action@v1
       - name: Install dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,10 +56,10 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@master
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v1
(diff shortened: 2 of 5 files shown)


schosterbarak:
i think that enforcing python 3.9 will increase the barrier of entrance for checkov users that are using python 3.8 or 3.7. 
would they still be able to install checkov for that version?
It's also defined that way on the setup.py https://github.com/bridgecrewio/checkov/blob/master/setup.py#L56

metahertz:
Linking this issue as @robeden had some concerns w/3.9 #688

chenrui333:
I don't see any problem of using the latest python (based on my usage as well as homebrew formulae), in fact the hybrid python versions would generally cause more confusions and some issues in the runtime.

If the end use do want to switch between different python versions, they can accomplish that by using something like `pyenv` and setup independent working environment.

Let me know if that makes sense.

robeden:
_Supporting_ 3.9 is great, but I wouldn't be able to run checkov at the moment if it _required_ it. Using it in the GH actions seems fine, but I'd rather not go there for the Pip requirements.

Pipfile
```diff
@@ -37,4 +37,4 @@ semantic_version = "*"
 packaging = "*"
 
 [requires]
-python_version = "3.7"
+python_version = "3.9"
```
robeden, line 40: Recommend: ">=3.7" or ">=3.8"

schosterbarak: @chenrui333  i agree with Rob here. can you please modify that?

schosterbarak:
Closing due to inactivity

Pipfile
chenrui333: I see, somehow this got lost in my box.

chenrui333: @schosterbarak can you reopen the PR and I can add the change?

schosterbarak: @chenrui333 sure. you might need to rebase due conflicts

chenrui333:
doing the change now.

Pipfile
chenrui333: absolutely.

antonblr:
If you don't want stick with particular python version, you could use job matrix for tests, something like:

```yaml
jobs:
  integration-tests:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ['3.7', '3.8', '3.9']     
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}
   ...
```

nimrodkor:
@chenrui333 Can we see this PR as resolved, as today checkov supports 3.7, 3.8 and 3.9?

njgibbon:
Hello,

I think we should be able to have a satisying closure on this given the work done and state of things as summarised below:

1. Given the matrix testing implemented this year Checkov can confidently say it supports use with Python 3.7, 3.8 and 3.9.

2. The setup.py file indicates this which is what the pip tool uses when downloading checkov from PyPi.
```
python_requires=">=3.7"
```
(Note there is no "<=3.9" because python 3.10 is not publicly released yet. And therefor this statement is semantically correct as checkov supports all versions after 3.7 After python 3.10 is released it  would make sense to test compatibility and such and perhaps change the setup.py version in some way)

3. The pipfile syntax is set as follows: 
```
[requires]
python_version = "3.7"
```
This is used if creating a pyenv environment.
The pipfile does not support greater than or less than suntax. You can set the python_version to '3' to indicate all versions of 3 are supported but this would not be accurate, as we do only ensure we support 3.7, 3.8 and 3.9. So, the above is correct. It makes sense to use the lowest version you support for development because there is less chance you will use some non-backwards compatible dependency. As we do matrix testing, an argument could still be made for using the latest you support (3.9) because anything backwards compatible would be caught by that. Anyway, the core point here is what currently there is makes sense and >= syntax is not supported here.

I encourage anyone interested in this to read this thread and specifically this comment to get a good grasp of this: https://github.com/pypa/pipenv/issues/1050#issuecomment-361485919

In looking into this I found one little thing - https://github.com/bridgecrewio/checkov/pull/1331

PR #749, opened 2020-12-24
Feature: Skip Pattern check

soufiane444:
Add a feature to sip checks using a pattern (eg --skip-pattern CKV_AWS* to skip all aws check). 
Useful when using custom policy to skip others checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,11 @@ Run all checks except 1 specified:
 checkov -d . --skip-check CKV_AWS_52
 ```
 
+Run all checks except specified patterns:
+```sh
+checkov -d . --skip-pattern CKV_AWS*
+```
+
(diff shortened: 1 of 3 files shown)


schosterbarak:
@soufiane444 great contribution. What do you think about adding this capability to the flag `--skip-check` vs creating a new flag? 

I prefer 1 flag that will support both scenarios if possible.

soufiane444:
> @soufiane444 great contribution. What do you think about adding this capability to the flag `--skip-check` vs creating a new flag?
> 
> I prefer 1 flag that will support both scenarios if possible.

Thanks.
I will do this in a new PR and close this one.

soufiane444:
Here is the new PR : https://github.com/bridgecrewio/checkov/pull/750

PR #750, opened 2020-12-28
skip check using pattern (wildcards)

soufiane444:
Add the ability to skip check using a pattern. (eg --skip-check CKV_AWS* to skip all aws check).
You can also use a deny list (--skip-check ) like before but with wildcards now like this --skip-check CKV_AWS*,CKV_AZURE*

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,11 @@ Run all checks except 1 specified:
 checkov -d . --skip-check CKV_AWS_52
 ```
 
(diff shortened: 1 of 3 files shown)


schosterbarak:
Thank you! can you please add a unit test?

schosterbarak:
Thank you @soufiane444 !
You are welcome to join: 
https://slack.bridgecrew.io/

Would love to talk more about checkov there :)

PR #758, opened 2020-12-31
Add evaluated keys to all terraform

arielkru:


checkov/common/checks/base_check.py
```diff
@@ -18,6 +19,7 @@ def __init__(self, name, id, categories, supported_entities, block_type):
         self.block_type = block_type
         self.supported_entities = supported_entities
         self.logger = logging.getLogger("{}".format(self.__module__))
+        self.evaluated_keys = []
```
robeden, line 22: Could some documentation be added to explain the expectations with this change? Describing the purpose of `evaluated_keys` here or the `get_evaluated_keys` function would be helpful. For one thing: is the syntax assumed to be `dpath` syntax? 
What is the impact if it's omitted (particularly for custom checks). I assume it's just not reported?

Also, documentation on the scan_resource_conf function (or some higher function?) would be helpful to let check implementors know it's something they're expected to update.

checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
```diff
@@ -18,8 +18,10 @@ def scan_resource_conf(self, conf):
         :param conf: aws_kms_key configuration
         :return: <CheckResult>
         """
+        self.evaluated_keys = []
```
robeden, line 21: This is done on all checks and seems redundant since it's done on the base. Is it necessary?

checkov/terraform/checks/resource/base_resource_value_check.py
```diff
@@ -94,3 +95,6 @@ def get_expected_value(self):
         Returns the default expected value, governed by provider best practices
         """
         return True
+
+    def get_evaluated_keys(self):
```
robeden, line 99: Is this necessary? I think it should pick the function up from base_check.

checkov/common/checks/base_check.py
arielkru: Hi @robeden , thanks for your comments and sorry for delay!

1. `self.evaluated_keys` should be set for checks that inherit directly from `BaseCheck`, and don't have today their paths of inspected keys, mostly since the checks are more dynamic than the ones that inherit from its child classes.
2. I've added an explanation under the `get_evaluated_keys` and `scan_resource_conf` functions, PTAL and LMK if it's clear enough.
3. If it's omitted the report would get the value of `self.inspected_keys`, so if a previous check in the same run set the value to certain path, and the current check didn't override it, the path would be wrong. That's why I've inserted the `self.evaluated_keys = []` in some of the checks, so if they don't have a relevant key their report doesn't show the previous value.

@nimrodkor Maybe `get_inspected_keys` could use the new `self.evaluated_keys` in `BaseCheck` so its simpler, I'm not sure we need both separated

checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
arielkru: Explained this on my third answer in previous comment, LMK what you think

checkov/terraform/checks/resource/base_resource_value_check.py
arielkru: They don't get the same values, here it takes `get_inspected_key()` and in base_check it takes `self.evaluated_keys`.
Anyway @nimrodkor this is relevant to my question in the first comment

checkov/common/checks/base_check.py
robeden: Documentation looks good. Thanks.

checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
nimrodkor: @arielkru I agree with @robeden . Let's put it in `BaseCheck`'s `run` method. Check out line 39

checkov/terraform/checks/resource/base_resource_value_check.py
nimrodkor: I think this is a nice shortcut to keep the implementation of `BaseResourceValueCheck` as simple as it gets. We know what the attribute is - no need to look for it and generate it in the scan

nimrodkor (commented):
Can we implement for `BaseResourceNegativeValueCheck` the same way we implemented for `BaseResourceValueCheck`?

checkov/terraform/checks/resource/aws/ALBListenerHTTPS.py
arielkru: I removed these from the checks and added it to base_check, @robeden thanks for the great input :)
 @nimrodkor PTAL

checkov/terraform/checks/resource/aws/ElasticsearchNodeToNodeEncryption.py
```diff
@@ -17,6 +17,7 @@ def scan_resource_conf(self, conf):
         :param conf: aws_elasticsearch_domain configuration
         :return: <CheckResult>
         """
+        self.evaluated_key = []
```
nimrodkor, line 20: @arielkru Unnecessary I believe

nimrodkor (approved):
@arielkru GJ!

PR #760, opened 2021-01-04
fix - tfplan - parsing where there are resources in root_module without values key

njgibbon:
**Description**
I have found an error when using the latest version of checkov on a terraform plan output.

If the terraform plan has elements in the root_module.resources which do not have 'values' then this error is output.
```
File "/usr/local/bin/checkov", line 5, in <module>
run()
File "/usr/local/lib/python3.6/site-packages/checkov/main.py", line 79, in run
guidelines=guidelines)
File "/usr/local/lib/python3.6/site-packages/checkov/common/runners/runner_registry.py", line 33, in run
runner_filter=self.runner_filter, collect_skip_comments=collect_skip_comments)
File "/usr/local/lib/python3.6/site-packages/checkov/terraform/plan_runner.py", line 60, in run
tf_definitions, template_lines = parse_tf_plan(file)
File "/usr/local/lib/python3.6/site-packages/checkov/terraform/plan_parser.py", line 68, in parse_tf_plan
resource_block[resource['type']][resource.get('name', "default")] = _hclify(resource['values'])
KeyError: 'values'
```

You can see that this case is already checked in the child_module part of the tf parser. This is just extending that logic backwards to fix my issue.

**Replication**
After some investigation into the specifics of the issue. This scenario can be replicated when using a data block in the root module where the values are 'known after apply'.... I've reduced the issue to this sort of config as a minimal example
```
# valid dummy data input
resource "aws_eks_cluster" "cluster" {
  name = "test"
  enabled_cluster_log_types = ["audit", "api"]
  role_arn = "arn:aws:iam::204481888888:role/test"
  vpc_config {
    subnet_ids = ["10.165.77.0/24"]
  }
}

data "aws_eks_cluster" "cluster" {
  name = aws_eks_cluster.cluster.id
}
```

```
# Do tf plan check
# 1st sort out aws config as partial configuration
terraform init
terraform plan --out tfplan.binary
terraform show -json tfplan.binary > tfplan.json
checkov -f tfplan.json
```
The above method was also used to generate data to use in testing.

**License**


schosterbarak:
Great one @njgibbon !
Can you add a test with a test file at https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/runner/test_plan_runner.py ?

njgibbon:
@schosterbarak  - Ah yes, didn't see where they were. I will write the test :)

schosterbarak:
Sweet! 
The current plan file is at https://github.com/bridgecrewio/checkov/tree/master/tests/terraform/runner/resources/plan
I think it would be easiest to add one right next to it

njgibbon:
@schosterbarak - Just working out the test data to use to replicate. It's an odd one. Terraform has to be a module in a module where the first module has an output but no standalone resources or something like that. So a bit of a weird one aha. Will probably finish tomorrow.

njgibbon:
@schosterbarak  - I've implemented the right testing and I've updated the description of the issue to properly reflect the root cause :)

tests/terraform/runner/resources/plan_root_module_resources_no_values/tfplan.json
```diff
@@ -0,0 +1,201 @@
...
+                    "constant_value":[
+                       "audit",
+                       "api"
+                    ]
+                 },
+                 "name":{
+                    "constant_value":"test"
+                 },
+                 "role_arn":{
+                    "constant_value":"arn:aws:iam::204481888888:role/test"
```
schosterbarak, line 169: Can you switch the account I'd with the value 12345678912 ?

njgibbon: Just replaced dummy account # in all places.

nimrodkor: Squashing this PR will leave the old commits as garbage, to be collected in the coming weeks by github 😄

schosterbarak:
Thank you @njgibbon

PR #768, opened 2021-01-06
fix - tfplan json scan of nested modules

njgibbon:
Hello again 🙂

Continuing down the line of enquiry from yesterday where we closed: https://github.com/bridgecrewio/checkov/pull/760

When running latest checkov it fails to find resources which are nested child_modules. My use case is a terraform module which is in turn made up of other versioned terraform modules.

This is because the parser checks all root_modules and then all child_modules and stops there. So it only goes one-level deep.

My use-case requires 2 levels deep. But I have implemented the search so that it will find all (if any) nested child_modules.

**License**


rotemavni (approved):
Thank you!

PR #770, opened 2021-01-06
fix - tfplan - rare - data block with partial values scanned to cause false negatives

njgibbon:
Hello again,

Hopefully this will conclude the journey I have been on this week which has raised and fixed:

https://github.com/bridgecrewio/checkov/pull/760

https://github.com/bridgecrewio/checkov/pull/768

For context I am debugging so that I can include terraform plan-time scanning into a pipeline. I am working on a _big_ complex set of terraform which includes many versioned child modules at one level and makes use of the aws, helm and kubernetes providers.

**Root Cause Description**
In rare circumstances a data resource with partial values in the plan could cause false negatives. 
Often 'data' does not even appear in the *_modules[x].resouces field within planned_values and is not scanned as expected.
It can occur when tf module B depends on tf module A.
And tf module A creates a resource that is used in a data block in tf module B.
So some values can be known but other are not at plan time.
This can cause the data block resource to be scanned as if it were a managed resource which is not configured correctly.
See 'Modes': https://www.terraform.io/docs/internals/json-format.html#values-representation.

In my implementation I have also used another helper method to reduce some duplication.

Being able to properly understand and reproduce the data for testing was the challenge here. And i'm satisfied that all regression tests are passing.

**Testing**
This test verifies that such a circumstance stops occurring.
There is a EKS Managed Resource and a EKS Data Resource.
The EKS Managed Resource should have 4 failures corresponding with EKS checks.
The EKS Data Resource should not be scanned. Previously this would cause 8 failures. 

Because the specifics of this are somewhat obscure and people will not be able to intuitively understand what is going on I have been very explicit in the 'why' in terms of comments in the unit testing added. This should make it more maintainable for the future.

**License**

### DIFF
--- a/checkov/terraform/plan_parser.py
+++ b/checkov/terraform/plan_parser.py
@@ -49,6 +49,25 @@ def _hclify(obj):
             ret_dict[key] = [child_dict]
     return ret_dict
 
+def _prepare_resource_block(resource):
+    """
+    hclify resource if pre-conditions met.
+    :type: resource: dict: tf resource block
+    :rtype: resource_block: dict: hclifyed if conditions met
+    :rtype: prepared: boolean: whether conditions met to prepare data
+    """
+    resource_block = {}
+    resource_block[resource['type']] = {}
+    prepared = False
+    mode = ""
+    if 'mode' in resource:
+        mode = resource.get("mode")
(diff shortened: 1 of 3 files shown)


schosterbarak (approved):
Good catch! @njgibbon

PR #780, opened 2021-01-12
Handle paths starting with / as local

nimrodkor:

I've ran into many failures while trying to load modules with an absolute path. I'm getting this error:
```
Unable to load module (source="/path/to/module" version="latest"): Invalid URL '/path/to/module': No schema supplied. Perhaps you meant http:///path/to/module/versions?
```
This should fix it

@robeden @rotemavni Would love your feedback!

### DIFF
--- a/checkov/terraform/module_loading/loaders/local_path_loader.py
+++ b/checkov/terraform/module_loading/loaders/local_path_loader.py
@@ -1,7 +1,7 @@
 import os
 
-from checkov.terraform.module_loading.loader import ModuleLoader
 from checkov.terraform.module_loading.content import ModuleContent
+from checkov.terraform.module_loading.loader import ModuleLoader
 
 
 class LocalPathLoader(ModuleLoader):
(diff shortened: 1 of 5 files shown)


schosterbarak:
test please

robeden:
So, I assume the TF CLI handles them that way? The [docs](https://www.terraform.io/docs/modules/sources.html#local-paths) are clear about “./“ and “../“ but don’t mention “/“. Certainly the “right” thing is to prefix with a dot, but I have no problem with Checkov being lenient due to the nature of the tool.

nimrodkor:
Playing with it a bit more, it wasn't loading the actual module via the local loader. So I tried another approach and wrote a test @schosterbarak 
@robeden `/` means you start out at the root dir (absolute path so to speak), which is different then `./`. It isn't noted in the docs, but it works 🤷

robeden:
> `/` means you start out at the root dir (absolute path so to speak)

That doesn't sound like a good practice... but okay. 🤷

checkov/terraform/module_loading/loaders/local_path_loader.py
```diff
@@ -10,7 +10,8 @@ def __init__(self) -> None:
         self.is_external = False
 
     def _is_matching_loader(self) -> bool:
-        return self.module_source.startswith("./") or self.module_source.startswith("../") or self.module_source.startswith(self.current_dir)
+        return self.module_source.startswith("./") or self.module_source.startswith("../") \
+               or self.module_source.startswith(self.current_dir) or self.module_source.startswith('/')
```
robeden, line 14: `self.module_source.startswith(self.current_dir)`

This might cause problems. If the current directory is called `s3` or `https` or something like that, it would conflict with other loaders. What case is that condition trying to cover?

robeden: Sorry, `current_dir` might be the full path? If so, ignore me.

PR #785, opened 2021-01-13
Helm v3 support and deps checking for external binaries.

metahertz:
**HELMv3 (Kubernetes templating) support for Checkov.**

HELM is written in go, however has a simple `helm template` function to produce a non-k8s (local only) rendering of the templates into kubernetes manifests which we can pass to the existing k8s runner.

- Also passing the chart name, to the runner as a new arg, which defaults to none. 
- When writing out the report items, the k8s_runner also now adds the chart to the `Record` object if the k8s_runner was called with a chart name. (Discussion point: There are no other runner specific fields in the `Record` object, but ideally want to surface this information for future use with the output).

On top of this, as the helm binary has no python-friendly bindings, so we're using  p.subprocess to exec out, which leads to external dependancies. My aim here was to ensure any runners that need to announce external dependancies get automatically checked, and excluded from the checkov run if they are missing on that system (in order to keep consistent the current behaviour of existing CI/Scripts/users who use the default of `--framework all`. )

 - To do this, i've introduced `--skip-framework`, and an optional `check_system_deps` method in a runner class. If the deps arent satisfied for a given runner, the runner is added to `--skip-framework`, along with any user-provided skipped frameworks. A resultant list of frameworks is generated from these (starting from either "all" for the default `--framework` arg, or a subset of frameworks passed in.
 
To support calling the check_system_deps and generating a resultant list of "active" frameworks for a given run, i'm listing the runner names and framework names in `main.py`:

```
checkov_runner_module_names = ['cfn', 'tf', 'k8', 'sls', 'arm', 'tf_plan', 'helm']
checkov_runners = ['cloudformation', 'terraform', 'kubernetes', 'serverless', 'arm', 'terraform_plan', 'helm']
```

As a completely important aside, since we're 1-1 mapping frameworks to runners at this point, may it be worth unifying the framework name and runner names?

**Very much open to suggestions, comments and improvements, especially on suitability of the `Record` updates and the new deps checking/handling, as well as less hacky solutions than calling out to `helm template`**

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -19,6 +19,7 @@ from checkov.common.util.banner import banner as checkov_banner
 from checkov.common.util.consts import DEFAULT_EXTERNAL_MODULES_DIR
 from checkov.common.util.docs_generator import print_checks
 from checkov.common.util.type_forcers import convert_str_to_bool
+from checkov.common.util.runner_dependency_handler import RunnerDependencyHandler
 from checkov.kubernetes.runner import Runner as k8_runner
 from checkov.logging_init import init as logging_init
 from checkov.runner_filter import RunnerFilter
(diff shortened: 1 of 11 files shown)


metahertz:
Removing addition of Helm chart details to the `Record`, as adding `helm_chart` to each runner unnecessarily is nasty.

checkov/main.py
```diff
@@ -22,27 +22,49 @@
...
+    except:
+        continue
+    
+    if globals()[f"{runner}_runner"]().system_deps:
+            result = globals()[f"{runner}_runner"]().check_system_deps()
+            if result is not None:
+                checkov_frameworks_unmatched_deps.append(result)
 
-def run(banner=checkov_banner, argv=sys.argv[1:]):
+def run(banner=checkov_banner, runners=checkov_runners, argv=sys.argv[1:]):
```
schosterbarak, line 47: you can access checkov_runners directly - it will keep the method signature :)

checkov/runner_filter.py
```diff
@@ -8,7 +9,7 @@ class RunnerFilter(object):
     #       logically a "static" concept anyway, so this makes logical sense.
     __EXTERNAL_CHECK_IDS = set()
 
-    def __init__(self, framework='all', checks=None, skip_checks=None, download_external_modules=False, external_modules_download_path=DEFAULT_EXTERNAL_MODULES_DIR, evaluate_variables=True):
+    def __init__(self, framework='all', skip_framework=None, checks=None, skip_checks=None, download_external_modules=False, external_modules_download_path=DEFAULT_EXTERNAL_MODULES_DIR, evaluate_variables=True, runners=None):
```
schosterbarak, line 12: add new params at the end - that way it's backward compatible

checkov/main.py
```diff
@@ -22,27 +22,49 @@
...
 
 outer_registry = None
 
 logging_init()
+checkov_runner_module_names = ['cfn', 'tf', 'k8', 'sls', 'arm', 'tf_plan', 'helm']
+checkov_runners = ['cloudformation', 'terraform', 'kubernetes', 'serverless', 'arm', 'terraform_plan', 'helm']
 
+#Check runners for system deps, append --skip-framework list for failing deps.
+checkov_frameworks_unmatched_deps = []
+for runner in checkov_runner_module_names:
```
schosterbarak, line 36: maybe add those to a python file - something like "runner_dependency_handler" with a short pydoc explaining how to use/ add

checkov/main.py
metahertz: Thanks, and good catch, wasnt using `runners` was left behind from a refactor :) - Fixed.

checkov/runner_filter.py
metahertz: Fixed, thanks!

checkov/main.py
metahertz: Todo

schosterbarak:
@metahertz pull from master should solve the error in integration tests

checkov/main.py
metahertz: Moved this to a little class as well as the logic that appends the args.skipped_framework, keeps `main.py` cleaner as discussed. also added pydoc.

nimrodkor (approved):
LGTM!
Would appreciate some testing on that runner filter changes :)

metahertz:
Woooo! passed checks!

metahertz:
Happy to merge @schosterbarak ?

PR #788, opened 2021-01-17
Handle tf file parsing errors

mikeurbanski1:
Handles Terraform parsing files when using `-f`. It already worked correctly at the directory level.

Also makes `report.parsing_errors` a map so that it stores the files and the error message.

Finally, adds them to the console output (previously only available with `-o json`). @schosterbarak maybe you intentionally left this out. I think we should show it. But another option would be to add a message like "use -o json to see the parsing errors", like this:

```
terraform scan results:

Passed checks: 0, Failed checks: 0, Skipped checks: 0, Parsing errors: 2 (use -o json to view all parsing errors)
```

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -21,13 +21,13 @@ class Report:
         self.skipped_checks = []
         self.parsing_errors = []
 
-    def add_parsing_errors(self, files):
-        for file in files:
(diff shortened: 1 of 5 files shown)


schosterbarak:
Looks great to me. Thank you :)

PR #796, opened 2021-01-19
Handle cases when soft_delete_enabled is not set or is false

sirlatrom:
As per https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-change, the attribute will no longer have any effect on new Key Vaults.
Furthermore, soft_delete_enabled now defaults to true as per the provider breaking changes: https://github.com/terraform-providers/terraform-provider-azurerm/releases/tag/v2.42.0.


schosterbarak:
Thank you @sirlatrom

PR #797, opened 2021-01-20
Allow TF module download tests to pass locally without github creds

metahertz:
Two tests check that a nonexistant github repo as a module resource doesn't interfere with the correct rendering of the output and correctly caches the non working url.

These tests when run by a developer locally ie `pipenv run pytest` prompt on the CLI for local creds.
As the repo is going to error anyway (it doesnt exist), i've embedded fake creds to allow the run to proceed automatically, internally the git client will be handling an auth error vs repo not existing, but I dont think that matters and gives a better testing experience for checkov contributors.

Also updated the names of the nonexistant repo's for readability.

cc/ @JamesWoolfenden 

### DIFF
--- a/checkov/common/output/record.py
+++ b/checkov/common/output/record.py
@@ -20,7 +20,7 @@ class Record:
     guideline = None
 
     def __init__(self, check_id, check_name, check_result, code_block, file_path, file_line_range, resource,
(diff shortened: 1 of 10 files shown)


JamesWoolfenden:
I think we can close this now @metahertz ?

PR #801, opened 2021-01-21
Add working dir relative path

mikeurbanski1:
The purpose of this PR is to add a relative path from the CWD into each record object. Generally, this will be the path from the repo root. However, this value may differ from the normal record `file_path` object, which is taken from the common prefix, not necessarily the working directory of the scan. The result is that the file_path of a given `-f` param in a scan is arbitrary compared to the value when you run using `-d .`

This is needed to guarantee a file path that matches the paths coming from suppressions, etc in the platform API responses.

I accidentally branched this off of the dev-api branch, so it has a few commits. They were isolated, so I just reverted them manually.


schosterbarak (approved):
LGTM

PR #806, opened 2021-01-21
Add compact CLI output option

julienduchesne:
Otherwise, there's just too much stuff with the code blocks

Here's what it looks like:

image (1)


metahertz:
Looks awesome @julienduchesne and thanks for the contribution! Would you mind keeping the banner? Still very neat and readable but also more "findable" in a large amount of CI logs/terminal output and keeps the project's branding across CLI output consistent :)

metahertz:
Thanks @julienduchesne, a very cool and useful contribution, will be using `--compact` myself!

PR #814, opened 2021-01-24
Add integration with Bridgecrew developer APIs

mikeurbanski1:
Enhanced API integration to get suppressions and generate fixes from the BC platform, with an API token.

TODO in future PRs:
- unit tests for fixes (I tested various scenarios and they seem to work well)
- determine how we want to show fixes in normal CLI output
- add some logging if we encounter problematic bits


mikeurbanski1:
cfn-lint issue is invalid https://github.com/aws-cloudformation/cfn-python-lint/issues/1577

checkov/cloudformation/runner.py
```diff
@@ -75,14 +76,40 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+                                                file_line_range=entity_lines_range, resource=resource_id,
+                                                evaluations=variable_evaluations,check_class=check.__class__.__module__,
+                                                file_abs_path=file_abs_path, entity_tags=tags)
                                 report.add_record(record=record)
-        return report
\ No newline at end of file
+        return report
+
+    @staticmethod
+    def get_resource_tags(entity):
```
schosterbarak, line 93: can we move both of those static methods to a new file?

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -197,17 +245,46 @@ def get_id_mapping(self) -> dict:
...
             response = json.loads(request.data.decode("utf8"))
             self.guidelines = response["guidelines"]
             self.bc_id_mapping = response.get("idMapping")
+            self.ckv_to_bc_id_mapping = {ckv_id: bc_id for (bc_id, ckv_id) in self.bc_id_mapping.items()}
             logging.debug(f"Got checkov mappings from Bridgecrew BE")
         except Exception as e:
             logging.debug(f"Failed to get the guidelines from {self.guidelines_api_url}, error:\n{e}")
             return {}
 
+    def get_platform_fixes(self, scan_results):
```
schosterbarak, line 265: let's move fixes flow to a new python class.
1st python class will handle vscode integration
2ns python class will handle CI integration

wdyt?

checkov/main.py
```diff
@@ -131,6 +149,13 @@ def add_parser_args(parser):
     parser.add_argument('-b', '--branch',
                         help="Selected branch of the persisted repository. Only has effect when using the --bc-api-key flag",
                         default='master')
+    parser.add_argument('--skip-fixes',
```
schosterbarak, line 152: maybe we should add `bc` prefix. since this is only possible using API integration?

mikeurbanski1: We don't do that for repo and branch today, though.

mikeurbanski1: I don't think it's necessary, with the argument clearly documented (along with extra docs that we'll have in the readme and elsewhere).

checkov/cloudformation/runner.py
mikeurbanski1: We could, but it would still be a CFN-specific logic, which is why I put them here. (In other words, getting CFN resource tags looks different from Terraform). And it seems to be like this is logic that is runner-centric.

Were you thinking just a utils file inside the CFN directory?

checkov/common/bridgecrew/platform_integration.py
mikeurbanski1: It's 99% independent of vscode. The only vscode specific logic is to skip the S3 stuff. Suppressions and fixes are just pure checkov features.

We should probably refactor all of these flows into new classes:

- dev API integrations (suppressions and fixes)
- old scanning logic (upload to S3)
- signup flow

And use the platform integration class just to hold the relevant attributes.

I'll try out refactoring just the dev API stuff to see how it looks. Maybe a "platform integration" class with abstract methods for setup, pre-scan, and post-scan logic? And the implementation classes just pull relevant state attributes from the platform_integration class.

mikeurbanski1: I used the check registry model to create a "platform integration registry". This made it trivial to convert the logic over.

It should work well to convert the S3 logic over as well, possible with more integration hooks (e.g., pre_scan). The only one I'm not sure of is the guidelines integration, because that one is a dependency for the runners. I am not sure what the best mechanism is to fetch guidelines and then make them available everywhere.

PR #815, opened 2021-01-25
Check - CF - CKV AWS 89 - Inline with TF

njgibbon:
Hey 🙂 ,

Writing the CloudFormation version of this check.

A key difference here is that in Terraform the default is false which results in a pass. But in CloudFormation the default is true which results in a fail.

I have also added a test case to the terraform check to test with the default where the field is not present as I think this is always good when working with checks on optional fields. 

CloudFormation Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dms-replicationinstance.html

Terraform version: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/DMSReplicationInstancePubliclyAccessible.py

**License**


schosterbarak:
Thanks @njgibbon

Issue #816, opened 2021-01-26
TF parser: support partial line resolution


robeden:
Kind of a bug, kind of a feature request...

Right now the TF parser won't pull out a var block (from `_find_var_blocks`) when it has unevaluated pieces inside it. For example: `${merge(local.common_tags,{'Name': 'my-bucket-${var.ENV}'})}` will not resolve while `${var.ENV}` is unresolved.

The problem comes up with this test case:
```
locals {
  common_tags = {
    Tag1 = "one"
    Tag2 = "two"
  }
}

variable "ENV" {}

resource "aws_s3_bucket" "bucket" {
  # var.ENV has no default, so need to evaluate the merge without the
  # fully resolved statement.
  tags = merge(local.common_tags, {Name = "my-bucket-${var.ENV}"})
}
```

Because `var.ENV` won't resolve, nothing resolves, which throws away a lot of data. In our case, this issue is causing necessary tags to be missed.

I'm not completely sure what the algorithm should be yet, but it should definitely allow for incomplete resolution where possible. Final expected result is:

```
{
  "main.tf": {
    "locals": [
      {
        "common_tags": [
          {
            "Tag1": "one",
            "Tag2": "two"
          }
        ]
      }
    ],
    "variable": [
      {
        "ENV": {}
      }
    ],
    "resource": [
      {
        "aws_s3_bucket": {
          "bucket": {
            "tags": [
              {
                "Tag1": "one",
                "Tag2": "two",
                "Name": "my-bucket-${var.ENV}"
              }
            ]
          }
        }
      }
    ]
  }
}
```

robeden:
I'm working on it...

PR #825, opened 2021-01-27
Allow nested, unresolved variables

robeden:
This is a fix for #816 which reworks the main logic which pulls out variables from a value (`find_var_blocks`). The new logic allows nested evaluations, returning the various pieces in order. In a nutshell, this allows outer evals to still occur even if inner ones don't. See the large comment in `process_items_helper` for more info.

In addition, that logic was pulled out of `parser.py`, into `parser_var_blocks.py`... just for organizational purposes.

Fixes: #816


nimrodkor (approved):
Nice and clean!

PR #826, opened 2021-01-28
Check Listener is using TLS 1.2

JamesWoolfenden:
This check examines an ALB listener, if it's using HTTPS it checks that it is using a policy that mandates the use of TLS 1.2.
Thats means policies starting:
ELBSecurityPolicy-FS-1-2 or ELBSecurityPolicy-TLS-1-2
See the policies here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html

### DIFF
--- a/checkov/terraform/checks/resource/aws/AppLoadBalancerTLS12.py
+++ b/checkov/terraform/checks/resource/aws/AppLoadBalancerTLS12.py
@@ -1,0 +1,33 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class AppLoadBalancerTLS12(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure that load balancer is using TLS 1.2"
+        id = "CKV_AWS_103"
+        supported_resources = ['aws_lb_listener']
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/aws/AppLoadBalancerTLS12.py
```diff
@@ -0,0 +1,33 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class AppLoadBalancerTLS12(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure that load balancer is using TLS 1.2"
```
schosterbarak, line 7: btw, why not 1.3? 1.2 had the heartbleed vulnerability as far as i remmember

JamesWoolfenden: i was going off their list https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html. I double checked on the UI there's no 1.3 option

schosterbarak: pff. I hoped AWS handled that. 
Thanks for verifying this one. merging :)

PR #827, opened 2021-01-28
add read to ci env variables for platform integration

schosterbarak:
add read to ci env variables for platform integration and send to the API in case of platform integration


checkov/common/bridgecrew/platform_integration.py
```diff
@@ -94,6 +93,7 @@ def setup_bridgecrew_credentials(self, bc_api_key, repo_id, skip_fixes=False, sk
 
         if self.bc_source != 'vscode':
             try:
+                self.skip_fixes = True  # no need to run fixes on CI integration
```
schosterbarak, line 96: @mikeurbanski1 FYI

PR #829, opened 2021-01-31
fix local modules path

schosterbarak:


checkov/terraform/runner.py
```diff
@@ -152,7 +152,7 @@ def run_block(self, entities,
                                 code_block=entity_code_lines, file_path=scanned_file,
                                 file_line_range=entity_lines_range,
                                 resource=entity_id, evaluations=entity_evaluations,
-                                check_class=check.__class__.__module__, file_abs_path=full_file_path, entity_tags=tags)
+                                check_class=check.__class__.__module__, file_abs_path=absolut_scanned_file_path, entity_tags=tags)
```
schosterbarak, line 155: @robeden FYI for visibility. Before that change, the full_file_path had the unstripped value - which caused an illegal file path on downstream project of ours for local modules.

robeden: Thanks for the heads up. Looks good to me.

PR #832, opened 2021-02-01
CF - Check - CKV_AWS_46 - EC2Credentials

njgibbon:
Hello 🙂 

Implementing this check as already done for terraform.

I have used the same key test data as is already in-use in the terraform check to verify the same outcome.

Also changed minor typo so that text is the same for both terraform and clloudformation checks.

Terraform Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/EC2Credentials.py

CloudFormation Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html

Examples of CF UserData (to verify formatting): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-ec2.html
https://gist.github.com/jeffbrl/abfd84d8d8586495632ba9b8489322e2
https://stackoverflow.com/questions/48197526/using-userdata-in-cloudformation

**License**

### DIFF
--- a/tests/cloudformation/checks/resource/aws/example_EC2Credentials/EC2Credentials-FAILED.yaml
+++ b/tests/cloudformation/checks/resource/aws/example_EC2Credentials/EC2Credentials-FAILED.yaml
@@ -3,7 +3,7 @@ Resources:
   Resource0:
     Type: AWS::EC2::Instance
     Properties: 
-      ImageID: ami-04169656fea786776
+      ImageId: ami-04169656fea786776
       UserData:
         Fn::Base64:
           !Sub |
--- a/tests/cloudformation/checks/resource/aws/example_EC2Credentials/EC2Credentials-FAILED.yaml
+++ b/tests/cloudformation/checks/resource/aws/example_EC2Credentials/EC2Credentials-FAILED.yaml
@@ -1,0 +1,18 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Resources:
+  Resource0:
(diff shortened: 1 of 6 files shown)


tests/cloudformation/checks/resource/aws/example_EC2Credentials/EC2Credentials-FAILED.yaml
```diff
@@ -0,0 +1,19 @@
...
+      UserData:
+        Fn::Base64:
+          !Sub |
+            #! /bin/bash
+            sudo apt-get update
+            sudo apt-get install -y apache2
+            sudo systemctl start apache2
+            sudo systemctl enable apache2
+            export AWS_ACCESS_KEY_ID
+            export AWS_ACCESS_KEY_ID=AKIAIOSFODNN7EXAMPLE
```
schosterbarak, line 16: Thanks @njgibbon . can you add this file to https://github.com/bridgecrewio/checkov/blob/master/.github/exclude-patterns.txt ?

njgibbon: @schosterbarak  Just added :)

schosterbarak: Thank you!

PR #836, opened 2021-02-02
add basic handling in dynamic values for terraform resources

schosterbarak:
Checkov will take the dynamic content and inject that into the resource configuration


tronxd (approved):
Super elegant!

PR #840, opened 2021-02-03
Use BC_XXX ID in output and -c/--skip

mikeurbanski1:
- Adds the BC_XXX ID to all checks (if the mapping exists) and Record objects
- Adds logic to use --check and --skip-check with BC IDs or CKV IDs
- Adds a flag to use the BC check ID instead of the CKV ID in CLI and gitlab output (the check ID is not used in the JUnit output, and the JSON output implicitly includes the BC ID)

These make it easier to use the platform alongside checkov. No default behavior is changed, other than including one new attribute in the JSON output.


checkov/main.py
```diff
@@ -80,6 +81,11 @@ def run(banner=checkov_banner, argv=sys.argv[1:]):
     external_checks_dir = get_external_checks_dir(args)
     url = None
 
+    all_checks = BaseCheckRegistry.get_all_registered_checks()
+    ckv_to_bc_id_mapping = bc_integration.get_ckv_to_bc_id_mapping()
```
schosterbarak, line 85: Maybe we should cache that mapping on a file with 1 day TTL. It will save a lot of REST API call time. WDYT?

checkov/serverless/runner.py
```diff
@@ -118,7 +118,7 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
                         results = cfn_registry.scan(sls_file, entity, skipped_checks, runner_filter)
                         tags = cfn_utils.get_resource_tags(entity, cfn_registry)
                         for check, check_result in results.items():
-                            record = Record(check_id=check.id, check_name=check.name, check_result=check_result,
+                            record = Record(check_id=check.id, bc_check_id=check.bc_id, check_name=check.name, check_result=check_result,
```
schosterbarak, line 121: i don't recommend changing the constructor of a basic class. let's add a setter.

checkov/common/runners/runner_registry.py
```diff
@@ -53,9 +53,9 @@ def print_reports(self, scan_reports, args, url=None):
                     junit_reports.append(report)
                     # report.print_junit_xml()
                 elif args.output == 'github_failed_only':
-                    report.print_failed_github_md()
+                    report.print_failed_github_md(use_bc_ids=args.use_bc_ids)
                 else:
-                    report.print_console(is_quiet=args.quiet, is_compact=args.compact)
+                    report.print_console(is_quiet=args.quiet, is_compact=args.compact, use_bc_ids=args.use_bc_ids)
```
schosterbarak, line 58: this method is also used upstream. let's use a setter.

mikeurbanski1: The only other usage is in tests.

checkov/main.py
mikeurbanski1: Yeah, I think the biggest improvement would be for VSCode. Otherwise I would say one API call per run is not that significant.

The call is already happening; we're not adding an additional call here. So I would prefer to merge this and then handle this change as a new PR (perhaps attached to my in-progress integrations refactor?). Does that work?

checkov/common/runners/runner_registry.py
mikeurbanski1: I think we should refactor the reports output into different classes, so that the Report class is just a data object. And probably runner_registry should not be having all the logic for determining how to print reports, so we can refactor that too.

I prefer to keep this for now and handle the refactor separately. I'm 99% sure that there is not another entrypoint to `print_reports` (it just takes `args` anyways) or `print_console`.

PR #842, opened 2021-02-04
Update LambdaEnvironmentCredentials.py

FredStober:
Ensure that the environment check is only applied when an environment is defined, solving a crash that happens when a lambda function does not define static environment variables, in which case the configuration looks like {..., 'environment': []}


schosterbarak:
Thanks @FredStober!

schosterbarak:
Hey @FredStober, congrats on your first merged PR 🎉 Nice work 👍 Please let me know if we can be helpful in any way.

PR #844, opened 2021-02-07
CF - Check - CKV_AWS_88

njgibbon:
Hello,

Implementing check for EC2 Public IP in CloudFormation as done in Terraform.

Terraform Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/EC2PublicIP.py

CloudFormation Docs: 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-network-iface-embedded.html#aws-properties-ec2-network-iface-embedded-associatepubip
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-networkinterface.html

Note (for if you take a closer look at the test data):
It is odd how they have chose to have 'NetworkInterfaces' as a slightly different type in instances and launch templates and how they have some of the same field names with different types and such. But that is the way it is.

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/EC2PublicIP.py
+++ b/checkov/cloudformation/checks/resource/aws/EC2PublicIP.py
@@ -1,0 +1,41 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_check import BaseResourceCheck
+
+class EC2PublicIP(BaseResourceCheck):
+    def __init__(self):
+        name = "EC2 instance should not have public IP."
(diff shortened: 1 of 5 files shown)


schosterbarak (approved):
Thank you!

Issue #793, opened 2021-01-18
Checkov doesn't evaluate simple ternary operation


maximmi:
**Describe the bug**
The conditional syntax (like ternary operation) is not evaluated by checkov at the moment, leading to force skipping checks.

**To Reproduce**
1. Define a resource and variable like this:
```
variable "metadata_http_tokens_required" {
  type        = bool
  default     = true
  description = "Whether or not the metadata service requires session tokens"
}

resource "aws_instance" "foo" {
  ami           = "ami-005e54dee72cc1d00" # us-west-2
  instance_type = "t2.micro"

  root_block_device {
    encrypted = true
  }

  metadata_options {
    http_tokens = (var.metadata_http_tokens_required) ? "required" : "optional"
  }
}
```
2. Run `checkov` in docker: `docker run -t -v $(pwd):/tf bridgecrew/checkov:latest -d /tf`

3. See error:
```
terraform scan results:

Passed checks: 3, Failed checks: 1, Skipped checks: 0

Check: CKV_AWS_79: "Ensure Instance Metadata Service Version 1 is not enabled"
	FAILED for resource: aws_instance.foo
	File: /main.tf:7-18
	Guide: https://docs.bridgecrew.io/docs/bc_aws_general_31
```

**Expected behavior**
`Checkov` evaluates condition to `"required"` value so all checks passed.

**Desktop (please complete the following information):**
 - OS: macOS Catalina 10.15.6 (19G73)
 - Checkov Version: 1.0.708

**Additional context**
There is an easy workaround for this case using `string` variable with values allowed for the parameter, but issue is about to support ternary evaluation.

robeden:
Thanks, @maximmi ... I'll see about adding that to the parser.

robeden:
Ticket for HCL2 parse error: https://github.com/amplify-education/python-hcl2/issues/55

nimrodkor:
@robeden we use our own checkov fork for these issues.. check out https://github.com/bridgecrewio/python-hcl2

robeden:
Yes, the issue occurs in both. I filed the ticket there as it seemed best to report at the lowest level.

PR #855, opened 2021-02-10
TF parser support for ternary conditions

robeden:
This introduces parser support for ternary (`${<condition> ? <then> : <else>}`) operations. If the `condition` portion can resolve to `true` or `false`, then the `then` or `else` portion is replaced accordingly. Conditions are extracted as proper variables to be resolved, so things like `var.foo` will be handled, if those can be resolved.

See the "ternary_793" and "ternaries" parser scenarios for a full test case (from #793).

Note that one case that doesn't currently work is mapping boolean string values:
```
foo = "true" ? "yes" : "no"
```
There is a bug with the HCL2 parser we use which currently blocks that functionality (see `test_hcl2_load_assumptions.py`'s `test_weird_ternary_string_clipping` case).

Fixes: #793


tests/terraform/parser/resources/parser_scenarios/ternary_793/main.tf
```diff
@@ -0,0 +1,18 @@
...
+resource "aws_instance" "foo" {
+  ami           = "ami-005e54dee72cc1d00" # us-west-2
+  instance_type = "t2.micro"
+
+  root_block_device {
+    encrypted = true
+  }
+
+  metadata_options {
+    http_tokens = (var.metadata_http_tokens_required) ? "required" : "optional"
```
nimrodkor, line 16: @robeden Can we test multi-line ternary? i.e.:
```terraform
 http_tokens = (var.metadata_http_tokens_required) ? 
    "required" : "optional"
```

nimrodkor (commented):
Looks really good @robeden !
One test case I'd like to make sure works

tests/terraform/parser/resources/parser_scenarios/ternary_793/main.tf
robeden: Looks like the HCL2 parser dies on that:

> Unexpected token Token(__ANON_0, '\\n') at line 10, column 35.
> Expected one of: 
>  * LSQB
>  * LBRACE
>  * LPAR
>  * NULL
>  * __ANON_11
>  * DECIMAL
>  * FALSE
>  * __ANON_3
>  * __ANON_10
>  * STRING_LIT
>  * MINUS
>  * TRUE
>  * BANG

robeden: I added this to the "ternaries" test case:
```
  local_true = true
  multiline = (local.local_true) ?
    "correct" : "wrong"
```

(Note: local evals inside locals isn't valid in terraform, but our parser would handle it... if the HCL2 parser would.)

nimrodkor: @robeden Is it on a commit you haven't pushed yet?

robeden: Yes, sorry. Pushed now.

tests/terraform/parser/resources/parser_scenarios/ternaries/main.tf
```diff
@@ -6,6 +6,11 @@ locals {
   bool_true  = true ? "correct" : "wrong"
   bool_false = false ? "wrong" : "correct"
 
+//  local_true = true
+  // TODO: HCL2 parser doesn't like the following line
+//  multiline = (local.local_true) ?
+//    "correct" : "wrong"
```
nimrodkor, line 12: @robeden I see it is commented out, so I'm guessing we don't support it? If not I'd like to have an issue on it

robeden: Filed #864. Note that #863 is also relevant to this PR.

nimrodkor (approved):

PR #870, opened 2021-02-16
Ensure retention is 90 days or greater

JamesWoolfenden:
The location of the audit policy for sql server is now a separate resource- there's an Azure policy that checks that the retention policy is 90 days or more, this is that check for the new resource.

### DIFF
--- a/checkov/terraform/checks/resource/azure/SQLServerAuditPolicyRetentionPeriod.py
+++ b/checkov/terraform/checks/resource/azure/SQLServerAuditPolicyRetentionPeriod.py
@@ -14,10 +14,11 @@ class SQLServerAuditPolicyRetentionPeriod(BaseResourceCheck):
     def scan_resource_conf(self, conf):
         if 'retention_in_days' in conf: 
             retention=force_int(conf['retention_in_days'][0])
-            if retention < 90:
-                return CheckResult.FAILED
-            else:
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/azure/SQLServerAuditPolicyRetentionPeriod.py
```diff
@@ -0,0 +1,24 @@
...
+        name = "Specifies a retention period of less than 90 days."
+        id = "CKV_AZURE_46"
+        supported_resources = ['azurerm_mssql_database_extended_auditing_policy']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if 'retention_in_days' in conf: 
+            retention=force_int(conf['retention_in_days'][0])
+            if retention < 90:
```
nimrodkor, line 17: force_int might return None if the value is not a number or a stringified number. Need to handle the None case

JamesWoolfenden: Good catch thanks!, added a catch for the none case.

nimrodkor (approved):

PR #888, opened 2021-02-18
CF - Check - ELBAccessLogs

njgibbon:
Hello,

Congrats all on the recent news! :)

Implementing this check as already done in TF. Able to use ValueCheck in this case because in CF world when AccessLoggingPolicy exists then Enabled is a required field and does not have a default value. So it must always be true or false.

TF Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/ELBAccessLogs.py
CF Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb-accessloggingpolicy.html

**License**


mikeurbanski1:
Thanks, @njgibbon ! LGTM.

mikeurbanski1 (approved):

PR #882, opened 2021-02-18
Make AWS secret checks more generic to match title

mikeurbanski1:
We previously updated the check titles to match the platform, which had broader checks than just AWS creds.

This PR updates the logic in the checks to look beyond AWS creds, as well, to bring it closer to what the platform looks for (prowler).

It also adds the Lambda check for CFN.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #889, opened 2021-02-18
Reworked: Make AWS secret checks more generic to match title

mikeurbanski1:
This is a reworked version of https://github.com/bridgecrewio/checkov/pull/882

It expands the secrets checks to be more broad and inline with the platform. It also removes overly aggressive Azure regexes which failed on nearly any length-32 or 88 "word".


JamesWoolfenden (approved):
LGTM

tests/cloudformation/checks/resource/aws/test_LambdaEnvironmentCredentials.py
```diff
@@ -0,0 +1,26 @@
...
+import unittest
+
+from checkov.cloudformation.checks.resource.aws.LambdaEnvironmentCredentials import check
+from checkov.cloudformation.runner import Runner
+from checkov.runner_filter import RunnerFilter
+
+
+class TestLambdaEnvironmentCredentials(unittest.TestCase):
+
+    def test_summary(self):
```
schosterbarak, line 11: @mikeurbanski1 can you add a test that will verify the previous FP won't come back?

PR #891, opened 2021-02-19
fix AWS_62 to check only for actual action/resource=*

mikeurbanski1:
Previously, this check failed when action and resource were a string containing *. For example, the following statement failed:

```
{
      "Action": "s3:PutObject*",
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::b1/*",
      "Sid": ""
    }
```

This PR updates the logic to only fail when a statement contains both an action and a resource that equal `*`.


JamesWoolfenden (approved):
LGTM

Issue #155, opened 2020-03-25
Enable correlation of multiple resources 


jonjozwiak:
Several checks such as VPC flow logs, S3 block public access, and CloudFront OAI require correlation of config across multiple resources.  Enable checkov to handle this in a single check

jare19:
VPC flow logs enabled for a newly created VPC is very important to my organization and would love to see this prioritized. Or an idea for a work around would be helpful.

christophetd:
One way to implement it would be to expose a function to custom checks allowing to retrieve resources of a specific type, e.g.

```python
from checkov.common.utils import find_resource_by_type # This does not exist

class MySampleCheck(BaseResourceCheck):
    def __init__(self):
        name = "Custom check"
        id = "SAMPLE"
        supported_resources = ['aws_s3_bucket']
        categories = [CheckCategories.GENERAL_SECURITY]
        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)

    def scan_resource_conf(self, conf):
      bucket_name = conf.get('bucket')[0]
      public_access_blocks = find_resource_by_type("aws_s3_public_access_block")
      for public_access_block in public_access_blocks:
            if public_access_block.get('bucket') == bucket_name:
                return CheckResult.PASSED
      return CheckResult.FAILED
```

robeden:
Flow log request ticket: #154

robeden:
It would be great if we could pass an addition argument in to the `scan_*` functions which would provide access to other parts of the model. That would prevent setting "global" state in the check itself and allow for future possibilities such as async scans (not that this is an imminent thing, but good to allow for it). I think this would be the cleanest and most intuitive place to put an accessor object so that devs know it exists.

However, assuming we don't want to change the function signature, the way to go about this would be to set a field in the base check during the duration of the `run` function:
```
def run(...):
   self.doc_access = ...
   try:
      ...
   finally:
      self.doc_access = None
```

Then there could be support functions such as the one @christophetd describes above.

robeden:
I'm working on this...

robeden:
Fixed by #1023

PR #892, opened 2021-02-19
WIP: Resource correlation support

robeden:
⚠️ Preview for feedback ⚠️

This adds a couple arguments to checks' `scan_entity_conf` function (etc) to access to the full model definition. New args are:
* `entity_name`: the name of the entity being checked
* `definition_access`: the real magic

`definition_access` is intended to be subclassed for more interesting, framework-specific functionality, but the base implementation currently provides this function:
```
    def full_definition(self) -> Mapping:
        """
        Return the full document definition. Note that the resulting mapping should not be modified. It
        may allow changes to be made, but they will not affect be seen by other checks.

        :return:        A mapping type of the entire definition structure. What exactly is contained at this
                        level is not strictly defined across all frameworks, but often (not always!!) starts
                        keyed by a file location. See your specific check class for more details.
        """
```

The terraform subclass currently adds this:
```
    def find_resource_by_name(self, resource_type: str, resource_name: str) -> Dict:
        """
        Searches definitions for a resource of a given type and name in the context of the current check and,
        if found, returns its configuration data. If not found, an empty dict will be returned.
        """
```

As an (not-quite-real) example, a check handling an `aws_s3_bucket_policy` could retrieve the corresponding bucket like this:
```
    def scan_resource_conf(self, conf, entity_type, entity_name, definition_access):
        bucket_conf = definition_access.find_resource_by_name("aws_s3_bucket", "my_bucket")
```
(Cheat notice: this isn't really a valid example because you probably wouldn't look it up like that, but the `bucket` field in this example would be something like `${aws_s3_bucket.my_bucket.id}... which I'm still figuring out how to deal with.)

I would imagine that other useful functions will be added over time for other purposes.

Adapters have been added for the previous signature for `scan_resource_conf` and `scan_entity_conf`, so there should be no breaking changes to existing code using the normal APIs.

Fixes: #155 


robeden:
@nimrodkor - Yup, I’m fine with that. 👍

PR #893, opened 2021-02-21
Malformed blocks handling

tronxd:
Handle malformed blocks (e.g. without a valid unicode identifier) gracefully, by excluding them from the parser output


rotemavni (approved):
LGTM :)

PR #894, opened 2021-02-22
update python-hcl2 to handle ternaries new line

schosterbarak:

### DIFF
--- a/tests/terraform/parser/resources/parser_scenarios/ternaries/main.tf
+++ b/tests/terraform/parser/resources/parser_scenarios/ternaries/main.tf
@@ -6,10 +6,9 @@ locals {
   bool_true  = true ? "correct" : "wrong"
   bool_false = false ? "wrong" : "correct"
 
-//  local_true = true
-  // TODO: HCL2 parser doesn't like the following line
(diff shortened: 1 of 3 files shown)


tests/terraform/parser/resources/parser_scenarios/ternaries/main.tf
```diff
@@ -6,10 +6,9 @@ locals {
   bool_true  = true ? "correct" : "wrong"
   bool_false = false ? "wrong" : "correct"
 
-//  local_true = true
-  // TODO: HCL2 parser doesn't like the following line
-//  multiline = (local.local_true) ?
-//    "correct" : "wrong"
+  local_true = true
+  multiline = (local.local_true) ?
```
schosterbarak, line 10: Thanks @arielkru now the parser is not breaking!
@robeden I think the method `_process_vars_and_locals` is removing the `multiline` local from the dictionary.  So it's missing currently at the expected value.

schosterbarak: nvm. found the issue.

PR #910, opened 2021-02-23
Check - Bug - EKSNode - TFPlan

njgibbon:
Hello,

Raising this to fix an edge case issue that I am seeing.

When the `aws_eks_node_group`  `remote_access` block is omitted then the check should pass.

All of the tests check the cases you would expect and it is all fine. However, when you omit this and run a terraform plan then the json ends up outputting the resource as `"remote_access": []". This ultimately causes the check to error out and crash.

Because it is an empty list, the `conf["remote_access"][0].keys()` throws an error because there is no list item which is a dict and so there you can not call `.keys()`.

It is not caught in tests because it is an issue in how terraform is setting an odd (and not possible to predict?) default when it goes from HCL to the json representation. It should be not there or even perhaps [{ }] would be better. Either way I can't replicate the issue by passing data in because at the HCL level all makes sense.

Checking that the list has items if it exists (even though it probably always does) is probably just a good defensive strategy anyways.

**Error log excerpt**
```
File "/usr/local/lib/python3.8/site-packages/checkov/terraform/checks/resource/base_resource_check.py", line 33, in wrapper
return wrapped(self, conf)
File "/usr/local/lib/python3.8/site-packages/checkov/terraform/checks/resource/aws/EKSNodeGroupRemoteAccess.py", line 15, in scan_resource_conf
if "ec2_ssh_key" in conf["remote_access"][0].keys() and not 'source_security_group_ids' in conf["remote_access"][0].keys():
File "/usr/local/lib/python3.8/site-packages/checkov/terraform/context_parsers/tf_plan/node.py", line 202, in __getattr__
raise TemplateAttributeError('%s.%s is invalid' % (self.__class__.__name__, name))
checkov.terraform.context_parsers.tf_plan.node.TemplateAttributeError: list_node.keys is invalid
```

**Code refs for ease of clicking**

Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/EKSNodeGroupRemoteAccess.py

Tests: https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/aws/test_EKSNodeGroupRemoteAccess.py

**License**

### DIFF
--- a/checkov/terraform/checks/resource/aws/EKSNodeGroupRemoteAccess.py
+++ b/checkov/terraform/checks/resource/aws/EKSNodeGroupRemoteAccess.py
@@ -12,6 +12,9 @@ class EKSNodeGroupRemoteAccess(BaseResourceCheck):
 
     def scan_resource_conf(self, conf):
         if "remote_access" in conf.keys():
+            if not conf["remote_access"]:
+                # If list is empty then x.[0].keys() will raise an error
+                return CheckResult.PASSED
             if "ec2_ssh_key" in conf["remote_access"][0].keys() and not 'source_security_group_ids' in conf["remote_access"][0].keys():
                 return CheckResult.FAILED
         return CheckResult.PASSED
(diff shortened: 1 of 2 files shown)


schosterbarak:
@njgibbon  i think the fix should be in the check itself and not in the plan runner:
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/EKSNodeGroupRemoteAccess.py
meaning verify it's a non-empty list.

nickgibbon-nationwide:
@schosterbarak  - Hey, if you look, the change is actually in the check its self. The stuff about the plan runner and all that is just an explanation of why i've done it :)

schosterbarak:
sweet! thanks @nickgibbon-nationwide

PR #914, opened 2021-02-24
MySQLPublicAccessDisabled - dont default to public

JamesWoolfenden:
By default the dbs are created public. Not a great option. This checks and errors if its is.


checkov/terraform/checks/resource/azure/MySQLPublicAccessDisabled.py
```diff
@@ -0,0 +1,24 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
+
+
+class MySQLPublicAccessDisabled(BaseResourceCheck):
```
schosterbarak, line 5: @JamesWoolfenden I think it be inherited from baseresourcevaluecheck

JamesWoolfenden: made it so

Issue #772, opened 2021-01-08
Feature request: report full module name if module instead of just the resource name


ismailyenigul:
**Describe the solution you'd like**
It  would be cool  to see the full module name if module instead of just the resource name

robeden:
@ismailyenigul - Would you mind posting a snippet of where you're looking in the report so I can verify we're looking at the same thing?

ismailyenigul:
Hi @robeden 
Thanks for prompt reply.
code:
```
module "my_sg" {
  source = "terraform-aws-modules/security-group/aws"
  name        = "user-service"
  description = "sg"
  vpc_id      = "vpc-123"
  ingress_with_cidr_blocks = [
    {
      from_port   = 22
      to_port     = 22
      protocol    = "tcp"
      description = "User-service ports"
      cidr_blocks = "0.0.0.0/0"
    },
    {
      from_port   = 80
      to_port     = 80
      protocol    = "tcp"
      description = "User-service ports 80"
      cidr_blocks = "0.0.0.0/0"
    },
    {
      from_port   = 443
      to_port     = 443
      protocol    = "tcp"
      description = "User-service ports 80"
      cidr_blocks = "0.0.0.0/0"
    }
  ]
}
```

check result:

```
Check: CKV_AWS_24: "Ensure no security groups allow ingress from 0.0.0.0:0 to port 22"
	FAILED for resource: aws_security_group_rule.ingress_with_cidr_blocks
	File: /tf.json:0-0
	Guide: https://docs.bridgecrew.io/docs/networking_1-port-security
```


it would be  
`FAILED for resource:  module.my_sg.aws_security_group_rule.ingress_with_cidr_blocks`

robeden:
Perfect. Thanks, @ismailyenigul

robeden:
Looking into this, I think it would also be helpful if the calling file were included in the report output.

For example, there's the parsed file key info for a test case that I have:
```
  "/<repo_path>/modules/kong/db_instance.tf[/<repo_path>/test/fixtures/bluegreen/main.tf#1]": { ... }
```

The reported file path is: `/modules/kong/db_instance.tf` ... which is true, but doesn't paint a great picture because there are 8 other references to that file from the "prod" locations which are all correct.

[...]

PR #917, opened 2021-02-24
Provide more detail for context of resource loaded from modules

robeden:
This records the full path for resources loaded from modules. For example, `module.test_module.aws_s3_bucket.inside` rather than simply `aws_s3_bucket.inside` to indicate that the bucket (named `inside`) was loaded from the module `test_module`. In addition the "Caller File" has been added to record output for both json and textual output.

Sample output can be seen [here](https://github.com/bridgecrewio/checkov/issues/772#issuecomment-785427909)

Fixes: #772

### DIFF
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -142,6 +142,7 @@ class Runner(BaseRunner):
                                                    "module." + referrer_id,
                                                    separator=".")
                     except KeyError:
+                        logging.debug("Unable to find caller context for: %s", abs_caller_file)
                         caller_context = None
 
                     if caller_context:
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -1,7 +1,7 @@
 import dataclasses
 import logging
 import os
-from typing import Dict
+from typing import Dict, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


checkov/terraform/runner.py
```diff
@@ -119,7 +121,34 @@ def run_block(self, entities,
...
+                    caller_file_path = f"/{os.path.relpath(abs_caller_file, root_folder)}"
+
+                    try:
+                        caller_context = dpath.get(definition_context[abs_caller_file],
+                                                   # HACK ALERT: module data is currently double-nested in
+                                                   #             definition context. If fixed, remove the
+                                                   #             addition of "module." at the beginning.
+                                                   "module." + referrer_id,
+                                                   separator=".")
+                    except KeyError:
```
schosterbarak, line 144: is thisi expected? do we want to log in debug such case?

robeden: Yes, good point. I’ll add that. One of the parser test cases hits that branch so I know it happens though I haven’t dug in to why.

robeden: Done

nimrodkor (approved):

Issue #924, opened 2021-02-26
BaseResourceValueCheck._is_variable_dependant incorrect


robeden:
**Describe the bug**
BaseResourceValueCheck is supposed to ignore values that have unresolved data, but its pattern matching isn't looking for the correct thing, so it doesn't work. For example, these values won't cause the check to be skipped: `${var.whatever}`, `something-${local.whatever}`.

It's looking (essentially, this is simplified a bit) for `var.*` or `local.*`, which isn't correct because the parser will provide values wrapped in eval markers.

**To Reproduce**
1. Create a check looking for a specific value
2. Provide terraform with a value of `var.SOMETHING_THAT_WONT_RESOLVE`. (Note: the check will see `${var.SOMETHING_THAT_WONT_RESOLVE`)
3. See that the check fails rather than passing

**Expected behavior**
Check should pass.

PR #925, opened 2021-02-26
Correct regex determining if a value if var dependant.

robeden:
Fixes: #924


checkov/terraform/checks/resource/base_resource_value_check.py
```diff
@@ -6,7 +6,7 @@
 from checkov.common.models.consts import ANY_VALUE
 from checkov.common.util.type_forcers import force_list
 
-VARIABLE_DEPENDANT_REGEX = r'(?:local|var)\.[^\s]+'
+VARIABLE_DEPENDANT_REGEX = re.compile(r'\${([^{}]+?)}')
```
tronxd, line 9: @robeden It looks like that the proposed regex would not match variable dependent expression such as 
`foo = var.bar` which is valid, so I suggest to let the curly braces be optional instead of mandatory.
Another issue is that it would match a non-dependent expression such as `foo={bar}`.

tronxd (changes requested):

checkov/terraform/checks/resource/base_resource_value_check.py
robeden: That's valid for Terraform, but not something the check should ever see. The HCL2 parser (`bc-python-hcl2` or `python-hcl2`) will always wrap it to `${var.bar}`.

The regex is only checking the value portion, so for the second, I don't think it applies?... unless I misunderstand you.

Assuming this terraform:
```
resource aws_s3_bucket test {
  foo = var.bar
  foo2 = { 
    "bar" = "baz"
  }
}
```

`foo` would be evaluated by the HCL parser to `["${var.bar}"]` and the value seen by the regex would be `${var.bar}`.
`foo2` would be evaluated to a map and so doesn't apply.

I could see a false hit with a TF string like `foo = "This eval has escapes: $${var.tricky}"`. But not requiring the eval brackets probably just makes that worse.

The regex actually comes from what I originally used when I redid the terraform parser. The problem is had was that it didn't real with really complex values (such as maps within the statement)... which is a consideration, I suppose. We could:

1. leave as-is which handles, I think, most of the likely cases but can miss some complex stuff.
2. change to `"${" in value`, which probably also handles most cases, frankly. It would misfire on escaped "${" values, which probably aren't that likely.
3. Use the full "var block" searching logic from the Terraform parser, which would work but seems overkill.

And of course, if I misunderstand you and missed a case, please let me know.

tronxd: @robeden Wasn't aware that the output of `var.foo` would be outputted from hcl2 as `${var.foo}`. If so, I tend to go with option 3 you mentioned, as expressions can be complex and we don't want to miss that.

robeden: @tronxd - Seems reasonable to me. Change pushed.

tronxd (approved):
GJ!

checkov/terraform/checks/resource/base_resource_value_check.py
robeden: @tronxd - BTW, I should have mentioned I've been putting tests for how the HCL2 parser behaves into [`test_hcl2_load_assumptions.py`](https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/parser/test_hcl2_load_assumptions.py). I don't have a simple case for variables in there, but you can see them wrapped in some more complex cases.

PR #939, opened 2021-03-03
remove funky char in update string causing crashes in windows installs.

metahertz:
Some windows users report repeatable crashes due to an unknown char for the arrow we were using in the "update available" output, debugging was done to work out if this was a version specific or colorama specific thing, however no specific commit/history item was found. 

Best thing for stability for these users is to remove the offending char.

### DIFF
--- a/checkov/common/util/banner.py
+++ b/checkov/common/util/banner.py
@@ -14,5 +14,5 @@ By bridgecrew.io | version: {} """.format(version)
 
 new_version = check_for_update("checkov", version)
 if new_version:
-    banner = "\n" + banner + "\nUpdate available " + colored(version,"grey") + " > " + colored(new_version, "green") + "\nRun " + colored(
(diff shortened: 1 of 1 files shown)


checkov/common/util/banner.py
```diff
@@ -14,5 +14,5 @@
 
 new_version = check_for_update("checkov", version)
 if new_version:
-    banner = "\n" + banner + "\nUpdate available " + colored(version,"grey",attrs=["blink"]) + " → " + colored(new_version, "green") + "\nRun " + colored(
+    banner = "\n" + banner + "\nUpdate available " + colored(version,"grey") + " > " + colored(new_version, "green") + "\nRun " + colored(
```
tronxd, line 17: I would use `->`

tronxd (approved):

checkov/common/util/banner.py
metahertz: Done, `->` it is!

Issue #941, opened 2021-03-03
Crashing in runner.py due to AttributeError raised by checking None for keys


KevinHock:
**Describe the bug**
I am getting a crash similar in appearance to #118 but for Terraform and not CloudFormation.

```bash
Scanning file: /some_directory/some-file.tf
full_file_path is /other_directory/some_directory/some-file.tf
Traceback (most recent call last):
  File "/my_path/bin/checkov", line 5, in <module>
    run()
  File "/my_path/lib/python3.7/site-packages/checkov/main.py", line 108, in run
    guidelines=guidelines, bc_integration=bc_integration)
  File "/my_path/lib/python3.7/site-packages/checkov/common/runners/runner_registry.py", line 35, in run
    runner_filter=self.runner_filter, collect_skip_comments=collect_skip_comments)
  File "/my_path/lib/python3.7/site-packages/checkov/terraform/runner.py", line 78, in run
    self.check_tf_definition(report, root_folder, runner_filter, collect_skip_comments)
  File "/my_path/lib/python3.7/site-packages/checkov/terraform/runner.py", line 100, in check_tf_definition
    scanned_file, runner_filter, abs_referrer)
  File "/my_path/lib/python3.7/site-packages/checkov/terraform/runner.py", line 104, in run_all_blocks
    for block_type in definition.keys():
AttributeError: 'NoneType' object has no attribute 'keys'
```

**To Reproduce**
edit: See https://github.com/bridgecrewio/checkov/issues/941#issuecomment-790803047

That's the thing 😢 It's real hard to reproduce, if I run `checkov` just on the file that it crashes on, things work fine! If I run it on just the last 10 files it scans before crashing, things work fine. For testing this, I'd recommend you simply run this on a big repo with a lot of output from `find . -type f -name "*.tf"` and then turn all the files into individual --file arguments, like my cool checkov wrapper package does.
e.g. `checkov --check CKV_AWS_88 --file fileA.tf --fileB.tf` (to save time, I just limited this to 1 check, but it could be any other check id)

1. Run cli command `checkov --check CKV_AWS_88 --file fileA.tf --fileB.tf` on a ton of --file arguments. Pray it triggers the bug.
2. See error

**Expected behavior**
No crashing.

**Desktop (please complete the following information):**
 - OS: OS X
 - Checkov Version: the latest as of today: `1.0.825`

robeden:
Checking...

robeden:
[...]

PR #942, opened 2021-03-03
Protection against None reference

robeden:
Fixes: #941

### DIFF
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -104,6 +104,7 @@ class Runner(BaseRunner):
     def run_all_blocks(self, definition, definitions_context, full_file_path, root_folder, report,
                        scanned_file, runner_filter, module_referrer: Optional[str]):
         if not definition:
+            logger.debug("Empty definition, skipping run")
             return
         for block_type in definition.keys():
             if block_type in CHECK_BLOCK_TYPES:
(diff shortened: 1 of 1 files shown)


checkov/terraform/runner.py
```diff
@@ -101,6 +103,8 @@ def check_tf_definition(self, report, root_folder, runner_filter, collect_skip_c
 
     def run_all_blocks(self, definition, definitions_context, full_file_path, root_folder, report,
                        scanned_file, runner_filter, module_referrer: Optional[str]):
+        if not definition:
+            return
```
schosterbarak, line 107: Thanks for the quick fix. 

I think we would want a debug log in that case.

robeden: Broke something. I'm away from my computer right now. I'll look at it when I get back.

robeden:
@schosterbarak @nimrodkor - Something seems to be up with integration tests. Both this and #943 are hitting the same issue and I don’t think either should have any effect on them? I don’t know much about those tests. Any advice?

checkov/terraform/runner.py
robeden: Looks like something broke in the recent `bc-python-hcl2` release. Error is:
> AttributeError("'Tree' object has no attribute 'items'")

schosterbarak:
thanks for getting this one. just pushed a fix to #944

tronxd (approved):

PR #943, opened 2021-03-05
Better handling of module data dependencies

robeden:
This fixes a resolution problem where module data dependencies didn't previously work. For example (from the new `module_output_reference` parser scenario):
```
module "common" {
  source = "./common"
}
module "bucket" {
  source = "./bucket"
  tags = module.common.tags
}
```
Previously, both modules would attempt to load in the same pass, before variable resolution, which means that the reference to "module.common.tags" would slip into the data for the "bucket" module. This would not be resolvable, so the broken reference would be all it would ever see.

This change recognizes that the data is unresolved and so delays loading of the "bucket" module. So now, "common" is loaded, then variables are resolved, then in the next pass "bucket" is loaded... this time with resolved tag data.

### DIFF
--- a/checkov/terraform/parser.py
+++ b/checkov/terraform/parser.py
@@ -36,8 +36,8 @@ class Parser:
         self._parsed_directories = set()
 
         # This ensures that we don't try to double-load modules
-        # Tuple is <file>, <module_index> (see _load_modules)
-        self._loaded_modules: Set[Tuple[str, int]] = set()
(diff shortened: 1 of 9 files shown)


robeden:
Integration test failure is caused by #944.

schosterbarak:
fixed for #944 pushed to master branch

rotemavni (approved):
@robeden 
Loved the implementation.

PR #962, opened 2021-03-09
CF - Check - ES Logging

njgibbon:
Hello,

Implementing this check as already done for tf.

TF Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/ElasticsearchDomainLogging.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticsearch-domain-logpublishingoption.html

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/ElasticsearchDomainLogging.py
+++ b/checkov/cloudformation/checks/resource/aws/ElasticsearchDomainLogging.py
@@ -11,6 +11,6 @@ class ElasticsearchDomainLogging(BaseResourceValueCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
-        return 'Properties/LogPublishingOption/Enabled'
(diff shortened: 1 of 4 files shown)


laurrentt:
I should probably open a separate issue for this, however this check does not work. Here's an example of how LogPublishingOptions should be declared (from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html#aws-resource-elasticsearch-domain--examples). 
```
      LogPublishingOptions:
        SEARCH_SLOW_LOGS:
          CloudWatchLogsLogGroupArn: >-
            arn:aws:logs:us-east-1:123456789012:log-group:/aws/aes/domains/es-slow-logs
          Enabled: 'true'
        INDEX_SLOW_LOGS:
          CloudWatchLogsLogGroupArn: >-
            arn:aws:logs:us-east-1:123456789012:log-group:/aws/aes/domains/es-index-slow-logs
          Enabled: 'true'
```

Issue #192, opened 2020-03-26
Add new check: Ensure Session Manager data is encrypted in transit and at rest 


jonjozwiak:
If document_type = Session, then check for the following within the document: 

	kmsKeyId	- This enabled encryption in transit
	s3EncryptionEnabled - Encryption at rest… only if s3BucketName had a value
	cloudWatchEncryptionEnabled - Encryption at rest… only if cloudWatchLogGroupName has a value

An example of the SSM document is here: https://docs.aws.amazon.com/systems-manager/latest/userguide/getting-started-sessiondocumentaccesscheck.html

https://www.terraform.io/docs/providers/aws/r/ssm_document.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-document.html

PR #980, opened 2021-03-14
Add CKV_AWS_112 check for aws_ssm_document session type

gruebel:

Adds the encryption in transit part of issue #192. I think it is better to have to different checks, this one and an other for checking if logging is enabled and encrypted, or should it be then 3 checks?

### DIFF
--- a/checkov/terraform/checks/resource/aws/SSMSessionManagerDocumentEncryption.py
+++ b/checkov/terraform/checks/resource/aws/SSMSessionManagerDocumentEncryption.py
@@ -8,7 +8,7 @@ from checkov.terraform.checks.resource.base_resource_check import BaseResourceCh
 class SSMSessionManagerDocumentEncryption(BaseResourceCheck):
     def __init__(self):
         name = "Ensure Session Manager data is encrypted in transit"
-        id = "CKV_AWS_107"
+        id = "CKV_AWS_112"
         supported_resources = ["aws_ssm_document"]
         categories = [CheckCategories.ENCRYPTION]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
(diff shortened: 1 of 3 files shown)


JamesWoolfenden (changes requested):
Is this conf.get("document_type") == ["Session"]: isnt true should this be a pass?  same for if content isnt found in keys?

gruebel:
Correct the `document_type` has couple of other possible values and the only relevant for this check is the `Session` one.

Actually the `content` argument for the Terraform resource is mandatory, but checking other implemented checks, it felt like you approach it very conservative, so did I 😄 

And I have an other question, is it possible to parse the `jsonencode()` function from Terraform with checkov somehow?

JamesWoolfenden:
if i run your check against
`resource "aws_ssm_document" "notsession" {
  name          = "test_document"
  document_type = "Command"

  content = <<DOC
  {
    "schemaVersion": "1.2",
    "description": "Check ip configuration of a Linux instance.",
    "parameters": {

    },
    "runtimeConfig": {
      "aws:runShellScript": {
        "properties": [
          {
            "id": "0.aws:runShellScript",
            "runCommand": ["ifconfig"]
          }
        ]
      }
    }
  }
DOC
}`

It will pass this check when it should be ignored.

gruebel:
Then the preferred return is `SKIPPED`?

JamesWoolfenden:
I talked to the source :
UNKNOWN - when we don't know what to return
SKIPPED - should not be used in a check. it is used in the suppression evaluation.
So its Unknown

JamesWoolfenden (approved):
LGTM

schosterbarak:
@gruebel CKV_AWS_107 already exists - can you please pull from master and choose the latest ID?

gruebel:
@schosterbarak I know 😄  that's why I changed it already to `CKV_AWS_112`
 I just didn't update the PR title https://github.com/bridgecrewio/checkov/blob/baf000fcf46f7ae919fc42d1fccfc93c3ac303fa/checkov/terraform/checks/resource/aws/SSMSessionManagerDocumentEncryption.py#L11

PR #983, opened 2021-03-15
Remove duplicate azure disk encryption and fix default fail

mikeurbanski1:
The checkov checks CKV_AZURE_2 and 51 are duplicates, with a small difference (the newer one accounts for a newer attribute).

Also, both give false positives. Managed disks are encrypted by default, but these policies fail by default.

https://docs.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview

So, the duplicate needs to be removed and the logic needs to be fixed.


mikeurbanski1:
The diff looks weird because I thinks I renamed a different file than I actually did.

schosterbarak (approved):

JamesWoolfenden (approved):
lgtm

PR #999, opened 2021-03-18
Update Seccomp check to work on non-list annotations

m-wynn:

In [get_skipped_checks](https://github.com/bridgecrewio/checkov/blob/f2268666d1b45655d98514d5bd7663e5238b3a65/checkov/kubernetes/runner.py#L232), the root-level `metadata["annotations"]` is transformed into a list.  This is later assumed to be a list in the Seccomp check.  However, annotations that appear at other levels of the tree may not be a list.

This ensures that `metadata["annotations"]`, wherever it might be, is a list.


nimrodkor (approved):
Thanks @m-wynn

PR #1003, opened 2021-03-21
Add CKV_AWS_113 check for aws_ssm_document session type

### LINKED ISSUE #192: Add new check: Ensure Session Manager data is encrypted in transit and at rest

gruebel:

Fixes #192 follow up to #980 adds a check for the logging configuration of Session Manager, if it is enabled and encrypted, either CloudWatch or S3 is possible.


checkov/terraform/checks/resource/aws/SSMSessionManagerDocumentLogging.py
```diff
@@ -0,0 +1,58 @@
...
+                return CheckResult.UNKNOWN
+
+        return CheckResult.UNKNOWN
+
+
+def is_json(data):
+    try:
+        json.loads(data)
+    except ValueError:
+        return False
```
schosterbarak, line 46: can you please add debug logging in that case (and the same for YAML)?

schosterbarak:
Thank you @gruebel

PR #1016, opened 2021-03-24
Improve performance

nimrodkor:


Pipfile
```diff
@@ -26,7 +26,7 @@ termcolor="*"
 junit-xml ="*"
 dpath = ">=1.5.0,<2"
 pyyaml = ">=5.4.1"
-boto3 = "*"
+boto3 = "==1.17.27"
```
robeden, line 29: Curious: why does this improve performance?

schosterbarak: we are not sure. trial and error. :(

nimrodkor: That's an excellent question. Still trying to figure that one out

schosterbarak: for me today felt like the commits at: https://twitter.com/gitlost

PR #1018, opened 2021-03-26
The check for Storage Account naming rules produced a false positive

Lebski:
re.match returns an object or None so we have to check if the object is not None

### DIFF
--- a/checkov/terraform/checks/resource/azure/StorageAccountName.py
+++ b/checkov/terraform/checks/resource/azure/StorageAccountName.py
@@ -21,7 +21,7 @@ class StorageAccountName(BaseResourceCheck):
         :param conf: azurerm_storage_account configuration
         :return: <CheckResult>
         """
-        return CheckResult.PASSED if conf.get('name') and re.match(STO_NAME_REGEX, conf['name'][0]) else CheckResult.FAILED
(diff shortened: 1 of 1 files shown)


checkov/terraform/checks/resource/azure/StorageAccountName.py
```diff
@@ -21,7 +21,7 @@ def scan_resource_conf(self, conf):
         :param conf: azurerm_storage_account configuration
         :return: <CheckResult>
         """
-        return CheckResult.PASSED if conf.get('name') and re.match(STO_NAME_REGEX, conf['name'][0]) else CheckResult.FAILED
+        return CheckResult.PASSED if conf.get('name') and (re.match(STO_NAME_REGEX, conf['name'][0]) != None) else CheckResult.FAILED
```
tronxd, line 24: I would use 're.findall` for cleaner code. On case there's no match, it would yield an empty array

Lebski: I agree

tronxd (approved):
LGTM

schosterbarak:
Thank you @Lebski

schosterbarak:
Thanks for your contribution @Lebski 🙌 Please let me know if you’d like some help to contribute further.

Also, contributors get swag. Feel free to reach out over slack  https://slack.bridgecrew.io/  to get yours :)

Issue #1007, opened 2021-03-22
Raising code quality


gruebel:
**Is your feature request related to a problem? Please describe.**
During developing my first Terraform checks for AWS resources I was surprised that their are no kind of code quality tools, like black or flake8 enforced.

**Describe the solution you'd like**
It would be great to add following tools:
- black (formatting)
- flake8 (linting)
- isort (import sorting)
- mypy (type checking)

**Additional context**
I'm happy to help you out here and also a GitHub action named `lint` should be added to check, if a new PR follows those new standards.

JamesWoolfenden:
i like gha but also pre-commit framework? something like that/both?

gruebel:
you probably mean https://github.com/pre-commit/pre-commit yeah that would also be a great addition.

nimrodkor:
Implemented in #2091 and the subsequent issues

PR #1019, opened 2021-03-27
ci: Add lint job; apply black, isort

antonblr:
An initial set of checks for #1007.

This PR adds:
* [pre-commit-config.yaml](https://github.com/antonblr/checkov/blob/lint-black-isort/.pre-commit-config.yaml) with [black](https://github.com/psf/black), [isort](https://github.com/PyCQA/isort), `end-of-file-fixer` and `trailing-whitespace` checks/hooks
* [.isort.cfg](https://github.com/antonblr/checkov/blob/lint-black-isort/.isort.cfg) with profile set to `black` (so `black` and `isort` won't fight each other)
* [.github/workflows/lint.yml](https://github.com/antonblr/checkov/blob/lint-black-isort/.github/workflows/lint.yml) to run pre-commit checks

This PR is large, but basically, all the changes are automatic via `pre-commit run --all-files` command. 

There is a room to improve code readability, while keep `black` satisfied, but I think improvements could be made as people working on those areas. 


gruebel:
Looks good, hoped to also work on that, but anyway great job 💪

gruebel:
@antonblr and next time please respect the issue creator, me. I clearly offered my help to do the task. Not a nice move by you at all!

antonblr:
@gruebel - oh, sorry. Didn't realize that you actually want to do it all. Would you be interested in doing `flake`, `mypy`, `markdownlint` still?

gruebel:
Yeah, would be great. Would be happy, if you can have a look then, because I still don't have much experience with  GitHub Actions.
But let's get this PR merged first, then I can follow up with flake8 and others.

gruebel:
An other thing I saw, you use the default chars per line for black, right? Should we raise it to 120, for me the default 88 is always so low 😄

antonblr:
>An other thing I saw, you use the default chars per line for black, right? Should we raise it to 120, for me the default 88 is always so low 😄

Here are the few reasons I prefer all defaults:
1. To not have a conversation about what length/style is better - people are different and it could be hard to get consensus here. It is also defeats the whole purpose of `black`.
2. The default length of 88 is chosen for a reason - [psf/black/docs/the_black_code_style.md#line-length](https://github.com/psf/black/blob/master/docs/the_black_code_style.md#line-length)
3. Upgrade to a newer black version should be seamless if you're on default setting. With non-default settings you may run into edge cases.
4. `isrot` has a [profile for black](https://pycqa.github.io/isort/docs/configuration/profiles/) with `line_length: 88`

With that said, I will fully accept the input from the repo owners.

gruebel:
Regarding your points:

1.  the de facto Python default is actually 79! No arguing about this.
2. I know the reasons and they are as good as mine saying 120 is the best! The video talks about 90ish, which is 85-95. Even pep8 says that you can raise the line length to 99.
3. This can be said about any tools you use
4. `isort` has profiles for couple of other tools and their default is 79 same for flake8.
5. This the only point I agree on, the repo owner should decide and as a contributer we should give him a choice and also talk about it.
6. Maybe you should also look into the number 88 and what kind of meaning it has https://en.wikipedia.org/wiki/88_(number)#In_white_nationalism_and_supremacism

schosterbarak:
@antonblr WOW. That's a huge PR. Was this code changes generated automatically?

checkov/arm/checks/SQLServerNoPublicAccess.py
```diff
@@ -19,13 +25,15 @@ def scan_resource_conf(self, conf):
...
-                                resource["type"] == "firewallrules":
+                        if (
+                            resource["type"] == "Microsoft.Sql/servers/firewallRules"
+                            or resource["type"].lower() == "firewallrules"
+                        ):
                             if "properties" in resource:
-                                if "startIpAddress" in resource["properties"] and \
-                                        resource["properties"]["startIpAddress"] in ['0.0.0.0', '0.0.0.0/0']:  # nosec
+                                start_ip = resource["properties"].get("startIpAddress")
+                                if start_ip in ["0.0.0.0", "0.0.0.0/0"]:  # nosec
```
antonblr, line 34: initially `black` formatted this into: 
```python
if "properties" in resource:
    if "startIpAddress" in resource[
        "properties"
    ] and resource["properties"]["startIpAddress"] in [
        "0.0.0.0",
        "0.0.0.0/0",
    ]:  # nosec
```
`# nosec` don't seem to work here as `bandit` check failed. I've updated it to be in-line. Alternatively it can be set as follows (let me know if you prefer this):
```python
 ] and resource["properties"]["startIpAddress"] in [  # nosec
   ...
```

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -1,48 +1,56 @@
...
 from colorama import Style
-# from git import Repo
-from json import JSONDecodeError
-from os import path
 from termcolor import colored
 from tqdm import trange
 from urllib3.exceptions import HTTPError
 
-from checkov.common.bridgecrew.ci_variables import *
+from checkov.common.bridgecrew import ci_variables
```
antonblr, line 19: removed `import *` to be more explicit - this module uses `os` lib, but it gets imported indirectly from `checkov.common.bridgecrew.ci_variables` module. 

All `import *` is bad in general and will have to be addressed anyways should we add a python linting tool, such as `flake8` or `pylint`

checkov/terraform/checks/resource/azure/SQLServerNoPublicAccess.py
```diff
@@ -1,22 +1,36 @@
...
+            id=id,
+            categories=categories,
+            supported_resources=supported_resources,
+        )
 
     def scan_resource_conf(self, conf):
-        if 'start_ip_address' in conf and conf['start_ip_address'][0] in ['0.0.0.0', '0.0.0.0/0']: # nosec
+        start_ip = conf.get("start_ip_address")
+        if start_ip and start_ip[0] in ["0.0.0.0", "0.0.0.0/0"]:  # nosec
             return CheckResult.FAILED
```
antonblr, line 32: initially `black` formatted this into: 
```python
if "start_ip_address" in conf and conf["start_ip_address"][0] in [
    "0.0.0.0",
    "0.0.0.0/0",
]:  # nosec
    return CheckResult.FAILED
```
`# nosec` don't seem to work here as `bandit` check failed. I've updated it to be in-line. Alternatively, I think, it can be set as follows (let me know if you prefer this):
```python
 if "start_ip_address" in conf and conf["start_ip_address"][0] in [  # nosec
   ...
```

antonblr (commented):
hi @schosterbarak, I've outlined (should have done this earlier) the things I've adjusted manually after an initial pass. The rest are all automatic changes.

schosterbarak:
Thank you @antonblr . I appreciate your work. 
Reviewing more than 1K files is too hard. 
Do you know a way to split it into several PR's? 
If yess, let's wait a little because i know @nimrodkor is preparing a PR that will delete some of the files you are working on.

antonblr:
@schosterbarak - the easiest for me would be to split this PR into multiple commits. 

For multiple PRs - I think I could limit changes by passing a glob pattern like `pre-commit run --files <path/glob>` and hardcode this glob in CI (and then expanding it as we go). There could be some confusions during that period, e.g. if someone's PR touches formatted and non-formatted areas. 

From my experience applying black/isort on enterprise level apps - black was harmless; with isort I ran into a few circular dependencies issues after the sorting, but that should be uncovered by unit tests.

gruebel:
@antonblr an other thing, wouldn't it be better to use a `pyproject.toml` for centralizing all the tool configurations, instead of having separate files for each tool?

antonblr:
@gruebel - I don't have objections. But keep in mind that not everything can be centralized in `pyproject.toml`. [flake8](https://gitlab.com/pycqa/flake8/-/issues/428#note_251982786) is against it, `bandit` does not support it yet etc. So as of now I'm not considering it.

antonblr:
Closing the PR (distracted by other projects - won't be able to follow-up). 

Also, at this point it will be easier to regenerate it from scratch (add configs mentioned in the description and run `pre-commit run --all-files`), rather than resolving the conflicts.

PR #1022, opened 2021-03-30
Reduce the Docker image size

JamesWoolfenden:

If the image is based on Alpine instead of ubuntu the image size drops from 325 to 182 MB. Faster downloads and response for the vscode integration.


schosterbarak (approved):
simple!

PR #1024, opened 2021-03-30
Further reduce Docker image size

syphernl:

In #1022 the image size went from 325MB to 182MB by switching from Debian to Alpine. 
I was able to reduce this even further to 165MB by disabling the caches of the APK and Pip installs. 

17MB is not as big of a reduction as the 143MB from #1022, but still it should make the download/usage a little faster. 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.8-alpine
 
-RUN apk update && apk add git
+RUN apk --no-cache add git
 
-RUN pip install -U checkov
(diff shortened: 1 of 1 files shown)


JamesWoolfenden:
Awesome!

schosterbarak:
Hey @syphernl, congrats on your first merged PR 🎉 Nice work 👍 Please let me know if we can be helpful in any way.
Also, we would love to see your org at: https://github.com/bridgecrewio/checkov/blob/master/INTHEWILD.md

PR #1030, opened 2021-03-31
CF - Extend Cloudsplaining to ManagedIAMPolicy

njgibbon:
Hello,

Whilst going through my CloudFormation coverage checks I see we have implemented Cloudsplaining which looks like it can play a larger role for IAM checks in CF and TF and across all ways of embedding policies. I think this is great work to get the integration working!

This is just addressing the TODO to extend to ManagedIAMPolicies and is an initial start in the area of extending this wider.

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
+++ b/checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
@@ -13,7 +13,7 @@ from checkov.terraform.checks.utils.iam_cloudformation_document_to_policy_conver
 class BaseCloudsplainingIAMCheck(BaseResourceCheck):
     def __init__(self, name, id):
         # todo: managedpolicy
-        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy"])
+        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy", "AWS::IAM::ManagedPolicy"])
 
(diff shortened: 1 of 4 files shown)


checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
```diff
@@ -13,7 +13,7 @@
 class BaseCloudsplainingIAMCheck(BaseResourceCheck):
     def __init__(self, name, id):
         # todo: managedpolicy
-        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy"])
+        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy", "AWS::IAM::ManagedPolicy"])
```
schosterbarak, line 16: @mikeurbanski1 FYI :)

schosterbarak: @njgibbon we can probably remove the todo from line 15 now

njgibbon: Removed. I think that we will be able to extend to other ways of embedding policies e.g https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-user.html#cfn-iam-user-policies but that requires a bit more thought because the fields are different and so you need to do different parsing to get to the policie(s).

Issue #988, opened 2021-03-17
CKV_AWS_107: aws_iam_policy for action ecr:GetAuthorizationToken with resource "*"


SoerenVoskuhl0042:
**Is your feature request related to a problem? Please describe.**
We have an aws iam policy with the following statement:
`  statement {
    actions = [
      "ecr:GetAuthorizationToken"
    ]
    resources = [
      "*"
    ]
  }`

checkov fails here with the message CKV_AWS_107: "Ensure IAM policies does not allow credentials exposure" but according to the aws documentation we found [(https://docs.aws.amazon.com/AmazonECR/latest/public/security_iam_id-based-policy-examples.html)](https://docs.aws.amazon.com/AmazonECR/latest/public/security_iam_id-based-policy-examples.html) we need to set resources = "*".

**Describe the solution you'd like**
Is it possible to change the check CKV_AWS_107 to allow the above action with resources = "*"?

acdha:
I think this is an incorrect application of the general guidance: unlike, say, access keys or STS tokens this is how ECRs are intended to be used — the Docker login step is required functionality.

schosterbarak:
adding @kmcquade  wdyt?

kmcquade:
@schosterbarak I can see how this request to change it makes sense. ecr:GetAuthorizationToken doesn't allow for ARN restriction, so that doesn’t mean that the developer is writing the IAM policy wrong. From an attacker perspective, it is useful to know who can get the ECR authorization token, but from the perspective of the person writing the IAM policy, they shouldn’t get dinged for it.

@schosterbarak I believe you leverage the Cloudsplaining library to determine credentials exposure. Is there a way to exclude this action only in the credentials exposure finding for Checkov, while keeping the ecr:GetAuthorizationToken the same in Cloudsplaining?

schosterbarak:
@kmcquade so we would probably want to exclude that if the resource is attached to ECR but to alarm for the rest. 
We don't have that capability yet, but should be available from soon where we will enable resource graph queries in checkov. I'll keep this ticket open for now and will update once the implementation is done

kmcquade:
@schosterbarak I think we are good to close this issue since it was addressed in #1032.

schosterbarak:
Thanks @kmcquade !

PR #1032, opened 2021-04-01
Allows resource=* for ecr:GetAuthorizationToken

nl-brett-stime:

There's not really a good way to constrain the AWS `ecr:GetAuthorizationToken` action by resource. The API just returns a token for use with e.g., Docker but the only argument it takes (and could use for constraint context) is deprecated (https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html#API_GetAuthorizationToken_RequestSyntax) and not usually passed by recent clients. The repo constraints are applied over other actions made via subsequent requests using the token.

This patch allows IAM policy statements like the following to pass without throwing `CKV_AWS_107` (credential exposure):

```
              statement {
                actions = [
                    "ecr:GetAuthorizationToken",
                ]
            
                resources = [
                  "*",
                ]
              }
```

Usually such statements are accompanied by something similar to:

```
  statement {
    actions = [
      "ecr:GetDownloadUrlForLayer",
      "ecr:BatchGetImage",
      "ecr:BatchCheckLayerAvailability"
    ]
    resources = [
      aws_ecr_repository.foo.arn
    ]
  }
```

Even though there's no constraint against the GetToken request itself, the returned token can only be used with the ECR service to authorize use of the Docker 'registry' APIs. The usage of the token is also automatically further constrained by whatever `ecr:` actions are granted to the entity that requested the token. E.g., callers using tokens derived from the example policy above would only be able to use them to download from `aws_ecr_repository.foo`. They wouldn't be able to write to that repo nor would they be able to read from other repos in the same account.

### DIFF
--- a/checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
+++ b/checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
@@ -9,7 +9,11 @@ class CloudSplainingCredentialsExposure(BaseCloudsplainingIAMCheck):
         super().__init__(name=name, id=id)
 
     def cloudsplaining_analysis(self, policy):
-        return policy.credentials_exposure
+        excluded_actions = {
+            "ecr:GetAuthorizationToken"
+        }
+        credentials_exposure_actions = policy.credentials_exposure
+        return [x for x in credentials_exposure_actions if x not in excluded_actions]
 
 
 check = CloudSplainingCredentialsExposure()
--- a/checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
+++ b/checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
@@ -2,6 +2,9 @@ from checkov.terraform.checks.data.BaseCloudsplainingIAMCheck import BaseCloudsp
 
 
 class CloudSplainingCredentialsExposure(BaseCloudsplainingIAMCheck):
+    excluded_actions = {
+        "ecr:GetAuthorizationToken"
+    }
 
     def __init__(self):
         name = "Ensure IAM policies does not allow credentials exposure"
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
```diff
@@ -9,7 +9,11 @@ def __init__(self):
         super().__init__(name=name, id=id)
 
     def cloudsplaining_analysis(self, policy):
-        return policy.credentials_exposure
+        excluded_actions = {
+            "ecr:GetAuthorizationToken"
```
tronxd, line 13: I would set the action as a file constant/class member

nl-brett-stime:
FWIW, example from AWS: https://docs.aws.amazon.com/AmazonECR/latest/userguide/security_iam_id-based-policy-examples.html#security_iam_id-based-policy-examples-access-one-bucket

schosterbarak:
@nl-brett-stime this looks great. 
seems it could fix https://github.com/bridgecrewio/checkov/issues/988 right?

nl-brett-stime:
> seems it could fix #988 right?

Yes. I didn't see that issue but count me among those who would like to see it fixed ;) .

checkov/terraform/checks/data/aws/IAMCredentialsExposure.py
nl-brett-stime: > I would set the action as a file constant/class member

Good call--patched.

schosterbarak:
Thanks for your contribution @nl-brett-stime 🙌 Please let me know if you’d like some help to contribute further.

Also, contributors get swag. Feel free to reach out over slack  https://slack.bridgecrew.io/  to get yours :)

PR #1046, opened 2021-04-08
Add CKV_DOCKER_6 check for existence of MAINTAINER

gruebel:

Happy to see Docker support by checkov. Also happy to help here with some best or worst practice 😄 

I'm still seeing from time to time the usage of the infamous `MAINTAINER` instruction. I hope someday it will be removed completely.

https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

### DIFF
--- a/checkov/dockerfile/checks/MaintainerExists.py
+++ b/checkov/dockerfile/checks/MaintainerExists.py
@@ -1,0 +1,20 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.dockerfile.base_dockerfile_check import BaseDockerfileCheck
+
+
+class MaintainerExists(BaseDockerfileCheck):
+    def __init__(self):
+        name = "Ensure that LABEL maintainer is used instead of MAINTAINER (deprecated)"
+        id = "CKV_DOCKER_6"
+        supported_instructions = ["MAINTAINER"]
+        categories = [CheckCategories.CONVENTION]
+        super().__init__(name=name, id=id, categories=categories, supported_instructions=supported_instructions)
+
+    def scan_entity_conf(self, conf):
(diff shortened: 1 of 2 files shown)


checkov/dockerfile/checks/MaintainerExists.py
```diff
@@ -0,0 +1,20 @@
...
+        name = "Ensure that LABEL maintainer is used instead of MAINTAINER (deprecated)"
+        id = "CKV_DOCKER_6"
+        supported_instructions = ["MAINTAINER"]
+        categories = [CheckCategories.CONVENTION]
+        super().__init__(name=name, id=id, categories=categories, supported_instructions=supported_instructions)
+
+    def scan_entity_conf(self, conf):
+        for instruction, content in conf.items():
+            if instruction == "MAINTAINER":
+                return CheckResult.FAILED, conf[instruction][0]
```
tronxd, line 16: I think the expression can be simplified into `content[0]`

tronxd (approved):

checkov/dockerfile/checks/MaintainerExists.py
gruebel: Yeah, you are right. Good point!

schosterbarak:
Thanks @gruebel . Learning here, why does MAITAINER action is a bad practices?

gruebel:
At some point Docker introduced the instruction `LABEL` which is much more flexible and if you want to add the maintainer information to the image, then you can do it via `LABEL maintainer="checkov"`. Also monitoring services like Datadog can use these labels to forward them to their platform and give you the possibility to query them.

schosterbarak:
Thank you @gruebel

PR #1058, opened 2021-04-10
CKV_AWS_143 - fix dynamic checks

mourlos:
- The check CKV_AWS_143 was only setup to handle static values. Has now been modified to handle dynamic values.

P.S (I closed the ones I raised yesterday and fixed all issues in this one)


tests/terraform/checks/resource/aws/test_S3BucketObjectLock.py
```diff
@@ -46,22 +44,35 @@ def test_success(self):
         scan_result = check.scan_resource_conf(conf=resource_conf)
         self.assertEqual(CheckResult.PASSED, scan_result)
 
-    def test_success_no_param(self):
+    def test_dynamic_value_object_lock(self):
```
nimrodkor, line 47: @mourlos Thanks!
Can we add another test, instead of replacing that test?

mourlos: Hi @nimrodkor , I had a look on the changes and looks like I am not handling the check correctly at this stage and the original check works fine (the issue was on my initial tf code that I was testing against).

Given the above, this PR doesn't really add any real value, apart from the extra test case, in which case I am not sure it's worth merging.

Feel free to close it in this case (or merge as part of an actual release)

Issue #559, opened 2020-09-07
junit ouptut format - add context information


jglapa:
**Is your feature request related to a problem? Please describe.**
We're using CI to run the cli checkov analysis. It works good but with bigger code bases it takes a lot of scrolling to find the actually violation. 
We started using junit report which can be parsed and presented in a nicer form.
![image](https://user-images.githubusercontent.com/742104/92414400-5db65a00-f154-11ea-8730-f44cd97baf7f.png)

The problem now is how to locate the violation. The junit report contains no context information where exactly should we look into. We end up again using grep/find

**Describe the solution you'd like**
It would be great if the junit report contained context information where exactly did the violation occur.

**Describe alternatives you've considered**
One option would be to use multiple output at the same time cli+junit but this doesn't work as expected since junit doesn't create a dedicated ouptut file but rather you have to pipe it yourself e.g.

`checkov  -o junitxml > analysis.xml`

tantweiler:
Hello! 

We have exactly the same use case and would be more than happy if the violated policy IDs would be visible in the junit reports. 

Thank you!

schosterbarak:
Hi @tantweiler @jglapa can you supply a mock of the suggestion?

tantweiler:
Hi @schosterbarak! Just an example --> It would be ok for me if I could see the ID (e.g. CKV_AZURE_6) in the suite column:

 
![image](https://user-images.githubusercontent.com/48676838/104896142-07b1b880-5977-11eb-8f22-e8d7749fc4ee.png)

schosterbarak:
as concatenated value? 

Btw what system is the screenshot from?

tantweiler:
Hi @schosterbarak,

That is GitLab! Concatenated value should work I would say.

schosterbarak:
I think that be done by adding modifying the following line:
https://github.com/bridgecrewio/checkov/blob/feea72d817db0f0a1fbb9689164858f356325213/checkov/common/output/report.py#L117
to have `record.check_id` concatenated as string. 

I haven't done the testing, if you want to give it a try - we would accept a PR there.

AFriemann:
I've been playing around with this a little together with tfsec

What's odd is that using warnings-ng to integrate this with github checks it detects findings from tfsec but none from checkov.

the format also differs slightly

tfsec:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="tfsec" failures="1" tests="1">
[...]

PR #1059, opened 2021-04-10
junit output - add check id, update deprecated method, use apostrophes

ketumbra:
This PR addresses some of the issues discussed in #559:

* adds the check ID to the junit test names so it's much easier to find the ID so the rule can be skipped etc.
* updates the deprecated `to_xml_string` method to the new `to_xml_report_string` method
* uses apostrophes instead of escaped quotes in the failure messages. The escapes were shown in gitlab output and made it harder to read.

NB: Some tests fail locally on windows as they expect the unix path separator. I don't think this is relevant to the issue and I look forward to seeing the CI results.

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -2,7 +2,7 @@ import json
 from collections import defaultdict
 
 from colorama import init
-from junit_xml import TestCase, TestSuite
+from junit_xml import TestCase, TestSuite, to_xml_report_string
 from termcolor import colored
 
 from checkov.common.models.enums import CheckResult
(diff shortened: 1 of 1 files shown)


tronxd (approved):
LGTM

schosterbarak:
@ketumbra can you please add unit tests?

ketumbra:
> @ketumbra can you please add unit tests?

Are there any existing tests on this functionality? I couldn't see any. Hard to think of any meaningful tests to add that aren't really tests of the junit_xml library.

PR #1060, opened 2021-04-10
Add CKV_DOCKER_7 check for latest version tag reference

gruebel:

It is always recommended to use a non latest version tag for your base image or multi stage builds, because of typical nature of the latest tag being used for the most up-to-date version, which can result in a major version upgrade.

Using a versioned tag makes also the build more reliable and consistent.

### DIFF
--- a/checkov/dockerfile/checks/ReferenceLatestTag.py
+++ b/checkov/dockerfile/checks/ReferenceLatestTag.py
@@ -1,0 +1,36 @@
+import re
+
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.dockerfile.base_dockerfile_check import BaseDockerfileCheck
+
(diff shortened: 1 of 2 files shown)


checkov/dockerfile/checks/ReferenceLatestTag.py
```diff
@@ -0,0 +1,36 @@
...
+            if instruction == "FROM":
+                for content in contents:
+                    base_image = content["value"]
+                    multi_stage = re.match(MULTI_STAGE_PATTERN, base_image)
+                    if multi_stage:
+                        base_image = multi_stage[1]
+                        stages.append(multi_stage[2])
+
+                    if ":" not in base_image and base_image not in stages:
+                        return CheckResult.FAILED, contents[0]
```
tronxd, line 30: Please verify contents is not empty

tronxd (approved):

Issue #1037, opened 2021-04-05
Managed IAM Policies Fail with any use of `*`


luckymike:
**Describe the bug**
In the latest version (1.0.865), it appears that a managed IAM policy will trigger a failure if either of `Resource` or `Action` is set to `*`. This is an overly naive check, as even a policy with `"Effect": "Deny"` will trigger a failure.

I believe this was introduced here: https://github.com/bridgecrewio/checkov/pull/1030

**To Reproduce**
1. Create a managed policy resource with either of `"Resource": "*"` or `"Action": "*"`, with the other scoped narrowly.
2. Run checkcov against this
3. One or more of `CKV_AWS_107`, `CKV_AWS_109`, and `CKV_AWS_111` will fail. 

**Expected behavior**
The policy should pass these checks.

**Additional context**
Even a policy with an explicit Deny all will fail now, e.g.:
```
Check: CKV_AWS_109: "Ensure IAM policies does not allow permissions management without constraints"
	FAILED for resource: AWS::IAM::ManagedPolicy.PlaceholderPolicy
	File: /miscellaneous/legacy_vpc.json:249-274
		249 |   "Resources": {
		250 |     "PlaceholderPolicy": {
		251 |       "Type": "AWS::IAM::ManagedPolicy",
		252 |       "Properties": {
		253 |         "Description": {
		254 |           "Fn::Join": [
		255 |             "",
		256 |             [
		257 |               "Placeholder policy for ",
		258 |               "staging",
		259 |               " legacy VPC"
		260 |             ]
		261 |           ]
		262 |         },
		263 |         "Path": "/placeholder/",
		264 |         "PolicyDocument": {
		265 |           "Version": "2012-10-17",
		266 |           "Statement": [
		267 |             {
		268 |               "Effect": "Deny",
		269 |               "Action": "*",
		270 |               "Resource": "*"
		271 |             }
		272 |           ]
		273 |         }
		274 |       }
```

njgibbon:
@luckymike  Hello, 

Thanks for raising this. 

Over the past month we have been adding the use of Cloudsplaining so that we re-use good work that is already being done out in the community ( https://github.com/salesforce/cloudsplaining ).

CloudFormation has lots of different ways a policy can get in to a Resource so we are working to support the scanning of all of them in an efficient way - hence the ref'd change where we extend to ManagedPolicies.

[...]

PR #1063, opened 2021-04-10
Cloudsplaining Testing Start

njgibbon:
Hello,

Start for organising testing for Cloudsplaining and CloudFormation checks.

Related to issue: https://github.com/bridgecrewio/checkov/issues/1037

This moves managed policy testing and covers the checks related to the above issue which will now pass and ignore 'Deny' Actions given 0.4.1 > Cloudsplaining usage.

**License**


Pipfile
```diff
@@ -34,7 +34,7 @@ tqdm = "*"
 update_checker = "*"
 semantic_version = "*"
 packaging = "*"
-cloudsplaining = ">=0.3.2"
+cloudsplaining = ">=0.4.1"
```
schosterbarak, line 37: thanks @njgibbon . Can you please updadte setup.py file too?

njgibbon: Done that now. Also moved the test file structure around so it's not nested after some cf-lint feedback on one of the previous commits.

schosterbarak:
Great work. should i close https://github.com/bridgecrewio/checkov/issues/1037 after merging?

njgibbon:
Yes I think this should cover all of 1037 as only these 2 checkov checks were affected by the cloudsplaining change. I'll be adding more test coverage and some more IAM checks for CF in the next week or so 🙂

Pipfile
schosterbarak: I think we need too update pipfile.lock to

njgibbon: Updating that file now

nimrodkor (approved):
GJ!

PR #1064, opened 2021-04-10
Add CKV_DOCKER_8 check for last USER being root

gruebel:

Quite often you need to use the `USER root` instruction to install packages or change permissions. Therefore the last `USER` instruction should not be root to make the container start with a non root user by default.


schosterbarak:
Thank you @gruebel gruebel

PR #1067, opened 2021-04-11
Various checks fixes

rotemavni:
1. Fixed cloudformation `CKV_AWS_58` - there access to conf was list-like rather than dict-like. Added UT as well.

2. Handled multiple cases in K8 where specs elements existed as `None` which caused errors while trying to access nested elements. Handled it by implementing `get_inner_spec` method in `base_spec_check`.

3. Fixed serveless runner accessing `cf_sub_template['Resources'].items()` when `cf_sub_template['Resources']` doesn't exists.
 
### DIFF
--- a/checkov/kubernetes/checks/Seccomp.py
+++ b/checkov/kubernetes/checks/Seccomp.py
@@ -57,10 +57,8 @@ class Seccomp(BaseK8Check):
                             if "metadata" in conf["spec"]["jobTemplate"]["spec"]["template"]:
                                 metadata = conf["spec"]["jobTemplate"]["spec"]["template"]["metadata"]
         else:
-            if "spec" in conf:
-                if "template" in conf["spec"]:
-                    if "metadata" in conf["spec"]["template"]:
-                        metadata = conf["spec"]["template"]["metadata"]
+            if conf.get("spec") and conf.get("spec").get("template"):
+                metadata = conf.get("spec").get("template").get("metadata", {})
 
         if metadata:
             if metadata.get('annotations'):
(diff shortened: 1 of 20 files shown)


checkov/kubernetes/checks/Seccomp.py
```diff
@@ -57,10 +57,8 @@ def scan_spec_conf(self, conf):
                             if "metadata" in conf["spec"]["jobTemplate"]["spec"]["template"]:
                                 metadata = conf["spec"]["jobTemplate"]["spec"]["template"]["metadata"]
         else:
-            if "spec" in conf:
-                if "template" in conf["spec"]:
-                    if "metadata" in conf["spec"]["template"]:
-                        metadata = conf["spec"]["template"]["metadata"]
+            if conf.get("spec") and conf.get("spec").get("template"):
+                metadata = conf.get("spec").get("template").get("metadata", {})
```
nimrodkor, line 61: @rotemavni Can we leverage the new `get_inner_spec`?

nimrodkor (approved):
Nice!

nimrodkor:
@rotemavni FYI - the CFN you added are not valid:
```
E3002 Property is an object instead of List at Resources/Resource1/Properties/EncryptionConfig
tests/cloudformation/checks/resource/aws/example_EKSSecretEncryption/EKSSecretEncryption-FAILED.yml:29:7
```

checkov/kubernetes/checks/Seccomp.py
rotemavni: @nimrodkor Good idea, renamed `get_inner_conf` to be `get_inner_entry` and added the entry name as param

rotemavni:
> @rotemavni FYI - the CFN you added are not valid:
> 
> ```
> E3002 Property is an object instead of List at Resources/Resource1/Properties/EncryptionConfig
> tests/cloudformation/checks/resource/aws/example_EKSSecretEncryption/EKSSecretEncryption-FAILED.yml:29:7
> ```

@nimrodkor Fixed the test files and the check

PR #1075, opened 2021-04-13
Fix update_inner_attribute index error

rotemavni:
When `nested_attributes` are a list, check that the current key is within the size of the list before accessing that index.


tronxd (approved):
LGTM. A UT that covers the edge case would be helpful

PR #1088, opened 2021-04-15
Enable CF Cloudsplaining Checks Where Policies Are Embedded In Resources

njgibbon:
Hello all,

Continuing on with the work we are doing with Cloudsplaining I have made it so that we are able to access and analyse policies that are embedded as part of other CF Resources - IAM Groups, Roles and Users.

As I've also been increasing the Cloudsplaining test coverage we can see that this does not disrupt the anything existing but it does enable both scanning of the new resource types and ensures that it will work where there are 1 or more policies in the policy lists for these resources.

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
+++ b/checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
@@ -12,30 +12,54 @@ from checkov.terraform.checks.utils.iam_cloudformation_document_to_policy_conver
 
 class BaseCloudsplainingIAMCheck(BaseResourceCheck):
     def __init__(self, name, id):
-        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy", "AWS::IAM::ManagedPolicy"])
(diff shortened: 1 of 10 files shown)


checkov/cloudformation/checks/resource/BaseCloudsplainingIAMCheck.py
```diff
@@ -12,28 +12,40 @@
 
 class BaseCloudsplainingIAMCheck(BaseResourceCheck):
     def __init__(self, name, id):
-        super().__init__(name=name, id=id, categories=CheckCategories.IAM, supported_resources=["AWS::IAM::Policy", "AWS::IAM::ManagedPolicy"])
+        super().__init__(name=name, id=id, categories=CheckCategories.IAM,
```
nimrodkor, line 15: @njgibbon Please note `categories` is plural, and supports multiple categories. Please wrap the `CheckCategories.IAM` in `[]` 🙏

nimrodkor (commented):
Thanks @njgibbon ! 1 small issue

schosterbarak:
great contribution @njgibbon

PR #1091, opened 2021-04-16
Add arm param check

mikeurbanski1:
This PR is much smaller than it looks. It includes a refactor of ARM checks to include `resource` and `parameter` checks. And then it adds a new check for this policy, by user request: https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/test-cases#secure-parameters-cant-have-hardcoded-default


checkov/arm/runner.py
```diff
@@ -92,13 +92,36 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+                                                             runner_filter)
                         for check, check_result in results.items():
                             record = Record(check_id=check.id, check_name=check.name, check_result=check_result,
                                             code_block=entity_code_lines, file_path=arm_file,
                                             file_line_range=entity_lines_range,
                                             resource=resource_id, evaluations=variable_evaluations,
                                             check_class=check.__class__.__module__, file_abs_path=file_abs_path)
                             report.add_record(record=record)
+
+            if isinstance(definitions[arm_file], dict_node) and 'parameters' in definitions[arm_file].keys():
```
nimrodkor, line 105: @mikeurbanski1 Suggestion - merge the first part of the 2 `if`s - `if isinstance(definitions[arm_file], dict_node)` and then have 2 inner `if`s for parameters and resources

nimrodkor (approved):
@mikeurbanski1 Very very nice!

checkov/arm/runner.py
mikeurbanski1: Good call, thanks. Initially when I saw the comment referencing "nested ARM resources" I thought, "nope, not touching that!"

But it's much better this way (and still did not require touching that).

PR #1099, opened 2021-04-18
allow the user to specify a CA bundle when running bridgecrew

tkishel:
Software (like GlobalProtect) generates 'ignore self signed certificate in certificate chain' errors when executing bridgecrew; and/or bridgecrew might be installed into a Python without any CA certificates.

This commit allows users to specify a CA bundle (including the 'Palo Alto Networks Inc Root CA' used by GlobalProtect, for example) via a --ca-certificate argument or a BC_CA_BUNDLE environment variable.

Hint: To generate a CA bundle, copy the CA bundle provided by the certifi module (locate via 'python -m certifi') and, if applicable, append other CA certificates (including the 'Palo Alto Networks Inc Root CA' used by GlobalProtect, for example).

This commit also replaces the global http variable with a self.http instance variable (representing a urllib3.ProxyManager or urllib3.PoolManager resource) used for network requests.

### DIFF
--- a/checkov/common/bridgecrew/platform_integration.py
+++ b/checkov/common/bridgecrew/platform_integration.py
@@ -41,12 +41,6 @@ SIGNUP_HEADER = {
     'Content-Type': 'application/json;charset=UTF-8'
 }
 
-try:
-    http = urllib3.ProxyManager(os.environ['https_proxy'])
-except KeyError:
-    http = urllib3.PoolManager()
-
-
 class BcPlatformIntegration(object):
     def __init__(self):
         self.bc_api_key = read_key()
@@ -73,6 +67,13 @@ class BcPlatformIntegration(object):
         self.ckv_to_bc_id_mapping = None
         self.use_s3_integration = False
         self.platform_integration_configured = False
+        self.setup_http_manager()
+
+    def setup_http_manager(self):
+        try:
+            http = urllib3.ProxyManager(os.environ['https_proxy'])
+        except KeyError:
+            http = urllib3.PoolManager()
 
     def setup_bridgecrew_credentials(self, bc_api_key, repo_id, skip_fixes=False, skip_suppressions=False, source=None,
                                      source_version=None):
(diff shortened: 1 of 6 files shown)


tkishel:
Here is an example of Azure detecting the same issue, and their solution. 

```
# az login

The default web browser has been opened at https://login.microsoftonline.com/common/oauth2/authorize. Please continue the login in the web browser. If no web browser is available or if the web browser fails to open, use device code flow with `az login --use-device-code`.
Certificate verification failed. This typically happens when using Azure CLI behind a proxy that intercepts traffic with a self-signed certificate. Please add this certificate to the trusted CA bundle. More info: https://docs.microsoft.com/cli/azure/use-cli-effectively#work-behind-a-proxy. 
Error detail: HTTPSConnectionPool(host='login.microsoftonline.com', port=443): Max retries exceeded with url: /common/oauth2/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1125)')))
```

```
# export REQUESTS_CA_BUNDLE=~/certs/certifi-globalprotect.cacerts.pem
# az login

[
  {
    "cloudName": "AzureCloud",
    "homeTenantId": "redacted",
    "id": "redacted",
    "isDefault": true,
    "managedByTenants": [],
    "name": "Example",
    "state": "Enabled",
    "tenantId": "redacted",
    "user": {
      "name": "redacted",
      "type": "user"
    }
  }
]
```

Because bridgecrew uses both `requests` and `urlib3` the user would have to set both `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` to the same file .. which seems excessive.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -73,6 +67,24 @@ def __init__(self):
         self.ckv_to_bc_id_mapping = None
         self.use_s3_integration = False
         self.platform_integration_configured = False
+        self.setup_http_manager(os.getenv('BC_CA_BUNDLE', None))
```
mikeurbanski1, line 70: I believe this call is unnecessary, because we will always call `setup_http_manager` directly from `main` with the CLI arg (which could be from the environment variable).

schosterbarak (approved):

checkov/common/bridgecrew/platform_integration.py
tkishel: I updated the PR to both check and skip if the `self.http` instance variable (a ProxyManager/PoolManager  object) is already defined, and to minimize the number of calls to `self.setup_http_manager` down to (hopefully) one ... while allowing the equivalent `BC_CA_BUNDLE` environment variable to establish a default if no `--ca/--ca-certificate` is specified. 

The complication here was: the former global `http` variable was defined when `bc_integration` was imported ... which occurs before `parser.parse_args` in `main.py` ... and also oocurs other integrations using based on the `BaseIntegrationFeature` class.

PR #1113, opened 2021-04-21
Don't fail if command is set

m-wynn:

Currently checkov will fail if `command` is on your pod, even if it is not set to `kube-controller-manager`.  This PR will tell these checks to not fail unless the command contains `kube-controller-manager`.


hadarpl (approved):

nimrodkor (approved):
Thanks @m-wynn !

PR #1120, opened 2021-04-24
Fix wrong check imports and add test to prevent future mistakes

gruebel:

During work on a new check I realized there was a gab between the last check number and the previous one, which actually surprised me, because there is a test which tries to prevent it. After further investigation, I found the culprit and I even discovered that one of the checks imported the wrong parent check class. Therefore I added a test, which checks, to prevent future mistakes. And the test even found two other miss configured checks 💪 

I also adjusted the check `KubernetesSecretsEncryptedUsingCMK` to return an `UNKNOWN` result, if the encryption config is pointing to an other resource, which at the moment is not possible, because the only possible value is `secrets`.


nimrodkor (approved):
Thanks @gruebel !

PR #1124, opened 2021-04-25
[DO NOT MERGE] testing matrix in PR for py3.6 support

metahertz:

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -37,6 +37,7 @@ packaging = "*"
 cloudsplaining = ">=0.4.1"
 networkx = "*"
 dockerfile-parse ="*"
+dataclasses = {version = "*", markers="python_version < '3.7'"}
 
 [requires]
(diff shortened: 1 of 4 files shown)


Pipfile
```diff
@@ -37,6 +37,7 @@ packaging = "*"
 cloudsplaining = ">=0.4.1"
 networkx = "*"
 dockerfile-parse ="*"
+dataclasses = {version = "*", markers="python_version < '3.7'"}
```
schosterbarak, line 40: @metahertz You'll need to update `Pipfile` and `Pipfile.lock` too

metahertz: Thanks!

Issue i'm having is there doesnt seem to be a way to specify a range with `>=` in `[requires]`
ie:
```
python_version = ">=3.6"
```

Still gives warnings you have a 3.7 pipenv, which isn't very neat.

PR #1125, opened 2021-04-26
Add jsonencode parsing

gruebel:

I added the possibility to parse `jsonencode` functions, which will be helpful to create checks, when users use this function instead of text blocks.

I also fixed couple of deprecation warnings, one is still popping up, but I can't figure out where it comes from 😞


checkov/terraform/variable_rendering/safe_eval_functions.py
```diff
@@ -176,3 +176,6 @@ def wrap_func(f, *args):
 SAFE_EVAL_DICT['tonumber'] = lambda arg: arg if type(arg) in [int, float] else wrap_func(tonumber, arg)
 SAFE_EVAL_DICT['toset'] = lambda origin: set(origin)
 SAFE_EVAL_DICT['tostring'] = lambda arg: arg if isinstance(arg, str) else wrap_func(tostring, str(arg))
+
+# encoding
+SAFE_EVAL_DICT['jsonencode'] = lambda arg: arg
```
schosterbarak, line 181: interesting. we had some challenges around encoding in the past. @nimrodkor @rotemavni @tronxd can I have another set of eyes and security guidelines on those? just to see we are not missing anything

nimrodkor: @schosterbarak Yep it's good :)

PR #1140, opened 2021-04-28
Add a major version image that is stable for the docker images

Deichindianer:
Hi,

We're using checkov quite a lot to check our Terraform configurations. It's great to see this repository getting a lot of releases and updates we hugely benefit from it!

However it is quite cumbersome to embed any version apart from `latest` in our deployment pipeline templates due to the high release cadence on the docker images. I've, naively, tried to update the Github actions here to additionally update not only the `latest` tag but create another tag with just the major version, this could be used to pull the latest release inside of a given major version.

I'm not quite sure with the tagging strategy but this script will of course break if the semver versioning I've seen sometimes is broken in the tags so it might not fit the repository entirely, happy to take suggestions on what to change :)

This would be a huge quality of life improvement for me and probably others!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,7 +179,9 @@ jobs:
         id: get_version
         run: |
           checkov_version=$(curl -sL https://api.github.com/repos/bridgecrewio/checkov/tags | jq -r '.[0]["name"]' )
+          checkov_major_version=$(echo "${checkov_version}" | head -c1)
           echo ::set-env name=RELEASE_VERSION::$(echo $checkov_version)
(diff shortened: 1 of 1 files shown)


nimrodkor (approved):
LGTM

schosterbarak:
@Deichindianer what does it mean on major releases? how will they get patched?

Deichindianer:
@schosterbarak Not quite sure what you're referring to, this change just picks the first sign of the git tag. That would mean if this repository would for example release `3.0.0` then a new tag `3` would be created next to the existing `2` tag. These tags will always point to the latest minor release inside of the respective major version.

Let me know if that helped 😄

schosterbarak:
superb! thanks @Deichindianer

PR #1150, opened 2021-04-29
Add CKV_AWS_152 check for cross-zone load balancing

gruebel:

Cross-zone load balancing is disabled by default for Network and Gateway load balancers. For Application LB is enabled by default and can't be changed.


nimrodkor (approved):
💯

PR #1152, opened 2021-05-01
Add type hints to Terraform base checks

gruebel:

Step by step I will go through the code and add type hints to make it easier to understand what the methods are actually doing. I didn't want to create a huge PR with a lot of changes at once to make it easier to review it :)

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy-dpath.*]
+ignore_missing_imports = True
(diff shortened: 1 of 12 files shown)


mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy-dpath.*]
+ignore_missing_imports = True
```
nimrodkor, line 2: @gruebel What's this?

nimrodkor (commented):
I LOVE this @gruebel !

Just one small comment

mypy.ini
gruebel: I use `mypy` for type checking. When more and more of the project has type hints added, then I would also add a github action to check the typing with `mypy` for each PR. And this tells `mypy` to ignore the `dpath` import, because that library doesn't have any type hints included.

nimrodkor (approved):

PR #1154, opened 2021-05-02
Add CKV_AWS_101 check to cloudformation

gruebel:

This check already exists for Terraform, so I also implemented it for CloudFormation. I also wanted to add `CKV_AWS_102`, but you can't create a public accessible Neptune Cluster via CloudFormation.

### DIFF
--- a/tests/cloudformation/checks/resource/aws/test_NeptuneClusterLogging.py
+++ b/tests/cloudformation/checks/resource/aws/test_NeptuneClusterLogging.py
@@ -1,0 +1,25 @@
+import os
+import unittest
+
+from checkov.cloudformation.checks.resource.aws.NeptuneClusterLogging import check
+from checkov.cloudformation.runner import Runner
+from checkov.runner_filter import RunnerFilter
+
+
+class TestNeptuneClusterLogging(unittest.TestCase):
+    def test_summary(self):
+        runner = Runner()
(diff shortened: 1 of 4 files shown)


tests/cloudformation/checks/resource/aws/test_NeptuneClusterLogging.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_summary(self):
+        runner = Runner()
+        current_dir = os.path.dirname(os.path.realpath(__file__))
+
+        test_files_dir = current_dir + "/example_NeptuneClusterLogging"
+        report = runner.run(root_folder=test_files_dir, runner_filter=RunnerFilter(checks=[check.id]))
+        summary = report.get_summary()
+
+        self.assertEqual(summary["passed"], 1)
+        self.assertEqual(summary["failed"], 2)
```
nimrodkor, line 19: Hey @gruebel !

To avoid inversion of logic (happened in the past...), can you set different cluster identifiers for the passed & failed, and check that the failed resources in the report should have failed and the passed resources should have passed?

gruebel: good point, similar to the tests for Terraform checks.

nimrodkor (approved):

PR #1155, opened 2021-05-02
Add type hints to terraform utils and provider

gruebel:

Next batch of type hints, this time for `terrafom.checks.provider` and `terrafom.checks.utils`.

I also moved `convert_cloudformation_conf_to_iam_policy.py` to the cloudformation code base, which is probably the correct place to have it 😉 

Other than that, I also renamed `CustomAttributes` to `CustomTerraformAttributes` in the terraform code base to better differentiate the difference between the original one and the subclass.

### DIFF
--- a/checkov/terraform/checks/provider/base_check.py
+++ b/checkov/terraform/checks/provider/base_check.py
@@ -1,30 +1,22 @@
 from abc import abstractmethod
+from typing import List, Dict, Any
 
 from checkov.common.checks.base_check import BaseCheck
-from checkov.common.multi_signature import multi_signature
+from checkov.common.models.enums import CheckCategories, CheckResult
 from checkov.terraform.checks.provider.registry import provider_registry
 
 
 class BaseProviderCheck(BaseCheck):
-    def __init__(self, name, id, categories, supported_provider):
-        super().__init__(name=name, id=id, categories=categories, supported_entities=supported_provider,
-                         block_type="provider")
+    def __init__(self, name: str, id: str, categories: List[CheckCategories], supported_provider: List[str]) -> None:
+        super().__init__(
+            name=name, id=id, categories=categories, supported_entities=supported_provider, block_type="provider"
+        )
         self.supported_provider = supported_provider
         provider_registry.register(self)
 
-    def scan_entity_conf(self, conf, entity_type):
-        return self.scan_provider_conf(conf, entity_type)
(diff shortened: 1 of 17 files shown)


checkov/terraform/checks/provider/base_check.py
```diff
@@ -1,30 +1,22 @@
...
         raise NotImplementedError()
-
-    @classmethod
-    @scan_provider_conf.add_signature(args=["self", "conf"])
-    def _scan_provider_conf_self_conf(cls, wrapped):
-        def wrapper(self, conf, provider_type=None):
-            # keep default argument for entity_type so old code, that doesn't set it, will work.
-            return wrapped(self, conf)
-
-        return wrapper
```
nimrodkor, line 30: @gruebel Not sure we want to delete this... Do we?

checkov/terraform/graph_builder/graph_components/attribute_names.py
```diff
@@ -4,10 +4,9 @@
 
 
 @dataclass
-class CustomAttributes(CustomAttributes):
-    SOURCE_MODULE = "source_module_"
+class CustomTerraformAttributes(CustomAttributes):
```
nimrodkor, line 7: @gruebel Can we avoid this renaming?

nimrodkor (commented):
Thanks @gruebel !

checkov/terraform/checks/provider/base_check.py
gruebel: I checked current code base and there is no check, which utilises the old way of handling it with `provider_type`. if someone wants to also do something with the `provider_type`, then he/she can use the method `scan_entity_conf()` instead. Or is there any reason, why you want to keep it inside the code base?

checkov/terraform/graph_builder/graph_components/attribute_names.py
gruebel: let me try it again, I had some issues with it, that's why I changed it.

gruebel: 😄 I found a way, please have a look into my new commit

nimrodkor (approved):

checkov/terraform/checks/provider/base_check.py
nimrodkor: There is no check and I find it hard to think of a valid usecase. It seems more relevant for resource checks than provider checks... Let's remove it for now. Thanks!

PR #1158, opened 2021-05-02
Check - CF - IAM Role Public Assume

njgibbon:
Hello,

Implementing this check as already done for terraform.

Terraform check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/IAMRoleAllowsPublicAssume.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html
IAM Assume Role Policy Doc: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html

**License**

### DIFF
--- a/tests/cloudformation/checks/resource/aws/test_IAMRoleAllowsPublicAssume.py
+++ b/tests/cloudformation/checks/resource/aws/test_IAMRoleAllowsPublicAssume.py
@@ -1,0 +1,32 @@
+import os
+import unittest
+
+from checkov.cloudformation.checks.resource.aws.IAMRoleAllowsPublicAssume import check
+from checkov.cloudformation.runner import Runner
+from checkov.runner_filter import RunnerFilter
+
+
+class TestIAMRoleAllowsPublicAssume(unittest.TestCase):
+
+    def test_summary(self):
+        runner = Runner()
+        current_dir = os.path.dirname(os.path.realpath(__file__))
+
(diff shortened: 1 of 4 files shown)


tests/cloudformation/checks/resource/aws/test_IAMRoleAllowsPublicAssume.py
```diff
@@ -0,0 +1,32 @@
...
+        for record in report.failed_checks:
+            self.assertEqual(record.check_id, check.id)
+        
+        for record in report.passed_checks:
+            self.assertEqual(record.check_id, check.id)
+
+        self.assertEqual(summary['passed'], 2)
+        self.assertEqual(summary['failed'], 2)
+        self.assertEqual(summary['skipped'], 0)
+        self.assertEqual(summary['parsing_errors'], 0)
```
nimrodkor, line 28: Hi @njgibbon !

To avoid inversion of logic (happened in the past...), can you set different cluster identifiers for the passed & failed, and check that the failed resources in the report should have failed and the passed resources should have passed?

See a recent example in #1154

njgibbon: Hey @njgibbon  - Ah yes, I like this! Will get on to it in a bit.

nimrodkor (approved):

PR #1161, opened 2021-05-03
Test - TF - IAM Checks - Clean

njgibbon:
Hello,

As I've been going through the CloudFormation and Terraform checks related to IAM I've just noticed a little test debt / cleaning that can be quickly actioned in some cases.

**Minor Changes**
* Changes some test class names to match up to expected name.
*  Removed 'test_IAMRoleAttachedToService.py' because it is duplicate testing IAMRoleAllowsPublicAssume with less coverage and no different cases exercised.
* Changed 'cloudWatchLogGroupRetention' to have upper case file and class name because it is the only check which doesn't.

Are there any other minor things bugging you that I should add in? Maybe specific to the terraform checks / tests? Let me know 🙂 

ItAintMuch

**License**


schosterbarak:
Thanks Nick!

PR #1165, opened 2021-05-04
[CKV_AZURE_109] Update KeyVaultEnablesFirewallRulesSettings.py to expect a properly formated value

htonkovac:
Hello,

I noticed this rule changed recently (it made no sense before). However, this rule is not working properly. It expects a "deny". However, azurerm expects "Deny". If "deny" is given to azurerm it will fail.

I'm not sure if there is anything else I should update. Please bring it to my attention and I will make the changes.

### DIFF
--- a/checkov/terraform/checks/resource/azure/KeyVaultEnablesFirewallRulesSettings.py
+++ b/checkov/terraform/checks/resource/azure/KeyVaultEnablesFirewallRulesSettings.py
@@ -14,7 +14,7 @@ class KeyVaultEnablesFirewallRulesSettings(BaseResourceValueCheck):
         return "network_acls/[0]/default_action"
 
     def get_expected_value(self):
-        return "deny"
+        return "Deny"
 
 
 check = KeyVaultEnablesFirewallRulesSettings()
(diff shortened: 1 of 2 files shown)


gruebel:
you would need to change the tests related to the test
https://github.com/bridgecrewio/checkov/blob/9ec5cb6568b0b4562a906c8a56ffc34424a35da6/tests/terraform/checks/resource/azure/test_KeyVaultEnablesFirewallRulesSettings.py#L75-L77
and this would be great to change from `allow` to `Allow`
https://github.com/bridgecrewio/checkov/blob/9ec5cb6568b0b4562a906c8a56ffc34424a35da6/tests/terraform/checks/resource/azure/test_KeyVaultEnablesFirewallRulesSettings.py#L113-L115

htonkovac:
@gruebel  Thanks. I've updated the unit tests. Is there anything else I should update?

schosterbarak:
Thank you @htonkovac for your contribution. You are welcome to join : https://slack.bridgecrew.io/  the slack community :)

PR #1179, opened 2021-05-06
Check: elasticsearch has node to node encrypted coms

JamesWoolfenden:
Ensure that node to node communication is encrypted - is not by default.
Addresses missing checks as tested here: https://github.com/iacsecurity/tool-compare


JamesWoolfenden:
This ID expects the previous PR to be merged and will pass the id check when that has happened

PR #1182, opened 2021-05-06
Check if multi-az is enabled for RDS. 

kartikp10:
Created CKV_AWS_157 for checking if RDS instances have Multi-AZ enabled. 

- Added the check for CFN and TF. 
- Added tests for default, disabled, and enabled values for CFN and for TF.

More on Multi-AZ deployments here: https://aws.amazon.com/rds/features/multi-az/


JamesWoolfenden (approved):
LGTM - need to fix your failing test for cf

mikeurbanski1 (approved):

PR #1184, opened 2021-05-07
fix ckv_aws_63 to handle conf dict properly

ac-square:

Upon testing the following resource:

```hcl
resource "aws_iam_policy" "policy" {
  name        = "my_policy-${var.aws_account_id}"
  path        = "/"
  description = "My test policy"
  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "*"
      ],
      "Effect": "Allow",
      "Resource": "arn:aws:iam::${var.aws_account_id}:role/admin"
    }
  ]
}
EOF
}
```
Checkov 2.x returns a false positive because it fails to parse the IAM policy properly. This is due to https://github.com/bridgecrewio/checkov/commit/44100303db08412b9ad28505d59de8b21aff2c30 as of Checkov 2.0 which changes the `conf` to a dict, in turn breaking the `json.loads` here in this check.

### DIFF
--- a/checkov/terraform/checks/resource/aws/IAMStarActionPolicyDocument.py
+++ b/checkov/terraform/checks/resource/aws/IAMStarActionPolicyDocument.py
@@ -16,7 +16,7 @@ class IAMStarActionPolicyDocument(BaseResourceCheck):
     def scan_resource_conf(self, conf):
         if 'policy' in conf.keys():
             try:
-                policy_block = json.loads(conf['policy'][0])
+                policy_block = conf['policy'][0]
                 if 'Statement' in policy_block.keys():
                     for statement in force_list(policy_block['Statement']):
                         if 'Action' in statement and \


ac-square:
image

nimrodkor:
Hey @ac-square !

Looking at the policy which failed - its actions are `["*"]`, which is what the policy should have found and failed on in the first place. Am I missing anything?

ac-square:
> Hey @ac-square !
> 
> Looking at the policy which failed - its actions are `["*"]`, which is what the policy should have found and failed on in the first place. Am I missing anything?

That's correct. Pre 2.0, the policy failed against the rule appropriately. Post 2.0, the policy reports a false positive against the rule. 

It didn't log correctly as an error, but if I print out the exception in this rule, it returned `the JSON object must be str, bytes or bytearray, not dict`.  This is seems due to this change https://github.com/bridgecrewio/checkov/commit/44100303db08412b9ad28505d59de8b21aff2c30 as of checkov 2.0.

nimrodkor:
@ac-square Got it, that's good. I'd love to merge it but it fails UT due to the fact these UTs still sending a string instead of the dict it now expects. Can you fix it?

nimrodkor (approved):
Will fix UTs in a separate PR

schosterbarak:
Hey @ac-square, Thank you so much for the contribution and your time debugging this one. If you would like some OSS contributor swag, feel free to reach out via DM at https://slack.bridgecrew.io/ with a delivery address and we’ll send those over :)

PR #1187, opened 2021-05-09
Add CKV2_AWS_22 check to find IAM users with console access

gruebel:

This check is looking for IAM users, who have console access, which is usually discouraged and instead something like AWS SSO or SAML should be used to manage human user access to AWS.


schosterbarak:
This is an awesome usage of checkov 2 query format. Good job @gruebel

PR #1195, opened 2021-05-11
Restrict string eval() calls in variable rendering

rotemavni:
In the variable rendering process we are trying to evaluate strings which might contain terraform function calls or python expressions.
In order to make sure the call to python's `eval` function is as safe as possible without losing the ability to evaluate terraform functions we can use the `compile` function to generate a code object, and check if that code object contains any names that are within python's builtin names (containing `__`).
If such a name exists, the string is not evaluated and a custom `BuiltinError` is raised.
Added unit tests to check some interesting cases.

### DIFF
--- a/checkov/terraform/variable_rendering/evaluate_terraform.py
+++ b/checkov/terraform/variable_rendering/evaluate_terraform.py
@@ -1,7 +1,9 @@
+import ast
 import re
+from copy import deepcopy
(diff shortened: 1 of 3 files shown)


nimrodkor (approved):
I like this

metahertz:
Thanks @rotemavni thats awesome :)

PR #1197, opened 2021-05-11
Skip scanning excluded paths in platform integration

rotemavni:
In platform-integration, added a function that calls for the repository settings and look for excluded paths. 
Added the excluded path as another member to the runner filter and use it to match paths in `filter_ignored_directories` function. 
This also prevents persisting the excluded paths to s3.

### DIFF
--- a/checkov/arm/runner.py
+++ b/checkov/arm/runner.py
@@ -32,7 +32,7 @@ class Runner(BaseRunner):
 
         if root_folder:
             for root, d_names, f_names in os.walk(root_folder):
-                filter_ignored_directories(d_names)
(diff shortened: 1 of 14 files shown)


schosterbarak (approved):

schosterbarak:
why can't we use the existing d_names?

rotemavni:
> why can't we use the existing d_names?

`d_names` are the sub directories we are iterating when scanning a directory.
Assuming a bridgecrew user has excluded `test` in the repo setting in bc platform, we want to be able to skip scanning and uploading the files under the matching excluded directories.

PR #1213, opened 2021-05-13
do equals comparisons as strings

mikeurbanski1:

This PR handles type conversion for Equals (and NotEquals) graph checks. For example, the HTTP -> HTTPS redirect check fails if you specify the port as an int. But this is valid in TF.

@rotemavni does this need to be updated somewhere in the platform?


rotemavni (approved):
@mikeurbanski1 I love this approach. No need to update in the platform.

PR #1214, opened 2021-05-13
do not copy line numbers during param replacement

mikeurbanski1:

This PR fixes CFN param replacement. It would copy in the line numbers of the parameter dict entry. The result was that the resource lines would range from the start of the param to the end of the resource.

@tronxd I think there might be nuances between CFN and JSON parsing that need to be handled here. I am not sure. Please take a look.


tronxd (approved):
This did not break buildtime remediations of both YML and JSON CFN templates. GJ!

PR #1215, opened 2021-05-14
Incorporate block reference in CKV_AWS_143 check

### LINKED ISSUE #1208: False positive for CKV_AWS_143 ("Ensure S3 bucket has lock configuration enabled by default")
**Describe the bug**

Checkov expects argument `object_lock_configuration` to be an object, i.e.

```hcl
object_lock_configuration = {
  object_lock_enabled = "Enabled"
}
```

Terraform works with the above configuration, but when also declaring rules for the object lock configuration, it expects a block instead, e.g.
```hcl
object_lock_configuration {
  object_lock_enabled = "Enabled"

  rule {
    default_retention {
      mode = "GOVERNANCE"
      days = 366
    }
  }
}
```

**Expected behavior**
Checkov should pass for a `object_lock_configuration` argument block.

 - OS: macOS Big Sur 11.3.1
 - Checkov Version: 2.0.135
 - Terraform version: v0.14.8

gruebel:

Fixes #1208. There is a high chance that also other checks have a similar problem. Sometimes configuration can be either written as an object or block. I also took a look into lark parser, but I don't think it is a good idea to change the logic there.

### DIFF
--- a/checkov/terraform/checks/resource/aws/S3BucketObjectLock.py
+++ b/checkov/terraform/checks/resource/aws/S3BucketObjectLock.py
@@ -1,25 +1,26 @@
+from typing import Dict, List, Any
+
 from checkov.common.models.enums import CheckCategories, CheckResult
 from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
 
 
 class S3BucketObjectLock(BaseResourceCheck):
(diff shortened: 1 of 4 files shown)


schosterbarak:
Thank you Anton!

PR #1226, opened 2021-05-18
boto3 dependency install latest patch version

### LINKED ISSUE #1211: boto3 is fixed at the patch level version
**Is your feature request related to a problem? Please describe.**
free boto3 dependency patch version.

**Describe the solution you'd like**
replace the line here:
https://github.com/bridgecrewio/checkov/blob/master/Pipfile#L29

with 
```
boto3 = "==1.17.*"
```
**Describe alternatives you've considered**
there are no alternatives as the patch version i don't see why is locked.
it can cause conflicts with already installed boto3 library 

**Additional context**

Raffaello:

Fixes #1211


nimrodkor (approved):
Thanks @Raffaello !

nimrodkor:
@Raffaello I believe we'll also need to update https://github.com/bridgecrewio/checkov/blob/master/setup.py for the package to be updated

nimrodkor:
I see UTs are failing on pipenv install. I'll have a go at this. Thanks for the effort @Raffaello !

PR #1237, opened 2021-05-23
Improve collect skip comments performance

nimrodkor:

1. Avoid double logging for no reason
2. Filter optional comment lines to reduce number of calls to re.match()
3. Use direct access instead of dpath.search inside the context to reduce run time
4. Handle case of non-str in evaluated_string_parts to succeed in joining to new string


rotemavni (approved):
LGTM

PR #1238, opened 2021-05-23
Performance improvements and bugfixes

rotemavni:
1. Fix CKV_AWS_1 (`AdminPolicyDocument`) to support list statements
2. Avoid removing leading zeros from numeric strings in graph.
3. Avoid inserting empty keys into graph vertices attributes

### DIFF
--- a/checkov/terraform/checks/utils/utils.py
+++ b/checkov/terraform/checks/utils/utils.py
@@ -206,7 +206,7 @@ def decode_graph_property_value(value: Any, leave_str: bool = False) -> Any:
     if value.startswith('"') and value.endswith('"'):
         value = value[1:-1]
     if not leave_str:
-        if value.isnumeric() and not value.startswith("0"):
+        if value.isnumeric():
             value = int(value)
         else:
             try:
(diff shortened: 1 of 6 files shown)


checkov/terraform/checks/utils/utils.py
```diff
@@ -206,7 +206,7 @@ def decode_graph_property_value(value: Any, leave_str: bool = False) -> Any:
     if value.startswith('"') and value.endswith('"'):
         value = value[1:-1]
     if not leave_str:
-        if value.isnumeric():
+        if value.isnumeric() and not value.startswith("0"):
```
nimrodkor, line 209: @rotemavni Can we leverage the `leave_str` bool instead?

nimrodkor (commented):
@rotemavni Nice! 1 small issue...

checkov/terraform/checks/utils/utils.py
rotemavni: I'm afraid we can't leverage it in that context.

nimrodkor (approved):

PR #1241, opened 2021-05-23
Add rds iam auth check

gruebel:

I first tried to put both checks into the same, but the handling is quite awkward, so I split them into two. In general it is recommended to use IAM authentication for RDS databases and cluster, because you don't need to inject the password anymore and have to temporarily invoke a session to connect to the database.

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/RDSIAMAuthentication.py
+++ b/checkov/cloudformation/checks/resource/aws/RDSIAMAuthentication.py
@@ -1,0 +1,26 @@
+from checkov.cloudformation.parser.node import dict_node
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class RDSIAMAuthentication(BaseResourceValueCheck):
+    def __init__(self) -> None:
+        name = "Ensure RDS database has IAM authentication enabled"
+        id = "CKV_AWS_161"
+        supported_resources = ["AWS::RDS::DBInstance"]
+        categories = [CheckCategories.IAM]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self) -> str:
+        return "Properties/EnableIAMDatabaseAuthentication"
+
+    def scan_resource_conf(self, conf: dict_node) -> CheckResult:
+        # IAM authentication is only supported for MySQL and PostgreSQL
+        engine = conf.get("Properties", {}).get("Engine", {})
+        if engine not in ("mysql", "postgres"):
+            return CheckResult.UNKNOWN
+
+        return super().scan_resource_conf(conf)
+
+
+check = RDSIAMAuthentication()
--- a/checkov/terraform/checks/resource/aws/RDSClusterIAMAuthentication.py
+++ b/checkov/terraform/checks/resource/aws/RDSClusterIAMAuthentication.py
@@ -1,0 +1,17 @@
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+from checkov.common.models.enums import CheckCategories
+
+
+class RDSClusterIAMAuthentication(BaseResourceValueCheck):
+    def __init__(self) -> None:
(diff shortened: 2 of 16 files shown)


checkov/cloudformation/checks/resource/aws/RDSIAMAuthentication.py
```diff
@@ -0,0 +1,26 @@
+from checkov.cloudformation.parser.node import dict_node
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class RDSIAMAuthentication(BaseResourceValueCheck):
+    def __init__(self) -> None:
+        name = "Ensure RDS database has IAM authentication enabled"
+        id = "CKV_AWS_161"
+        supported_resources = ["AWS::RDS::DBInstance"]
```
schosterbarak, line 10: if a cluster has it enabled, does the instance get it too automatically?

gruebel: yeah correct, in a cluster it is only possible to configure this on cluster level not instance level. Therefore I check for the engine type. Additionally it is only supported for mysql und postgresql.

schosterbarak: so we should only test an instance if the cluster has not configured it already? is that a graph query?

gruebel: no need, I added one check for the cluster and one for normal RDS Instanzen. In Terraform it is a bit more clear, what is what, because Terraform differentiate between them. In CloudFormation the main difference between a cluster instance and a normal instance is the cluster identifier and the engine type, if set.

nimrodkor: Hey @gruebel !

I'm hesitant to create this check for the DB instances - you can't create an instance without a cluster, and this should be on the cluster I believe. Am I wrong?

gruebel: Hey @nimrodkor 😄 
I think my explanation was not clear enough 😅
CloudFormation works a bit different compared to Terraform here. A single RDS instance and a cluster instance are provisioned with the same CloudFormation resource. When you want to deploy an instance to a cluster you have to set the cluster identifier property to reference the cluster. If you don't then it is a single RDS instance.
For cluster instances the engine type can only be something with aurora.

nimrodkor (approved):
After a short call with @gruebel , I understood we need to handle 3 different cases:
1. Aurora clusters - should have it set
2. Aurora cluster db instances - should be skipped, handled by the cluster
3. Non-Aurora RDS DB instances - should have it set

in tf the 3 resources are different, so we can check only the `aws_db_instance` and the `aws_rds_cluster` skipping the `aws_rds_cluster_instance`.
In CF the 3 resources are actually 2 - `AWS::RDS::DBCluster` which should always have that flag and `AWS::RDS::DBInstance` - which might or might not. 

That's the reason for 4 checks that should be 2. LGTM 👏

checkov/terraform/checks/resource/aws/RDSClusterIAMAuthentication.py
```diff
@@ -0,0 +1,17 @@
...
+class RDSClusterIAMAuthentication(BaseResourceValueCheck):
+    def __init__(self) -> None:
+        name = "Ensure RDS cluster has IAM authentication enabled"
+        id = "CKV_AWS_162"
+        supported_resources = ["aws_rds_cluster"]
+        categories = [CheckCategories.IAM]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self) -> str:
+        return "iam_database_authentication_enabled"
```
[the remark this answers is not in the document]
gruebel: thanks, you are right 🙈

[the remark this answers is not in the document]
gruebel: retiring `CKV_AWS_128` #4350

PR #1248, opened 2021-05-26
Config file feature

kartikp10:
This pull request adds the ability to use checkov with a configuration file.

Noteworthy changes:

1. ConfigArgParse is now the default argument parsing library. 
2. Adds support for --config-file flag to specify a yaml configuration file. 
3. Adds support for --show-config to show where the values are being parsed from (inline args / confg file / env variables)
4. Adds support for --create-config
5. --check and --skip-check can now be used together. 

TODOs:

- [x] Fix new unit tests.
- [x] Add integration tests.
- [x] Update docs.


JamesWoolfenden (approved):
LGTM

schosterbarak:
very elegant. I love it.  can we add an integration test too?

schosterbarak:
can we also update docs as part of this change?

kartikp10:
@schosterbarak Thanks! Yeah, I'll do both those things. Adding to 'TODOs'

schosterbarak:
@kartikp10 I'm not sure multi env is a must since you can just create multiple files

checkov/common/util/config_utils.py
```diff
@@ -0,0 +1,22 @@
...
+from pathlib import Path
+
+
+def config_file_paths(dir_path):
+    return [os.path.join(dir_path, '.checkov.yaml'), os.path.join(dir_path, '.checkov.yml')]
+
+
+def get_default_config_paths(argv):
+    """
+    Checkov looks for .checkov.yml or .checkov.yaml file in the directory (--directory) against which it is run.
```
mikeurbanski1, line 12: What about checking the current working directory? I know we talked about this, but I can't remember if we decided it made sense or not.

kartikp10: Hey @mikeurbanski1, I was thinking about this use case some more. CWD might be handy for IDE extensions. Thoughts?

mikeurbanski1: Just closing the loop on this - Kartik will add CWD. Priority will be:

--config-file arg > --directory > CWD > home

schosterbarak:
@gruebel I think you'll like this one too :)

gruebel:
Yeah, definitely waiting for this feature 🎉

nimrodkor:
@kartikp10 If it is ready for review, please mark it accordingly :)

checkov/common/util/config_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    If that does not have the config file, the current working directory is checked followed by checking the user's
+    home directory is searched.
+    :param argv: List of CLI args from sys.argv.
+    :return: List of default config file paths.
+    """
+    home_paths = config_file_paths(Path.home())
```
nimrodkor, line 18: @kartikp10 Can we parse this **after** we parsed the directory arg using configParse, instead of knitpicking through argv by ourselves?

nimrodkor (commented):
Looks really good, just 1 small comment

checkov/common/util/ext_argument_parser.py
```diff
@@ -0,0 +1,47 @@
+import configargparse
+
+from checkov.common.util.type_forcers import convert_str_to_bool
+
+
+class ExtArgumentParser(configargparse.ArgumentParser):
+
+    def write_config_file(self, parsed_namespace, output_file_paths, exit_after=False):
```
mikeurbanski1, line 8: +100 points

mikeurbanski1 (approved):

checkov/common/util/config_utils.py
kartikp10: Hey @nimrodkor, that was my thought initially as well. The downsides of adding new "default_config_files" after the parser has been initialized are as follows:
- _default_config_files is a protected attribute ( not a blocker, we can define a setter in the extended class )
-  the --show-config command will not be able to print accurate information if the user runs it like `checkov -d terragoat --show-config` . For example:
```
# existing design. default paths are defined when the parser is initialized
❯ ckv --show-config -d terragoat
Command Line Args:   --show-config -d terragoat
Config File (terragoat/.checkov.yml):
  branch:            master
  check:             ['CKV_AWS_144']
  download-external-modules:False
  evaluate-variables:True
  external-modules-download-path:.external_modules
  framework:         all
  output:            cli

# if the default paths are added after arguments are parsed 
❯ ckv --show-config -d terragoat
Command Line Args:   --show-config -d terragoat
Defaults:
  --output:          cli
  --framework:       all
  --branch:          master
  --download-external-modules:False
  --external-modules-download-path:.external_modules
  --evaluate-variables:True
```
This is why I'd like to define default paths before the parser is initialized. Let me know if this makes sense.

nimrodkor: @kartikp10 I'm not sure I follow... Won't it include simply switching the order in `main.py` of lines 47 & 48? I think I'm still missing something, let's talk 🙂

README.md
```diff
@@ -303,6 +303,70 @@ Default is `CKV_IGNORED_DIRECTORIES=node_modules,.terraform,.serverless`
...
+
+Users can also use the `--show-config` flag to view all the args and settings and where they came from i.e. commandline, config file, environment variable or default. For example:
+```sh
+checkov --show-config
+```
+Will display:
+```sh
+Command Line Args:   --show-config
+Environment Variables:
+  BC_API_KEY:        your-api-key
```
schosterbarak, line 358: this is really cool. how does it know where to take this key from?

kartikp10: Hey Barak, when the option is added to the parser, the environment variable name can be defined in ConfigArgParse like so:
`parser.add('--bc-api-key', help='Bridgecrew API key', env_var='BC_API_KEY')`

PR #1249, opened 2021-05-26
fix:Check neg on attribute allow_blob_public_access for Azure storage account

JamesWoolfenden:
https://docs.bridgecrew.io/docs/ensure-that-storage-accounts-disallow-public-access, the “Ensure that storage accounts disallow public access” should flag on allow_blob_public_access = true, NOT enable_https_traffic_only.  You can see that the below 2 issues are flagging on the same line.

attribute checko was wrong and of the wrong type. not sure how the test passed.


mikeurbanski1 (approved):

nimrodkor (approved):
The tests match the behavior of `allow_blob_public_access` with the desired behavior of `enable_https_traffic_only`

PR #1255, opened 2021-05-27
Handle empty templates

mikeurbanski1:

Handles a case common with Helm where generated files look like the example below. The blank blocks made the parser exit and no scan was performed.

```
---
# Source: a/templates/pod_disruption_budget.yaml
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
...
---
# Source: a/templates/service.yaml
apiVersion: v1
kind: Service
...

---
# Source: a/templates/deployment.yaml
---

---
# Source: a/templates/cron_job.yaml
```


JamesWoolfenden (approved):
LGTM

PR #1259, opened 2021-05-30
Add type hints to terraform.checks_infra

gruebel:

This time it is the complete `terraform.checks_infra` part, I think the biggest change was done in `and_connection_solver.py`, because half of the code is not needed 😄

### DIFF
--- a/checkov/common/graph/checks_infra/registry.py
+++ b/checkov/common/graph/checks_infra/registry.py
@@ -1,20 +1,32 @@
+from typing import List, Dict, Any
+
+from networkx import DiGraph
+
+from checkov.common.graph.checks_infra.base_check import BaseGraphCheck
+from checkov.common.graph.checks_infra.base_parser import BaseGraphCheckParser
 from checkov.common.models.enums import CheckResult
 from checkov.runner_filter import RunnerFilter
 
 
 class BaseRegistry:
-    def __init__(self, parser):
-        self.checks = []
+    def __init__(self, parser: BaseGraphCheckParser) -> None:
+        self.checks: List[BaseGraphCheck] = []
         self.parser = parser
 
     def load_checks(self):
         raise NotImplementedError
 
-    def run_checks(self, graph_connector, runner_filter: RunnerFilter):
+    def run_checks(
+        self, graph_connector: DiGraph, runner_filter: RunnerFilter
+    ) -> Dict[BaseGraphCheck, List[Dict[str, Any]]]:
         check_results = {}
         for check in self.checks:
-            if runner_filter.checks and check.id not in runner_filter.checks \
-                    or runner_filter.skip_checks and not runner_filter.should_run_check(check.id):
+            if (
+                runner_filter.checks
+                and check.id not in runner_filter.checks
+                or runner_filter.skip_checks
+                and not runner_filter.should_run_check(check.id)
+            ):
                 continue
             passed, failed = check.run(graph_connector)
             check_result = self._process_check_result(passed, [], CheckResult.PASSED)
@@ -25,5 +37,5 @@ class BaseRegistry:
     @staticmethod
     def _process_check_result(results, processed_results, result):
         for vertex in results:
-            processed_results.append({'result': result, 'entity': vertex})
+            processed_results.append({"result": result, "entity": vertex})
         return processed_results
--- a/checkov/common/graph/checks_infra/registry.py
+++ b/checkov/common/graph/checks_infra/registry.py
@@ -21,11 +21,8 @@ class BaseRegistry:
     ) -> Dict[BaseGraphCheck, List[Dict[str, Any]]]:
         check_results = {}
         for check in self.checks:
-            if (
-                runner_filter.checks
-                and check.id not in runner_filter.checks
-                or runner_filter.skip_checks
-                and not runner_filter.should_run_check(check.id)
+            if (runner_filter.checks and check.id not in runner_filter.checks) or (
+                runner_filter.skip_checks and not runner_filter.should_run_check(check.id)
             ):
                 continue
             passed, failed = check.run(graph_connector)
--- a/checkov/terraform/checks_infra/checks_parser.py
+++ b/checkov/terraform/checks_infra/checks_parser.py
@@ -73,10 +73,8 @@ class NXGraphCheckParser(BaseGraphCheckParser):
             resource_type = raw_check.get("resource_types", [])
             if (
                 not resource_type
-                or isinstance(resource_type, str)
-                and resource_type.lower() == "all"
-                or isinstance(resource_type, list)
-                and resource_type[0].lower() == "all"
+                or (isinstance(resource_type, str) and resource_type.lower() == "all")
+                or (isinstance(resource_type, list) and resource_type[0].lower() == "all")
             ):
                 check.resource_types = resources_types
             else:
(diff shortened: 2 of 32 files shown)


checkov/common/graph/checks_infra/registry.py
```diff
@@ -1,20 +1,32 @@
...
+    ) -> Dict[BaseGraphCheck, List[Dict[str, Any]]]:
         check_results = {}
         for check in self.checks:
-            if runner_filter.checks and check.id not in runner_filter.checks \
-                    or runner_filter.skip_checks and not runner_filter.should_run_check(check.id):
+            if (
+                runner_filter.checks
+                and check.id not in runner_filter.checks
+                or runner_filter.skip_checks
+                and not runner_filter.should_run_check(check.id)
```
nimrodkor, line 28: @gruebel `and` and `or` inline is not a good idea... Mind adding parenthesis to make the order more explicit?

checkov/terraform/checks_infra/checks_parser.py
```diff
@@ -1,114 +1,127 @@
...
 
         else:
             resource_type = raw_check.get("resource_types", [])
-            if not resource_type or isinstance(resource_type, str) and resource_type.lower() == 'all' or\
-                    isinstance(resource_type, list) and resource_type[0].lower() == 'all':
+            if (
+                not resource_type
+                or isinstance(resource_type, str)
+                and resource_type.lower() == "all"
+                or isinstance(resource_type, list)
```
nimrodkor, line 78: @gruebel Sam here

checkov/terraform/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -7,50 +10,59 @@
 
 
 class BaseAttributeSolver(BaseSolver):
-    operator = ''
+    operator = ""
 
-    def __init__(self, resource_types, attribute, value):
+    def __init__(self, resource_types: List[str], attribute: Optional[str], value: Any) -> None:
```
nimrodkor, line 15: @gruebel Why is `attribute` optional? I believe it should be required here

checkov/terraform/checks_infra/solvers/attribute_solvers/exists_attribute_solver.py
```diff
@@ -1,13 +1,14 @@
...
     operator = Operators.EXISTS
 
-    def __init__(self, resource_types, attribute, value):
-        super().__init__(resource_types=resource_types,
-                         attribute=attribute, value=value)
+    def __init__(self, resource_types: List[str], attribute: Optional[str], value: Any) -> None:
+        super().__init__(resource_types=resource_types, attribute=attribute, value=value)
 
-    def _get_operation(self, vertex, attribute):
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:
```
nimrodkor, line 13: @gruebel Pretty sure attribute shouldn't be Optional in this method and its inheritances too

checkov/terraform/checks_infra/solvers/complex_solvers/and_solver.py
```diff
@@ -7,15 +10,13 @@
 class AndSolver(BaseComplexSolver):
     operator = Operators.AND
 
-    def __init__(self, solvers, resource_types):
+    def __init__(self, solvers: List[BaseSolver], resource_types: List[str]) -> None:
         super().__init__(solvers, resource_types)
 
-    def _get_operation(self, *args):
+    def _get_operation(self, *args: Any) -> Any:
```
nimrodkor, line 16: @gruebel `Any` doesn't say much... What are the cases?

checkov/terraform/checks_infra/solvers/complex_solvers/base_complex_solver.py
```diff
@@ -1,28 +1,31 @@
...
         self.solvers = solvers
         self.resource_types = resource_types
         super().__init__(SolverType.COMPLEX)
 
-    def _get_operation(self, *args, **kwargs):
+    def _get_operation(self, *args: Any, **kwargs: Any) -> Any:
         raise NotImplementedError()
 
-    def _get_negative_op(self, *args):
+    def _get_negative_op(self, *args: Any) -> bool:
```
nimrodkor, line 22: @gruebel Pretty sure this doesn't return true / false

checkov/terraform/checks_infra/solvers/connections_solvers/and_connection_solver.py
```diff
@@ -1,43 +1,36 @@
...
-        failed.extend(failed_connections)
-        passed = [p for p in passed if
-                              p[CustomAttributes.ID] not in [f[CustomAttributes.ID] for f in failed]]
-        passed = [p for p in passed if
-                  p[CustomAttributes.ID] in [pt[CustomAttributes.ID] for pt in passed_attributes] and p[
-                      CustomAttributes.ID] in [pt[CustomAttributes.ID] for pt in passed_connections]]
+            passed.extend(passed_solver)
+            failed.extend(failed_solver)
+
+        passed = [p for p in passed if p[CustomAttributes.ID] not in [f[CustomAttributes.ID] for f in failed]]
```
nimrodkor, line 31: @gruebel Nice!
I see we're using this twice.
1. Can we reuse?
2. Can we compute `[f[CustomAttributes.ID] for f in failed]` once, instead of for every passed vertex?

nimrodkor (commented):
Nice work @gruebel !

checkov/terraform/checks_infra/solvers/connections_solvers/and_connection_solver.py
gruebel: much better, I can just remove the first time it is used 😄

nimrodkor: @gruebel That's not exactly what I meant... we need it twice. I would however split this:
```python
```

to 
```python
```

checkov/terraform/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
gruebel: as I mentioned in the call, this is related to connection solver, where `attribute` is `None`.

checkov/terraform/checks_infra/solvers/attribute_solvers/exists_attribute_solver.py
gruebel: same here, this is related to connection solver, where `attribute` is `None`.

checkov/terraform/checks_infra/solvers/complex_solvers/base_complex_solver.py
gruebel: yeah, you are right 😄

checkov/terraform/checks_infra/solvers/complex_solvers/and_solver.py
gruebel: the typing should be refined here for all the `_get_operation` methods, but to do that it would be great to actually. not use `*args` and `**kwargs` at all. I would recommend to do it in separate step.

checkov/terraform/checks_infra/solvers/connections_solvers/and_connection_solver.py
gruebel: I adjusted it and had to a bit of extra work

gruebel:
@nimrodkor please have a look at my comments and the code change :)

nimrodkor (approved):
Nice!

PR #1261, opened 2021-05-31
CF - Check - ECRImageScanning

njgibbon:
Hello,

Implementing this check as already done for Terraform. There was a placeholder because a previous version of CloudFormation did not include this functionality.

* Resolved Check ID conflict. Made sure TF and CF check for this could both live with the same ID.

TF Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/ECRImageScanning.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecr-repository-imagescanningconfiguration.html

**License**

### DIFF
--- a/checkov/terraform/checks/resource/aws/ECRImageScanning.py
+++ b/checkov/terraform/checks/resource/aws/ECRImageScanning.py
@@ -5,7 +5,7 @@ from checkov.common.models.enums import CheckCategories
 class ECRImageScanning(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure ECR image scanning on push is enabled"
-        id = "CKV_AWS_162"
+        id = "CKV_AWS_163"
         supported_resources = ['aws_ecr_repository']
         categories = [CheckCategories.GENERAL_SECURITY]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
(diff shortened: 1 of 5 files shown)


checkov/terraform/checks/resource/aws/ECRImageScanning.py
```diff
@@ -5,7 +5,7 @@
 class ECRImageScanning(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure ECR image scanning on push is enabled"
-        id = "CKV_AWS_161"
+        id = "CKV_AWS_163"
```
nimrodkor, line 8: @njgibbon Is `CKV_AWS_161` used for any other check? I'm not sure it is

njgibbon: Hey @nimrodkor  - There actually isn't if you check the generated docs in the main branch. But, turns out we are in a slightly awkward state somehow as the docs haven't updated correctly...

I believe changes went in yesterday that all individually passed their checks. But, caused a conflict as they went in. 

https://github.com/bridgecrewio/checkov/pull/1241/files

By putting this change through. This should all be resolved with no conflict on 161, 162 or 163.

nimrodkor (approved):

PR #1264, opened 2021-06-03
Fix dynamo encryption check for default encryption

mikeurbanski1:

Dynamo tables are encrypted by default, so this check should only look for `enabled = false`.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table

### DIFF
--- a/checkov/terraform/checks/resource/aws/DynamoDBTablesEncrypted.py
+++ b/checkov/terraform/checks/resource/aws/DynamoDBTablesEncrypted.py
@@ -1,8 +1,8 @@
 from checkov.common.models.enums import CheckCategories
-from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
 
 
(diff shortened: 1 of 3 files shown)


nimrodkor:
@mikeurbanski1 According to the docs, when it is set to `false` it will still be encrypted IMO:
> enabled - (Required) Whether or not to enable encryption at rest using an AWS managed KMS customer master key (CMK).

[source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#enabled)

mikeurbanski1:
> @mikeurbanski1 According to the docs, when it is set to `false` it will still be encrypted IMO:
> 
> > enabled - (Required) Whether or not to enable encryption at rest using an AWS managed KMS customer master key (CMK).
> 
> [source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#enabled)

You are right. This check doesn't really do anything. We need to rename it to specifically mention KMS, and then lower the severity I guess.

mikeurbanski1:
Oh, it's already named like this in Checkov. We just need to fix the platform. :|

PR #1273, opened 2021-06-07
Bugfix -  Relative file path calculation in plan runner

tronxd:


checkov/terraform/plan_runner.py
```diff
@@ -59,19 +59,19 @@ def run(self, root_folder=None, external_checks_dir=None, files=None, runner_fil
                         continue
                     self.tf_definitions = tf_definitions
                     self.template_lines = template_lines
-                    self.check_tf_definition(report, runner_filter)
+                    self.check_tf_definition(report,root_folder, runner_filter)
```
nimrodkor, line 62: @tronxd Please format this file

Pipfile
```diff
@@ -28,7 +28,7 @@ dpath = ">=1.5.0,<2"
 pyyaml = ">=5.4.1"
 boto3 = "==1.17.*"
 GitPython = "*"
-six = "==1.15.0"
+six = "==1.16.0"
```
nimrodkor, line 31: @tronxd Why is this required?

nimrodkor (commented):
@tronxd Seems to be breaking the `test_plan_runner` tests. Can you fix those?

Pipfile
tronxd: @nimrodkor It has a conflict in pipenv when trying to install with 1.15 (guess that six 1.16 is pinned in one of our deps)

nimrodkor (changes requested):
I prefer not to modify the behavior of the existing runners. Can we avoid that?

PR #1274, opened 2021-06-08
added normalisation of api-key and repo-id for file and dir scan

eurogig:
Due to fixing some of the oddities in the onboarding, the multiple sources of where a bc_api_key could be retrieved from AND the multiple ways a repo_id could be either passed in or defaulted to made it difficult to track what it would be when and where.

Added functions to bc_integration to make that decision early in main.

Also when using the cli with out a github integration the flow was going down a custom path.  Getting the above correct should mean it always uses the same paths regardless for file or directory specified scan.  

This is just adding and commenting out.  If it works consistently there is a bunch of stuff in bc_integration that can be striped out.

This needs a new test in the workflow but I cannot add a secret in order to test the api_key part so I'll submit a new PR with the tests once that is in place.  Please review but don't approve until I can submit the tests for this in advance.

### DIFF
--- a/checkov/common/bridgecrew/platform_integration.py
+++ b/checkov/common/bridgecrew/platform_integration.py
@@ -357,6 +357,34 @@ class BcPlatformIntegration(object):
             if self.is_integration_configured():
                 self._upload_run(args, scan_reports)
 
+# Added this to generate a default repo_id for cli scans for upload to the platform 
+# whilst also persisting a cli repo_id into the object
+    def persist_bc_api_key(self, args):
+        if args.bc_api_key:
+            self.bc_api_key=args.bc_api_key
+        else: 
+            # get the key from file
+            self.bc_api_key=read_key()
+        return self.bc_api_key    
+
+# Added this to generate a default repo_id for cli scans for upload to the platform 
+# whilst also persisting a cli repo_id into the object
+    def persist_repo_id(self, args):
+        if args.repo_id is None:
+            if BC_FROM_BRANCH:
+                self.repo_id = BC_FROM_BRANCH
+            if args.directory:
+                basename = path.basename(os.path.abspath(args.directory[0]))
+                self.repo_id = "cli_repo/" + basename
+            if args.file:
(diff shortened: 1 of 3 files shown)


checkov/common/bridgecrew/platform_integration.py
```diff
@@ -357,6 +357,34 @@ def get_report_to_platform(self, args, scan_reports):
...
+            self.bc_api_key=read_key()
+        return self.bc_api_key    
+
+# Added this to generate a default repo_id for cli scans for upload to the platform 
+# whilst also persisting a cli repo_id into the object
+    def persist_repo_id(self, args):
+        if args.repo_id is None:
+            if BC_FROM_BRANCH:
+                self.repo_id = BC_FROM_BRANCH
+            if args.directory:
```
nimrodkor, line 376: @eurogig Should this be `elif`?

checkov/common/runners/runner_registry.py
```diff
@@ -85,8 +85,8 @@ def print_reports(self, scan_reports, config, url=None):
                 print(json.dumps(report_jsons[0], indent=4))
             else:
                 print(json.dumps(report_jsons, indent=4))
-        if config.output == "cli":
-            bc_integration.get_report_to_platform(config,scan_reports)
+        #if config.output == "cli":
```
nimrodkor, line 88: @eurogig On purpose? If so - delete it, don't leave it commented out

nimrodkor (approved):
👍

checkov/common/bridgecrew/platform_integration.py
eurogig: Hmmm... that would be clearer perhaps.

checkov/common/runners/runner_registry.py
eurogig: Ok will do

mikeurbanski1 (approved):

kartikp10:
Looks good!

kartikp10 (approved):

PR #1275, opened 2021-06-08
Added tests for bc-api-key

eurogig:
Two tests. One -d and one -f to test with bc_api_key


metahertz (commented):
I don't think these will correctly test the bc_integration logic if `-o json` as it will fail to run past the `if output==cli` statements?

kartikp10:
Hey @eurogig, I don't think the code review link is stored in the json output. I suppose a good test would be to ensure that the cli output has a link to code reviews. Let me know what you think.

PR #1283, opened 2021-06-10
handle case when from_port and to_port are both 0 (all ports)

mikeurbanski1:

Handles the port checks when you specify `0` for both from_port and to_port.


JamesWoolfenden (approved):
LGTM, but its not valid to call an sg "sg-" anything but i dont think it matters here.

PR #1285, opened 2021-06-11
CF - Check - APIGatewayCacheEnable

njgibbon:
Hello,

Implementing this check for CF as already done in TF.

Terraform check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/APIGatewayCacheEnable.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-cacheclusterenabled

**License**


checkov/cloudformation/checks/resource/aws/APIGatewayCacheEnable.py
```diff
@@ -0,0 +1,19 @@
+from checkov.cloudformation.checks.resource.base_resource_value_check import BaseResourceValueCheck
+from checkov.common.models.enums import CheckCategories
+from checkov.common.models.consts import ANY_VALUE
+
+
+class APIGatewayCacheEnable(BaseResourceValueCheck):
+    
+    def __init__(self):
+        name = "Ensure API Gateway caching is enabled"
```
schosterbarak, line 9: Thanks Nick! 
Does api gateway have a default behavior for this one?

njgibbon: Hey @schosterbarak . In CloudFormation it's not a required field. The default, if not set, is for the stage cache to be disabled. https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-caching.html

schosterbarak: thanks!

PR #1287, opened 2021-06-12
refactor external module loading

gruebel:

The biggest change I did was to restructure on where to clone the module repositories. Instead of just using one folder for the best suitable version of module, now each module version, which is used will get its own folder and be checked. The result is something like this:
```shell
github.com/terraform-aws-modules/terraform-aws-security-group/v3.18.0
github.com/terraform-aws-modules/terraform-aws-security-group/v4.0.0
github.com/terraform-aws-modules/terraform-aws-security-group/HEAD
```
the `HEAD` version of a module is the latest commit on the default branch.

Module references with SSH should also work, but I will add some tests to make sure it really works 😄

### DIFF
--- a/checkov/terraform/module_loading/loaders/git_loader.py
+++ b/checkov/terraform/module_loading/loaders/git_loader.py
@@ -1,4 +1,7 @@
 import os
+import re
+from dataclasses import dataclass
+from pathlib import Path
 
 from checkov.common.goget.github.get_git import GitGetter
 from checkov.terraform.module_loading.content import ModuleContent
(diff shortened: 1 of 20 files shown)


gruebel:
Would be great to get some feedback on how you like it or if I should change some parts? I also still have some problems with a graph check test, because it looks under the wrong folder name for the Terraform template files.

checkov/terraform/module_loading/loaders/local_path_loader.py
```diff
@@ -10,8 +10,12 @@ def __init__(self) -> None:
         self.is_external = False
 
     def _is_matching_loader(self) -> bool:
-        return self.module_source.startswith("./") or self.module_source.startswith("../") \
-               or self.module_source.startswith(self.current_dir) or self.module_source.startswith('/')
+        return (
+            self.module_source.startswith("./")
+            or self.module_source.startswith("../")
+            or self.module_source.startswith(self.current_dir)
+            or self.module_source.startswith("/")
```
nimrodkor, line 17: @gruebel Merge from master, as this line was removed. Abs paths will be wrong in different machines so we don't try to initiallize them

checkov/terraform/module_loading/loaders/git_loader.py
```diff
@@ -1,31 +1,103 @@
...
+            version = module_version_components[1]
+        else:
+            version = "HEAD"
+
+        if len(module_source_components) == 2:
+            inner_module = ""
+        elif len(module_source_components) == 3:
+            inner_module = module_source_components[2]
+        else:
+            raise Exception("invalid git url")
```
nimrodkor, line 68: @gruebel In general, in this area (module loading) we prefer logging warnings and skipping rather than raising exceptions and failing, Is this caught anywhere?

nimrodkor (commented):
Overall looks good, had 2 really minor comments

checkov/terraform/module_loading/loaders/git_loader.py
gruebel: yeah, it is caught in the `parser.py` with an appropriate message
https://github.com/bridgecrewio/checkov/blob/0f63f8e5be52d2a4b9c1b46d7952eddb717e673c/checkov/terraform/parser.py#L396-L397
https://github.com/bridgecrewio/checkov/blob/0f63f8e5be52d2a4b9c1b46d7952eddb717e673c/checkov/terraform/parser.py#L453-L455

checkov/terraform/module_loading/loaders/local_path_loader.py
gruebel: actually the absolute path is needed, otherwise it won't work, because in the `parser.py` the module source is changed to an abs path.
https://github.com/bridgecrewio/checkov/blob/0f63f8e5be52d2a4b9c1b46d7952eddb717e673c/checkov/terraform/parser.py#L389-L391

rotemavni (approved):
@gruebel Wonderful work!

PR #1290, opened 2021-06-14
Don't fail checks for vars (or any reference type string) not evaluated

Saarett:
Ensures that AWS checks return UNKOWN if they don't have the variable (or any other reference type string like module.something... , locals.unknown, etc...) to evaluate, instead of failing the checks in some cases.

Affected checks include at least:
BaseResourceValueCheck
BaseResourceNegativeValueCheck


checkov/terraform/checks/resource/base_resource_negative_value_check.py
```diff
@@ -39,6 +41,8 @@ def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
             value = dpath.get(conf, inspected_key)
             if isinstance(value, list) and len(value) == 1:
                 value = value[0]
+            if get_referenced_vertices_in_value(value=value, aliases={}, resources_types=resources_types.get("aws")):
```
schosterbarak, line 44: we want it to work just for AWS?

tests/graph/terraform/runner/test_graph_builder.py
```diff
@@ -15,7 +15,7 @@ def test_build_graph(self):
         runner = Runner()
         report = runner.run(None, None, files=list(map(lambda f: f'{resources_path}/{f}', source_files)))
         tf_definitions = runner.tf_definitions
-        self.assertEqual(6, len(report.failed_checks))
+        self.assertEqual(4, len(report.failed_checks))
```
schosterbarak, line 18: why?

rotemavni (approved):
@Saarett Great Job!

PR #1291, opened 2021-06-14
run dependency checks after the runner filter to skip unneeded checks

mikeurbanski1:

Applies the runner filter (--frameworks or --skip-frameworks) before checking for dependencies. This greatly speeds up the runtime if helm is installed but the cluster is not reachable (results in waiting for a timeout).


checkov/common/runners/runner_registry.py
```diff
@@ -105,6 +105,9 @@ def filter_runner_framework(self):
         self.runners = filtered_runners
         return
 
+    def remove_runner(self, runner):
+        del self.runners[self.runners.index(runner)]
```
gruebel, line 109: using `remove()` here is no need for checking the index, but this method can result in a `ValueError` if the runner is not in the list.
```suggestion
    def remove_runner(self, runner):
        if runner in self.runners:
            self.runners.remove(runner)
```
or catch the `ValueError` with try ... except ... 😄

checkov/common/util/runner_dependency_handler.py
```diff
@@ -1,64 +1,47 @@
...
+    def __init__(self, runner_registry: RunnerRegistry):
         """
         RunnerDependencyHandler
-        :param checkov_runner_module_names: list of runner module names to check
-        :param calledGlobals: The main-scoped dict output of globals(), so we can access the registered runners.
+        :param runner_registry: a populated runner registry
         """
-        self.checkov_runner_module_names = checkov_runner_module_names
-        self.calledGlobals = calledGlobals
+        # if frameworks:
```
gruebel, line 17: is it some left over code or a real comment?

mikeurbanski1:
Thanks @gruebel . Not sure what I was thinking with `del` 🤦

nimrodkor (approved):

PR #1293, opened 2021-06-15
Adding a rule to detect if an AWS Transfer Server is public

ac-square:

https://aws.amazon.com/premiumsupport/knowledge-center/aws-sftp-endpoint-type/

### DIFF
--- a/tests/terraform/checks/resource/aws/test_TransferServerIsPublic.py
+++ b/tests/terraform/checks/resource/aws/test_TransferServerIsPublic.py
@@ -1,35 +1,38 @@
+import os
 import unittest
 
-from checkov.common.models.enums import CheckResult
+from checkov.runner_filter import RunnerFilter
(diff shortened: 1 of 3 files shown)


tests/terraform/checks/resource/aws/test_TransferServerIsPublic.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+
+from checkov.common.models.enums import CheckResult
+from checkov.terraform.checks.resource.aws.TransferServerIsPublic import check
+import hcl2
+
+
+class TestTransferServerIsPublic(unittest.TestCase):
```
gruebel, line 8: Please change the tests to use the newer set up like here #1182

And also please add a test without specifying the `endpoint_type` should result in a failure.

ac-square:
@gruebel updated per your comments

gruebel:
nice, looks good 👍

schosterbarak:
Thanks guys!

PR #1294, opened 2021-06-15
serialize set as list for json

mikeurbanski1:

This google module produced a JSON serialization error when running with platform integration and --download-external-modules.

https://github.com/terraform-google-modules/terraform-google-service-accounts

This is because it has sets in the definition, which don't get serialized to JSON.

This PR adds a JSON encoder to convert them to lists first.

### DIFF
--- a/checkov/common/util/json_utils.py
+++ b/checkov/common/util/json_utils.py
@@ -4,6 +4,6 @@ import json
 class CustomJSONEncoder(json.JSONEncoder):
     def default(self, o):
         if isinstance(o, set):
-            return str(list(o))
+            return json.JSONEncoder.default(self, list(o))
         else:
             return json.JSONEncoder.default(self, o)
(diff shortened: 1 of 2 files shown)


checkov/common/util/json_utils.py
```diff
@@ -0,0 +1,9 @@
+import json
+
+
+class CustomJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, set):
+            return json.dumps(list(o))
```
mikeurbanski1, line 7: I tried

`json.JSONEncoder.default(self, list(o))`

here, and I got "Object of type list is not JSON serializable". Not sure why.

gruebel: I think you can just change ti to
```python
return list(o)
```
for further info https://docs.python.org/3/library/json.html#json.JSONEncoder.default

PR #1295, opened 2021-06-16
Adding a rule to detect if an AWS Backup Vault has server-side encryption enabled

ac-square:

### DIFF
--- a/checkov/terraform/checks/resource/aws/BackupVaultEncrypted.py
+++ b/checkov/terraform/checks/resource/aws/BackupVaultEncrypted.py
@@ -5,10 +5,10 @@ from checkov.common.models.consts import ANY_VALUE
 
 class BackupVaultEncrypted(BaseResourceValueCheck):
     def __init__(self):
-        name = "Ensure Backup Vault server-side encryption is enabled."
+        name = "Ensure Backup Vault is encrypted at rest using KMS CMK"
         id = "CKV_AWS_165"
         supported_resources = ['aws_backup_vault']
-        categories = [CheckCategories.BACKUP_AND_RECOVERY]
+        categories = [CheckCategories.ENCRYPTION]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/aws/BackupVaultEncrypted.py
```diff
@@ -0,0 +1,20 @@
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+from checkov.common.models.consts import ANY_VALUE
+
+
+class BackupVaultEncrypted(BaseResourceValueCheck):
+    def __init__(self):
+        name = "Ensure Backup Vault server-side encryption is enabled."
```
gruebel, line 8: the text is actually misleading, because AWS Backup will use an AWS managed KMS key, if you don't set it. So something more like
`Ensure Backup Vault is encrypted at rest using KMS CMK`

checkov/terraform/checks/resource/aws/BackupVaultEncrypted.py
```diff
@@ -0,0 +1,20 @@
...
+        id = "CKV_AWS_165"
+        supported_resources = ['aws_backup_vault']
+        categories = [CheckCategories.BACKUP_AND_RECOVERY]
```
gruebel, line 11: the category should be different 😉 
```suggestion
        categories = [CheckCategories.ENCRYPTION]
```

tests/terraform/checks/resource/aws/test_BackupVaultEncrypted.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+
+from checkov.common.models.enums import CheckResult
+from checkov.terraform.checks.resource.aws.BackupVaultEncrypted import check
+import hcl2
+
+
+class TestBackupVaultEncrypted(unittest.TestCase):
```
gruebel, line 8: Please change the tests to use the newer set up like here #1182

gruebel (commented):
@schosterbarak In general this check feels like it is misleading, because even if you set it up, it doesn't mean your backup will be encrypted in general or especially with the used KMS key.
https://docs.aws.amazon.com/aws-backup/latest/devguide/encryption.html

ac-square:
@gruebel updated per your comments

JamesWoolfenden (changes requested):
ignore

JamesWoolfenden (approved):

nimrodkor (approved):

PR #1297, opened 2021-06-16
CF Check - CKV_AWS_165

shubh-cs29:
Hello,

Implementing check for point in time recovery (Backup) for DynamoDB Global Table  in CloudFormation.

CloudFormation Docs:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-globaltable.html

Note:
1. If we have not pass the "PointInTimeRecoveryEnabled" parameter in CloudFormation template then the check fails, as by default it will not provide backup facility.
2. If we have pass the "PointInTimeRecoveryEnabled" parameter with value "false" in CloudFormation template then the check fails, means we have turned off the backup facility.

License


tests/cloudformation/checks/resource/aws/test_DynamodbGlobalTableRecovery.py
```diff
@@ -0,0 +1,26 @@
...
+
+        test_files_dir = current_dir + "/example_DynamodbGlobalTableRecovery"
+        report = runner.run(root_folder=test_files_dir,runner_filter=RunnerFilter(checks=[check.id]))
+        summary = report.get_summary()
+
+        self.assertEqual(summary['passed'], 1)
+        self.assertEqual(summary['failed'], 2)
+        self.assertEqual(summary['skipped'], 0)
+        self.assertEqual(summary['parsing_errors'], 0)
+
```
gruebel, line 23: can you also check for which resource exactly fails, like here
https://github.com/bridgecrewio/checkov/blob/4fa8fbb2bfa7931a01e9da3fc1c46d1e49d18d5d/tests/cloudformation/checks/resource/aws/test_RDSMultiAZEnabled.py#L19-L36
I think it would be also great, if you give them a proper name too 😄

gruebel:
@shubh-cs29 and please have a look into the cfn-lint errors

tests/cloudformation/checks/resource/aws/test_DynamodbGlobalTableRecovery.py
shubh-cs29: @gruebel I have fixed the cfn-lint issue.

gruebel: this comment has nothing to do with the cfn-lint errors. It would be great if you can check for the exact passed & failed resources, like in the example I added.

shubham-cldcvr: @gruebel Yes as per your suggestion I have modified the test case for passed and failed resources.

shubh-cs29: @gruebel Yes as per your suggestion I have modified the test case for passed and failed resources.

PR #1298, opened 2021-06-16
CF - Check -  ALBDropHttpHeaders

njgibbon:
Hello,

Contributing this check for CF as in TF.

Could not use ValueCheck as done in terraform check because of how the LoadBalancerAttributes are a list of objects in CloudFormation but a top level field for terraform.

The true and false for the LoadBalancerAttributes need to be strings and not booleans.

* Terraform check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/ALBDropHttpHeaders.py
* CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html

**License**


tests/cloudformation/checks/resource/aws/example_ALBDropHttpHeaders/FAIL.yaml
```diff
@@ -0,0 +1,42 @@
...
+    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
+    Properties:
+      Name: test
+      Subnets: 
+        - test-0
+        - test-1
+      Type: application
+      LoadBalancerAttributes:
+        - Key: routing.http.drop_invalid_header_fields.enabled
+          Value: "false"
```
gruebel, line 31: There is no difference between both resources, I think you wanted to the set `Type: application` without defining `LoadBalancerAttributes`, right?

checkov/cloudformation/checks/resource/aws/ALBDropHttpHeaders.py
```diff
@@ -0,0 +1,41 @@
...
+                    if isinstance(lb_attributes, list):
+                        for item in lb_attributes:
+                            if 'Key' in item.keys() and 'Value' in item.keys():
+                                key = item['Key']
+                                value = item['Value']
+                                if key == 'routing.http.drop_invalid_header_fields.enabled' and value == "true":
+                                    return CheckResult.PASSED
+                return CheckResult.FAILED
+
+        return CheckResult.PASSED
```
gruebel, line 38: In general the default should be here `CheckResult.UNKNOWN` because you can't set it for NLBs and GLBs. Also adding them as an example resource would be helpful to be sure, they are not considered as a passed resource.

checkov/cloudformation/checks/resource/aws/ALBDropHttpHeaders.py
```diff
@@ -0,0 +1,41 @@
...
+    def scan_resource_conf(self, conf):
+        # alb is default loadbalancer type if not explicitly set
+        alb = True
+
+        if 'Properties' in conf.keys():
+            properties = conf['Properties']
+            if 'Type' in properties.keys():
+                lb_type = properties['Type']
+                if lb_type != 'application':
+                    alb = False
```
gruebel, line 23:
```suggestion
        properties = conf.get("Properties", {})
        lb_type = properties.get("Type")
        if lb_type != "application":
            alb = False
```

checkov/cloudformation/checks/resource/aws/ALBDropHttpHeaders.py
```diff
@@ -0,0 +1,41 @@
...
+                if 'LoadBalancerAttributes' in properties.keys():
+                    lb_attributes = properties['LoadBalancerAttributes']
+                    if isinstance(lb_attributes, list):
+                        for item in lb_attributes:
+                            if 'Key' in item.keys() and 'Value' in item.keys():
+                                key = item['Key']
+                                value = item['Value']
+                                if key == 'routing.http.drop_invalid_header_fields.enabled' and value == "true":
+                                    return CheckResult.PASSED
+                return CheckResult.FAILED
```
gruebel, line 36:
```suggestion
            if alb:
                lb_attributes = properties.get("LoadBalancerAttributes")
                if isinstance(lb_attributes, list):
                    for item in lb_attributes:
                        key = item.get("Key")
                        value = item.get("Value")
                        if key == 'routing.http.drop_invalid_header_fields.enabled' and value == "true":
                            return CheckResult.PASSED
                return CheckResult.FAILED
```

JamesWoolfenden (approved):
LGTM

PR #1316, opened 2021-06-22
Add skip file logic

mikeurbanski1:

This PR adds an argument to skip a file using a regular expression path. It also corrects the way in which skip paths are applied from the platform. It preserves the legacy skip logic (mostly used to skip specific problematic directories, like .terraform, node_modules, etc), as explained in the comments.

The biggest change is to the logic in base_runner, which used to only look at directory basenames. This was problematic, as shown in the following example:

The old way to filter looked like this:
```
    filter_ignored_directories(d_names, runner_filter.excluded_paths)
```
`excluded_paths` contains the paths from the repo config screen.
Here, `d_names` will always be the directory basenames without any paths.
So if we iterate this tree:
```
.
  ./dir1
     ./dir1/dir3
  ./dir2
     ./dir2/dir3
```
`d_names` would be `['dir1', 'dir2']` for the first iteration, and then `['dir3']` (within dir1) for the second iteration, and then `['dir3']` (within dir2) for the third iteration. So it would be impossible to exclude dir1/dir3 but not dir2/dir3.

Additionally, this logic would never filter out a specific file (only directories).

The new logic will check the path relative to the current working directory.

It may have an issue with running on Windows, because it will be unable to apply paths specified in the platform due to `/` vs `\`. However, the local argument filter logic should work natively.


mikeurbanski1:
@rotemavni I added you to sanity check the Terraform logic, since it's different than the rest.

kartikp10 (approved):
Looks good!

gruebel:
@mikeurbanski1 just a general suggestion, why not use `pathlib.Path.glob()` function to iterate through the filesystem? https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob
it will take care for you the typical problems between supporting Windows and Linux filesystems.

mikeurbanski1:
> @mikeurbanski1 just a general suggestion, why not use `pathlib.Path.glob()` function to iterate through the filesystem? https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob
> it will take care for you the typical problems between supporting Windows and Linux filesystems.

The issue is not with the new CLI arg itself. That should work fine with native paths. The regexes will be a little painful to write, though. Globs would be easier, but...

The challenge is in how this used the Bridgecrew platform, which also offers this exclusion logic via a config page. This also gets pulled into checkov if you run with the API key. The platform uses regexes, and we are making this arg consistent. You could argue globs would be better, but it doesn't solve the problem. Since the platform backend is all linux based, the values you enter on the platform configuration have to use forward slashes. So if you fetch those values on Windows, they will never match the paths on your filesystem. We could and will handle it in the future by converting \ to / in those paths.

gruebel:
> The issue is not with the new CLI arg itself. That should work fine with native paths. The regexes will be a little painful to write, though. Globs would be easier, but...
> 
> The challenge is in how this used the Bridgecrew platform, which also offers this exclusion logic via a config page. This also gets pulled into checkov if you run with the API key. The platform uses regexes, and we are making this arg consistent. You could argue globs would be better, but it doesn't solve the problem. Since the platform backend is all linux based, the values you enter on the platform configuration have to use forward slashes. So if you fetch those values on Windows, they will never match the paths on your filesystem. We could and will handle it in the future by converting \ to / in those paths.

ah ok, I think the usage of `pathlib` can usually solve issues with being compatible with Windows and Linux, but because the input comes from the outside, it won't be that helpful here. But thanks for giving some more insights 💪

checkov/common/runners/base_runner.py
```diff
@@ -18,7 +18,32 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+    #   ./dir2
+    #      ./dir2/dir33
+    #      /.dir2/hello.yaml
+    #
+    # if excluded_paths = ['dir1/dir33', 'dir2/hello.yaml'], then we would scan dir1, but we would skip its subdirectories. We would scan
+    # dir2 and its subdirectory, but we'd skip hello.yaml.
+
+    # first handle the legacy logic - this will also remove files starting with '.' but that's probably fine
+    # mostly this will just remove those problematic directories hardcoded above.
+    [names.remove(path) for path in list(names) if path in ignored_directories or path.startswith(".")]
```
gruebel, line 41: is there any reason, why you use list comprehension here without assigning the resulted list to a variable? feels kind of redundant :) using a plain loop is favourable, because you don't get any performance benefits, just creating an unused list.

checkov/common/runners/base_runner.py
```diff
@@ -18,7 +18,32 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+
+    # now apply the new logic
+    # TODO this is not going to work well on Windows, because paths specified in the platform will use /, and
+    #  paths specified via the CLI argument will presumably use \\
+    if excluded_paths:
+        compiled = [re.compile(p) for p in excluded_paths]
+        [names.remove(path) for path in list(names) if any(pattern.search(os.path.join(root_dir, path)) for pattern in compiled)]
```
gruebel, line 48: same here

checkov/terraform/parser.py
```diff
@@ -24,9 +24,9 @@
 external_modules_download_path = os.environ.get('EXTERNAL_MODULES_DIR', DEFAULT_EXTERNAL_MODULES_DIR)
 
 
-def _filter_ignored_directories(d_names, excluded_paths):
-    filter_ignored_directories(d_names, excluded_paths)
-    [d_names.remove(d) for d in list(d_names) if d in [default_ml_registry.external_modules_folder_name]]
+def _filter_ignored_paths(root, paths, excluded_paths):
+    filter_ignored_paths(root, paths, excluded_paths)
+    [paths.remove(path) for path in list(paths) if path in [default_ml_registry.external_modules_folder_name]]
```
gruebel, line 29: same here

checkov/common/runners/base_runner.py
mikeurbanski1: I inherited it from before. You're right, it's just used as a looping mechanism to modify the path list in place (which then ensures they get skipped later on during os.walk)

checkov/main.py
```diff
@@ -104,8 +106,8 @@ def run(banner=checkov_banner, argv=sys.argv[1:]):
                                                         skip_fixes=config.skip_fixes,
                                                         skip_suppressions=config.skip_suppressions,
                                                         source=source, source_version=source_version, repo_branch=config.branch)
-            excluded_paths = bc_integration.get_excluded_paths()
-            runner_filter.excluded_paths = excluded_paths
+            platform_excluded_paths = bc_integration.get_excluded_paths()
+            runner_filter.excluded_paths = (runner_filter.excluded_paths + platform_excluded_paths) if runner_filter.excluded_paths else platform_excluded_paths
```
rotemavni, line 110: @mikeurbanski1 Isn't `runner_filter.excluded_paths + platform_excluded_paths` enough? if `runner_filter.excluded_paths` is empty it should have the same effect as just using platform_excluded_paths IMO

rotemavni (approved):
@mikeurbanski1 LGTM

checkov/main.py
mikeurbanski1: I wanted to guard against it somehow being set to `None` in the future, but you are right that it would work currently. I'll change it.

nimrodkor: @rotemavni Nope
```sh
Python 3.7.10 (default, Mar  3 2021, 14:09:10) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
None + ['a']
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'
```

nimrodkor (approved):

PR #1322, opened 2021-06-23
Scan files for secrets concurrently and disable unnecessary heavy regex

nimrodkor:

Improved performance of a use case from >20 minutes to ~5 by disabling a filter (mostly relevant for a disabled plugin, `secret keyword`).
Add logging in case of a long scan for a specific file (over 10 seconds)


checkov/secrets/runner.py
```diff
@@ -103,13 +109,31 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
-                    continue
+                    secrets.scan_file(file_path)
+                except Exception as err:
+                    logging.warning(f"Secret scanning:could not process file {file_path}, {err}")
+                end = time.time()
+                scan_time = end - start
+                if scan_time > 10:
+                    logging.info(f'Scanned {file_path}, took {scan_time} seconds')
+
+            with concurrent.futures.ThreadPoolExecutor(max_workers=os.cpu_count()) as executor:
```
rotemavni, line 127: @nimrodkor Why not use `checkov/terraform/checks/utils/utils.py/run_function_multithreaded`?

rotemavni (approved):

PR #1334, opened 2021-06-28
add type hints to terraform.graph_builder

gruebel:

Added type hints to `terraform.graph_builder` and changed condtional checks with `type()` to `isinstance()` this is more preferred and has better performance. I also introduced a couple of times the usage of `setdefault()` to make the code simpler and a bit more performant :) 

I also added a `TODO` in `local_graph.py` because I think the code part can never be reached, but making it properly work makes tests fail. It is no problem to adjust the tests, but the problem is the result becomes worse, therefore I didn't want to just fix them.


checkov/terraform/graph_builder/graph_components/generic_resource_encryption.py
```diff
@@ -1,9 +1,16 @@
...
 from checkov.common.graph.graph_builder import EncryptionTypes
 from checkov.terraform.checks.utils.utils import decode_graph_property_value
 
 
 class GenericResourceEncryption:
-    def __init__(self, resource_type: str, attribute_values_map: dict, enabled_by_default=False):
+    def __init__(
+        self,
+        resource_type: str,
+        attribute_values_map: Dict[str, Union[List[bool], List[str]]],
```
nimrodkor, line 11: @gruebel Shouldn't it be `Dict[str, List[Union[bool, str]]]`?

checkov/terraform/graph_builder/graph_to_tf_definitions.py
```diff
@@ -1,38 +1,33 @@
...
             tf_path = f"{block_path}[{vertex.module_dependency}#{vertex.module_dependency_num}]"
         block_type = vertex.block_type.value
         if block_type == BlockType.TF_VARIABLE:
             continue
-        if tf_definitions.get(tf_path) is None:
-            tf_definitions[tf_path] = {}
-        if tf_definitions.get(tf_path).get(block_type) is None:
-            tf_definitions.get(tf_path)[block_type] = []
-        tf_definitions.get(tf_path)[block_type].append(vertex.config)
+        tf_definitions.setdefault(tf_path, {}).setdefault(block_type, []).append(vertex.config)
```
nimrodkor, line 24: @gruebel tnx!

nimrodkor (approved):

checkov/terraform/graph_builder/graph_components/generic_resource_encryption.py
gruebel: I checked and there shouldn't be a `list` with `bool` and `str` values mixed, just one or the other. But if you think this could happen, then I can adjust it.

PR #1336, opened 2021-06-29
Added AWS provider check for default tags

bdausses:
Signed-off-by: Brad Dausses <brad.dausses@gmail.com>

This check was requested by Mike Urbanski as part of my onboarding activities.  Please let me know if anything additional is needed.

### DIFF
--- a/checkov/terraform/checks/provider/aws/defaulttags.py
+++ b/checkov/terraform/checks/provider/aws/defaulttags.py
@@ -1,0 +1,28 @@
+import re
+from typing import Dict, List, Any
+
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.provider.base_check import BaseProviderCheck
+
+
+class AWSDefaultTags(BaseProviderCheck):
+    def __init__(self) -> None:
+        name = "AWS provider should specify default tags"
+        id = "CKV_AWS_166"
+        supported_provider = ["aws"]
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_provider=supported_provider)
+
+    def scan_provider_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
+        """
+            AWS provider should specify default tags:
+            https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags
+        :param conf: aws provider configuration
+        :return: <CheckResult>
+        """
+        if "default_tags" in conf.keys():
+            return CheckResult.PASSED
+        return CheckResult.FAILED
+
+
+check = AWSDefaultTags()
--- a/tests/terraform/checks/provider/aws/test_defaulttags.py
+++ b/tests/terraform/checks/provider/aws/test_defaulttags.py
@@ -1,0 +1,20 @@
+import unittest
+
+from checkov.terraform.checks.provider.aws.defaulttags import check
+from checkov.common.models.enums import CheckResult
+
+
+class TestDefaultTags(unittest.TestCase):
+
+    def test_success(self):
+        provider_conf = {'region': ['us-west-2'], 'default_tags': {'tags': {'Environment': 'Test', 'Owner': 'TestOwner'}}}
+        scan_result = check.scan_provider_conf(conf=provider_conf)
+        self.assertEqual(CheckResult.PASSED, scan_result)
+
+    def test_failure(self):
+        provider_conf = {'region': ['us-west-2']}
+        scan_result = check.scan_provider_conf(conf=provider_conf)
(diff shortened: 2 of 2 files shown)


tests/terraform/checks/provider/aws/test_defaulttags.py
```diff
@@ -0,0 +1,20 @@
+import unittest
+
+from checkov.terraform.checks.provider.aws.defaulttags import check
+from checkov.common.models.enums import CheckResult
+
+
+class TestDefaultTags(unittest.TestCase):
+
+    def test_success(self):
```
mikeurbanski1, line 9: This needs to use the test format described here: https://github.com/bridgecrewio/checkov/blob/master/CONTRIBUTING.md#tests-for-new-checks

This is the way it says NOT to do it

checkov/terraform/checks/provider/aws/defaulttags.py
```diff
@@ -0,0 +1,28 @@
...
+        super().__init__(name=name, id=id, categories=categories, supported_provider=supported_provider)
+
+    def scan_provider_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
+        """
+            AWS provider should specify default tags:
+            https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags
+        :param conf: aws provider configuration
+        :return: <CheckResult>
+        """
+        if "default_tags" in conf.keys():
```
mikeurbanski1, line 23: This should check to see whether there is an actual tag (it could be an empty map)

checkov/terraform/checks/provider/aws/defaulttags.py
```diff
@@ -0,0 +1,28 @@
+import re
```
gruebel, line 1: this import is unused and should be removed

checkov/terraform/checks/provider/aws/defaulttags.py
```diff
@@ -0,0 +1,28 @@
...
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.provider.base_check import BaseProviderCheck
+
+
+class AWSDefaultTags(BaseProviderCheck):
+    def __init__(self) -> None:
+        name = "AWS provider should specify default tags"
+        id = "CKV_AWS_166"
+        supported_provider = ["aws"]
+        categories = [CheckCategories.GENERAL_SECURITY]
```
gruebel, line 13: just a general question, wouldn't `CONVENTION` fit better here or does it make sense to even create a new category like `BEST_PRACTICE`?

nimrodkor: Interesting idea. WDYT @schosterbarak @GiladMark

tests/terraform/checks/provider/aws/test_defaulttags.py
bdausses: @mikeurbanski1 This should be resolved using the alternate test method you suggested.

checkov/terraform/checks/provider/aws/defaulttags.py
bdausses: Check does now verify the presence of actual entries under `default_tags { tags {`.  Also added a test case for this as well.

checkov/terraform/checks/provider/aws/defaulttags.py
bdausses: Removed.  Thanks!

checkov/terraform/checks/provider/aws/defaulttags.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict, List, Any
+
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.provider.base_check import BaseProviderCheck
+
+
+class AWSDefaultTags(BaseProviderCheck):
+    def __init__(self) -> None:
+        name = "Ensure AWS provider uses default_tags"
+        id = "CKV_AWS_166"
```
nimrodkor, line 10: @bdausses Please use an unused check ID, this one is already taken and it fails the tests

checkov/terraform/checks/provider/aws/defaulttags.py
schosterbarak: CONVENTION sounds good

JamesWoolfenden:
Isnt this going to raise a significant number of failures?

schosterbarak:
just had a sync with @JamesWoolfenden 
@nimrodkor is it possible to do a graph query for "is resources is having tags" both on the resource itself and the connected provider?

nimrodkor:
@schosterbarak We do not create edges today between the provider and the resource, so not as of today...

PR #1340, opened 2021-06-30
Custom policies from platform

mikeurbanski1:

Pulls custom policies from the platform when using `--bc-api-key`. Provides a flag to skip the download.

Also updates platform suppression logic to work with custom policies.

Updates --list to show the custom policies that get fetched.


checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -0,0 +1,57 @@
...
+    def _get_policies_from_platform(self):
+        headers = merge_dicts(get_default_get_headers(self.bc_integration.bc_source, self.bc_integration.bc_source_version),
+                              get_auth_header(self.bc_integration.bc_api_key))
+        response = requests.request('GET', self.policies_url, headers=headers)
+
+        if response.status_code != 200:
+            error_message = extract_error_message(response)
+            raise Exception(f'Get custom policies request failed with response code {response.status_code}: {error_message}')
+
+        policies = json.loads(response.content).get('data')
```
gruebel, line 53: the `response` object from `request` has a method to directly return JSON encoded content and I think it would be good to add a default for the `get()` so if it doesn't exist it won't raise an exception in the `pre_scan()` method.
```suggestion
        policies = response.json().get('data', [])
```

checkov/main.py
```diff
@@ -127,6 +129,8 @@ def run(banner=checkov_banner, argv=sys.argv[1:]):
             parser.error("The check ids specified for '--check' and '--skip-check' must be mutually exclusive.")
             return
 
+    integration_feature_registry.run_pre_scan()
```
nimrodkor, line 132: @mikeurbanski1 Is this protected if running without api-key?

nimrodkor (commented):
Nice!

checkov/main.py
mikeurbanski1: Each "integration feature" has logic to determine whether it runs, so nothing will happen if integration is not configured.

nimrodkor (approved):

PR #1320, opened 2021-06-22
Adding 3 new checks for * principals in Glacier Vaults, SNS Topic Policies, and SQS Queue Policies

ac-square:
This PR introduces three new checks:

* CKV_AWS_167 - Ensure Glacier Vault access policy is not public by only allowing specific services or principals to access it
* CKV_AWS_168 - Ensure SQS queue policy is not public by only allowing specific services or principals to access it
* CKV_AWS_169 - Ensure SNS topic policy is not public by only allowing specific services or principals to access it

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

These checks use Netflix's [policyuniverse](https://github.com/Netflix-Skunkworks/policyuniverse) to parse IAM and resource policies. I recommend using this for all future checks that deal with these policies, as it's as easy as calling `policy.is_internet_accessible()`.

PR #1341, opened 2021-07-01
Support un-rendered policy in SNSTopicPolicy / GlacierVaultAnyPrincipal

aristotelesneto:
Support a case where there is no json policy available to avoid throwing an exception.


aristotelesneto:
This was introduced by #1320 and results in an exception when the policy is a data source. It was addressed for SQS in #1338.

nimrodkor:
Thanks! It was actually just fixed on master, so closing it...

aristotelesneto:
Thanks @nimrodkor. Worth noting that GlacierVaultAnyPolicy also had the same issue but it doesn't appear to have been fixed

PR #1343, opened 2021-07-01
Python 3 test improvements. Better asserts and subTest

atombrella:

I used `flake8-assertive` to find these. There are more things that could be adapted to https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests


nimrodkor (approved):
Thanks!

Issue #466, opened 2020-07-23
Always run custom checks


robeden:
**Is your feature request related to a problem? Please describe.**
To keep results stable, it makes sense for us to opt-in to Checkov built-in checks. This requires using the `--check` flag. However, I have a number of custom checks that make sense to always run. Currently, their IDs also have to be specified with the `--check` flag, which becomes quite cumbersome as their number grows.

**Describe the solution you'd like**
After talking it out on slack, we believe it would make sense to always run checks specified via `--external-checks-dir`. (Bonus points, perhaps, if they could be disabled if flagged via `--skip-check`.)

**Describe alternatives you've considered**
There could be an extra flag to control this behavior (something like `--check-all-custom` or `--skip-all-custom`), but this doesn't seem necessary.

**Additional context**
([Slack discussion](https://codifiedsecurity.slack.com/archives/C016DJYKALU/p1595504569017500))

robeden:
Pseudo-code for checks to run would be:
```
if has_check_flag_specified():
    checks_to_run = checks_specifically_included
else:
    checks_to_run = all_built_in_checks

if has_checks_dir_specified():
   checks_to_run += checks_from_external_dir

for skipped_check in skip_check_flags():
    checks_to_run.remove(skipped_check)
```

KevinHock:
For maintainers: Somewhat unrelated, but I recently wrote some dynamic import code very similar to the checkov dynamic importing code without the `sys` hacking stuff, might be of interest, might not, but figured I'd share https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/common/util.py

PR #1348, opened 2021-07-02
include external checks in --check flag

mikeurbanski1:

Treats externally loaded checks like all other checks in the `--check` flag. This changes the previous default behavior, in which all external checks would always run (unless skipped with `--skip-check`), even if they were not included in `--check`.

As discussed in [this issue](https://github.com/bridgecrewio/checkov/issues/466), some users rely on this old default behavior to ensure that any external checks added to a repo always get run, without modifying the `--check` list. To preserve this behavior, this change also introduces a new flag: `--run-all-external-checks`


kartikp10 (approved):
LGTM!

PR #1367, opened 2021-07-11
CF - Check - CloudWatchLogGroupKMSKey

njgibbon:
Hello,

Adding the CF check as already done for Terraform.

Terraform Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/CloudwatchLogGroupKMSKey.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html 

Also, for the 4 checks across TF and CF I've made consistent naming and messaging and updated the testing approach to the most modern we use for all. Looks like a lot of changes but just a rename and refactor for the other CloudWatch related checks and tests really.

**License**


nimrodkor (approved):
Nice! Tnx @njgibbon

PR #1374, opened 2021-07-13
Graph classes refactor and basic cloudformation graph

rotemavni:
**Refactor**
1. Created a base `BlockType` that has only 'resource' in it. It is extended by both terraform and cfn `BlockType`. I transformed this entity from an enum to dataclass, which removed the call to `.value`.
2. Created a base `Block` with most of the functionality in it, and `CloudformationBlock` and `TerraformBlock` extend it.
3. Removed all encoding/decoding functions and parameters relating to it.
4. Created a base `LocalGraph`, `CloudformationLocalGraph` and `TerraformLocalGraph` extend it.
5. Removed the `utils` file that was nested under `terraform/checks` because it contained many unrelated functions to the checks dir. The unused functions were deleted and the rest was moved to more appropriate locations.
6. Refactored `GraphManager`: created a base class and updated signatures to be more general.

**Cloudformation graph**
Created `CloudformationLocalGraph` that builds only resource vertices using `cfn_definitions`, and created a UT to demonstrate building the graph.


checkov/cloudformation/graph_builder/local_graph.py
```diff
@@ -0,0 +1,53 @@
...
+                                        )
+            self.vertices.append(block)
+
+
+def get_only_dict_items(origin_dict: Dict) -> Dict:
+    return {key: origin_dict[key] for key in origin_dict.keys() if isinstance(origin_dict[key], dict)}
+
+
+def filter_excluded_dict_attributes(origin_dict: Dict) -> Dict:
+    return {key: origin_dict[key] for key in origin_dict.keys() if key not in DEFINITIONS_EXCLUDED_ATTRIBUTES}
```
nimrodkor, line 53: @rotemavni Won't we need these attributes for the back-to-definitions method?

checkov/common/graph/graph_builder/local_graph.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self) -> None:
+        self.vertices: List[Block] = []
+        self.edges: List[Edge] = []
+        self.in_edges: Dict[int, List[Edge]] = {}  # map between vertex index and the edges entering it
+        self.out_edges: Dict[int, List[Edge]] = {}  # map between vertex index and the edges exiting it
+        self.vertices_by_block_type: Dict[BlockType, List[int]] = {}
+        self.vertex_hash_cache: Dict[int, str] = {}
+        self.vertices_block_name_map: Dict[BlockType, Dict[str, List[int]]] = {}
+
+    def build_graph(self, render_variables: bool) -> None:
```
nimrodkor, line 18: @rotemavni Please mark as `abstractmethod`

checkov/common/graph/graph_manager.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, db_connector, parser, source=''):
+        self.db_connector = db_connector
+        self.source = source
+        self.parser = parser
+
+    def build_graph_from_source_directory(self, source_dir, render_variables=True, local_graph_class=LocalGraph,
+                                          parsing_errors=None, download_external_modules=False, excluded_paths: List[str]=None):
+        pass
+
+    def build_graph_from_definitions(self, definitions, render_variables=True):
```
nimrodkor, line 16: @rotemavni Please add `@abstractmethod` here as well

tests/terraform/graph/utils/test_utils.py
```diff
@@ -183,21 +180,3 @@ def test__attribute_has_nested_attributes_list(self):
         self.assertTrue(attribute_has_nested_attributes(attribute_key='filter.1.values', attributes=attributes))
         self.assertFalse(attribute_has_nested_attributes(attribute_key='filter.1.values.0', attributes=attributes))
 
-    def test_malicious_eval(self):
```
nimrodkor, line 186: @rotemavni Did we move this somewhere?

tests/cloudformation/graph_builder/test_local_graph.py
```diff
@@ -0,0 +1,27 @@
...
+        (definitions[relative_file_path], definitions_raw) = parse(file)
+        local_graph = CloudformationLocalGraph(definitions)
+        local_graph.build_graph(render_variables=False)
+        self.assertEqual(1, len(local_graph.vertices))
+        resource_vertex = local_graph.vertices[0]
+        self.assertEqual("AWS::ApiGateway::Stage.MyStage", resource_vertex.name)
+        self.assertEqual("AWS::ApiGateway::Stage.MyStage", resource_vertex.id)
+        self.assertEqual(BlockType.RESOURCE, resource_vertex.block_type)
+        self.assertEqual("CloudFormation", resource_vertex.source)
+        self.assertEqual("TestDeployment", resource_vertex.attributes.get("DeploymentId"))
```
nimrodkor, line 25: @rotemavni Let's use `self.assertDictEqual` and ensure the values are unchanged. I want rendering and stuff to be caught by this, and `DeploymentId` is hardcoded so it's only 1 case...

nimrodkor (commented):
@rotemavni Very nice! minor comments

tests/terraform/graph/utils/test_utils.py
rotemavni: @nimrodkor no, we removed the `decode_graph_property_value` so this logic doesn't need to be tested.
We still have the logic of avoid evaluating strings with double underscores, which protects us even better.

checkov/cloudformation/graph_builder/local_graph.py
rotemavni: @nimrodkor good point, I'm not sure yet. I'll remove the exclusion for now

nimrodkor (approved):

PR #1381, opened 2021-07-14
Python empty init

atombrella:

I found a couple more, compared to #1342. 

https://github.com/hadolint/hadolint checks for the missing `rm -rf /var/lib/apt/lists/*` instruction. `checkov` doesn't do that. The Dockerfile also misses a tag, instead of the implied `latest`.


schosterbarak:
thanks @atombrella

PR #1382, opened 2021-07-14
Enable running yaml on graph checks

rotemavni:
Created two CFN yaml checks:
* MSKClusterLogging: Ensure MSK Cluster logging is enabled
* SagemakerNotebookEncryption: Ensure all data stored in the Sagemaker Notebook is securely encrypted at rest

Created also infrastructure to test yaml policies using the CloudformationGraphManager.

### DIFF
--- a/checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
+++ b/checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
@@ -1,6 +1,6 @@
 metadata:
   id: "CKV2_AWS_24"
-  name: "Ensure all data stored in the Sagemaker Notebook is securely encrypted at rest"
+  name: "Ensure SageMaker Notebook is encrypted at rest using KMS CMK"
   category: "ENCRYPTION"
 definition:
     cond_type: attribute
--- a/checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
+++ b/checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
@@ -1,0 +1,10 @@
+metadata:
+  id: "CKV2_AWS_24"
+  name: "Ensure all data stored in the Sagemaker Notebook is securely encrypted at rest"
+  category: "ENCRYPTION"
+definition:
+    cond_type: attribute
+    attribute: KmsKeyId
+    operator: exists
+    resource_types:
+      - AWS::SageMaker::NotebookInstance
--- a/checkov/common/graph/graph_builder/graph_components/blocks.py
+++ b/checkov/common/graph/graph_builder/graph_components/blocks.py
@@ -144,7 +144,7 @@ class Block:
     def get_export_data(self) -> Dict[str, Union[bool, str]]:
         return {"type": self.block_type, "name": self.name, "path": self.path}
 
-    def get_base_attributes(self) -> Dict[str, Union[str, List[str], Dict[str, Any]]]:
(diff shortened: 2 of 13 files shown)


checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
```diff
@@ -0,0 +1,10 @@
+metadata:
+  id: "CKV2_AWS_24"
+  name: "Ensure all data stored in the Sagemaker Notebook is securely encrypted at rest"
```
gruebel, line 3: the name is misleading here, the notebook instances are encrypted by default. a text like
```
"Ensure SageMaker Notebook is encrypted at rest using KMS CMK"
```
would be nicer :)

checkov/common/graph/graph_builder/graph_components/blocks.py
```diff
@@ -41,7 +41,7 @@ def _extract_inner_attributes(self) -> Dict[str, Any]:
         attributes_to_add = {}
         for attribute_key in self.attributes:
             attribute_value = self.attributes[attribute_key]
-            if type(attribute_value) is list and len(attribute_value) > 0 and type(attribute_value[0]) is dict:
+            if isinstance(attribute_value, dict) or (isinstance(attribute_value, list) and len(attribute_value) > 0 and isinstance(attribute_value[0], dict)):
```
gruebel, line 44: `isinstance(attribute_value[0], dict))` wouldn't this break, if `attribute_value` is a dict?

checkov/common/graph/graph_builder/graph_components/blocks.py
```diff
@@ -144,6 +144,17 @@ def update_inner_attribute(
     def get_export_data(self) -> Dict[str, Union[bool, str]]:
         return {"type": self.block_type, "name": self.name, "path": self.path}
 
+    def get_base_attributes(self) -> Dict[str, Union[str, List[str], Dict[str, Any]]]:
```
gruebel, line 147:
```suggestion
    def get_base_attributes(self) -> Dict[CustomAttributes, Union[str, List[str], Dict[str, Any]]]:
```

checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
rotemavni: @gruebel I agree with you, I do want to keep it uniform with the matching terraform check - https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/SagemakerNotebookEncryption.py
Do you think I should prefer clarity over consistency?

checkov/common/graph/graph_builder/graph_components/blocks.py
rotemavni: @gruebel It wouldn't break because it either a dict or it's a list, and if so - check `attribute_value[0]`

checkov/cloudformation/checks/graph_checks/aws/SagemakerNotebookEncryption.yaml
rotemavni: I renamed both checks :)

nimrodkor (approved):
Almost too simple @rotemavni

checkov/common/graph/graph_builder/graph_components/blocks.py
gruebel: Yeah, you are right. Didn't see the or at the beginning 😅

PR #1383, opened 2021-07-14
Fix entropy-keyword plugin to use analyze_line

nimrodkor:

detect-secret's plugins require you to create a `analyze_string`, but this specific plugin needs to leverage the logic in the underlying plugins' `analyze_line`. Changed implementation accordingly.


rotemavni (approved):
@nimrodkor Great unit tests!

PR #1387, opened 2021-07-15
Relax argument requirements when using --list

mikeurbanski1:

This PR fixes an annoyance where using `--list` with an API key (which will download custom policies from the platform) still requires a repo-id, even though it will not be used.

It also refactors the logic for how different BC_SOURCE values affect the behavior to make it a little more generic. This refactor was required to make this change work (otherwise, with the platform integration, the repo-id was required to set up S3 credentials, but those creds wouldn't get used).

### DIFF
--- a/checkov/common/bridgecrew/bc_source.py
+++ b/checkov/common/bridgecrew/bc_source.py
@@ -1,14 +1,30 @@
-from typing import Optional
-
-# Helper methods for determining behavior based on different BC_SOURCE values. Python enums are limiting here, and
-# using a class is overkill
+from dataclasses import dataclass
 
 
-def should_upload_results(source: Optional[str]) -> bool:
-    """Whether scan results should be uploaded to the platform. Default for unknown sources is False."""
-    if source == 'vscode':
-        return False
-    elif source == 'cli':
-        return True
+class SourceType:
+    def __init__(self, name: str, upload_results: bool):
+        self.name = name
+        self.upload_results = upload_results
+
+
+@dataclass
+class BCSourceType:
+    VSCODE = 'vscode'
+    CLI = 'cli'
+    DISABLED = 'disabled'  # use this to indicate that #TODO
+
+
+SourceTypes = {
+    BCSourceType.VSCODE: SourceType(BCSourceType.VSCODE, False),
+    BCSourceType.CLI: SourceType(BCSourceType.CLI, True),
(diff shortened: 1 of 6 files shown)


checkov/common/bridgecrew/bc_source.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def should_upload_results(source: Optional[str]) -> bool:
+    """Whether scan results should be uploaded to the platform. Default for unknown sources is False."""
+    if source == 'vscode':
+        return False
+    elif source == 'cli':
+        return True
+    else:
+        return False
```
gruebel, line 14: not sure, why there should be an extra `vscode` check?
```suggestion
    if source == 'cli':
        return True
    return False
```
or an `else` at the end, whichever style you prefer 🙂

mikeurbanski1: Because vscode is a valid source.

gruebel: But it will still return `False` like on the `else` case. So, there is no real benefit for checking the `source` is `vscode`. When you change the method, then all the tests will still pass.

mikeurbanski1: Fair enough, but I am using this as an ad hoc enum (since Python's enums seem limiting here), and it seems better to me to be explicit.

mikeurbanski1: I plan to take a dataclass approach that Nimrod showed me, to make it more explicit for each source

gruebel: Ok, sounds good.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -58,7 +58,7 @@ def __init__(self):
         self.timestamp = None
         self.scan_reports = []
         self.bc_api_url = os.getenv('BC_API_URL', "https://www.bridgecrew.cloud/api/v1")
-        self.bc_source = os.getenv('BC_SOURCE', 'cli')
```
mikeurbanski1, line 61: This is not needed anymore, because we use a singleton instance of this class. These used to be a (really bad) workaround for the 2-3 different paths for initializing this class.

nimrodkor (approved):

PR #1388, opened 2021-07-15
Use BC Check IDs throughout user interface

mikeurbanski1:

This PR adds an option to display BC platform ID numbers in the output.

It also adds the ability to use BC IDs in the --check/skip and --hard/soft-fail-on flags (no need to use the option; it accepts both).

Finally, it modifies the suppression parsing logic of all runners to accept BC IDs (previously this only applied to Terraform).


nimrodkor (approved):
Looks really good. However, we do have a lot of duplicating logic around the find suppressions. Perhaps after this, another PR that puts it all in one central place is in order?

mikeurbanski1:
> Looks really good. However, we do have a lot of duplicating logic around the find suppressions. Perhaps after this, another PR that puts it all in one central place is in order?

Agreed.

PR #1385, opened 2021-07-15
Definitions context builder for CFN

Saarett:
Added definitions context build function for CFN and added UTs

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1397, opened 2021-07-18
Running CloudFormation graph checks

rotemavni:
This PR allows running YAML checks on CloudFormation graph!

At the moment, the graph contains resource vertices, and has no edges. The current form of value evaluation is using `cf_context_parser.evaluate_default_refs()`, the same as it was used in CFN runner until now.

The process of building the graph can be found in the following PRs: https://github.com/bridgecrewio/checkov/pull/1386, https://github.com/bridgecrewio/checkov/pull/1385

Changes in this PR:
1. Added more methods and members to `BaseRunner` to allow creating graphs for more than one framework.
2. Extracted the logic of fetching the `graph_checks_registry`
3. Enabled passing to `cfn_runner` external definitions and context

**Enabling CFN graph checks to run**
In order to enable running graph check, one must set the following environment variable:
`CHECKOV_CLOUDFORMATION_GRAPH=true`


Saarett:
Wonderful refactor and new implementation!

checkov/cloudformation/runner.py
```diff
@@ -63,5 +70,42 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
                                 report.add_record(record=record)
+
+        if self.should_create_graph():
+            graph_report = self.get_graph_checks_report(root_folder, runner_filter)
+            merge_reports(report, graph_report)
+
         return report
 
+    def get_graph_checks_report(self, root_folder, runner_filter: RunnerFilter):
+        local_graph = self.graph_manager.build_graph_from_definitions(self.definitions)
```
Saarett, line 81: _local_graph_ is needed in the _run_ method in order to create the definitions from vertices and use them in the python checks, so I suggest to put the _build_graph_ and _save_graph_ in the run method

nimrodkor: let's create a `self.cfn_definitions` - same as the tf runner, and use it in both places

checkov/common/runners/base_runner.py
```diff
@@ -12,11 +12,43 @@
...
+
+    def get_graph_checks_report(self, root_folder, runner_filter: RunnerFilter):
+        pass
+
+    def should_create_graph(self):
+        # This allows to check for a specific check type whether we want to build a graph
+        # The expected format of BUILD_GRAPH is BUILD_GRAPH=`cloudformation,terraform`
+        # A runner will limit building graph only if it uses this method
+        graphs_to_build = os.environ.get('BUILD_GRAPH')
+        return graphs_to_build and self.check_type in graphs_to_build.lower()
```
nimrodkor, line 42: @rotemavni Let's use a more explicit env var name please - `CHECKOV_CLOUDFORMATION_GRAPH` = true.

nimrodkor (commented):
Very cool!

checkov/cloudformation/runner.py
rotemavni: @nimrodkor I created `self.definitions` which are the same as `cfn_definitions`, but I agree with saar, the build and save graph should be in `run`, the same as in terraform

checkov/common/runners/base_runner.py
rotemavni: @nimrodkor I thought that it could be a good practice to have this method reused for all frameworks in the future

nimrodkor: We want it to be temp, not consistent. Hopefully it will be removed soon, once we feel more comfortable with performance etc

nimrodkor (approved):
LGTM

checkov/common/runners/base_runner.py
```diff
@@ -12,11 +12,36 @@
...
+    def set_external_data(self, definitions: Optional[Dict], context: Optional[Dict], breadcrumbs: Optional[Dict]):
+        self.definitions = definitions
+        self.context = context
+        self.breadcrumbs = breadcrumbs
+
+    def load_external_checks(self, external_checks_dir: List[str]):
+        pass
+
+    def get_graph_checks_report(self, root_folder, runner_filter: RunnerFilter):
+        pass
```
gruebel, line 35: should these be abstract methods?

checkov/common/runners/base_runner.py
```diff
@@ -12,11 +12,36 @@
...
+
+    def run_graph_checks_results(self, runner_filter):
+        checks_results = {}
+        for r in self.external_registries + [self.graph_registry]:
```
gruebel, line 39: this can be written without creating an extra list
```suggestion
        for r in itertools.chain(self.external_registries, [self.graph_registry]):
```

checkov/cloudformation/runner.py
```diff
@@ -1,48 +1,58 @@
...
+                for directory in external_checks_dir:
+                    cfn_registry.load_external_checks(directory)
 
-        # Filter out empty files that have not been parsed successfully, and filter out non-CF template files
-        definitions = {k: v for k, v in definitions.items() if v and isinstance(v, dict_node) and v.__contains__("Resources") and isinstance(v["Resources"], dict_node)}
-        definitions_raw = {k: v for k, v in definitions_raw.items() if k in definitions.keys()}
+        self.context = build_definitions_context(self.definitions, self.definitions_raw, root_folder)
 
-        for cf_file in definitions.keys():
+        for cf_file in self.definitions.keys():
```
gruebel, line 49:
```suggestion
        for cf_file, definition in self.definitions.items():
```
then you don't need to use `self.definitions[cf_file]` further on.

checkov/cloudformation/runner.py
```diff
@@ -63,5 +73,44 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+    def get_graph_checks_report(self, root_folder, runner_filter: RunnerFilter):
+        report = Report(self.check_type)
+        checks_results = self.run_graph_checks_results(runner_filter)
+
+        for check, check_results in checks_results.items():
+            for check_result in check_results:
+                entity = check_result['entity']
+                entity_file_abs_path = create_file_abs_path(root_folder, entity.get(CustomAttributes.FILE_PATH))
+                entity_name = entity.get(CustomAttributes.BLOCK_NAME).split(".")[1]
+                entity_context = self.context.get(entity_file_abs_path).get(CloudformationTemplateSections.RESOURCES).get(entity_name)
```
gruebel, line 97: if you try to retrieve the corresponding values, without any default, then it can result in ugly `NoneType` errors. I think it would be better just to directly access the value, if they will be there for sure via `entity[key]`.

checkov/common/checks_infra/registry.py
```diff
@@ -49,3 +51,8 @@ def load_external_checks(self, dir: str) -> None:
     def _get_resource_types(check_json: Dict[str, Dict[str, Any]]) -> Optional[List[str]]:
         provider = check_json.get("scope", {}).get("provider", "").lower()
         return resources_types.get(provider)
+
+
+def get_graph_checks_registry(check_type):
+    return Registry(parser=NXGraphCheckParser(),
+                    checks_dir=f"{Path(__file__).parent.parent.parent}/{check_type}/checks/graph_checks")
```
gruebel, line 58: that's some great grand parent 😄 always a big fan of `pathlib.Path` 👍

gruebel (commented):
Nice work 💪 
I also prepared a PR regarding the CloudFormation graph code, I cleaned up a couple of things and added a lot of type hints 🙂 but gonna wait till your changes are merged and then I probably have to do quite some rebasing or cherry picking 🙈

checkov/common/runners/base_runner.py
rotemavni: No, otherwise I'll have to force every runner to implement them, even if they don't need them yet. This way each runner can "choose" whether to implement them or not.

PR #1403, opened 2021-07-19
use a singleton registry per check type

mikeurbanski1:

Fixes an issue that prevented platform custom policies from being saved in the registry, which prevented them from running.

### DIFF
--- a/checkov/common/checks_infra/registry.py
+++ b/checkov/common/checks_infra/registry.py
@@ -56,6 +56,11 @@ class Registry(BaseRegistry):
         return resources_types.get(provider)
 
 
+_registry_instances = {}
+
+
 def get_graph_checks_registry(check_type):
-    return Registry(parser=NXGraphCheckParser(),
-                    checks_dir=f"{Path(__file__).parent.parent.parent}/{check_type}/checks/graph_checks")
+    if not _registry_instances.get(check_type):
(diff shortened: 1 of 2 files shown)


rotemavni (commented):
@mikeurbanski1 I love this idea! 
However, I noticed you have two failing UTs. I modified these and reduced the number of failing checks because I discovered these tests were loading external checks that were loaded in previous tests in that class. I think we should prevent the registry from loading unwanted checks before merging this PR.

kartikp10:
Hey @rotemavni, I tested this further to check if the singleton is causing "unexpected" checks to be added to the registry, however, in my testing I found that this is working as expected. I've tried to explain why I think this should be okay - 

1. The test cases are executed in alphabetical order. 
2. When the 2 failing UTs are run (`test_runner_specific_file` and `test_runner_valid_tf`) using a new instance of the runner, the `self.graph_registry = get_graph_checks_registry(self.check_type)` still holds the value of the singleton created by previous instances of runners. 

For the sake of sanity in the `TestRunnerValid` class, I think it makes sense to simply remove the custom checks from `runner.graph_registry.checks` in the test cases where they're added.
Let me know if that makes sense.

rotemavni (approved):
@kartikp10 Looks good to me!

PR #1405, opened 2021-07-20
Created vertices to definitions function

Saarett:

1. Created the function "convert_graph_vertices_to_definitions" for CFN
2. The function is used by the CFN runner
3. Changed the vertex name to the resource name and not resource_type.resource_name (Affects the reports)
4. Refactored tests to use the new vertex name
4. Vertex attribute's start line and end line were the Propertie's and now are the resource's start and end lines


checkov/cloudformation/runner.py
```diff
@@ -46,11 +46,19 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
 
         self.context = build_definitions_context(self.definitions, self.definitions_raw, root_folder)
 
+        # run graph checks only if environment variable CHECKOV_CLOUDFORMATION_GRAPH='true'
+        should_create_graph = os.environ.get("CHECKOV_CLOUDFORMATION_GRAPH")
+        if should_create_graph and should_create_graph.lower() == "true":
```
nimrodkor, line 51:
```suggestion
        if os.getenv("CHECKOV_CLOUDFORMATION_GRAPH", "false").lower() == "true":
```

tests/cloudformation/graph/checks/resources/MSKClusterLogging/expected.yaml
```diff
@@ -1,9 +1,9 @@
...
-  - "AWS::MSK::Cluster.ClusterAllGood"
+  - "ClusterCloudWatchLogsGood"
+  - "ClusterFirehoseGood"
+  - "Clusters3Good"
+  - "ClusterAllGood"
 fail:
-  - "AWS::MSK::Cluster.ClusterBadNoLoggingInfo"
-  - "AWS::MSK::Cluster.ClusterBadNoLoggingDisabled"
+  - "ClusterBadNoLoggingInfo"
+  - "ClusterBadNoLoggingDisabled"
```
nimrodkor, line 8: @Saarett Resource type should be part of the resource ID

nimrodkor (commented):
@Saarett Very nice!

checkov/cloudformation/runner.py
```diff
@@ -93,8 +98,8 @@ def get_graph_checks_report(self, root_folder, runner_filter: RunnerFilter):
             for check_result in check_results:
                 entity = check_result['entity']
                 entity_file_abs_path = create_file_abs_path(root_folder, entity.get(CustomAttributes.FILE_PATH))
-                entity_name = entity.get(CustomAttributes.BLOCK_NAME).split(".")[1]
-                entity_context = self.context[entity_file_abs_path][CloudformationTemplateSections.RESOURCES][entity_name]
+                entity_name = entity.get(CustomAttributes.BLOCK_NAME)
+                entity_context = self.context.get(entity_file_abs_path).get(CloudformationTemplateSections.RESOURCES).get(entity_name)
```
gruebel, line 102: how it was before, is better, because if it fails you know exactly, that a specific key was not present, now you would get errors telling you that `NoneType` doesn't have a method `get()`

tests/cloudformation/graph/checks/resources/MSKClusterLogging/expected.yaml
Saarett: @nimrodkor Resource type is still part of the vertex id, I've changed the vertex name from _resource_type.resource_name_ to _resource_name_ so it affects the expected results of this test

nimrodkor (approved):

PR #1406, opened 2021-07-20
Fix start and end line numbers of CFN definitions.

YaaraVerner:
There was a problem in the calculation of the start and end line of CFN definitions.
in yaml file - the end line was the last line before the next definition (including an empty line or comments).
for resources in yaml file - the start line was the line above the "Type" line, even if it was a comment or empty line.
in json file - the start and end lines were the line above the correct line.

I fixed the line numbers in the build_definitions_context function in cfn_utils file and added or fixed tests to check the correct values.

### DIFF
--- a/checkov/cloudformation/cfn_utils.py
+++ b/checkov/cloudformation/cfn_utils.py
@@ -128,7 +128,14 @@ def build_definitions_context(definitions, definitions_raw, root_folder):
                         start_line = attr_value.start_mark.line
                         end_line = attr_value.end_mark.line
                         # fix lines number for yaml and json files
-                        if file_path.endswith(".yaml") or file_path.endswith(".yml"):
+                        first_line_index = 0
+                        while not str.strip(definitions_raw[file_path][first_line_index][1]):
+                            first_line_index += 1
+                        # check if the file is a json file
+                        if str.strip(definitions_raw[file_path][first_line_index][1])[0] is "{":
+                            start_line += 1
+                            end_line += 1
+                        else:
                             current_line = str.strip(definitions_raw[file_path][start_line - 1][1])
                             while not current_line or current_line[0] is YAML_COMMENT_MARK:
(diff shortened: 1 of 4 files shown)


checkov/cloudformation/cfn_utils.py
```diff
@@ -126,6 +126,19 @@ def build_definitions_context(definitions, definitions_raw, root_folder):
                     if isinstance(attr_value, dict_node):
                         start_line = attr_value.start_mark.line
                         end_line = attr_value.end_mark.line
+                        # fix lines number for yaml and json files
+                        if file_path.endswith(".yaml") or file_path.endswith(".yml"):
+                            current_line = str.strip(definitions_raw[file_path][start_line - 1][1])
+                            while not current_line or current_line[0] is "#":
```
tronxd, line 132: `#` can be moved to a constant (consts file)

tronxd (approved):

nimrodkor (commented):
Very nice!
However, it is common practice for cfn templates to have the `.template` extension, i.e. `s3.template` - and this file can either be JSON or yaml. Can you add handling for that case as well?

checkov/cloudformation/cfn_utils.py
```diff
@@ -126,6 +127,19 @@ def build_definitions_context(definitions, definitions_raw, root_folder):
                     if isinstance(attr_value, dict_node):
                         start_line = attr_value.start_mark.line
                         end_line = attr_value.end_mark.line
+                        # fix lines number for yaml and json files
+                        if file_path.endswith(".yaml") or file_path.endswith(".yml"):
```
tronxd, line 131: this can be rewritten as:
`if file_ending in CF_POSSIBLE_ENDINGS:`

YaaraVerner: but I want only files in yaml format

PR #1410, opened 2021-07-21
Fix module->variable eval to select correct source variable

nimrodkor:

Also, moving cfn context building into "build only if not given from external source" `if`

### DIFF
--- a/tests/terraform/graph/graph_builder/test_local_graph.py
+++ b/tests/terraform/graph/graph_builder/test_local_graph.py
@@ -178,3 +178,23 @@ class TestLocalGraph(TestCase):
         local_graph.build_graph(render_variables=True)
 
         self.assertEqual(12, len(local_graph.edges))
+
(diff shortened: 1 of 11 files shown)


tests/terraform/graph/graph_builder/test_local_graph.py
```diff
@@ -178,3 +178,23 @@ def test_vertices_from_local_graph_module(self):
         local_graph.build_graph(render_variables=True)
 
         self.assertEqual(12, len(local_graph.edges))
+
+    def test_variables_same_name_different_modules(self):
+        resources_dir = os.path.realpath(os.path.join(TEST_DIRNAME, '../resources/modules/same_var_names'))
```
rotemavni, line 183: @nimrodkor Love your test case! 
What happens if you have a module that is not nested under `modules/same_var_names`, perhaps a sibling of `same_var_names`?

rotemavni (approved):

PR #1426, opened 2021-07-27
Fix connection check for checkov check CKV2_AZURE_17

rdenchev:
Change connection check to be based on PostgreSQL server key instead of PostgreSQL server.


nimrodkor:
Hey @rdenchev !
Since connections are from one node to another, I don't expect this to matter. Does this make a difference?
If so - would love to have a UT for it :)

rdenchev:
I've further investigated the issue. The above check only fails in case of module call and Terraform plan scan. I've created a new issue #1469 and will close this pull request.

PR #1427, opened 2021-07-27
Cloudformation graph bugfixes

rotemavni:
* Use empty dictionary when "Properties" doesn't exist in resource
* Get the correct resource name when the vertex block name contains more than one dot

### DIFF
--- a/checkov/cloudformation/graph_builder/graph_to_definitions.py
+++ b/checkov/cloudformation/graph_builder/graph_to_definitions.py
@@ -14,7 +14,7 @@ def convert_graph_vertices_to_definitions(
     for vertex in vertices:
         block_path = vertex.path
         block_type = CloudformationTemplateSections.RESOURCES.value if vertex.block_type == 'resource' else vertex.block_type
-        block_name = vertex.name.split('.')[1]  # vertex.name is "type.name" so type.name -> [type, name]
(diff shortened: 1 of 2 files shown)


checkov/cloudformation/graph_builder/graph_to_definitions.py
```diff
@@ -14,7 +14,7 @@ def convert_graph_vertices_to_definitions(
     for vertex in vertices:
         block_path = vertex.path
         block_type = CloudformationTemplateSections.RESOURCES.value if vertex.block_type == 'resource' else vertex.block_type
-        block_name = vertex.name.split('.')[1]  # vertex.name is "type.name" so type.name -> [type, name]
+        block_name = vertex.name.split('.')[-1]  # vertex.name is "type.name" so type.name -> [type, name]
```
Saarett, line 17: What’s the purpose of this change?

Saarett (commented):
Looks good

checkov/cloudformation/graph_builder/graph_to_definitions.py
rotemavni: It handles a special case where a resource's `Type: Boto3::CodeCommit.put_file`
When we create the block name it will be: `Boto3::CodeCommit.put_file.<name>`
If we use `[1]` we will get `block_name='put_file'`, if we use `[-1]` instead it will use the last index in the array

Saarett (approved):

PR #1431, opened 2021-07-27
Remove ContextParser instance from CFN runner

Saarett:

There are 2 reasons for this change:

1. The CFN runner used the ContextParser instance because it needed the non static method `extract_cf_resource_code_lines` of this class but it doesn't use it anymore (after a recent refactor), it uses only static methods
2. When invoking the `run` method of the CFN runner with external `context`, `breadcrumbs` and `definitions`, the `definitions_raw` is not calculated, and then checkov crashes for `self.definitions_raw[cf_file]` of `                cf_context_parser = ContextParser(cf_file, definition, self.definitions_raw[cf_file])
`

### DIFF
--- a/checkov/cloudformation/runner.py
+++ b/checkov/cloudformation/runner.py
@@ -70,7 +70,7 @@ class Runner(BaseRunner):
             file_abs_path = create_file_abs_path(root_folder, cf_file)
 
             if isinstance(definition, dict) and CloudformationTemplateSections.RESOURCES in definition.keys():
-                cf_context_parser = ContextParser(cf_file, definition, self.definitions_raw[cf_file])
+                cf_context_parser = ContextParser(cf_file, definition)
                 for resource_name, resource in definition[CloudformationTemplateSections.RESOURCES].items():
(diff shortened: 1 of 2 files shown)


checkov/cloudformation/runner.py
```diff
@@ -70,9 +70,8 @@ def run(
             file_abs_path = create_file_abs_path(root_folder, cf_file)
 
             if isinstance(definition, dict) and CloudformationTemplateSections.RESOURCES in definition.keys():
-                cf_context_parser = ContextParser(cf_file, definition, self.definitions_raw[cf_file])
                 for resource_name, resource in definition[CloudformationTemplateSections.RESOURCES].items():
-                    resource_id = cf_context_parser.extract_cf_resource_id(resource, resource_name)
+                    resource_id = ContextParser.extract_cf_resource_id(resource, resource_name)
```
tronxd, line 74: can be just `self.extract_cf_resource_id` :)

tronxd (approved):

PR #1434, opened 2021-07-28
Check public alb has waf

JamesWoolfenden:
This checks that aws_lb have an association to a waf, but it only if its a public lb.


JamesWoolfenden:
@rotemavni This pr is failing incorrectly those check id are being used for CF graph, what do i need to do to fix?

checkov/terraform/checks/graph_checks/aws/ALBProtectedByWAF.yaml
```diff
@@ -0,0 +1,26 @@
+metadata:
+  id: "CKV2_AWS_28"
```
rotemavni, line 2: @JamesWoolfenden If you rename the check to be `CKV2_AWS_24` the check will pass. The failing check is checking for a sequence of check ids

rotemavni (approved):

PR #1435, opened 2021-07-28
Use threads for running the BaseAttributeSolver.

YaaraVerner:
Change the run function in BaseAttributeSolver to use threads in order to make it faster.

### DIFF
--- a/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
@@ -6,6 +6,8 @@ from networkx import DiGraph
 from checkov.common.graph.checks_infra.enums import SolverType
 from checkov.common.graph.checks_infra.solvers.base_solver import BaseSolver
 
+import threading
+
 WILDCARD_PATTERN = re.compile(r"(\S+[.][*][.]*)+")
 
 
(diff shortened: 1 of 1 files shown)


checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -19,11 +21,17 @@ def __init__(self, resource_types: List[str], attribute: Optional[str], value: A
...
+        jobs = []
+        passed_vertices = []
+        failed_vertices = []
+        for _, data in graph_connector.nodes(data=True):
+            thread = threading.Thread(target=self._process_node(data, passed_vertices, failed_vertices))
+            jobs.append(thread)
+            thread.start()
+
+        for job in jobs:
+            job.join()
```
tronxd, line 33: I would use https://docs.python.org/3/library/concurrent.futures.html instead of the more raw `threading` API. From the docs it is more thread safe and efficient on memory management, and with less code

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -45,6 +53,14 @@ def get_operation(self, vertex: Dict[str, Any]) -> bool:
     def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:
         raise NotImplementedError
 
+    def _process_node(self, data, passed_vartices, failed_vertices):
+        if not self.resource_type_pred(data, self.resource_types):
+            return
+        if self.get_operation(vertex=data):
+            passed_vartices.append(data)
+        else:
+            failed_vertices.append(data)
```
tronxd, line 62: It is a best practice to return new modified lists instead of inline modifying passed lists. This way you avoid hard to debug side effects (like append)

tronxd (approved):
LGTM! had comments which are only opinions :)

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
YaaraVerner: I need to use inline modifying because multiple threads change the list in parallel.

tronxd: @YaaraVerner Do you rely on some kind of lock when modifying? Or rely on the GIL?

YaaraVerner: I rely on the GIL. I read that append function for list is thread safe.

PR #1436, opened 2021-07-28
Remove duplicate records for the same resource

rotemavni:
When having multiple modules calling the same module we create a vertex for each module. 
This can lead to having duplicate records for the same resource in the scan report.

Added a dedup function that verifies we don't have more than one record with the same:
* check id
* resource
* file path
* lines

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -1,5 +1,4 @@
 import json
-import logging
 from collections import defaultdict
 from typing import List, Dict, Union, Any, Optional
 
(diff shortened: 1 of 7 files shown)


checkov/common/output/report.py
```diff
@@ -1,4 +1,5 @@
 import json
+import logging
```
gruebel, line 2: some leftover from debugging? 😄

rotemavni: Yes, thank you!

tronxd (approved):
Beautiful!

PR #1437, opened 2021-07-28
GitLab ci example allow_failure can not be a var

dfredell:
In the hosted gitlab.com having a variable for allow_failure fails the CI Lint. Also remove the junit file when the output mode is text.

image

### DIFF
--- a/docs/4.Integrations/GitLab CLI.md
+++ b/docs/4.Integrations/GitLab CLI.md
@@ -95,6 +95,11 @@ checkov:
     # Use `script` to emulate `tty` for colored output.
     - script -q -c 'checkov -d . ; echo $? > CKVEXIT'
     - exit $(cat CKVEXIT)
+  artifacts:
+    reports:
+      junit: "checkov.test.xml"
+    paths:
+      - "checkov.test.xml"
 ```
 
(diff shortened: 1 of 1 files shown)


docs/4.Integrations/GitLab CLI.md
```diff
@@ -99,11 +95,6 @@ checkov:
     # Use `script` to emulate `tty` for colored output.
     - script -q -c 'checkov -d . ; echo $? > CKVEXIT'
     - exit $(cat CKVEXIT)
-  artifacts:
-    reports:
-      junit: "checkov.test.xml"
```
schosterbarak, line 104: why are we deleting the artifacts?

dfredell: I guess it isn't mandatory to fix the allow_failure issue. I'm removing it because it is not used as this script `checkov -d . ` outputs the console and not junit xml.

schosterbarak: can we keep it then and not delete it?

dfredell: Sounds good, restored.

schosterbarak: thanks!

PR #1442, opened 2021-07-29
CF - Check - TransferServerIsPublic

njgibbon:
Hello,

CF check for this as already done for TF.

TF Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/TransferServerIsPublic.py
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-transfer-server.html

**License**

### DIFF
--- a/tests/cloudformation/checks/resource/aws/example_TransferServerIsPublic/FAIL.yaml
+++ b/tests/cloudformation/checks/resource/aws/example_TransferServerIsPublic/FAIL.yaml
@@ -4,7 +4,5 @@ Resources:
     Type: AWS::Transfer::Server
     Properties: 
       EndpointType: "PUBLIC"
-  VPCENDPOINT:
+  NONE:
     Type: AWS::Transfer::Server
-    Properties: 
-      EndpointType: "VPC_ENDPOINT"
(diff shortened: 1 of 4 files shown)


tests/cloudformation/checks/resource/aws/example_TransferServerIsPublic/FAIL.yaml
```diff
@@ -0,0 +1,10 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Resources: 
+  PUBLIC:
+    Type: AWS::Transfer::Server
+    Properties: 
+      EndpointType: "PUBLIC"
+  VPCENDPOINT:
+    Type: AWS::Transfer::Server
+    Properties: 
+      EndpointType: "VPC_ENDPOINT"
```
gruebel, line 10: just a minor thing, wouldn't it be better to move this resource to the `PASS.yaml` and the `NONE` resource to the `FAIL.yaml`?

schosterbarak:
Thanks @njgibbon !

PR #1470, opened 2021-08-04
fix case when left side of within is null

mikeurbanski1:

Fixes a case when `get(attribute)` returns `None` and when does an `in <string>` op, which crashes.

Also adds a debug log prior to each graph check to make it easier to see which graph check crashes a run.


kartikp10 (approved):
LGTM

PR #1471, opened 2021-08-05
Support Terraform git subdirectories

laurensknoll:

This change adds support for go-getter urls such as `test.com/repo//repo/subpath` next to the existing support for urls such as `test.com/repo.git//repo/subpath`.

Parsing of these urls is implemented according to the go-getter [implementation](https://github.com/hashicorp/go-getter/blob/main/source.go). The url scheme is ignored, and the first '//' occurence is used to split the url. To validate the implementation, this PR adds basic unit tests for parsing git repository source urls.

Second to this change, the git_loader is updated to use the GitGetter result directory. Without this change, the parser would consume the invalid repo folder: the base repo path instead of the repo sub path.


nimrodkor (approved):
Very nice! thanks @laurensknoll

PR #1492, opened 2021-08-13
DynamoDB Encrypted using Customer Managed CMK - Terraform

njgibbon:
Hello,

**License**
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1493, opened 2021-08-13
Fix - Python 3.7 Integration Test - Checkov Report Step

njgibbon:
Hello,

Whilst testing https://github.com/bridgecrewio/checkov/pull/1492 I found some issues and after looking into it could seethe env var didn't match up to the container env var name.

I wanted to raise as an isolated issue to make it easier to explain.

**License**


njgibbon:
Closed as I thought I could quickly resolve the problem but I don't think I can actually control it.

PR #1504, opened 2021-08-16
Update entrypoint.sh for API key support and new args. 

kartikp10:

PR addresses the following:
1. Adds ability to use API key with checkov action. 
2. Fixes the config file flag. 
3. Refactors problem matcher logic for consistency with the bridgecrew action. 
4. Adds support for --hard-fail-on and --soft-fail-on args. 

Note - Updates to the checkov-action will be made once this PR is merged. Branch: https://github.com/bridgecrewio/checkov-action/blob/add-missing-args/action.yml


nimrodkor (approved):
LGTM

PR #1509, opened 2021-08-17
CFN Runner crashes for non dict template file and file paths that are missing in the definitions object

Saarett:

1. In some cases the template is not a dict so it crashes for template.get
2. In some cases the file path does not exist in the definitions object


rotemavni (approved):
@Saarett LGTM, please give the PR a more informative title for future reference.

PR #1510, opened 2021-08-17
CFN runner crashes during edges creation and IAMRoleAllowAssumeFromAccount check

Saarett:
1. During edges creation, when checked if a value is Parameter/Mapping/Condition/Resource, if the checked value was not a string, checkov crashed
2. During the check of IAMRoleAllowAssumeFromAccount, if the attribute AssumeRolePolicyDocument did not exist in the resource's properties, checkov crashed


checkov/cloudformation/checks/resource/aws/IAMRoleAllowAssumeFromAccount.py
```diff
@@ -23,6 +23,8 @@ def scan_resource_conf(self, conf):
                 assume_role_block = json.loads(assume_role_policy_doc)
             else:
                 assume_role_block = assume_role_policy_doc
+        else:
+            return CheckResult.FAILED
```
rotemavni, line 27: @Saarett Should be `CheckResult.UNKNOWN` IMO

rotemavni (approved):
GJ, Please review my comment

checkov/cloudformation/checks/resource/aws/IAMRoleAllowAssumeFromAccount.py
Saarett: Basically this attribute is a required property, AWS::IAM::Role resources which are missing this attribute are invalid, but I could agree it's unknown in this case

PR #1523, opened 2021-08-20
Add lambda dlq check

gruebel:

This check already existed for Terraform and I also cleaned up the tests for the check `LambdaEnvironmentCredentials` to reflect the newer way of testing.

Other than that I also adjusted the type hints for the BaseCheck classes and I probably will rework them later again, because the `@multi_signature` decorator is still a mystery for me to type hint it correctly 😵


nimrodkor (approved):
Great!

PR #1524, opened 2021-08-20
make the path used for suppressions a unix path always

mikeurbanski1:

Fixes the value for `repo_file_path` on Windows to always be a unix path. This value is never used for actual filesystem operations; it's only used to match suppressions with the platform.


kartikp10 (approved):
LGTM

PR #1529, opened 2021-08-24
Assign placeholder map and list default values when no value is available

mikeurbanski1:

This PR updates logic for `map` and `list` variables to assign a default value of `{}` or `[]`, which allows functions like `merge` and `concat` to still be evaluated.

For example, prior to this change, the following code would fail `CKV_GCP_32`:

```
variable "metadata" {
  type = map
}

resource "google_compute_instance" "default" {
  metadata = merge(var.metadata, {block-project-ssh-keys = true})
}
```

But the following code would pass:

```
variable "metadata" {
  type = map
  default = {}
}

resource "google_compute_instance" "default" {
  metadata = merge(var.metadata, {block-project-ssh-keys = true})
}
```

This PR allows the first code to pass, as well. It does not affect the case when a default value is provided or when a value is actually passed to a var.


tests/terraform/graph/variable_rendering/test_render_scenario.py
```diff
@@ -169,21 +172,24 @@ def match_blocks(self, expected_block_val, different_expected, got_block_type_li
...
     def match_resources(self, expected_block_val, different_expected, got_block_type_list, expected_block_name):
+        found = False
         for got_block_dict in got_block_type_list:
             for got_block_name, got_block_val in got_block_dict.items():
                 if got_block_name == expected_block_name:
-                    expected_resource_name = list(expected_block_val.keys())[0]
+                    # expected_resource_name = list(expected_block_val.keys())[0]
                     got_resource_name = list(got_block_val.keys())[0]
-                    if expected_resource_name != got_resource_name:
+                    if got_resource_name not in expected_block_val:
```
mikeurbanski1, line 181: FYI, dear reviewers: These changes are to support evaluating more than one resource of the same type (as in my test main.tf below).

rotemavni (approved):
Great work! Really loved the default values map.

PR #1544, opened 2021-08-26
Added functionality for multiple output formats

jkesterson:

- Treated the --output arg the same way as --external-checks-dir with multiple values accepted. Still defaults to cli. 
- Moved the junitxml print to the last action undertaken in the print function as it seemed to lag behind the other formats.
- Used '---OUTPUT DELIMITER---' as the delimiter between output types. I have not been able to find a better alternative but feel free to make the delimited whatever you all please.


nimrodkor (approved):
Thanks @jkesterson !

PR #1545, opened 2021-08-26
Specify tfvars files using an argument

mikeurbanski1:

- Adds a new argument `--var-file` to specify `x.tfvars[.json]` files that get evaluated as with native Terraform https://www.terraform.io/docs/language/values/variables.html#variable-definition-precedence
- Fixes a graph issue where the var -> tfvar edge evaluation precedence was opposite (the logic to parse the variables applies them in the correct order, but then the logic to evaluate the edges just took the first one).
- Updated test to cover full tfvar evaluation precedence: `specified var-file` > `*.auto.tfvars` [alphabetical] > `terraform.tfvars` > `variable.default`

@rotemavni there may be a better way to do the graph fix. One option would be to overwrite tfvar vertices with the same path instead of creating new ones (probably [here](https://github.com/bridgecrewio/checkov/blob/2b11680e100c7452b385ff980462f6b04d5047f0/checkov/terraform/graph_builder/graph_components/module.py#L178)). Maybe there are others. PTAL.


rotemavni (approved):
Great job, loved the test case!
Had two small comments, please take a look.

PR #1546, opened 2021-08-27
aws lambda terraform - checks encryption settings for env vars

JamesWoolfenden:
Environmental variables for lambdas are supposed to be encrypted, this checks for that. 
Also if there no env vars theyre shouldn't be a key set, this also checks for that.


mikeurbanski1 (approved):

mikeurbanski1:
I spoke to @JamesWoolfenden . tfsec has this, hence the inclusion. James, if you are good adding this, then I am. I feel like it will be very noisy, though.

Issue #1312, opened 2021-06-21
Hidden directories are ignored, regardless of CKV_IGNORED_DIRECTORIES


ribbybibby:
Opening a new issue seen as #1292 was closed.

**Describe the bug**
Checkov will not recurse into hidden directories like `.terraform`, even when the directory is omitted from `CKV_IGNORED_DIRECTORIES`.

**To Reproduce**
1. Setup a terraform module in `/tmp/test/module` with some resources that will fail a check
2. Setup another terraform module with failures, but in `/tmp/test/.terraform/modules/module`
2. Run `CKV_IGNORED_DIRECTORIES="node_modules" checkov -d /tmp/test`
3. Note that checkov will not scan and produce failures for `/tmp/test/.terraform/modules/module`

**Expected behavior**
I would expect Checkov to scan modules under `.terraform` when it is omitted from `CKV_IGNORED_DIRECTORIES`.

**Desktop (please complete the following information):**
 - Checkov Version: 2.0.210

**Additional context**
Pretty sure this is a result of the `d.startswith(".")` check here:
https://github.com/bridgecrewio/checkov/blob/0c52d0468bc76ba8ef09b915b7648ae57327decf/checkov/common/runners/base_runner.py#L23-L24

PR #1553, opened 2021-08-31
move hidden dir to: IGNORED_DIRECTORIES_ENV

schosterbarak:
this should fix: https://github.com/bridgecrewio/checkov/issues/1312

### DIFF
--- a/checkov/common/runners/base_runner.py
+++ b/checkov/common/runners/base_runner.py
@@ -3,13 +3,30 @@ import os
 import re
 from abc import ABC, abstractmethod
 from typing import List, Dict, Optional, Any
-from distutils.util import strtobool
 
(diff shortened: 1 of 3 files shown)


nimrodkor (approved):

checkov/common/runners/base_runner.py
```diff
@@ -3,12 +3,14 @@
 import re
 from abc import ABC, abstractmethod
 from typing import List, Dict, Optional, Any
+from distutils.util import strtobool
```
gruebel, line 6: Just a heads up, `distutils` will be deprecated with Python 3.10.
https://www.python.org/dev/peps/pep-0632/#specification

schosterbarak: good point. copying the specific method to our code base

PR #1560, opened 2021-09-01
Prevent checkov crash when rendering Fn::Sub

Saarett:

1. checkov crashes while rendering Fn::Sub value which is a list and not a string. For now, skip values of list type
2. changed `logging.info` to `debug` level in some places


tronxd (approved):
LGTM

PR #1564, opened 2021-09-01
CFN breadcrumbs

Saarett:

For CFN breadcrumbs we have to insert to each breadcrumb a new field - the `attribute_key` field, which points to the specific attribute in the associated `parameter/resource/mappings/etc`.
In TF breadcrumbs this is not the case as the `attribute_key` is not relevant, we always take the default value.
For the purpose of this change I've created a new class called `BreadcrumbMetadata` as previously we held a list of vertex ids and now we need a list of pairs of vertex ids & their appropriate attribute key. I would appreciate a better name for this class :)


rotemavni (approved):
GJ @Saarett

PR #1565, opened 2021-09-01
Fail gracefully on BC integration feature errors

kartikp10:

This PR handles errors caused when the integration feature registry is run. An example of this would be, if an API token gets an error response while fetching custom policies from the platform, checkov will still complete the scan without erroring out.

### DIFF
--- a/checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
+++ b/checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
@@ -31,7 +31,7 @@ class CustomPoliciesIntegration(BaseIntegrationFeature):
             logging.debug(f'Found {len(self.policies)} custom policies from the platform.')
         except Exception as e:
             self.integration_feature_failures = True
-            logging.error(f'{e} \n Scanning without applying custom policies from the platform.')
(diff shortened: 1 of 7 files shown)


checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -17,16 +17,21 @@ def __init__(self, bc_integration):
...
+            self.policies = self._get_policies_from_platform()
+            for policy in self.policies:
+                converted_check = self._convert_raw_check(policy)
+                resource_types = Registry._get_resource_types(converted_check['metadata'])
+                check = self.platform_policy_parser.parse_raw_check(converted_check, resources_types=resource_types)
+                get_graph_checks_registry("terraform").checks.append(check)
+            logging.debug(f'Found {len(self.policies)} custom policies from the platform.')
+        except Exception as e:
+            self.integration_feature_failures = True
+            logging.debug(f'{e} \nScanning without applying custom policies from the platform.')
```
mikeurbanski1, line 34: Use `logging.debug('...', exc_info=True)`, which will include the full stack trace and exception message.

mikeurbanski1 (approved):

Issue #1538, opened 2021-08-25
Graph checks defined in checkov are being considered as external checks


meghaddn:
Hello Checkov Team,

Using checkov==2.0.360

All graph_checks defined at https://github.com/bridgecrewio/checkov/tree/master/checkov/terraform/checks/graph_checks/aws are being considered external Checks.

**What I've done**
Looking into the code I noticed load_checks function which is called for external and graph checks at https://github.com/bridgecrewio/checkov/blob/master/checkov/common/runners/base_runner.py#L55 basically considers both external and graph checks as external based on https://github.com/bridgecrewio/checkov/blob/master/checkov/common/checks_infra/registry.py#L47

**What I expect**
I expect checkov to not consider graph_checks as external checks since only the ones specified as --external-checks-dir directory are external checks.

Is this a design decision?

**Use Case**
As a security team, we maintain an allowlist of checks which we alert on for all the teams. External checks are allowed by default so teams can add external checks specific for their use case. Since these graph checks are being considered as external checks they are getting triggered even though they are not in the allowlist.

Thanks for your help and guidance!

meghaddn:
[https://github.com/bridgecrewio/checkov/pull/1575](url) PR created for this. Already merged

PR #1575, opened 2021-09-03
Updated code to not consider graph checks as external

meghaddn:

This PR is raised for [#1538](url)
It adds some logic to ensure that when internal graph_checks are loaded they are not added as external checks. Checks loaded via load external_checks are considered as external checks.


mikeurbanski1:
Hi @meghaddn - thanks for your contribution! This was a great catch.

Can you please move your tests to `tests/common/checks` directory? The `integration_features` relates to integration with the platform repo, so this isn't the best spot for them.

meghaddn:
Sure, Thanks. Updated tests to tests/common/checks.

mikeurbanski1 (approved):

PR #1577, opened 2021-09-05
Remove unwanted encoding to unicode null value `\u0000`

rotemavni:
When evaluating strings, the eval method encodes and decodes strings with unicode.
This may cause generating a null value `\u0000`. 

This PR removes this null value.


Saarett (approved):
LGTM

PR #1572, opened 2021-09-03
Fixes Github upload sarif results failure

harringj:
Closes #1521.

This PR adds some logic to ensure the rules in the rules array are unique, which is required by the SARIF schema. It also adds a unit test to validate the change. The test checks to see that a case where two records have the same rule is handled properly. If you add the test to the current master branch it fails; with the logic changes to report.py in this PR it passes.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1587, opened 2021-09-07
Bugfix/value error on retrieval of rule index

### LINKED ISSUE #1586: Fix for #1521 (GH CodeQL Upload Sarif) contains a bug
As it turns out, my fix for this issue contains a bug: 
image
As the screenshot indicates, the cause is https://github.com/bridgecrewio/checkov/blob/507340f8bc7e9d9b58263c9a66bbdd05752dae3a/checkov/common/output/report.py#L212 The call to rules.index(rule) is searching for the current rule object in the rules array, but no such object will exist whenever the values of record.check_id are the same. There will, however, be another rule object in the array with the same id.

I'm opening a PR, which I'll link to this issue, which resolves the error by comparing the value of rule['id'] with each rule in the array until it finds the index of the rule with the matching id. Apologies for letting this slip past me in the first place!

_Originally posted by @harringj in https://github.com/bridgecrewio/checkov/issues/1521#issuecomment-914581519_

harringj:
Fixes #1586.

This PR fixes I bug I introduced in PR #1572. In the current state, when multiple findings reference the same Checkov rule, get_sarif_json() will fail with a ValueError.

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -209,7 +209,9 @@ class Report:
                 ruleset.add(record.check_id)
                 rules.append(rule)
             else:
-                idx = rules.index(rule)
+                for r in rules:
+                    if r['id'] == rule['id']:
+                        idx = rules.index(r)
             if record.file_line_range[0] == 0:
                 record.file_line_range[0] = 1
             if record.file_line_range[1] == 0:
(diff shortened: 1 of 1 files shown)


mikeurbanski1:
LGTM; thanks @harringj

PR #1590, opened 2021-09-08
Bugfix/incorrect index in rules array

harringj:
One small problem still exists with the way get_sarif_json() functions: it does not guarantee that the index of a rule is unique and correct within the SARIF results. This PR adds logic to fix this, as well as two test cases to guard against regression.

Relates to #1586 and #1587.

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -194,7 +194,8 @@ class Report:
         rules = []
         results = []
         ruleset = set()
-        for idx, record in enumerate(self.failed_checks):
+        idx = 0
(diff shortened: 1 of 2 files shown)


checkov/common/output/report.py
```diff
@@ -194,7 +194,8 @@ def get_sarif_json(self) -> Dict[str, Any]:
         rules = []
         results = []
         ruleset = set()
-        for idx, record in enumerate(self.failed_checks):
+        idx = 0
```
schosterbarak, line 197: what impact does this change make?

Issue #1585, opened 2021-09-07
Plan file scan slowness - json.loads custom decoder


larryboymi:
**Describe the bug**
With a larger plan JSON to check, seeing times of 20-30 minutes due to slowness using [custom decoder](https://github.com/bridgecrewio/checkov/blob/3081a8560f6369465314ee8f4ac8a8ec01649d68/checkov/terraform/context_parsers/tf_plan/tf_plan_json.py#L346) in the plan parsing section of the code.

When I remove the custom parser, and use either the regular `json` or `ujson` decoder without a custom class, the time to decode drops to almost no time.

Is there anything I can do to improve the performance? My plan example contains data I can't necessarily share here, but is about a 9-10MB JSON file

Note: I am still actively debugging which part of the custom decoder is the culprit, might just its use in general. Running a profile on it to see what might be the biggest issue.

Deleted cumulative times less than a second from call list (note for line numbers I have added a few debug logging statements so they probably don't line up with master currently).

```txt
   ncalls  tottime  percall  cumtime  percall   filename:lineno(function)
        1     0.000     0.000   1355.540   1355.540   <string>:1(<module>)
        1     0.000     0.000   1353.681   1353.681   __init__.py:15(parse)
       55     0.000     0.000   1353.469     24.609   __init__.py:299(loads)
12011/2041     0.101     0.000   1249.348     0.612   decoder.py:217(JSONArray)
       55     0.000     0.000   1353.469     24.609   decoder.py:332(decode)
       55     0.001     0.000   1353.468     24.609   decoder.py:343(raw_decode)
        1     0.001     0.001   1355.540   1355.540   main.py:52(run)
        1     0.000     0.000   1354.567   1354.567   plan_parser.py:119(parse_tf_plan)
        1     0.000     0.000   1354.567   1354.567   plan_runner.py:30(run)
        1     0.000     0.000   1355.131   1355.131   runner_registry.py:38(run)
  32731/1     1.329     0.000   1353.467   1353.467   tf_plan_json.py:149(cfn_json_object)
 193722/1     0.338     0.000   1353.467   1353.467   tf_plan_json.py:272(_scan_once)
 168372/8     0.299     0.000   1353.432    169.179   tf_plan_json.py:311(scan_once)
   213114     1.169     0.000   1347.026     0.006   tf_plan_json.py:321(get_beg_end_mark)
        1     0.100     0.100   1353.681   1353.681   tf_plan_json.py:333(load)
[...]

PR #1596, opened 2021-09-09
Terraform Plan performance enhancement

larryboymi:

The goal of this PR is to help achieve the spirit of #1585 . 

I believe both the Cloudformation JSON [decoder](https://github.com/bridgecrewio/checkov/blob/3081a8560f6369465314ee8f4ac8a8ec01649d68/checkov/cloudformation/parser/cfn_json.py#L323) as well as the ARM JSON [decoder](https://github.com/bridgecrewio/checkov/blob/3081a8560f6369465314ee8f4ac8a8ec01649d68/checkov/arm/parser/cfn_json.py#L323) utilize similar logic and could probably also use this performance boost. I did not modify those however as it was not my current use case.

I haven't contributed to this project before, but the unit tests are passing and I am open to review.

In my case, I reduced a 22 minute scan of a 3MB terraform plan down to 32 seconds, and a 26 minute scan of a 9MB plan down to 14 seconds.

Thanks!


checkov/terraform/context_parsers/tf_plan/tf_plan_json.py
```diff
@@ -317,14 +213,41 @@ def scan_once(string, idx):
 
     return _scan_once
 
+def find_indexes(s, ch='\n'):
+    """Finds all instance of given char and returns list of indexes """
```
gruebel, line 217:
```suggestion
    """Finds all instances of given char and returns list of indexes """
```

checkov/terraform/context_parsers/tf_plan/tf_plan_json.py
```diff
@@ -317,14 +213,41 @@ def scan_once(string, idx):
...
+        if (arr[mid] <= key):
+            count = mid + 1
+            left = mid + 1
+        else:
+            right = mid - 1
+    return count
+
+def largest_less_than(indexes, line_num, pos):
+    """Replacement func for python str.rfind using indexes """
+    return indexes[line_num-1] if count_occurrences(indexes, pos) and len(indexes) > 1 else -1
```
gruebel, line 239:
```suggestion
    return indexes[line_num-1] if count_occurrences(indexes, pos) and indexes else -1
```
This should be a bit faster, since you don't need to calculate the length. I'm not sure, if it would be better to switch the `and` conditional sides, depends how often `indexes` is empty 🤔

checkov/terraform/context_parsers/tf_plan/tf_plan_json.py
```diff
@@ -317,14 +213,41 @@ def scan_once(string, idx):
...
     """Get the Start and End Mark """
-    beg_lineno = s.count('\n', 0, start)
-    beg_colno = start - s.rfind('\n', 0, start)
+    beg_lineno = count_occurrences(indexes, start)
+    beg_colno = start - largest_less_than(indexes, beg_lineno, start)
     beg_mark = Mark(beg_lineno, beg_colno)
-    end_lineno = s.count('\n', 0, end)
-    end_colno = end - s.rfind('\n', 0, end)
+
+    offset = 1 if len(indexes) > 1 else 0
```
gruebel, line 248:
```suggestion
    offset = 1 if indexes else 0
```
No need to check the length

checkov/terraform/context_parsers/tf_plan/tf_plan_json.py
```diff
@@ -354,16 +277,126 @@ class CfnJSONDecoder(json.JSONDecoder):
...
         self.parse_string = py_scanstring
         self.memo = {}
         self.object_pairs_hook = check_duplicates
         self.scan_once = py_make_scanner(self)
+        self.newline_indexes = []
+
+    def decode(self, s):
+        """Overridden to retrieve indexes """
+        self.newline_indexes = find_indexes(s)
+        obj = super(CfnJSONDecoder, self).decode(s)
```
gruebel, line 291:
```suggestion
        obj = super().decode(s)
```
No need to write it here, it will call the correct super type `decode`.

gruebel (commented):
great improvement 💪

larryboymi:
Thanks @gruebel ... modified per your comments.

schosterbarak:
Hey @larryboymi this is pretty awesome. i'd love to send some swag your way! I'll send a DM via linkedin. :)

PR #1598, opened 2021-09-10
Basic support for generic json document scanning

BrentSouza:
My company is currently evaluating Bridgecrew and one of the things that we're looking for in a policy as code framework is the ability to scan arbitrary json input. The thinking here is that:

1. It's generally easy to get some kind of configuration from various components in our stack as JSON output.
2. It's relatively simple to write checkov custom checks against these config docs 
3. It's considerably more effort to create a fully-fledged runner implementation for each component of our stack that we want to check

Using the json parser from the terraform plan implementation, this turns out to be pretty trivial to implement with a very basic feature set. This is a very quick first pass at providing the following functionality:

* Parse JSON documents into a relevant python dict, list, etc
* A base JSON check that should be used to write custom checks against documents
* Narrow the scope of a check to relevant properties within a JSON document using JMESpath-lite syntax, e.g. `a.b.c[0]`
* Run checks against all items in a JSON array
* Run checks against properties of a JSON object
* Run checks against an entire JSON document

I've tested this locally with some checks we would like to run against our internal GitLab repositories, which can help show how this feature is useful.

Two checks we would like to run are:

1. Users can't push changes directly into a default branch 
2. Merge Requests require at least 2 approvals

Here's an example of the MR approvals check:

```py
from checkov.common.models.enums import CheckCategories, CheckResult
from checkov.json.base_json_check import BaseJsonCheck
from checkov.json.enums import BlockType


class MergeRequestRequiresApproval(BaseJsonCheck):
    def __init__(self):
        name = "Merge requests should require at least 2 approvals"
        id = "CKV_GITLAB_2"
        categories = [CheckCategories.CONVENTION]
        super().__init__(
            name=name,
            id=id,
            categories=categories,
            supported_entities=["approvals"],
            block_type=BlockType.OBJECT,
        )

        if conf["approvals"]["approvals_before_merge"] < 2:
            for rule in conf["approval_rules"]:
                if rule["approvals_required"] >= 2:
                    return CheckResult.PASSED, rule
            return CheckResult.FAILED, conf
        return CheckResult.PASSED, conf["approvals"]


check = MergeRequestRequiresApproval()
```

Before we go much further with building this out, I wanted to get the checkov team's feedback on the concept of a generic json runner. Is this something that makes sense to include with the project?


checkov/json/runner.py
```diff
@@ -0,0 +1,73 @@
...
+                    # lambda k: f"{os.sep}{os.path.relpath(k, os.path.commonprefix((root_folder, k)))}"
+                )
+
+        for json_file_path in definitions.keys():
+            results = registry.scan(
+                json_file_path, definitions[json_file_path], [], runner_filter
+            )
+            for check, result in results.items():
+                result_config = result["results_configuration"]
+                start = result_config.start_mark.line
```
schosterbarak, line 55: are we populating this var with a value somewhere?

BrentSouza: Yes. This is using the json parser that is already in the codebase for terraform plans which captures line numbers along with the deserialized values. The custom checks should return the passing/failing configuration along with the check status. That configuration will have the start/end mark properties as a result of the tf plan parser. Does that make sense?

BrentSouza: I was also wondering if you would want to do something like move the tf plan parser into common as a general JSON parser? There doesn't seem to be anything terraform specific about it, but I didn't want to move something like that without discussing it first.

schosterbarak: i think it is the right thing to do. you'll need to rebase from master because we had a contribution with performance enhancement there a day ago

BrentSouza:
@schosterbarak I started pulling all of the json parsing out into `common.parsers` and `common.parsers.json` and updated arm, cloudformation, serverless, and terraform runners to use the common json parser. Please let me know if this looks ok!

checkov/json/runner.py
```diff
@@ -0,0 +1,73 @@
...
+        parsing_errors = {}
+
+        def load_files(files_to_load, filename_fn=None, key_fn=None):
+            for file in files_to_load:
+                f = filename_fn(file) if filename_fn else file
+                key = key_fn(f) if key_fn else f
+                (definitions[key], definitions_raw[key]) = parse(f)
+
+        if not external_checks_dir:
+            logging.error("The json runner requires that external checks are defined.")
```
nimrodkor, line 30: @BrentSouza Can we lower this to a warning log?

checkov/json/runner.py
```diff
@@ -0,0 +1,73 @@
...
+        report = Report(self.check_type)
+        definitions = {}
+        definitions_raw = {}
+        parsing_errors = {}
+
+        def load_files(files_to_load, filename_fn=None, key_fn=None):
+            for file in files_to_load:
+                f = filename_fn(file) if filename_fn else file
+                key = key_fn(f) if key_fn else f
+                (definitions[key], definitions_raw[key]) = parse(f)
```
nimrodkor, line 27: @BrentSouza Any reason to put this as a nested function instead of a function of `Runner`?

checkov/terraform/plan_runner.py
```diff
@@ -85,7 +85,7 @@ def get_entity_context_and_evaluations(self, entity):
 
     def check_tf_definition(self, report, runner_filter):
         for full_file_path, definition in self.tf_definitions.items():
-            scanned_file = f"/{os.path.relpath(full_file_path)}"
+            scanned_file = f"{os.path.sep}{os.path.relpath(full_file_path)}"
```
nimrodkor, line 88: @BrentSouza We're using this convention across all our runners, wouldn't change this for now

checkov/json/base_registry.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+        results[check] = {}
+        if result["result"] == CheckResult.SKIPPED:
+            results[check]["result"] = result["result"]
+            results[check]["suppress_comment"] = result["suppress_comment"]
+            results[check]["results_configuration"] = None
+        else:
+            results[check]["result"] = result["result"][0]
+            results[check]["results_configuration"] = result["result"][1]
```
nimrodkor, line 148: @BrentSouza Why do we get the conf back from the check? IF it's the one `update_result` gets, can we leverage that?

tests/generic_json/checks/object/PropHasValue.py
```diff
@@ -0,0 +1,27 @@
...
+            id=id,
+            categories=categories,
+            supported_entities=["foo"],
+            block_type=BlockType.OBJECT,
+        )
+
+    def scan_entity_conf(self, conf):
+        if "prop" in conf and conf["prop"] == "value":
+            return CheckResult.PASSED, conf
+        return CheckResult.FAILED, conf
```
nimrodkor, line 24: @BrentSouza This is matching my previous comment - usually we don't return the `conf` as part of the scan result

nimrodkor (commented):
Very nice @BrentSouza !

Some minor comments

checkov/terraform/plan_runner.py
BrentSouza: Ah yeah this leaked into a commit from changes I was making to get the unit tests to pass on Windows. I'll revert it.

Sidenote: It seems there is a mix between code using `os.path` functions to get os specific paths and hardcoded `/`'s. That causes a bit of noise when running things from Windows. I did find a possible bug as a result of this behavior but I didn't want to overload this pull request.

checkov/json/base_registry.py
BrentSouza: @nimrodkor I was following along with the Dockerfile implementation, which does something similar. The conf that gets returned back from the check could be a property of the conf that is passed into `update_result`. For example, the check gets invoked for the following json node:

```json
[
  { "prop": true },
  { "prop": false}
]
```
Let's say we have a check that ensures all `prop` values are `true`:

```py
for item in conf:
    if not item["prop"]:
        return CheckResult.FAILED, item
return CheckResult.PASSED  
```

In this case the failing check returns the JSON node that failed, which we can then use to print the relevant code snippet since the node contains the line information as well as the json values.

Does this make sense or would you prefer I follow a different runner implementation?

checkov/json/runner.py
BrentSouza: Sure

checkov/json/runner.py
BrentSouza: I felt it was only relevant in the context of the call to `run` and not so much for the Runner itself so I nested it. I can move it if you want to keep a consistent style.

checkov/json/base_registry.py
BrentSouza: @nimrodkor I thought about this a bit more and in the general case, I think users wouldn't need to return the config, so I made some changes to make returning the failed configuration node optional. Now if you just return the CheckResult, it will use the `entity_configuration` value in `update_result` as you noted.

Another option I can think of to completely get rid of returning a node could be to do something with `get_evaluated_keys` so that it returns the path to the failed key in the configuration. Ideally, this would happen within the base check or base_registry but I haven't fully thought it through yet. What do you think?

checkov/json/runner.py
nimrodkor: @BrentSouza Got it. Would prefer if you moved it 🙏 
Make it private `_load_files` if you feel it is relevant only inside the class

checkov/json/base_registry.py
nimrodkor: Love it!

nimrodkor (approved):

schosterbarak:
thank you @BrentSouza don't forget to apply for hacktober fest 👍 https://bridgecrew.io/blog/happy-hacktoberfest-2021/

metahertz:
Epic contribution @BrentSouza!!!!

🎉

Issue #1109, opened 2021-04-21
AttributeError: 'NoneType' object has no attribute 'get' with an empty yaml file in custom dir


dwilliams782:
**Describe the bug**
When an empty yaml file is declared in an external checks dir, you get this AttributeError

`AttributeError: 'NoneType' object has no attribute 'get'`

**To Reproduce**
1. Create a directory with an empty yaml file
2. Point checkov at that custom directory: `checkov -f <your-file>.tf --external-checks-dir <your-path>`

**Expected behavior**
The empty file should be ignored

**Desktop (please complete the following information):**
 - OS: os X
 - Checkov Version: 2.0.64

njgibbon:
I'll work on resolving this :)

njgibbon:
@gruebel - another one to close.

gruebel:
Thanks to @njgibbon it is fixed since version `2.0.418`

PR #1600, opened 2021-09-10
Fix External-Checks-Dir Incorrect YAML Crash

njgibbon:
Hello,

Raising to resolve user issue: https://github.com/bridgecrewio/checkov/issues/1109

When using `--external-checks-dir` the program will crash if it loads an empty file or it loads a file which can not be serialised to a `dict`.

The later downstream code is resilient to being the wrong type of `dict`. Subsequent methods and functions will try to `get` fields and will fail gracefully if they do not exist or are `None`. However, they are not resilient about having `check_json` passed in to them which is not a `dict`. 

I first starting checking if we were dealing with a `dict` in each case and catching the `AttributeError` exception however I realised this would involve several changes.

This solution is the minimal change to fix the issue and is an improvement.

Arguably we could check for more than it just being a dict `here`. I.e use a helper function to validate that all of the fields we may later want to use exist. However, as I mention, all of the downstream functions are already resilient to some expected fields not existing so I wouldn't want to over-engineer.

**License**


schosterbarak:
thanks @njgibbon can we add a UT for it?

njgibbon:
@schosterbarak - yes can do, will look at this

njgibbon:
@schosterbarak  - I've added tests related to the conditions that i've added.

These are the first UTs in this area so there is a lot of room to add more.

JamesWoolfenden (approved):
LGTM

PR #1613, opened 2021-09-14
strips leading slash from record filepath

### LINKED ISSUE #1599: Bug: file preview not working with SARIF results uploaded from Checkov to Github
Being able to view the code responsible for a security alert in Github's code scanning feature is crucial for developers. The file preview provides important contextual information for devs that helps them arrive at a fix without a lot of context switching.

Other tools, like CodeQL and tfsec, currently display file previews in code scanning without issue. The individual result view for Checkov, however, currently looks like this. 

image

Note the odd double slash in front of the filename. My hunch is that the double slash is preventing code scanning's rendering engine from displaying a file preview because, in this example, //backend.tf is not a valid filepath, as it needs to be per https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#physicallocation-object.

The bug seems to be affecting the individual result view specifically, since the list view does have a valid filepath, as can be seen here:

image

I don't know Checkov's codebase well enough to suggest a fix -- and it looks like Bridgecrew is committed to absolute URIs, which may make a fix trickier -- but `record.file_path` is somehow ending up with an additional prepended slash it shouldn't have:

https://github.com/bridgecrewio/checkov/blob/a217a1a04d92ed0ca6456ed5e3e13f4403f0eef1/checkov/common/output/report.py#L230

harringj:
Fixes #1599.

This PR removes the prepended '/' from the record filepath in the SARIF results. By doing so, I hope to avoid a double slash in the filepath, which I believe may be causing file previews not to display in Github's code scanning results.

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -227,7 +227,7 @@ class Report:
                 "locations": [
                     {
                         "physicalLocation": {
-                            "artifactLocation": {"uri": record.file_path},
+                            "artifactLocation": {"uri": record.file_path.lstrip("/")},
                             "region": {
                                 "startLine": int(record.file_line_range[0]),
                                 "endLine": int(record.file_line_range[1]),


mikeurbanski1:
Thanks @harringj . This is one of our more unfortunate pieces of tech debt, as this slash is EVERYWHERE. Stripping it just in the relevant output makes the most sense at this point.

PR #1614, opened 2021-09-14
Restore broken BASELINE_FLAG option in entrypoint.sh

cr-latacora:
The baseline input option is ignored for GitHub Action runs because the BASELINE_FLAG command line argument was dropped in commit 0b6a21e6f8127abc4f7d31e91f59266c8e79dee. This commit restores BASELINE_FLAG to the options used for command display and the checkov run.


schosterbarak:
thank you @cr-latacora

Issue #1594, opened 2021-09-09
feat(operators): Add less_than, less_than_or_equals, greater_than, and greater_than_or_equals for graph


AjkayAlan:
## The Problem

We would like to write tests which can assert that numbers are less than our greater than a value. For example, given we have the following CloudFormation:

```json
{
  "Resources": {
    "Vpc8378EB38": {
      "Type": "AWS::EC2::VPC",
      "Properties": {
        "CidrBlock": "10.250.0.0/16"
      }
    },
    "Acl": {
      "Type": "AWS::EC2::NetworkAcl",
      "Properties": {
        "VpcId": {
          "Ref": "Vpc8378EB38"
        }
      }
    },
    "BadAclEntry": {
      "Type": "AWS::EC2::NetworkAclEntry",
      "Properties": {
        "NetworkAclId": {
          "Ref": "Acl"
        },
        "Egress": false,
        "PortRange": {
          "From": 21,
          "To": 23
        }
      }
    }
  }
}
```

We would like to write a rule that makes sure that the `PortRange` on `AWS::EC2::NetworkAclEntry` does not encompass port 22 for ingress. Right now, we have a rule like the following for ingress

```yaml
metadata:
  name: "vpc network acls should not allow public access to ssh"
  id: "CUSTOM_VPC_NETWORK_ACL_NO_SSH"
  category: "GENERAL_SECURITY"
definition:
  or:
    - cond_type: attribute
      resource_types:
        - AWS::EC2::NetworkAclEntry
      attribute: Egress
      operator: equals
      value: true
    - and:
      - cond_type: attribute
        resource_types:
          - AWS::EC2::NetworkAclEntry
        attribute: PortRange.From
        operator: not_equals
        value: 22
      - cond_type: attribute
        resource_types:
          - AWS::EC2::NetworkAclEntry
        attribute: PortRange.To
        operator: not_equals
        value: 22
```

This rule will pass, since From is 21, and To is 23.

## The Solution

Allow new operators of `greater_than`, `greater_than_or_equals`, `less_than`, and `less_than_or_equals`. This would allow me to write a policy similar to the following (in Pseudocode):

```
For AWS::EC2::NetworkAclEntry
OR
   If Egress equals true
   AND
      From < 22
      AND To < 22
   AND
      From > 22
      To > 22
```

## Alternatives

Right now this can be written using Python rules instead of the graph based YAML ones.

## Other

Thanks to @vangundy-jason-pfg for the research and example code/policy

[...]

PR #1618, opened 2021-09-15
gt, lt, gte, lte operators

mikeurbanski1:

Adds >, >=, <, <= operators

In all cases, it will try to operate on the values as numbers (floats) if possible, otherwise as strings


AjkayAlan:
Closes #1594

checkov/common/graph/checks_infra/enums.py
```diff
@@ -33,6 +33,10 @@ class Operators:
     NOT_EQUALS = 'not_equals'
     REGEX_MATCH = 'regex_match'
     NOT_REGEX_MATCH = 'not_regex_match'
+    GREATER_THAN = 'greater_than'
```
schosterbarak, line 36: can we add a test that validates what happens with those operators on string, int and map values?

mikeurbanski1: I have tests for all the combinations of string / int / float.

Anything that can't be converted to a float is compared as a string. The behavior is undefined for if it's a map or list. I will add a test to sanity check that it doesn't break.

PR #1604, opened 2021-09-13
add regex_match and not_regex_match operators

mikeurbanski1:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Adds regex_match and not_regex_match operators

PR #1023, opened 2021-03-30
Checkov 2.0

nimrodkor:
# Checkov 2.0
Dear community!
​
We're happy to announce the upcoming version of checkov, which includes a major upgrade to checkov's capabilities and content!
 
TL;DR - Checkov v2 will include:
1. Graph-based variable resolution, for terraform only at this stage. This is an upgrade from our older, iterative variable rendering algorithm (kudos @Robeden!). [More details...](#graph-based-variable-resolution)
2. Performance Improvement - with the replacement of the old, iterative algorithm with the graph comes significant performance improvements. [More details...](#performance-improvement)
3. Support for a brand new, multi-resource query language on top of the graph. This allows for simple, yaml-based check writing over resources and their connections. [More details...](#yaml-based-checks)
4. 230+ new policies across AWS, GCP, Azure and K8s resources. [More details...](#over-230-new-policies-across-aws-gcp-azure-and-k8s)
5.  5 new policies analyzing `Dockerfile` for best practices and CIS related checks. [More details](https://github.com/nimrodkor/checkov/pull/224)

We hope this version will help you find and avoid even more security misconfigurations!
 
Contributor? Questions? Feel free to comment on this PR, or jump into the conversation in our Slack channel [Slack channel](https://slack.bridgecrew.io).
 
We are also hosting contributor sessions for one on one Q&A with the Bridgecrew Checkov 2.0 development team, for those looking to deep dive into the code paths, new graph-powered checks format and any other development questions! [Pick a slot via Calendly](https://calendly.com/metahertz/checkov-2-developer-deepdive)
	
​
===========
​
## Graph-based variable resolution
One of the most repeated, important requests we've been receiving is the fact that variable resolution, as one of the main features of IaC, needs to be handled. In this version we introduce the usage of [networkx](https://networkx.org/), an open source in-memory graph library.
Using this library we create a graph linking the variable values with the relevant attributes.
This allows for faster evaluation and back-tracing the source of the misconfiguration.
The process is as follows:
1. All HCL files are parsed, as we've done before.
2. A local graph is built, and we then resolve references between the existing nodes to evaluate the values of the different variables.
3. The graph is saved as a networkx graph, which allows us to run queries.
​
## Performance Improvement
We've tested this out over many open source repositories and seen a great performance improvement.
Let's take [terragoat](https://github.com/bridgecrewio/terragoat) as a benchmark. We've ran v1 and v2 over terragoat 20 times:
Checkov 1.x average run time: 3086401 ms / 3 seconds
[...]

PR #1623, opened 2021-09-17
docs: Add missing operators

AjkayAlan:
This adds newer operators to the docs, or ones that weren't originally there on checkov 2.0's release but were missed in the docs. 

- Regex was added in #1604 
- gt, gte, lt, lte was added in #1618 
- Any and Within was added in #1023


nimrodkor (approved):
TY!

Issue #1589, opened 2021-09-08
Scan is slow since 2.0.357


TLaborde:
**Describe the bug**
Running the scan at the root of the repo is 6 times slower than running it for each subfolder. The issue appeared around version 2.0.357.

**To Reproduce**
1. downgrade to 2.0.356, time a scan of the repo
2. update to current version, time a scan of the repo
3. Compare

**Expected behavior**
The scan should be faster.

**Screenshots**
version -> time in sec to run scan
version 2.0.352 : 106.2096581
version 2.0.353 : 103.2969057
version 2.0.354 : 83.3623289
version 2.0.355 : 84.4924343
version 2.0.356 : 76.3720223
version 2.0.357 : 1086.3829189
version 2.0.358 : 778.0082139
version 2.0.359 : 1442.6692375
version 2.0.402 : 674.2031332

on 2.0.402, running on 22 modules one by one, time in seconds:
```
Count    : 22
Average  : 4.40065963181818
Sum      : 96.8145119
Maximum  : 5.3321704
Minimum  : 3.8702054
```
The sum of module time is lower than running checkov at the root

**Desktop (please complete the following information):**
 - OS: tested on windows 2019 and windows 10

schosterbarak:
@TLaborde we had a contribution by @gruebel fixing a few performance issues that should be released in the next 30minutes. please reopen the ticket if it still an issue

PR #1626, opened 2021-09-18
Improve performance

gruebel:

Relates #1589 

I still didn't figure out, why the scans became so much slower like described in the issue, but I did some profiling on the project, while scanning one of my Terraform projects.

I started with following
initial_time
and finished with this 🙂 (profiling slows the run down dramatically, without it takes ~ 8 sec)
final_time

Taking a look on the worst methods (top 10), I started with this
initial_own_time
and finished with this 💪 
final_own_time

For the `<listcomp>` and `fnmatchcase` I will create a separate PR, which will dramatically reduce it usage, it is related to usgae of the `cloudsplaining` lib.

After improving `resource_type_pred` I got following time
improve_2
It looks like `resource_type` is not present so often, so it makes to first check its existence. Also removing the length calculation reduced the time by 500ms.

After improving `_get_dest_module_path` I got following time
improve_3
reason behind is, if a module can't be found an empty list will be assigned to the cache and if this module is referenced several times through the Terraform config, then it will be again checked. So, I changed the condition check to just check the existence of the module name in the cache.

After removing a couple of `deepcopy` and optimizing the usage of it, if it was really needed (`deepmerge` is also using `deepcopy` quite a lot) I got following time
improve_6

I also fixed some other minor things, which are not really worth mentioning, like incorrect `inspected_key` definitions.

### DIFF
--- a/checkov/common/graph/checks_infra/solvers/base_solver.py
+++ b/checkov/common/graph/checks_infra/solvers/base_solver.py
@@ -26,4 +26,4 @@ class BaseSolver:
 
     @staticmethod
     def resource_type_pred(v: Dict[str, Any], resource_types: List[str]) -> bool:
-        return len(resource_types) == 0 or v.get("resource_type") in resource_types
+        return not resource_types or ("resource_type" in v and v["resource_type"] in resource_types)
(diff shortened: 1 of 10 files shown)


schosterbarak:
Thanks anton!

PR #1627, opened 2021-09-18
CF - Check - LambdaEnvironmentEncryptionSettings

njgibbon:
Hello,

Added CF check as done for TF.

No special case where kms key without env vars causes a problem like with Terraform.

**CF Docs**: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html
**Terraform Check**: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/LambdaEnvironmentEncryptionSettings.py

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/LambdaEnvironmentEncryptionSettings.py
+++ b/checkov/cloudformation/checks/resource/aws/LambdaEnvironmentEncryptionSettings.py
@@ -1,0 +1,36 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.cloudformation.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class LambdaEnvironmentEncryptionSettings(BaseResourceCheck):
+    def __init__(self):
+        name = "Check encryption settings for Lambda environmental variable"
+        id = "CKV_AWS_173"
+        supported_resources = ['AWS::Lambda::Function']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        env_vars_in_use = False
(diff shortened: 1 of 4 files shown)


checkov/cloudformation/checks/resource/aws/LambdaEnvironmentEncryptionSettings.py
```diff
@@ -0,0 +1,36 @@
...
+                if len(env_vars_config) > 0:
+                    env_vars_in_use = True
+
+            if kms_config is not None:
+                kms_in_use = True
+
+        if env_vars_in_use and not kms_in_use:
+            return CheckResult.FAILED
+            
+        return CheckResult.PASSED
```
gruebel, line 33:
```suggestion
    def scan_resource_conf(self, conf):
        properties = conf.get('Properties')
        
        if properties:
            env_config = properties.get('Environment')
            
            if env_config and env_config.get('Variables') and not properties.get('KmsKeyArn'):
                return CheckResult.FAILED

        return CheckResult.PASSED
```
How about this? If I'm not mistaken, this should be equivalent to yours.

njgibbon: Hiya. Re-check now i've streamlined. 

Not equivalent because I was checking that env variables was a list > 0 as that's done in tf check and then because of that it was getting quite nested so I wanted to get the values and bring comparison to first function indent level.

I think it should be more than sufficient to just see that the env.variables exists and take that as an indicator that the user would be trying to use env vars.

gruebel: Yeah, I get it. Sometimes you go for specific use cases and make it work, but it is always good to step back at the end and look again, if you maybe overachieved it 😄

Issue #777, opened 2021-01-11
Feature request: Support checkov:skip in modules


ismailyenigul:
**Describe the solution you'd like**
The current skip only works with terraform `resources`. I would like to be able to skip checks for modules. If I add  a skip comment into  terraform `module(s)`, then the checks should be skipped for all resources under that module.

Thanks

metahertz:
Hey @ismailyenigul 

Is this basically "Ignore this module", or looking to skip specific checks only across a module by adding a list of checks to skip once in the module definition?

ismailyenigul:
Hi @metahertz 
The latter option. Don't check specific X,Y,Z checks for  this module resources.
something like the following:

```
module "sg" {
  #checkov:skip=CKV_AWS_24,CKV_AWS_XYZ:The is testing sg
  source      = "./sg"
  vpc_id      = "vpc-123"
  company     = "c1"
  environment = "dev"
  application = "test"
  cidrs       = ["0.0.0.0/0"]
}
```

codepay4999:
this is very useful. We are using third party modules and this will be helpful for us. 

At the moment, checkov will be helpful since it is flagging third party modules which we have no control over it

metahertz:
Thanks fo the the confirmation @ismailyenigul.

I'll take a look at the current flow (not played with the new module scanning code much!), have a sneaking suspicion we wont be holding state to easily pass skips to a specific module, but it's a needed enhancement/feature nontheless!

ismailyenigul:
Hi @metahertz 
Is there any ETA for this feature? :)
Thanks

camaeel:
But it works for modules as well.

schosterbarak:
fixed at: https://github.com/bridgecrewio/checkov/pull/1629

PR #1629, opened 2021-09-20
feat: skip rules when defined in a module

josh-barker:
Hi,

This is a potential solution #777.

I found that the `skipped_checks` property is already set on all entity types, i.e. modules, resources, etc.
So, I did not need to change the parse algorithm specifically for modules.

From here, I considered 2 potential solutions:
1. Push the skipped checks from one layer down to the next
2. When processing a resource, look up the graph to find other skip checks.

I implemented option 1 as it seemed simple and logical.
Note: This implementation will push skipped checks from a module to all entity types - modules, data, resources and provider.

Looking forward to your feedback.

### DIFF
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -264,6 +264,9 @@ class Runner(BaseRunner):
                 entity_code_lines = None
                 skipped_checks = None
 
+            if block_type == "module" and skipped_checks is not None:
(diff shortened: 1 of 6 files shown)


schosterbarak:
hey @josh-barker  
 thanks for the contribution. looks like some tests are still failing.

josh-barker:
Hey @schosterbarak , I've fixed the failing test.

schosterbarak:
thank you josh!

Issue #822, opened 2021-01-27
Unbalanced eval brackets causes TF parser loop


robeden:
**Describe the bug**
From this ([link 1](https://codifiedsecurity.slack.com/archives/CSMKB8XPH/p1611741949003000), [link 2](https://codifiedsecurity.slack.com/archives/CSMKB8XPH/p1611743126004100)) slack conversation, a malformed eval statement with mismatched brackets can cause a loop in the TF parser. Example statement (note missing closing bracket):
```
s3_access_logs_prefix = "${replace(var.cdn_logging_prefix, "cdn", "s3")/${var.bucket_name}"
```

It is currently unknown if it is an infinite loop, but it's at least very slow.

**To Reproduce**
I have reproduced with this parse scenario:
_main.tf_:
```
locals {
  # This is intentionally missing the closing bracket
  s3_access_logs_prefix = "${replace(var.cdn_logging_prefix, "cdn", "s3")/${var.bucket_name}"
}
```

**Expected behavior**
_expected.json_:
```
{
  "main.tf": {
    "locals": [
      {
        "s3_access_logs_prefix": ["${replace(var.cdn_logging_prefix, 'cdn', 's3')/${var.bucket_name}"]
      }
    ]
  }
}
```
I wouldn't expect extra looping in the parser.

robeden:
I'm looking into this...

robeden:
@schosterbarak - Looks like the loop is in the hcl2 parser. It never returns from the `hcl2.load(f)` call.

robeden:
I've put a test case here ☝️ . The problem is in the `python-hcl2` library (or at least the BC fork of it), so we'll need some fix there.

metahertz:
🤘 Rock on! Thanks Rob, know you've been speaking with Barak+Ariel on the issue.

Adding a bit of context for other watchers on the issue:
Parser in use by checkov is here: https://github.com/bridgecrewio/python-hcl2 Assigning @arielkru and work-in-progress

arielkru:
Hi @robeden & @metahertz !
I've been looking for several ways to solve this issue, but didn't find a way that wouldn't break other well-written tf files, and since this issue is based on a broken tf format anyway, I think we shouldn't change the parser, but would like to know if you have other ideas, here are the details:
[...]

PR #1640, opened 2021-09-22
Timeout TF file parsing when stuck

Saarett:

Fixes issue #822 

In some cases when a TF file is malformed, the hcl2 parser hangs "forever" and therefore checkov hangs too. This solution puts the parsing method in a child process under a 60 seconds timeframe.
1. If the parsing succeeds before the time is up, the process is finished and closed automatically (as usual)
2. If the parsing fails before the time is up, the process is finished and closed automatically, the file is added to the "parsing error" dict (as usual)
3. If the parsing isn't finished under the 60 seconds cap, the process will be killed and the file will be added to the "parsing error" dict

### DIFF
--- a/checkov/terraform/parser.py
+++ b/checkov/terraform/parser.py
@@ -668,7 +668,12 @@ def _hcl2_load_with_timeout(f: io.TextIOWrapper) -> Dict:
     q = Queue()  # used by the child process to return its result
     p = Process(target=_hcl2_load, args=(q, f))
     p.start()
-    raw_data = q.get(block=True, timeout=60)  # Wait until the file is parsed, up to 60 seconds. A timeout will raise a queue.Empty exception
+    try:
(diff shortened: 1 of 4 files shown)


checkov/terraform/parser.py
```diff
@@ -661,6 +662,24 @@ def _load_or_die_quietly(file: os.PathLike, parsing_errors: Dict,
...
+    p.join()  # Make sure the process is finished and closed
+    return raw_data
+
+
+def _hcl2_load(q: Queue, f: io.TextIOWrapper) -> None:
+    try:
+        raw_data = hcl2.load(f)
+        q.put(raw_data)
+    except Exception:
+        q.put(None)
```
schosterbarak, line 681: can you please log which file have failed and on what exception?

schosterbarak: + are those added to the report parsing errors dictionary?

Saarett: Yes, they are added to the report

schosterbarak (approved):

Issue #1584, opened 2021-09-07
Failure scanning plan of aws_autoscaling_group resource with tags argument


aucheun-tu:
**Describe the bug**
When scanning the terraform plan output of an aws_autoscaling_group resource that utilizes the tags argument. I receive a "checkov.terraform.context_parsers.tf_plan.node.TemplateAttributeError: str_node.keys" is invalid error. 

**To Reproduce**
Create a main.tf file using the following template.
```
resource "aws_autoscaling_group" "example" {
  max_size             = 1
  min_size             = 1
  health_check_type    = "ELB"
  vpc_zone_identifier  = ["test"]
  launch_configuration = "test"

  tags = [{
    key                 = "test"
    value               = "test"
    propagate_at_launch = true
    },
    {
      key                 = "test2"
      value               = "test2"
      propagate_at_launch = true
  }]
}
```
Run the following commands.
```
terraform init
terraform plan -out test.binary
terraform show -json test.binary > test.json
checkov -f test.json
```
Below is the error that occurs.
```
Traceback (most recent call last):
  File "/venv/bin/checkov", line 8, in <module>
    exit(run())
  File "/venv/lib/python3.9/site-packages/checkov/main.py", line 146, in run
    scan_reports = runner_registry.run(external_checks_dir=external_checks_dir, files=config.file,
  File "/venv/lib/python3.9/site-packages/checkov/common/runners/runner_registry.py", line 39, in run
    scan_report = runner.run(root_folder, external_checks_dir=external_checks_dir, files=files,
  File "/venv/lib/python3.9/site-packages/checkov/terraform/plan_runner.py", line 57, in run
    tf_definitions, template_lines = parse_tf_plan(file)
  File "/venv/lib/python3.9/site-packages/checkov/terraform/plan_parser.py", line 137, in parse_tf_plan
    resource_block, prepared = _prepare_resource_block(resource, conf)
  File "/venv/lib/python3.9/site-packages/checkov/terraform/plan_parser.py", line 92, in _prepare_resource_block
    resource_block[resource["type"]][resource.get("name", "default")] = _hclify(resource["values"], expressions)
  File "/venv/lib/python3.9/site-packages/checkov/terraform/plan_parser.py", line 54, in _hclify
    child_list.append(_hclify(internal_val, internal_conf_val))
  File "/venv/lib/python3.9/site-packages/checkov/terraform/plan_parser.py", line 63, in _hclify
    for conf_key in conf.keys() - obj.keys():
[...]

Issue #1522, opened 2021-08-19
Terraform Plan Parsing Error when running checkov -f tfplan.json on multi-provider Terraform Configuration


CCOLLOT:
Hi, I am currently facing an issue while passing a terraform plan (json format) to the checkov -f command (parsing ERROR) as follows:

terraform plan  -var-file=values.prod.tfvars --out=tfplan.binary
terraform show -json tfplan.binary > tfplan.json 
checkov -f tfplan.json
```
Traceback (most recent call last):
  File "/home/ccollot/.local/bin/checkov", line 8, in <module>
    exit(run())
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/main.py", line 186, in run
    scan_reports = runner_registry.run(external_checks_dir=external_checks_dir, files=config.file,
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/common/runners/runner_registry.py", line 49, in run
    scan_report = runner.run(
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/plan_runner.py", line 62, in run
    tf_definitions, template_lines = parse_tf_plan(file)
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/plan_parser.py", line 137, in parse_tf_plan
    resource_block, prepared = _prepare_resource_block(resource, conf)
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/plan_parser.py", line 92, in _prepare_resource_block
    resource_block[resource["type"]][resource.get("name", "default")] = _hclify(resource["values"], expressions)
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/plan_parser.py", line 54, in _hclify
    child_list.append(_hclify(internal_val, internal_conf_val))
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/plan_parser.py", line 63, in _hclify
    for conf_key in conf.keys() - obj.keys():
  File "/home/ccollot/.local/lib/python3.8/site-packages/checkov/terraform/context_parsers/tf_plan/node.py", line 39, in __getattr__
    raise TemplateAttributeError('%s.%s is invalid' % (self.__class__.__name__, name))
checkov.terraform.context_parsers.tf_plan.node.TemplateAttributeError: str_node.keys is invalid
```

[...]

PR #1642, opened 2021-09-22
Fix plan parser edge case and tag processing

bpoole:
There are essentially two changes in this PR

1 - add some sanity checks to make sure we don't try to work on dict's when they are str's. I made it work but someone more familiar with it definitely needs a closer look. Prevents fatal exceptions when parsing terraform plans it doesn't like (ASG w/ tags in my case)
2 - encapsulate Tags in an extra [] for a certain code path. This fixes a problem I observed where Tags specified on an aws_autoscaling_group were returned as a [[{}]] by terraform's parser and [{}] by terraform_plan's parser. This makes them both [[{}]]. 

This change works for me with some reasonable testing. Note, I added a partial test case for #1 but not for #2. The existing test case hurt my brain too much to try and fully emulate.

### DIFF
--- a/checkov/terraform/plan_parser.py
+++ b/checkov/terraform/plan_parser.py
@@ -51,15 +51,19 @@ def _hclify(obj: dict_node, conf: Optional[dict_node] = None, parent_key: Option
             child_list = []
             conf_val = conf.get(key, []) if conf else []
             for internal_val, internal_conf_val in itertools.zip_longest(value, conf_val):
-                child_list.append(_hclify(internal_val, internal_conf_val))
-            ret_dict[key] = child_list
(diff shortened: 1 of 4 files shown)


bpoole:
I believe this PR will also address the following two issues:

https://github.com/bridgecrewio/checkov/issues/1522
https://github.com/bridgecrewio/checkov/issues/1584

schosterbarak:
this is awesome. thanks @bpoole

schosterbarak:
@bpoole  thanks a lot! please fill-up the form and we'll send some swag your way: https://bridgecrew.io/event/2021-hacktoberfest/

PR #1662, opened 2021-09-27
add env var to also scan .hcl files as if they were tf

mikeurbanski1:

This may allow us to usefully process .hcl files in certain cases that would otherwise be ignored


kartikp10 (approved):
Looks good!

PR #1669, opened 2021-09-28
use fork for multiprocessing

mikeurbanski1:

This fixes an issue that users on Mac with Python 3.8 and 3.9 started reporting, where all TF file parsing threw errors. The root cause was a change in `multiprocessing` introduced in Python 3.8, which deprecates the old logic in favor of forking.

Source for the fix: https://github.com/pytorch/pytorch/pull/36542/files

Relevant threads:
https://bugs.python.org/issue33725
https://github.com/pytorch/pytorch/issues/36515
https://github.com/GoogleCloudPlatform/gsutil/issues/961


Saarett (approved):
Looks great! Please remove the old commented functions just to clean it up.

nimrodkor:
It looks good, but the performance on this is not good enough yet

mikeurbanski1:
> It looks good, but the performance on this is not good enough yet

@nimrodkor agreed

PR #1672, opened 2021-09-29
fix: do not raise error when parsing empty local terraform blocks

josh-barker:
When terraform code has an empty `local` block, the parser errors.

This PR ensures that empty blocks are skipped.


schosterbarak:
Thanks @josh-barker ! don't forget to apply for hacktoberfest with this PR and previous ones too :)

PR #1676, opened 2021-09-29
Improve runtime of variable rendering

rotemavni:
This PR includes several changes to make the variable rendering run faster:

1. Added an optional parameter for `Block/get_attribute_dict` method, that allows skip calculating the block hash. This value is not needed in the variable rendering algorithm and it can take time to calculate in large graphs.

2. Added a tracking list for the previously evaluated edges. The list helps to track loops from evaluating:
Let's say we have these edges:
```
a->b
b->a
```
We render `a->b` first, then `b->a` and we'll reach an infinite loop. The algorithm detects if you had more than 90% identical edges 2 loops ago, and if it happens more than 4 times we will break the rendering loop.

3. Computing the next `edges_to_render` without having duplicate edges (using sets instead of lists)


schosterbarak (approved):

checkov/common/graph/graph_builder/variable_rendering/renderer.py
```diff
@@ -17,6 +17,8 @@ def __init__(self, local_graph: "LocalGraph") -> None:
         self.local_graph = local_graph
         self.run_async = True if os.environ.get("RENDER_VARIABLES_ASYNC") == "True" else False
         self.max_workers = int(os.environ.get("RENDER_ASYNC_MAX_WORKERS", 50))
+        self.duplicate_percent = int(os.environ.get("RENDER_EDGES_DUPLICATE_PERCENT", 90))
+        self.duplicate_iter_count = int(os.environ.get("RENDER_EDGES_DUPLICATE_ITER_COUNT", 4))
```
nimrodkor, line 21:
```suggestion
        self.duplicate_percent = int(os.getenv("RENDER_EDGES_DUPLICATE_PERCENT", 90))
        self.duplicate_iter_count = int(os.getenv("RENDER_EDGES_DUPLICATE_ITER_COUNT", 4))
```

Issue #1595, opened 2021-09-09
feat(metadata): Allow custom policies to provide guideline urls


AjkayAlan:
## The Problem

Currently if I write a custom policy, and someone's resource violates that policy, they get the feedback in the console which states that they failed, but don't get a "guide" url. It would be hugely beneficial if my custom policies can incorporate guides to help engineers fix their resources.

## The Solution

This can be tackled in a number of ways. A few ideas:

- Allow "guide" in the metadata of a custom policy. If provided, it should be used in terminal output.
- Allow an "external-checks-guide" CLI arg that can be used with "external-checks-git"
- Create a repo template that people can use with "external-checks-git", perhaps with a "guidelines.json" file at the root, showing how you can map your ids to guidelines. Allow docs to be inlined in the git repo (markdown?) or allow external urls

njgibbon:
This has also been requested as part of this issue: https://github.com/bridgecrewio/checkov/issues/1433 which will be closed because the bug part can be easily worked around.

njgibbon:
Also requested here: https://github.com/bridgecrewio/checkov/issues/1115

njgibbon:
The current implementation obtains the IDs dynamically with an API call which has all of the Check ID to Check Guideline mappings.

I think that adding it as an optional field to the base class would be a good thing. If it's set then the guideline is used. If it's not set then nothing will change and the guideline will be obtained dynamically if it exists.

I think that I will look in to implementing this.

schosterbarak:
@njgibbon @AjkayAlan currently centralized guideline management is a platform feature at bridgecrew.cloud . you can start using it for free :)

gruebel:
thanks to @vangundy-jason-pfg this is possible since version `2.0.484` 🚀

PR #1682, opened 2021-09-29
allow custom policies to use a custom guideline url

vangundy-jason-pfg:
This PR addresses: https://github.com/bridgecrewio/checkov/issues/1595

The majority of the change and the reason this touches many files here is purely plumbing.  `Record` already contained the `guideline` property and the existing code was handling including the guideline url in the output / report.  Prior to this PR those guidelines were exclusively being set (separately) by https://github.com/bridgecrewio/checkov/blob/master/checkov/common/runners/runner_registry.py#L162.  This change allows subclasses of `BaseCheck` and `BaseGraphCheck` to propagate a custom guideline into the check itself, whereby the `Runner` implementations can then leverage `check.guideline` to include in the `Record`.

The hurdle here was that there are so many subclasses of `BaseCheck` and each one (prior) had explicit keyword args without allowing for variable kwargs.  I have added that, future-proofing subclasses from explicitly needing to know what kwargs `BaseCheck` may need.   

I believe I have modified each of the relevant `Runner`s to handle adding the guideline to the `Record`.  Notable exceptions include the runners for _helm_ and _secrets_.  Helm delegates to the _kubernetes_ runner and the _secrets_ implementation does not follow the same inheritance pattern with `BaseCheck` etc.  Presumably the intent was not for users of checkov to use the baked in secrets tools for scanning and not necessarily implement their own.

I have two questions around your desired testing around this functionality, given that it's mostly plumbing.  
- I wrote one example test showing how I would test the propagation of the `**kwargs` back to the `BaseCheck`.  Do you want me to include this?  Do you want me to write this for the other subclasses as well?
- Likewise, I wrote one example test showing how I would test that a runner is setting `guideline=check.guideline` on each `Record`.  Do you want me to include this?  For all of them?  This one is a little more tricky given that some of the runners have a number of logic paths that each end up crafting `Record`s in a loop.  Thoughts on this? 

### DIFF
--- a/checkov/arm/base_parameter_check.py
+++ b/checkov/arm/base_parameter_check.py
@@ -6,9 +6,9 @@ from checkov.common.multi_signature import multi_signature
 
 
 class BaseParameterCheck(BaseCheck):
-    def __init__(self, name, id, categories, supported_resources):
+    def __init__(self, name, id, categories, supported_resources, **kwargs):
         super().__init__(name=name, id=id, categories=categories, supported_entities=supported_resources,
-                         block_type="parameter")
+                         block_type="parameter", **kwargs)
         self.supported_resources = supported_resources
         arm_parameter_registry.register(self)
 
--- a/checkov/arm/base_resource_check.py
+++ b/checkov/arm/base_resource_check.py
@@ -6,9 +6,9 @@ from checkov.common.multi_signature import multi_signature
 
 
 class BaseResourceCheck(BaseCheck):
-    def __init__(self, name, id, categories, supported_resources):
+    def __init__(self, name, id, categories, supported_resources, **kwargs):
         super().__init__(name=name, id=id, categories=categories, supported_entities=supported_resources,
-                         block_type="resource")
+                         block_type="resource", **kwargs)
         self.supported_resources = supported_resources
         arm_resource_registry.register(self)
 
--- a/checkov/arm/runner.py
+++ b/checkov/arm/runner.py
@@ -103,7 +103,8 @@ class Runner(BaseRunner):
                                                 code_block=entity_code_lines, file_path=arm_file,
                                                 file_line_range=entity_lines_range,
                                                 resource=resource_id, evaluations=variable_evaluations,
-                                                check_class=check.__class__.__module__, file_abs_path=file_abs_path)
+                                                check_class=check.__class__.__module__, file_abs_path=file_abs_path,
+                                                guideline=check.guideline)
                                 report.add_record(record=record)
 
                 if 'parameters' in definitions[arm_file].keys():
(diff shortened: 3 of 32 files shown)


schosterbarak:
thanks @vangundy-jason-pfg integration tests are failing. can you please fix them before we'll merge?

vangundy-jason-pfg:
> thanks @vangundy-jason-pfg integration tests are failing. can you please fix them before we'll merge?

absolutely. my apologies - it took me a bit to figure out what i did wrong here.  hopefully got it now!

gruebel:
I don't think it is a good idea to introduce `kwargs` here, it makes it really hard to interact with the Class. You could use any parameter you like now and don't know what will actually happen downstream. Also if you misspell the `guideline` parameter you won't know, why it doesn't work.

schosterbarak:
I understand what anton is saying. @vangundy-jason-pfg  can you add additional param instead of kwargs?

nimrodkor:
Hey @vangundy-jason-pfg !

This would be a great addition. Might I suggest making changes in the runner_registry, to add the guideline from the check if it exists?

Should be somewhere around here: https://github.com/bridgecrewio/checkov/blob/65f23c31decdae13b5286fddbb30639073bc6840/checkov/common/runners/runner_registry.py#L162

vangundy-jason-pfg:
> I don't think it is a good idea to introduce `kwargs` here, it makes it really hard to interact with the Class. You could use any parameter you like now and don't know what will actually happen downstream. Also if you misspell the `guideline` parameter you won't know, why it doesn't work.

Thanks for the feedback. IIUC correctly you are asking me to remove the `**kwargs` from `BaseCheck` and subclasses?  I do understand the downside that you're expressing, but to me there is a very significant downside of not having the kwargs.  With all of the Check subclassing and inheritance, in this case of adding guideline I needed to change all of the subclasses to propagate it to the BaseCheck.  If we have another use case like this we'll have to deal with all of that class coupling again.  In my view this is / was worth the change.  You feel strongly that I should change it to an explicit kwarg still though @gruebel ?

gruebel:
@vangundy-jason-pfg as the Zen of Python states

_Explicit is better than implicit._ 

Using `kwargs` is very error prone and usually discouraged. There are of course valid use cases, like decorators or wrapping third party functions. This parameter will be mainly used by `checkov` users, so I think and prefer to be here very explicit 🙂

vangundy-jason-pfg:
@gruebel I believe I have what you're asking for here: https://github.com/bridgecrewio/checkov/pull/1682/commits/0cc7e17413e140dacfe649b735420f7176cdf21f

celik0311:
Is this available for YAML-defined checks? I am unable to generate a custom URL for checks defined as YAML.

PR #1688, opened 2021-09-30
Change CKV_AWS_149 (secretsmanager KMS) to look for CMK

mikeurbanski1:

Changes the secrets manager encryption check to explicitly check for a CMK, as encryption using an AWS key is the default.


JamesWoolfenden (approved):
LGTM I think we need a full review of all KMS key checks to may sure they look for CMK and not AWS managed keys

nimrodkor (approved):

Issue #1262, opened 2021-06-01
Dockerfile Framework Only Scans files named "Dockerfile"


knoxknot:
**Is your feature request related to a problem? Please describe.**
The tool cannot scan file names other than Dockerfile for the dockerfile framework 
Ex. checkov -f Dockerfile --framework dockerfile #works
      checkov -f Dockerfile.prod --framework dockerfile  #does not work

**Describe the solution you'd like**
Passing a framework flag should be able to scan the file irrespective of whether it is the default naming for that framework

njgibbon:
@gruebel - This can be closed as PR resolved issue.

gruebel:
thanks to @Xitric it is fixed since version `2.0.473`

PR #1689, opened 2021-10-01
Support name variations for Dockerfiles

Xitric:
This PR addresses: #1262

It is very common for organizations to use variations of `Dockerfile` for their Docker files, especially with the advent of mono repositories. Such alternative names could include:

- `Dockerfile.prod`
- `Dockerfile.Product1`
- `dev.Dockerfile`
- `team1.product.dockerfile`

Furthermore, the VS Code plugin for Docker has also long recognized these patterns as valid Docker files and provided syntax highlighting: microsoft/vscode-docker#1908.

For this reason, it seems reasonable for Checkov to include these variations of names when scanning for Dockerfiles to analyze. This PR changes the behavior of the Docker runner to use the following regular expression for detecting Dockerfiles:

```regex
^(?:.+\.)?[Dd]ockerfile(?:\..+)?$
```

A unit test was added to ensure that the behavior is as intended. It is entirely by intention that `.Dockerfile` is not considered a valid name. With this change, Checkov is now capable of scanning a wider range of Dockerfiles:

```
dockerfile scan results:

Passed checks: 0, Failed checks: 1, Skipped checks: 0

Check: CKV_DOCKER_7: "Ensure the base image uses a non latest version tag"
        FAILED for resource: Dockerfile.prod.FROM
        File: Dockerfile.prod:0-0
        Guide: https://docs.bridgecrew.io/docs/ensure-the-base-image-uses-a-non-latest-version-tag
```

### DIFF
--- a/checkov/dockerfile/runner.py
+++ b/checkov/dockerfile/runner.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from dockerfile_parse.constants import DOCKERFILE_FILENAME
+import re
 
 from checkov.common.output.record import Record
 from checkov.common.output.report import Report
@@ -9,11 +9,15 @@ from checkov.dockerfile.parser import parse, collect_skipped_checks
 from checkov.dockerfile.registry import registry
 from checkov.runner_filter import RunnerFilter
 
-DOCKER_FILE_MASK = [DOCKERFILE_FILENAME]
+DOCKER_FILE_MASK = r"^(?:.+\.)?[Dd]ockerfile(?:\..+)?$"
 
 
 class Runner(BaseRunner):
     check_type = "dockerfile"
+
+    @staticmethod
+    def __is_docker_file(file):
+        return re.search(DOCKER_FILE_MASK, file) != None
 
     def run(self, root_folder=None, external_checks_dir=None, files=None, runner_filter=RunnerFilter(),
             collect_skip_comments=True):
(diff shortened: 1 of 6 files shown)


tronxd (approved):
LGTM! Thanks @Xitric

checkov/dockerfile/runner.py
```diff
@@ -9,12 +9,16 @@
...
-DOCKER_FILE_MASK = [DOCKERFILE_FILENAME]
+DOCKER_FILE_MASK = r"^(?:.+\.)?[Dd]ockerfile(?:\..+)?$"
 
 
 class Runner(BaseRunner):
     check_type = "dockerfile"
 
+    @staticmethod
+    def __is_docker_file(file):
+        return re.search(DOCKER_FILE_MASK, file) != None
```
gruebel, line 20:
```suggestion
        return re.match(DOCKER_FILE_MASK, file) is not None
```
a conditional check against `None` should be done with `is`, because this checks the identity and `None` is a singleton.
`match` fits better here, because it checks the string from the beginning against the pattern, compared to `search`, which looks actually somewhere inside the string. There is even `fullmatch`, which would fit this scenario even better 😄

checkov/dockerfile/runner.py
```diff
@@ -9,12 +9,16 @@
...
 
-DOCKER_FILE_MASK = [DOCKERFILE_FILENAME]
+DOCKER_FILE_MASK = r"^(?:.+\.)?[Dd]ockerfile(?:\..+)?$"
 
 
 class Runner(BaseRunner):
     check_type = "dockerfile"
 
+    @staticmethod
+    def __is_docker_file(file):
```
gruebel, line 19:
```suggestion
    def _is_docker_file(file):
```
one underscore is enough to symbolize this method is private, a double underscore does a bit more, which is called name mangling . This is only used to avoid naming conflicts with subclasses, which is not the case here.

gruebel (changes requested):

checkov/dockerfile/runner.py
Xitric: Goes to show my inexperience with Python 😉 Thanks a lot for the suggestions, I will push the changes as soon as the tests pass locally!

checkov/dockerfile/runner.py
Xitric: Learned something new today. Will definitely change this!

gruebel (approved):
glad you could learn something new about Python 🙂

PR #1692, opened 2021-10-01
Added working directory command in dockerfile

Falguni-M:
Inserted a line in Dockerfile, confirming the working directory.

### DIFF
--- a/kubernetes/Dockerfile
+++ b/kubernetes/Dockerfile
@@ -14,6 +14,7 @@ RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
 
 COPY kubernetes/run_checkov.sh /app
+WORKDIR /app
 RUN chown checkov:checkov /app/run_checkov.sh && chmod +x /app/run_checkov.sh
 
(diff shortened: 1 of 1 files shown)


kubernetes/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
 
 COPY kubernetes/run_checkov.sh /app
+WORKDIR /app
```
schosterbarak, line 17: why is this important?

Falguni-M:
The WORKDIR instruction sets the working directory for any command(RUN, CMD, etc) that follows it in the Dockerfile. If the WORKDIR doesn’t exist, it will be created even if it’s not used in any subsequent Dockerfile instruction.
Docker best practices recommend you to use it. It's totally fine to not use it as well since the "RUN chown checkov:checkov /data /app /home/checkov" command will do it automatically.

schosterbarak:
got ya. just read about it. thanks @Falguni-M

Falguni-M:
Thanks for letting me contribute! :)

PR #1702, opened 2021-10-03
CKV_AZURE_119: fix: public ip is not configured when public_ip_address_id is null

josh-barker:
This PR fixes a bug when CKV_AZURE_119 fails and the `public_ip_address_id` is null.

However the check should pass because no public interface is configured.


schosterbarak:
Thanks @josh-barker

PR #1704, opened 2021-10-04
Incomplete URL substring sanitization

c0d3x27:

Sanitizing untrusted URLs is a common technique for preventing attacks such as to request forgeries and malicious redirections. Usually, this is done by checking that the host of a URL is in a set of allowed hosts.

However, treating the URL as a string and checking if one of the allowed hosts is a substring of the URL is very prone to errors. Malicious URLs can bypass such security checks by embedding one of the allowed hosts in an unexpected location.

Even if the substring check is not used in a security-critical context, the incomplete check may still cause undesirable behaviors when the check succeeds accidentally.

### DIFF
--- a/checkov/terraform/module_loading/loaders/bitbucket_loader.py
+++ b/checkov/terraform/module_loading/loaders/bitbucket_loader.py
@@ -1,16 +1,14 @@
 from checkov.terraform.module_loading.loaders.git_loader import GenericGitLoader
+from flask import Flask, request, redirect
 
 
 class BitbucketLoader(GenericGitLoader):
-    def _is_matching_loader(self):
+    def _is_matching_loader(resquest):
+        allowlist = ["bitbucket.org/product", "bitbucket.org/login"]
         # https://www.terraform.io/docs/modules/sources.html#bitbucket
-        if 'bitbucket.org' in self.module_source:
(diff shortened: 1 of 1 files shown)


gruebel:
Thanks for the contribution, but this part is tackled in a different PR (#1287) already and `checkov` doesn't have any dependency towards `flask` and will probably never have.

checkov/terraform/module_loading/loaders/bitbucket_loader.py
```diff
@@ -1,16 +1,14 @@
 from checkov.terraform.module_loading.loaders.git_loader import GenericGitLoader
+from flask import Flask, request, redirect
 
 
 class BitbucketLoader(GenericGitLoader):
-    def _is_matching_loader(self):
+    def _is_matching_loader(resquest):
+        allowlist = ["bitbucket.org/product", "bitbucket.org/login"]
```
schosterbarak, line 7: i think this will block self hosted bitbucket instances

schosterbarak:
thanks @c0d3x27 I am closing this PR since as @gruebel  has mentioned is already managed on another contribution.

c0d3x27:
Oh I see, I'm sorry for not properly checking this out. Thank you guys, I'm deleting the branch

PR #1712, opened 2021-10-05
fix: handle nested blocks at any layer of configuration

josh-barker:
Currently, dynamic blocks are only handled when they are defined at the root level of a resource.
When a dynamic block is defined a nested block, rules will return a fail.
When a dynamic block is inside another dynamic block, rules will fail.

This PR implements a recursive search to search for dynamic blocks and process them.


schosterbarak:
Thanks @josh-barker you are a star! don't forget to apply for hacktoberfest :)

PR #1718, opened 2021-10-05
Fix `dict` verification in the loop, use `NotImplementedError` and remove unneeded default `return`

donno2048:
 - Make sure `statement` is `dict` the right way (24cc9bd)
 - `raise NotImplemented()` gives `TypeError` instead of `NotImplementedError` since `NotImplemented`'s type is `NotImplementedType` (9d799b1)
 - Remove unneeded default `return` (633812b)


schosterbarak:
what is the impact of those changes?

PR #1719, opened 2021-10-06
Updated deprecated method assertEquals

zatch3301:

`assertEquals` has been deprecated. Changed it to standardized `assertEqual` instead.


zatch3301:
Here, `exit()` is used and will fail if the python is run with the -s option.
The exit or quit function don't exist at top-level if python is started with -s flag, and will raise an error, So I changed it to `sys.exit()` instead and this works, regardless of the interpreter options,

gruebel:
good catch with the `assertEquals` and please also change the `exit` call in
https://github.com/bridgecrewio/checkov/blob/f37875ef0f444fa7c26d0fba6ca52ed3c65a37ba/bin/checkov#L8

zatch3301:
If there are no other changes required, can you merge my pr.

PR #1737, opened 2021-10-07
Added .dockerignore

yutachaos:
Change to prevent unnecessary files for build from being read by .dockerignore.


schosterbarak:
does it work without those files?

Issue #1542, opened 2021-08-26
Checkov no longer analyses local modules on windows


JamesWoolfenden:
Clone https://github.com/JamesWoolfenden/terraform-aws-dynamodb

```
checkov -d . 
```

You see multiple check results.

run 
```
checkov -d .\example\examplea
```

Expected: 1st pass results +1 more. 
Actual: just the contents of examplea folder and not the local module referenced

This used to work, works as expect on wsl and mac/linux

v 2.0.367

gruebel:
thanks to @njgibbon this should be fixed since version `2.0.490` 🏅

PR #1751, opened 2021-10-12
Fix - Local Module Loading on Windows

njgibbon:
Hello,

Work to resolve the issue where Checkov regressed to stop loading local modules on Windows.

@JamesWoolfenden 
It's only a one-line change but took a few hours and 2 computers aha. 

## Description
A special case was added to `parser.py` which is the ultimate root cause. To stop double parsing if the module is in a child directory (I think) this stops obtaining the module path using a relative link and replaces with only a full absolute path. On windows, this means that  `source` is set from `../../` to `C:\<path-to-module>`. 
```
# Special handling for local sources to make sure we aren't double-parsing
if source.startswith("./") or source.startswith("../"):
    source = os.path.normpath(os.path.join(os.path.dirname(_remove_module_dependency_in_path(file)), source))
```
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/parser.py#L389

As a consequence the `local_path_loader` doesn't understand the path because it's only configured to understand unix paths and  so `_is_matching_loader` evaluates to `false`.
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/module_loading/loader.py#L61

And ultimately the module is not loaded:
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/parser.py#L399

## Testing
I've done a lot of local functional testing but it is a bit of a challenge to add a _new_ unit test because of the OS dependent nature of the issue. 

This test already verifies that the expected windows path is being used when converted but naturally it's only proven when running in a Windows environment.
https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/module_loading/loaders/test_local_path_loader.py#L12

## Issue
https://github.com/bridgecrewio/checkov/issues/1542

## License

### DIFF
--- a/tests/terraform/module_loading/loaders/test_local_path_loader.py
+++ b/tests/terraform/module_loading/loaders/test_local_path_loader.py
@@ -18,12 +18,12 @@ class TestLocalPathLoader(unittest.TestCase):
     def test_bad_source(self):
         current_dir = os.path.dirname(os.path.realpath(__file__))
         with self.assertRaises(FileNotFoundError):
-            loader.load(current_dir, "./path_that_doesnt_exist", None, '', '')
+            loader.load(current_dir, current_dir, "./path_that_doesnt_exist", None, '', '')
 
     def test_absolute_path(self):
         # Generate absolute path dynamically so test is OS agnostic
         current_dir = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


JamesWoolfenden:
Much appreciated @njgibbon !

njgibbon:
Will do some time today~ :)

PR #1758, opened 2021-10-14
Improve runners performance

YaaraVerner:
Change `renner_registry` to run the runners in different processes using multiprocessing (only for unix os).
Refactor `node.py` to define classes instead creating the classes using functions (this way the classes are pickable). 

### DIFF
--- a/checkov/common/parsers/node.py
+++ b/checkov/common/parsers/node.py
@@ -184,4 +184,4 @@ class ListNode(list):
 
     @staticmethod
     def __name__():
-        return f'{list.__name__}_node' % list.__name__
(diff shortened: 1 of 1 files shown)


gruebel:
Good work with finally creating proper classes for `DictNode`, `ListNode` and `StrNode` 💪  I'm not sure, if it is out of scope here, but it is usually recommended to not subclass the builtin types and rather subclass `collections.UserDict`, `collections.UserList` and `collections.UserString`. For some more background about the reasons, here is a great read https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/

PR #1862, opened 2021-10-27
expanding terraform plan example

jsolis:

The example in the Terraform Plan section didn't have an example showing the use of `jq` which is actually pretty important when scanning plan files. Also, there was a broken example image which I replaced with a code snippet which I felt was better than nothing.

### DIFF
--- a/docs/4.Integrations/Terraform Scanning.md
+++ b/docs/4.Integrations/Terraform Scanning.md
@@ -24,7 +24,6 @@ Note: The Terraform show output file `tf.json` will be a single line. For that r
 If you have installed jq, you can convert a JSON file into multiple lines making it easier to read the scan result.
 
 ```json
-terraform plan --out tfplan.binary
 terraform show -json tfplan.binary | jq '.' > tfplan.json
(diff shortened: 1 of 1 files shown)


docs/4.Integrations/Terraform Scanning.md
```diff
@@ -20,8 +20,29 @@ terraform show -json tfplan.binary > tfplan.json
 checkov -f tfplan.json
 ```
 
+Note: The Terraform show output file `tf.json` will be a single line. For that reason Checkov will report all findings as line number 0.
+If you have installed jq, you can convert a JSON file into multiple lines making it easier to read the scan result.
+
+```json
+terraform plan --out tfplan.binary
```
schosterbarak, line 27: why arer we adding this example? (lines 26 to 30) - it is already in line 16.

jsolis: only because I felt a more complete example makes more sense, but it's not the way the rest of the docs do it, so I'll remove this

PR #1870, opened 2021-10-29
CloudFormation check ALB HTTP to HTTPS redirect

pmalkki:

This same control exists for TerraForm with control ID CKV2_AWS_20. I'm under the impression that prefix CKV2 is reserved for graph checks so chose a new CKV_-prefixed control ID for this. Not sure if that's the correct thing to do.


schosterbarak (approved):

gruebel (approved):
nice, thanks for changing it into a YAML check 💪

PR #1872, opened 2021-10-29
Add Codestar Connections Hosts to SG connection check for CKV2_AWS_5

benglerHIG:
Add AWS Codestar Connections Hosts to the list of resource types checked to confirm if a Security Group is attached and not orphaned
 
### DIFF
--- a/checkov/terraform/checks/graph_checks/aws/SGAttachedToResource.yaml
+++ b/checkov/terraform/checks/graph_checks/aws/SGAttachedToResource.yaml
@@ -16,6 +16,7 @@ definition:
         - aws_batch_compute_environment
         - aws_cloudwatch_event_target
         - aws_codebuild_project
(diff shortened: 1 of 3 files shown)


gruebel:
@benglerHIG nice find, could also add a test case here https://github.com/bridgecrewio/checkov/tree/master/tests/terraform/graph/checks/resources/SGAttachedToResource
you need to add the resources for the test case into the `main.tf` and then add the passing resource name into the `expected.yaml`, if you need more help, just write here 🙂

benglerHIG:
@gruebel oh, my bad, I forgot the test. Thank you for the reminder. I updated the PR with the test. This is my first PR on this project. Please let me know if I missed anything else. Thank you!

gruebel (approved):
thanks looks good!

schosterbarak (approved):

Issue #1777, opened 2021-10-18
`CKV_GCP_69` needs to be updated to be compliant with the new provider changes around `workload_metadata_config` configuration block


calexandre:
**Describe the bug**
The code block `workload_metadata_config` has changed and checkov needs to follow up with an upgrade, so that `CKV_GCP_69` is compliant to the new model, as described here: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#workload_metadata_config

**To Reproduce**

First update the google provider to `3.88.0` (in my case I bumped from `3.82.0 -> 3.88.0`).

The following will throw a deprecation notice on terraform's side:

```hcl
workload_metadata_config {
      node_metadata = "GKE_METADATA_SERVER" # https://docs.bridgecrew.io/docs/ensure-the-gke-metadata-server-is-enabled
    }
```

After updating the code to the new mode, checkov will throw a violation error and needs to be explicitly skipped;

```hcl
workload_metadata_config {
      #checkov:skip=CKV_GCP_69:Ensure the GKE Metadata Server is Enabled - false positive (waiting for checkov update) before was GKE_METADATA_SERVER and now is GKE_METADATA
      mode = "GKE_METADATA"
    }
```

**Expected behavior**
Checkov needs to be updated to be compliant with the new provider.

**Desktop (please complete the following information):**
 - OS: ubuntu (wsl)
 - Checkov Version `2.0.465`

gruebel:
thanks for reporting, yeah this should definitely be updated.

mzwennes:
@calexandre @gruebel Hopefully this will help: https://github.com/bridgecrewio/checkov/pull/1877

gruebel:
thanks to @mzwennes it should be fixed in version `2.0.559` 🚀

PR #1877, opened 2021-11-01
Edit CKV_GCP_69 to support both types of workload_metadata_config

mzwennes:
Fix for https://github.com/bridgecrewio/checkov/issues/1777.

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GKEMetadataServerisEnabled.py
+++ b/checkov/terraform/checks/resource/gcp/GKEMetadataServerisEnabled.py
@@ -1,5 +1,5 @@
 from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
-from checkov.common.models.enums import CheckCategories
+from checkov.common.models.enums import CheckResult, CheckCategories
 
 
 class GKEMetadataServerisEnabled(BaseResourceValueCheck):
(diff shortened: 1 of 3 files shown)


schosterbarak:
> I’ve noticed the failed test, will look at those later this day. In regards to the implementation, would we either go with backwards compatibility, or should the check only take into account the latest provider version?
> 
> if the check should only look at the latest provider we can simplify the implementation by just validating the new key/value

better to be backward compatible

mzwennes:
@schosterbarak Thanks for the reply. I've altered the check in order to fix the tests and kept the part where it's backwards compatible.

checkov/terraform/checks/resource/gcp/GKEMetadataServerisEnabled.py
```diff
@@ -1,20 +1,37 @@
...
+            if isinstance(node, dict) and 'workload_metadata_config' in node:
+                workload_metadata = node["workload_metadata_config"][0]
+              
+                if workload_metadata.get("mode", None) == ["GKE_METADATA"]:
+                    return CheckResult.PASSED
+                # deprecated in newer google provider
+                elif workload_metadata.get("node_metadata", None) == ["GKE_METADATA_SERVER"]:
+                    return CheckResult.PASSED
+                else:
+                    return CheckResult.FAILED
```
gruebel, line 32: this part is actually not needed, because it will be handled two lines lower 🙂

gruebel (commented):
nice job 💪

checkov/terraform/checks/resource/gcp/GKEMetadataServerisEnabled.py
mzwennes: Yeah, you are absolutely right.

gruebel:
@mzwennes changes look good, but there is a problem with `pipenv` looks like they botched their new release 😄 will wait, if they publish a new version till tomorrow, otherwise I will pin the old version.

PR #1878, opened 2021-11-01
Fix cyclonedx ref

kartikp10:

This PR overrides the `output_as_string` method in the `Xml` class so that the `ref` for every vulnerability of a component can be unique.


kartikp10:
@tsmithv11 FYI

schosterbarak:
@kartikp10 see security tests

PR #1881, opened 2021-11-02
Multiprocessing bug fixes

rotemavni:
Fixed a number of different bugs:

1. General multiprocessing - pass on EOF errors on yield when running a multiprocess function.
2. Secrets - added a safe running function to catch errors in scanning without causing run timeouts.
3. K8s - Fixed warning message to use correct param.
4. Terraform - unrelated bugfix in check.

### DIFF
--- a/checkov/common/parallelizer/parallel_runner.py
+++ b/checkov/common/parallelizer/parallel_runner.py
@@ -38,7 +38,10 @@ class ParallelRunner:
 
         for process, parent_conn, group_len in processes:
             for _ in range(group_len):
-                yield parent_conn.recv()
+                try:
+                    yield parent_conn.recv()
+                except EOFError:
+                    pass
 
     def _run_function_multithreaded(self, func: Callable[[Any], Any], items: List[Any]) -> Iterator:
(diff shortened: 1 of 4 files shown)


checkov/common/parallelizer/parallel_runner.py
```diff
@@ -38,7 +38,10 @@ def func_wrapper(original_func, items_group, connection):
 
         for process, parent_conn, group_len in processes:
             for _ in range(group_len):
-                yield parent_conn.recv()
+                try:
+                    yield parent_conn.recv()
+                except EOFError:
+                    pass
```
nimrodkor, line 44: @rotemavni When is this error thrown?

nimrodkor (approved):

checkov/common/parallelizer/parallel_runner.py
rotemavni: @nimrodkor 
It's being thrown by recv():
Return an object sent from the other end of the connection using send(). Blocks until there is something to receive. Raises EOFError if there is nothing left to receive and the other end was closed.
https://docs.python.org/3/library/multiprocessing.html

Issue #1416, opened 2021-07-25
CKV2_GCP_3 seems to be doing the opposite of what it says it is doing


jbratto:
**Describe the bug**
Hello, it appears that this check (Ensure there are only GCP-managed service account keys for each service account) is in fact failing unless you actually create a user managed service account key.  Either I'm misunderstanding something, or this check is setup to do the opposite of what it says it is doing.

**To Reproduce**
To demonstrate this, I used the sample code provided by the check documentation and then ran *gcloud iam service-accounts keys list* to show the details about the service account keys.

**Expected behavior**
A google_service_account_key resource in Terraform creates a user managed key, so I would expect this check to be trigger when one is created, rather than when one is not created.

**Desktop (please complete the following information):**
 This example was run from Google Cloud Shell using the latest checkov container (7/25/2021)

**Additional context**

Using this sample Terraform code:
```
provider "google" {}
provider "local" {}

resource "google_service_account" "account_ok" {
  account_id = "pass-account"
}

resource "google_service_account_key" "ok_key" {
  service_account_id = google_service_account.account_ok.name
}

resource "local_file" "foo" {
    content     = google_service_account_key.ok_key.private_key
    filename    = "${path.module}/key.json"
}
```

Executing checkov gets the following results:
```
docker run --tty --volume /home/xxxxxxx/test:/tf bridgecrew/checkov --directory /tf
terraform scan results:

Passed checks: 1, Failed checks: 0, Skipped checks: 0
```

However, if I check for details on the service account key, I see that there are now actually 2 service account keys. One is managed by the system, and the second one is a user managed key. 
```
xxxxxx@cloudshell:~/test (xxxxxx)$ gcloud iam service-accounts keys list --iam-account=pass-account@xxxxxx.iam.gserviceaccount.com
KEY_ID                                    CREATED_AT            EXPIRES_AT
e6af1e75f8a8801b34440951edd76365414e2147  2021-07-25T14:59:49Z  9999-12-31T23:59:59Z
8c635682666032bd3c215fd47f41db0f1bda745e  2021-07-25T14:59:49Z  2023-07-26T05:13:00Z
xxxxxx@cloudshell:~/test (xxxxxx)$ gcloud iam service-accounts keys list --iam-account=pass-account@xxxxxx.iam.gserviceaccount.com --managed-by=user
KEY_ID                                    CREATED_AT            EXPIRES_AT
[...]

PR #1884, opened 2021-11-02
CKV2_GCP3: Reverse test logic for GCP service account key check

mzwennes:
Fix for https://github.com/bridgecrewio/checkov/issues/1416.

As mentioned in issue above, it's indeed the case that Google does not recommend using service account keys if possible. However looking at the code I am wondering if the checks are looking for [user-managed service account keys](https://cloud.google.com/iam/docs/service-accounts#user-managed_keys). Either way, they should be avoided.

### DIFF
--- a/checkov/terraform/checks/graph_checks/gcp/ServiceAccountHasGCPmanagedKey.yaml
+++ b/checkov/terraform/checks/graph_checks/gcp/ServiceAccountHasGCPmanagedKey.yaml
@@ -10,7 +10,7 @@ definition:
         - google_service_account
       connected_resource_types:
         - google_service_account_key
-      operator:  exists
+      operator:  not_exists
       cond_type: connection
     - cond_type: filter
       attribute: resource_type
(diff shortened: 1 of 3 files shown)


mzwennes:
Docs still need an update.

gruebel:
@mzwennes thanks for the contribution, yeah it feels like the first approach is the correct one, but it shouldn't rely on any connection between both resources. So, basically check that the `public_key_data` field is not filled with anything. @schosterbarak any thoughts?

mzwennes:
> @mzwennes thanks for the contribution, yeah it feels like the first approach is the correct one, but it shouldn't rely on any connection between both resources. So, basically check that the `public_key_data` field is not filled with anything. @schosterbarak any thoughts?

@Gruebel Agreed, it feels like an overkill to completely fail checks on service account keys since people still need those outside of GCP without enforcing Workload Identity Federation on them. 

I've changed the test so that it just looks for the `public_key_data` attribute. I did not find any other examples with just a single check so I hope it's correct.

gruebel (approved):
nice, looks good 👍

nimrodkor (approved):

PR #1889, opened 2021-11-03
CF - Check - SecretManagerSecretEncrypted

njgibbon:
Hello,

I'm back to slowly trying to catch the CF checks up with TF.

Tried to follow latest conventions but just lemme know if need to change anything.

TF Check: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/SecretManagerSecretEncrypted.py
TF Doc: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret
CF Doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html

If you see i'm instead checking if the field matches the pattern `"aws/"` instead of `"alias/aws/"` because I am under the impression that if you miss out the 'alias' part CF / TF / the AWS CLI still infer that it should use aws/* KMS Key.

I could be wrong but I figured this was a safer bet because otherwise `aws/secretsmanager` would be accepted (and I think it would work!).

No CMK can use `aws/` in a user defined alias: https://docs.aws.amazon.com/kms/latest/developerguide/alias-manage.html#alias-create

**License**

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/SecretManagerSecretEncrypted.py
+++ b/checkov/cloudformation/checks/resource/aws/SecretManagerSecretEncrypted.py
@@ -16,8 +16,7 @@ class SecretManagerSecretEncrypted(BaseResourceCheck):
         properties = conf.get("Properties")
         if properties:
             kms_key_id = properties.get("KmsKeyId")
-            if kms_key_id:
-                if aws_kms_alias not in kms_key_id:
+            if kms_key_id and aws_kms_alias not in kms_key_id:
                     return CheckResult.PASSED
         return CheckResult.FAILED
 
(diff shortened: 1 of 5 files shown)


JamesWoolfenden:
keep up the good work @njgibbon (although i'm just adding more checks for Tf now so....)

checkov/cloudformation/checks/resource/aws/SecretManagerSecretEncrypted.py
```diff
@@ -0,0 +1,25 @@
...
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf: DictNode) -> CheckResult:
+        aws_kms_alias = "aws/"
+        properties = conf.get("Properties")
+        if properties:
+            kms_key_id = properties.get("KmsKeyId")
+            if kms_key_id:
+                if aws_kms_alias not in kms_key_id:
```
gruebel, line 20:
```suggestion
            if kms_key_id and aws_kms_alias not in kms_key_id:
```

gruebel (commented):
just one minor thing, but nice job 👍

checkov/cloudformation/checks/resource/aws/SecretManagerSecretEncrypted.py
njgibbon: Will update in a mo'.

gruebel (approved):
🍻

PR #1892, opened 2021-11-03
Handle None limits for k8s checks

rotemavni:
Supported a case where `conf["resources"]["limits"]` is None to avoid failures for `CKV_K8S_11` and `CKV_K8S_13`.


Saarett (approved):
Nice catch! LGTM

PR #1893, opened 2021-11-03
Check/waf has rules

JamesWoolfenden:
You can create an AWS WAF without any rules, so it's worth SFA. This stops that.

### DIFF
--- a/checkov/terraform/checks/resource/aws/WAFHasAnyRules.py
+++ b/checkov/terraform/checks/resource/aws/WAFHasAnyRules.py
@@ -11,13 +11,11 @@ class WAFHasAnyRules(BaseResourceCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf):
-        if "rules" not in conf.keys() and "rule" not in conf.keys():
-            return CheckResult.FAILED
-        if "rules" in conf.keys() and conf["rules"] == [{}]:
-            return CheckResult.FAILED
-        if "rule" in conf.keys() and conf["rule"] == [{}]:
-            return CheckResult.FAILED
-        return CheckResult.PASSED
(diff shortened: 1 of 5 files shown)


checkov/terraform/checks/resource/aws/WAFHasAnyRules.py
```diff
@@ -0,0 +1,23 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class WAFHasAnyRules(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure WAF has any rules"
```
mikeurbanski1, line 7:
```suggestion
        name = "Ensure WAF has associated rules"
```

checkov/terraform/checks/resource/aws/WAFHasAnyRules.py
```diff
@@ -0,0 +1,23 @@
...
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if "rules" not in conf.keys() and "rule" not in conf.keys():
+            return CheckResult.FAILED
+        if "rules" in conf.keys() and conf["rules"] == [{}]:
+            return CheckResult.FAILED
+        if "rule" in conf.keys() and conf["rule"] == [{}]:
+            return CheckResult.FAILED
+        return CheckResult.PASSED
```
gruebel, line 20:
```suggestion
        if "rules" in conf.keys() and conf["rules"] != [{}]:
            return CheckResult.PASSED
        if "rule" in conf.keys() and conf["rule"] != [{}]:
            return CheckResult.PASSED
        return CheckResult.FAILED
```
how about this way?

gruebel (approved):
add a possible adjustment, but nothing critical

PR #1894, opened 2021-11-03
Write sarif to results.sarif

kartikp10:

This PR changes the behavior of the sarif output flag as follows:

1. Running `checkov -d xxx -o sarif` will print the CLI output to the console. 
2. A file `results.sarif` will be created in the directory where checkov is called. 
3. Default error level for every rule in the SARIF output is `error`. However, for results, level will depend on the outcome of the check. `error` for `CheckResult.FAILED` and `warning` for `CheckResult.SKIPPED`

Note that the SARIF output will not have `PASSED` checks but the console output will.  

Fixes: https://github.com/bridgecrewio/bridgecrew-action/issues/24

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -327,8 +327,13 @@ class Report:
         return sarif_template_report
 
     def write_sarif_output(self) -> None:
-        with open("results.sarif", "w") as f:
-            f.write(json.dumps(self.get_sarif_json()))
+        try:
+            with open("results.sarif", "w") as f:
+                f.write(json.dumps(self.get_sarif_json()))
+                print("\nWrote output in SARIF format to the file 'results.sarif'")
+        except EnvironmentError as e:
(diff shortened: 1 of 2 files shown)


mikeurbanski1 (approved):

fproulx-boostsecurity:
Can I understand why this breaking change of behavior. 
I depend on the fact that I can get SARIF as standard out.
I do not want a file written on the file system.... 
Can you add another output flag for this maybe ?
Thanks,

fabiopires:
This PR seems to have introduced an issue in checkov `>=v2.0.538`.

The output sent to the console is no longer in a SARIF format, meaning that we can no longer redirect the output to a file. This wouldn't be a problem if `results.sarif` was still being created (as mentioned by the output message). However, after running checkov and seeing the message `Wrote output in SARIF format to the file 'results.sarif'`  I can't find any `results.sarif` in the current directory.

I'm I missing something?

Thanks

PR #1907, opened 2021-11-05
Add checks for NGINX Ingress CVE-2021-25742

metahertz:
This PR Contains two new Kubernetes checks based on detecting CVE-2021-25742.
(More details on the CVE Here: https://github.com/kubernetes/ingress-nginx/issues/7837).

The first check fails if there are annotations containing configuration snippets **AND** it looks like those snippets contain lua-related directives or access the kubernetes.io secrets/config paths.

The second check is wider, and will fail if there are any annotations related to nginx configuration snippets.

This gives users a choice, as conversations in https://github.com/kubernetes/ingress-nginx/issues/7837 suggest that fully blocking snippets in annotations works well, but for others they need snippets and just want to block lua script directives.

A user could chose which they need by either `--skip-checks CKV_K8S_152` or `--skip-checks CKV_K8S_153` globally on the `checkov` command line, or use the `SKIP_CHECK` annotations for checkov directly in their k8s manifests or helm/kustomize charts.

@eurogig Would love some feedback/other examples for the tests directory if you have anything more creative ;)

### DIFF
--- a/checkov/kubernetes/checks/NginxIngressCVE202125742AllSnippets.py
+++ b/checkov/kubernetes/checks/NginxIngressCVE202125742AllSnippets.py
@@ -23,16 +23,11 @@ class NginxIngressCVE202125742AllSnippets(BaseK8Check):
     def scan_spec_conf(self, conf):
 
         if conf["metadata"]:
-            if conf["metadata"]["annotations"]:
-                if conf["metadata"].get('annotations'):
-                    for annotation in force_list(conf["metadata"]["annotations"]):
-                        for key in annotation:
-                            if "snippet" in key:
-                                return CheckResult.FAILED
-                            else:
-                                return CheckResult.PASSED
-                else:
-                    CheckResult.PASSED
+            if conf["metadata"].get('annotations'):
+                for annotation in force_list(conf["metadata"]["annotations"]):
+                    for key in annotation:
+                        if "snippet" in key:
+                            return CheckResult.FAILED
         return CheckResult.PASSED
 
 check = NginxIngressCVE202125742AllSnippets()
--- a/checkov/kubernetes/checks/NginxIngressCVE202125742AllSnippets.py
+++ b/checkov/kubernetes/checks/NginxIngressCVE202125742AllSnippets.py
@@ -1,0 +1,38 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.common.util.data_structures_utils import find_in_dict
+from checkov.kubernetes.base_spec_check import BaseK8Check
+from checkov.common.util.type_forcers import force_list
+import re
+
+
+class NginxIngressCVE202125742AllSnippets(BaseK8Check):
+
+    def __init__(self):
+        name = "Prevent All NGINX Ingress annotation snippets. See CVE-2021-25742"
+        id = "CKV_K8S_153"
+        supported_kind = ['Ingress']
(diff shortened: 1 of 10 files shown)


checkov/kubernetes/checks/NginxIngressCVE202125742AllSnippets.py
```diff
@@ -0,0 +1,38 @@
...
+                if conf["metadata"].get('annotations'):
+                    for annotation in force_list(conf["metadata"]["annotations"]):
+                        for key in annotation:
+                            if "snippet" in key:
+                                return CheckResult.FAILED
+                            else:
+                                return CheckResult.PASSED
+                else:
+                    CheckResult.PASSED
+        return CheckResult.PASSED
```
gruebel, line 36:
```suggestion
        if conf["metadata"]:
            if conf["metadata"].get('annotations'):
                for annotation in force_list(conf["metadata"]["annotations"]):
                    for key in annotation:
                        if "snippet" in key:
                            return CheckResult.FAILED
        return CheckResult.PASSED
```
There are couple of things
- `conf["metadata"]["annotations"]` is doubled, once is more than enough 😄 
- the last else has no return, and actually is not needed
- the current check, only checks if the first annotation has `snippet` in the name. you can test it by changing the order of the annotations in the passed check
This is how it should look like from my point of view 🙂

checkov/kubernetes/checks/NginxIngressCVE202125742Lua.py
```diff
@@ -0,0 +1,40 @@
...
+                    for annotation in force_list(conf["metadata"]["annotations"]):
+                        for key in annotation:
+                            if "snippet" in key:
+                                if re.match(badInjectionPatterns ,annotation[key]):
+                                    return CheckResult.FAILED
+                            else:
+                                return CheckResult.PASSED
+                else:
+                    CheckResult.PASSED
+        return CheckResult.PASSED
```
gruebel, line 38:
```suggestion
        if conf["metadata"]:
            if conf["metadata"].get('annotations'):
                for annotation in force_list(conf["metadata"]["annotations"]):
                    for key, value in annotation.items():
                        if "snippet" in key and  re.match(badInjectionPatterns, value):
                            return CheckResult.FAILED
        return CheckResult.PASSED
```
same points as before 🙂

gruebel (changes requested):

metahertz:
Added a third check on the following logic after walking through @eurogig's example malicious ingress: 
- The "All snippets" check may be commonly skipped as snippets are a common usecase.
- The LUA check does not prevent abuse of filesystem access using alias configuration.
- The third Alias check is to address this non-lua exploit path

This check will prevent against the following style of abuse:

```
kind: Ingress
metadata:
  name: example-ingress
  namespace: developer
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/rewrite-target: /$2
    nginx.ingress.kubernetes.io/server-snippet: |
      location ^~ "/test" {
        default_type 'text/plain';
        alias /var/run;
      }
spec:
  rules:
  - http:
      paths:
        - path: /test
          pathType: Prefix
          backend:
            service:
              name: web
              port:
                number: 8080
```

eurogig:
Awesome!  This is looking pretty solid!

schosterbarak (approved):

PR #1908, opened 2021-11-05
Add custom policy guidelines to integration

mikeurbanski1:

Inserts guidelines from the platform when pulling custom checks.

There is supposed to be an API call that returns guidelines for custom checks, but it is not doing so. This works around that more reliably.


kartikp10 (approved):
Looks good! Pretty sure UTs won't break

schosterbarak (changes requested):
@mikeurbanski1 please add UT under integration tests directory . We have ones that are running with API key

integration_tests/test_checkov_cli_integration_report.py
```diff
@@ -8,6 +9,15 @@
 
 class TestCheckovJsonReport(unittest.TestCase):
 
+    def test_guideline_in_custom_policy(self):
```
nimrodkor, line 12: @mikeurbanski1 This test should run only when we have the API key, which is embodied in this:
```python
if sys.version_info[1] == 7 and platform.system() == 'Linux':
```

nimrodkor:
@mikeurbanski1 is this still needed?

PR #1912, opened 2021-11-07
Update Dockerfile

schosterbarak:
update k8 docker image to slim - takes less space and fewer vulnerabilities

### DIFF
--- a/kubernetes/Dockerfile
+++ b/kubernetes/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM python:3.9-slim
 
 COPY kubernetes/requirements.txt /requirements.txt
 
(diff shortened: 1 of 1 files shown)


gruebel (approved):
🥇

gruebel:
I tested it and it looks like the `slim` images doesn't have curl installed anymore, so you also need to add it on line 7 after `git`
```Dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends git curl && \
```

gruebel (changes requested):
see comment

schosterbarak:
> see comment

@gruebel PTAL :)

gruebel (approved):
👍

PR #1919, opened 2021-11-09
Set max length of 10,000 characters for terraform statement evaluation

nimrodkor:

The rendering can cause very, very long strings to get created. Most of them are not useful so it's a waste of time to try and evaluate them - this sets a configurable limit (with a friendly log message)

### DIFF
--- a/checkov/terraform/graph_builder/variable_rendering/renderer.py
+++ b/checkov/terraform/graph_builder/variable_rendering/renderer.py
@@ -1,3 +1,5 @@
+import logging
+import os
 import re
 from collections.abc import Hashable
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -279,6 +283,12 @@ def evaluate_non_rendered_values(self) -> None:
             self.local_graph.update_vertex_config(vertex, changed_attributes)
 
     def evaluate_value(self, val: Any) -> Any:
+        val_length: int = len(str(val))
+        if CHECKOV_RENDER_MAX_LEN and 0 < CHECKOV_RENDER_MAX_LEN < val_length:
```
gruebel, line 287: either
```python
if 0 < CHECKOV_RENDER_MAX_LEN < val_length:
```
or 
```python
if CHECKOV_RENDER_MAX_LEN and CHECKOV_RENDER_MAX_LEN < val_length
```
and then `0` is the value to render any length

gruebel (changes requested):

checkov/terraform/graph_builder/variable_rendering/renderer.py
nimrodkor: First option is no good - CHECKOV_RENDER_MAX_LEN can be `None`. I'll use the second

gruebel: ah ok, then how about
```python
if CHECKOV_RENDER_MAX_LEN is not None and 0 < CHECKOV_RENDER_MAX_LEN < val_length:
```

gruebel (approved):
🍻

Issue #727, opened 2020-12-09
False positive on CKV_GCP_14 - Requires backup configuration enabled on read replica


pritho:
**Describe the bug**
CKV_GCP_14 requires a backup configuration, but it does not take into consideration read replicas.
A read replica cannot have backup enabled in GCP.

**To Reproduce**
1. Create a google_sql_database_instance with master_instance_name and replica_configuration
2. Chekov will pop up error CKV_GCP_14

**Expected behavior**
Read replicas should not be considered needing a backup in GCP CloudSQL

**Desktop (please complete the following information):**
 - Checkov Version 1.0.660

mzwennes:
This issue can be closed.

gruebel:
Thanks to @mzwennes this now fixed in version `2.0.566` 🎉

PR #1920, opened 2021-11-09
CKV_GCP_14: Add check that ignores Cloud SQL replica instances

mzwennes:
Solves the false positive mentioned in https://github.com/bridgecrewio/checkov/issues/727.

The `master_instance_name` attribute is used to [point the replica to a master instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#master_instance_name). So the existence of this attribute means the current resource is a replica.

This PR adds a little check that checks if the attribute exists and passes it if it does.

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GoogleCloudSqlBackupConfiguration.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleCloudSqlBackupConfiguration.py
@@ -1,28 +1,21 @@
 from checkov.common.models.enums import CheckCategories, CheckResult
-from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
-class GoogleCloudSqlBackupConfiguration(BaseResourceCheck):
+class GoogleCloudSqlBackupConfiguration(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure all Cloud SQL database instance have backup configuration enabled"
         id = "CKV_GCP_14"
         supported_resources = ['google_sql_database_instance']
-        categories = [CheckCategories.NETWORKING]
+        categories = [CheckCategories.BACKUP_AND_RECOVERY]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf):
         if 'master_instance_name' in conf:
-           return CheckResult.PASSED
-        
-        if 'settings' in conf:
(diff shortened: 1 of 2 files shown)


checkov/terraform/checks/resource/gcp/GoogleCloudSqlBackupConfiguration.py
```diff
@@ -1,17 +1,28 @@
...
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceCheck
 
 
-class GoogleCloudSqlBackupConfiguration(BaseResourceValueCheck):
+class GoogleCloudSqlBackupConfiguration(BaseResourceCheck):
     def __init__(self):
         name = "Ensure all Cloud SQL database instance have backup configuration enabled"
         id = "CKV_GCP_14"
         supported_resources = ['google_sql_database_instance']
         categories = [CheckCategories.NETWORKING]
```
gruebel, line 10:
```suggestion
        categories = [CheckCategories.BACKUP_AND_RECOVERY]
```
could you change the category, this fits better 🙂

checkov/terraform/checks/resource/gcp/GoogleCloudSqlBackupConfiguration.py
```diff
@@ -1,17 +1,28 @@
...
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
-    def get_inspected_key(self):
-        return 'settings/[0]/backup_configuration/[0]/enabled'
+    def scan_resource_conf(self, conf):
+        if 'master_instance_name' in conf:
+           return CheckResult.PASSED
```
gruebel, line 15: you can just leave the subclass as it is and adjust the check like this
```python
        if 'master_instance_name' in conf:
            return CheckResult.UNKNOWN

    def get_inspected_key(self):
        return "settings/[0]/backup_configuration/[0]/enabled"
```
similar like here
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/ALBDropHttpHeaders.py

mzwennes: Way cleaner! Thanks for the sugggestion.

checkov/terraform/checks/resource/gcp/GoogleCloudSqlBackupConfiguration.py
mzwennes: 👍

schosterbarak:
Thanks @mzwennes!

PR #1921, opened 2021-11-09
Auth error handling

mikeurbanski1:

Adds some better error handling and messaging around auth errors, which are causing users a lot of trouble in the beta.

Also trims a trailing slash off the Prisma API URL, which causes 404s.


nimrodkor (approved):
Nice!

PR #1923, opened 2021-11-10
Performance tests

tronxd:
- Use `pytest-benchmark` plugin to evaluate execution times of different IaC runners on known benchmarks.

- Performance tests will fail if they deviate by a configurable percecnt from the maximum duration of the tested branch against the data. 
- Thresholds are set according to current execution on the data.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -98,13 +98,11 @@ jobs:
       - uses: actions/setup-node@v2
       - uses: azure/setup-helm@v1
       - uses: dschep/install-pipenv-action@v1
-      - name: Build & install checkov package
+      - name: Install dependencies
         run: |
-          pipenv --python ${{ matrix.python }}
+          pipenv --python 3.7
           pipenv run pip install --upgrade pip==21.1.1
           pipenv run pip install pytest pytest-benchmark
-          pipenv run python setup.py sdist bdist_wheel
-          bash -c 'pipenv run pip install dist/checkov-*.whl'
       - name: Clone terraform-aws-components
         run: git clone --branch 0.182.0 https://github.com/cloudposse/terraform-aws-components.git
         working-directory: ${{ env.working-directory }}
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -91,10 +91,9 @@ jobs:
       fail-fast: true
       matrix:
         python: ["3.7", "3.8", "3.9", "3.10"]
-        os: [ubuntu-latest, macos-latest, windows-latest]
     env:
       working-directory: ./performance_tests
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-test.yml
```diff
@@ -85,3 +85,43 @@ jobs:
       - name: Run integration tests
         run: |
           pipenv run pytest integration_tests -k 'not api_key'
+
+  performance-tests:
+    strategy:
+      fail-fast: true
+      matrix:
+        python: ["3.7", "3.8", "3.9", "3.10"]
```
nimrodkor, line 93: @tronxd Won't the runtime be a little different in each python version? What does running it on all python runtimes give us?

I'd expect we would run only on 1 runtime, doesn't matter which one IMO

.github/workflows/pr-test.yml
```diff
@@ -85,3 +85,43 @@ jobs:
...
+        with:
+          # version 2021.11.5, 2021.11.5.post0 do not work
+          version: 2021.5.29
+      - name: Build & install checkov package
+        run: |
+          pipenv --python ${{ matrix.python }}
+          pipenv run pip install --upgrade pip==21.1.1
+          pipenv run pip install pytest pytest-benchmark
+          pipenv run python setup.py sdist bdist_wheel
+          bash -c 'pipenv run pip install dist/checkov-*.whl'
```
nimrodkor, line 114: @tronxd I checked the tests out - we're not using the packaged checkov. So why do we need to build & install checkov?

.github/workflows/pr-test.yml
```diff
@@ -85,3 +85,43 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+      - uses: actions/setup-node@v2
+      - uses: azure/setup-helm@v1
+      - uses: dschep/install-pipenv-action@v1
+        with:
+          # version 2021.11.5, 2021.11.5.post0 do not work
```
nimrodkor, line 106: @tronxd Why does it not work? How old is the version we're using?

nimrodkor (commented):
@tronxd Nice!!

.github/workflows/pr-test.yml
gruebel: there was an issue with both of the version, I pinned it on the weekend. They published a new one, probably worth a try. Remove `with` block -> run it -> works then great, if not put it back

.github/workflows/pr-test.yml
tronxd: @nimrodkor The tests measure different checkov runners duration. so checkov is needed

.github/workflows/pr-test.yml
tronxd: @gruebel Worked, thanks!

.github/workflows/pr-test.yml
nimrodkor: Checkov the code exists - why do we need the packaged version?
Are the tests bashing out and running `checkov ...`? I didn't see that

.github/workflows/pr-test.yml
```diff
@@ -85,3 +85,40 @@ jobs:
...
+      matrix:
+        python: ["3.7"]
+    env:
+      working-directory: ./performance_tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
```
nimrodkor, line 101:
```suggestion
  performance-tests:
    env:
      working-directory: ./performance_tests
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v2
        with:
          python-version: 3.7
```

PR #1930, opened 2021-11-12
add package file scanning

gruebel:

- added a couple of new packages to leverage asyncio + multiprocessing, also also for testing them
- created a `twistcli` base class
- image scanning is still using the normal process, but it could be changed to also use asyncio
- there was no test coverage in that part, so I tried to add as much as possible, but still more is possible


checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
```diff
@@ -0,0 +1,53 @@
...
+from checkov.common.bridgecrew.vulnerability_scanning.integrations.twistcli import TwistcliIntegration
+from checkov.common.util.str_utils import removeprefix
+
+
+class PackageScanningIntegration(TwistcliIntegration):
+    # not used for package scanning
+    def _create_report(self, **kwargs: Any) -> Dict[str, Any]:
+        pass
+
+    def _create_report_async(
```
nimrodkor, line 16: @gruebel Can you explain why we need these 2 methods, and they can't be 1?

tests/common/bridgecrew/vulnerability_scanning/test_package_scanner.py
```diff
@@ -0,0 +1,237 @@
...
+from mock import AsyncMock, MagicMock
+from pytest_mock import MockerFixture
+
+from checkov.common.bridgecrew.bc_source import SourceType
+from checkov.common.bridgecrew.platform_integration import BcPlatformIntegration, bc_integration
+from checkov.common.bridgecrew.vulnerability_scanning.package_scanner import PackageScanner
+
+
+@pytest.fixture()
+def mock_bc_integration() -> BcPlatformIntegration:
```
nimrodkor, line 15: @gruebel This seems like a duplication of the `test_package_scanning` file. Why do we need both?

Also - since you rewrote the image scanning - can we test the parts which are not shared with the package scanning as well?

nimrodkor (commented):
Very nice @gruebel !
Just a few comments

tests/common/bridgecrew/vulnerability_scanning/test_package_scanner.py
gruebel: good point moved them now to a `conftest` file

checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
gruebel: merged them together, I did it to make it more clear, that those private methods are used in a bit different context 🙂

nimrodkor (approved):
Nice!

PR #1934, opened 2021-11-12
Fix custom guidelines

mikeurbanski1:

Fixes an issue where GraphRecords overwrite guidelines that get set on the record.


kartikp10 (approved):
Good catch 💡

PR #1935, opened 2021-11-14
add CKV2_AWS_32 for CloudFront security header

gruebel:

I used AWS example as a great base for creating the custom CloudFront response header policy, but was a bit more strict regards to the `override` 👮 

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/creating-response-headers-policies.html


nimrodkor (approved):
👏

nimrodkor:
Let's see if we get any feedback on it, it is easy to skip / suppress... @gruebel

PR #1951, opened 2021-11-17
changed re.match to re.search

eurogig:
Changed the regex check for the lua check from re.match to re.search as the snippets are multiline.  It worked for the test because it's a single line snippet.  I'll change the test shortly as well.  


gruebel (approved):
🏅

PR #1983, opened 2021-11-17
Fix tf plan entity lookup

mikeurbanski1:

Works around a larger issue where TF plan methods get called for TF files and crash on Windows


kartikp10 (approved):
Makes sense ✅

PR #1991, opened 2021-11-18
Updated HELM documentation

metahertz:
Added information about scanning both charts automatically through auto-detection of `Chart.yaml` files and also howto scan resultant kubernetes manifests using your own helm values files.

### DIFF
--- a/docs/4.Integrations/Helm.md
+++ b/docs/4.Integrations/Helm.md
@@ -1,0 +1,65 @@
+---
+layout: default
+published: true
+title: Helm
+nav_order: 9
+---
+
+# Scan Helm charts with Checkov
+
+Checkov is able to autodetect helm charts by the presence of a `Chart.yaml` file, if found, the helm framework will automatically be used to template out the helm chart (with it's default values) into resulting Kubernetes manifests, which will then be scanned by all Checkovs' Kubernetes policies.
+
+
+```
+       _               _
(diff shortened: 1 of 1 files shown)


docs/4.Integrations/Helm.md
```diff
@@ -0,0 +1,65 @@
...
+
+In this case, you can use your custom values from your values file (`gocd.yml` in this case) to generate kubernetes output with Helm, and pass the templated output to checkov to receive the same policy information. This would also be trivial to set up in CI for automated scanning of this scenario:
+
+```
+helm template gocd/gocd -f gocd.yaml > k8s-template.yaml
+checkov -f k8s-template.yaml --framework kubernetes --skip-check CKV_K8S_21
+```
+
+Note we skip check `CKV_K8S_21` for this process, which alerts on default namespace usage within Kubernetes manifests. 
+Since HELM manages our namespaces, we always skip this internally when using the HELM framework, so we want to replicate the same behaviour here.
```
gruebel, line 64: most of the time `helm` is written with lower caps except here, whatever you choose, just be consistent 😉

gruebel (approved):

docs/4.Integrations/Helm.md
metahertz: Fair, changed! thanks

PR #2010, opened 2021-11-23
fix in IAMPasswordLength check

YaaraVerner:
check if minimum_password_length attribute is int before compere it

### DIFF
--- a/checkov/terraform/checks/resource/oci/IAMPasswordLength.py
+++ b/checkov/terraform/checks/resource/oci/IAMPasswordLength.py
@@ -16,7 +16,7 @@ class IAMPasswordLength(BaseResourceCheck):
             rules = conf.get("password_policy")[0]
             if 'minimum_password_length' in rules:
                 passwordlength = rules.get("minimum_password_length")
-                if passwordlength[0] < 14:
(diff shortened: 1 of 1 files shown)


checkov/terraform/checks/resource/oci/IAMPasswordLength.py
```diff
@@ -16,7 +16,7 @@ def scan_resource_conf(self, conf):
             rules = conf.get("password_policy")[0]
             if 'minimum_password_length' in rules:
                 passwordlength = rules.get("minimum_password_length")
-                if passwordlength[0] < 14:
+                if isinstance(passwordlength[0], int) and passwordlength[0] < 14:
```
nimrodkor, line 19: @YaaraVerner Please use `force_int` method 🙏

YaaraVerner: the use-case that an exception was thrown is when the attribute is assign with variable, force_int won't work for this

nimrodkor: Resolving and writing it again 😓

nimrodkor (approved):

PR #2011, opened 2021-11-23
Build k8s graph

YaaraVerner:
Implementing Kubernetes `local_graph`, `graph_manager` and `block`.
Extract getting Kubernetes definitions to `kubernetes_utils`.
Add usage of the graph in the runner (without using it yet).


checkov/kubernetes/runner.py
```diff
@@ -4,49 +4,58 @@
...
-                            files_list.append(full_path)
-
-            _parse_files(files_list, definitions, definitions_raw, filepath_fn)
+            definitions, self.definitions_raw = get_files_definitions(files)
+        elif root_folder:
+            definitions, self.definitions_raw = get_folder_definitions(root_folder, runner_filter.excluded_paths)
+        else:
+            return report
+
+        # TODO: uncomment it in order to build the graph
```
Saarett, line 53: Why is this commented out?

checkov/kubernetes/kebernetes_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    for root, d_names, f_names in os.walk(root_folder):
+        filter_ignored_paths(root, d_names, excluded_paths)
+        filter_ignored_paths(root, f_names, excluded_paths)
+
+        for file in f_names:
+            file_ending = os.path.splitext(file)[1]
+            if file_ending in K8_POSSIBLE_ENDINGS:
+                full_path = os.path.join(root, file)
+                if "/." not in full_path and file not in ['package.json', 'package-lock.json']:
+                    # skip temp directories
```
Saarett, line 27: What do you mean by temp dirs?

checkov/kubernetes/runner.py
```diff
@@ -4,49 +4,58 @@
...
+    def __init__(
+        self,
+        graph_class=KubernetesLocalGraph,
+        db_connector=NetworkxConnector(),
+        source="Kubernetes",
+        graph_manager=None,
+    ):
+        self.check_type = "kubernetes"
+        self.graph_class = graph_class
+        # TODO: uncomment it in order to create the graph
```
Saarett, line 32: Why is this commented out?

checkov/cloudformation/cfn_utils.py
```diff
@@ -103,12 +103,7 @@ def get_folder_definitions(
             if file_ending in CF_POSSIBLE_ENDINGS:
                 files_list.append(os.path.join(root, file))
 
-    definitions, definitions_raw = get_files_definitions(
-        files_list, out_parsing_errors, lambda f: f'/{os.path.relpath(f, os.path.commonprefix((root_folder, f)))}')
-
-    definitions = {create_file_abs_path(root_folder, file_path): v for (file_path, v) in definitions.items()}
-    definitions_raw = {create_file_abs_path(root_folder, file_path): v for (file_path, v) in definitions_raw.items()}
```
Saarett, line 110: why are those 2 lines removed?

tests/kubernetes/graph/test_local_graph.py
```diff
@@ -0,0 +1,30 @@
+import os
+
+from checkov.kubernetes.graph_builder.local_graph import KubernetesLocalGraph
+from checkov.kubernetes.parser.parser import parse
+from tests.kubernetes.graph.base_graph_tests import TestGraph
+
+TEST_DIRNAME = os.path.dirname(os.path.realpath(__file__))
+
+
+class TestKubernetesLocalGraph(TestGraph):
```
Saarett, line 10: Can/Should we add more tests for failing k8s files?

Saarett (commented):
Great job! Comments in the code :)

checkov/kubernetes/runner.py
YaaraVerner: I'll do it in the next task, and already implemented some staff so I don't want to delete

checkov/kubernetes/kebernetes_utils.py
YaaraVerner: actually I didn't write this code, just moved it to a different file

checkov/kubernetes/runner.py
YaaraVerner: I'll do it in the next task, and already implemented some staff so I don't want to delete

checkov/cloudformation/cfn_utils.py
YaaraVerner: I saw we change the file path when calling get_files_definitions function and than change it back, so I removed the two changes.

tests/kubernetes/graph/test_local_graph.py
YaaraVerner: I think we should add tests for failing k8s files for the runner (when implemented), this part isn't effected by this because we are not running any checks here.

Saarett (approved):
Approved by me, but I think we should get another opinion here (especially @gruebel )

checkov/kubernetes/runner.py
```diff
@@ -191,6 +199,41 @@ def run(self, root_folder, external_checks_dir=None, files=None, runner_filter=R
...
+        report = Report(self.check_type)
+        checks_results = self.run_graph_checks_results(runner_filter)
+
+        for check, check_results in checks_results.items():
+            for check_result in check_results:
+                entity = check_result["entity"]
+                entity_file_abs_path = entity.get(CustomAttributes.FILE_PATH)
+                entity_file_path = f"/{os.path.relpath(entity_file_abs_path, root_folder)}"
+                start_line = entity.get('__startline__')
+                end_line = entity.get('__endline__')
```
gruebel, line 212: is there a case when either one them is not set? Because it would break the lower part of the code

tests/kubernetes/graph/base_graph_tests.py
```diff
@@ -0,0 +1,40 @@
...
+
+from checkov.common.graph.graph_builder.graph_components.block_types import BlockType
+from checkov.common.graph.graph_builder.graph_components.blocks import get_inner_attributes
+
+
+class TestGraph(TestCase):
+    def assert_vertex(self, resource_vertex, resource):
+        resource_name = f'{resource["kind"]}.{resource["metadata"].get("namespace", "default")}.{resource["metadata"]["name"]}'
+        config = resource["spec"]
+        attributes = config.copy()
```
gruebel, line 11: also here `deepcopy`

checkov/kubernetes/runner.py
YaaraVerner: there shouldn't be a case that one of them not set, you can see another use of them in the `run` method

gruebel: ok, then just access them directly, `entity['__startline__']` so if they shouldn't exist for some reason, then it will fail here and not a few lines lower 🙂

gruebel (approved):
🏅

PR #2013, opened 2021-11-23
add AWS SAM support

gruebel:

- adjusted Lambda checks to support Serverless resource types
- adjusted API Gateway checks to support Serverless resource types
- SAM `Globals` have specific rules on property overrides https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html#sam-specification-template-anatomy-globals-overrideable
- actually not all properties can be overridden with `Globals`, but I didn't want to maintain the list, which is growing with new features added to SAM


checkov/cloudformation/checks/resource/aws/LambdaDLQConfigured.py
```diff
@@ -1,21 +1,30 @@
...
+        properties = conf.get("Properties")
+        if properties:
+            self.evaluated_keys = ["Properties/DeadLetterConfig/TargetArn"]
+            dlc = properties.get("DeadLetterConfig")
+            if dlc and dlc.get("TargetArn"):
+                return CheckResult.PASSED
 
-    def get_expected_value(self) -> str:
-        return ANY_VALUE
+            self.evaluated_keys.append("Properties/DeadLetterQueue/TargetArn")
```
nimrodkor, line 22: @gruebel Why append and not just set the evaluated keys to this?

checkov/common/graph/graph_builder/graph_components/blocks.py
```diff
@@ -44,7 +44,10 @@ def _extract_inner_attributes(self) -> Dict[str, Any]:
             if isinstance(attribute_value, dict) or (
                     isinstance(attribute_value, list) and len(attribute_value) > 0 and isinstance(attribute_value[0],
                                                                                                   dict)):
-                inner_attributes = get_inner_attributes(attribute_key, attribute_value)
+                inner_attributes = self.get_inner_attributes(
+                    attribute_key=attribute_key,
+                    attribute_value=attribute_value,
+                )
```
nimrodkor, line 50: @gruebel Unnecessary change?

nimrodkor (approved):
Very, very nice. 👏  @gruebel

checkov/common/graph/graph_builder/graph_components/blocks.py
gruebel: before `get_inner_attributes` was a function inside the file, now it is a `@classmethod` that's why there is now a `self.` at the beginning and it is reformated, because I implemented it differently before and them came up with a better approach 😄

checkov/cloudformation/checks/resource/aws/LambdaDLQConfigured.py
gruebel: in general I was not sure, how to best handle `evaluated_keys` here, because it actually depends on the resource type. Do we want to add it as an attribute when the base class is invoked and passed the resource type?

nimrodkor (approved):

schosterbarak:
very nice. please update README.md too :)

PR #2014, opened 2021-11-24
Fix rendering interpolation of non-rendered data

rotemavni:
If we remove an interpolation from a non-rendered data such as `${var.x}`, wrap it with `'` to avoid creating malformed strings to render.


gruebel (approved):
🚀

PR #2015, opened 2021-11-24
fix in module loader

YaaraVerner:
update module loader to wait for all modules loading when running parallel

### DIFF
--- a/checkov/terraform/module_loading/module_finder.py
+++ b/checkov/terraform/module_loading/module_finder.py
@@ -87,7 +87,7 @@ def load_tf_modules(path: str, should_download_module: Callable[[str], bool] = s
     distinct_modules = list({m.address: m for m in modules_to_load}.values())
 
     if run_parallel:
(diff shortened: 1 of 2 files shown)


rotemavni (approved):

checkov/terraform/module_loading/module_finder.py
```diff
@@ -87,7 +87,7 @@ def _download_module(m):
     distinct_modules = list({m.address: m for m in modules_to_load}.values())
 
     if run_parallel:
-        parallel_runner.run_function(_download_module, distinct_modules)
+        list(parallel_runner.run_function(_download_module, distinct_modules))
```
gruebel, line 90:
```suggestion
        deque(parallel_runner.run_function(_download_module, distinct_modules), maxlen=0)
```
I think `collections.deque` could be used here, it is faster to evaluate a generator, but I don't think we really need it 😄

gruebel (approved):

PR #2021, opened 2021-11-25
add try-catch block for hcl2.loads failure in ProviderContextParser

YaaraVerner:
return False for `_is_block_signature` function in `ProviderContextParser` if the terraform file isn't valid

### DIFF
--- a/checkov/terraform/context_parsers/parsers/provider_context_parser.py
+++ b/checkov/terraform/context_parsers/parsers/provider_context_parser.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Dict, Any, List
 
 import hcl2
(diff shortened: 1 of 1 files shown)


checkov/terraform/context_parsers/parsers/provider_context_parser.py
```diff
@@ -24,11 +24,14 @@ def _is_block_signature(self, line_num: int, line_tokens: List[str], entity_cont
...
-                map(lambda obj: obj[1], self.file_lines[line_num - 1 : end_line if end_line > line_num else line_num])
-            )
-        )["provider"][0]
+        try:
+            provider_obj = hcl2.loads(
+                "\n".join(
+                    map(lambda obj: obj[1], self.file_lines[line_num - 1 : end_line if end_line > line_num else line_num])
+                )
+            )["provider"][0]
+        except Exception:
```
gruebel, line 33: I think it would be good to log here something, when it fails, either info or debug level

gruebel (approved):
🚀

PR #2043, opened 2021-12-01
Fix loading of module sources without explicit ssh://

pascal-hofmann:
This also removes a duplicated test function and adds tests for modules
that are loaded via ssh in general.


schosterbarak:
thanks @pascal-hofmann!

PR #2044, opened 2021-12-01
Use full resource address as resource name when parsing terraform plan  (#1116)

### LINKED ISSUE #1116: When scanning a terraform plan, module resources are not fully qualified in the resource name
**Describe the bug**
image

each instance here of `google_storage_bucket.this` is created by a module, but we are unable to see which module is invoking this because the resource doesn't include `module.x.` before it.

**To Reproduce**
1. Scan a plan that contains resources created with modules, and use the github_failed_only output to see this table

**Expected behavior**
Any module created resources should be explicit in the output `module.x.resource.name`

 - OS: OS X
 - Checkov Version: 2.0.67

pascal-hofmann:
This makes sure that there are no colliding resources when a module is
used multiple times. Without this fix, line numbers and affected code
reported was incorrect. This also fixes #1116.

Also fixes some typos in variable name (separate commit, in order to ease reviewing).

### DIFF
--- a/checkov/terraform/plan_parser.py
+++ b/checkov/terraform/plan_parser.py
@@ -123,9 +123,9 @@ def parse_tf_plan(tf_plan_file: str) -> Tuple[Optional[Dict[str, Dict[str, Any]]
     :type tf_plan_file: str - path to plan file
     :rtype: tf_definition dictionary
     """
-    tf_defintions: Dict[str, Dict[str, Any]] = {}
-    tf_defintions[tf_plan_file] = {}
-    tf_defintions[tf_plan_file]["resource"] = []
+    tf_definitions: Dict[str, Dict[str, Any]] = {}
+    tf_definitions[tf_plan_file] = {}
+    tf_definitions[tf_plan_file]["resource"] = []
     template, template_lines = parse(tf_plan_file)
     if not template:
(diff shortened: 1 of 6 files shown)


nimrodkor:
Hey @pascal-hofmann , can you fix the UTs please? 🙏

pascal-hofmann:
> Hey @pascal-hofmann , can you fix the UTs please? 🙏

I'm a bit lost here. The unit tests work for me locally. 😵🤷

pascal-hofmann:
I'll rebase my work on #2048 and will then try to find a good fix for the things I broke.

mikeurbanski1:
@nimrodkor I think this might break stuff in the platform, no?

pascal-hofmann:
I tried to tackle this problem properly, but realized it requires too much understanding of, and work with the internals of checkov. Thus I'm closing this PR. The new test conditions are pretty much useless, but I think the test files (in `tests/terraform/parser/resources/plan_modules`) will be helpful for anyone seeking to build proper `terraform plan` support for checkov.

For now we won't use checkov in the project I'm currently working on (because of this issue and also because there is no way to configure suppressing/skipping without code annotations - a configuration file with resource addresses would be great for this).

Issue #2047, opened 2021-12-01
Module suppressions with plan file enrichment does not work


mikeurbanski1:
**Describe the bug**
In [this](https://github.com/bridgecrewio/checkov/issues/777) issue, there was a request for the ability to suppress all resources in a module using a comment in the `module` block. It was implemented here: https://github.com/bridgecrewio/checkov/pull/1629

However, the capability does not extend to using the `--repo-root-for-plan-enrichment` with a plan file - the enrichment does not apply the logic down to the resource level.

**To Reproduce**

`main.tf`:

```
module "m" {
  # checkov:skip=CKV_AWS_19: ADD REASON
  source = "../mod1"
}
```

`../mod1/main.tf`:

```
resource "aws_s3_bucket" "b1" {
  bucket = "bc-test-bucket-1"
}
```

Commands: 
```
<generate tfplan.json>
checkov -f tfplan.json --repo-root-for-plan-enrichment .
```

**Expected behavior**
CKV_AWS_19 is suppressed

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macOS catalina 10.15.7
 - Checkov Version [e.g. 22] 2.0.619

**Additional context**

PR #2048, opened 2021-12-02
Handle module suppression enrichments

mikeurbanski1:

Fixes https://github.com/bridgecrewio/checkov/issues/2047 by saving the resource address in a TF plan run and then using that address to look up suppressions for the module during the enrichment phase.

I could think of one alternative approach to this that seemed like a little less of a hack, but was also probably overkill. That approach would be to replace this line with logic to repeat the entire parsing and graph building logic and reconstruct the entire module structure from scratch. Then use the logic introduced in that PR I reference in the issue to push suppressions down to the resources in the same way.

However, that seemed like overkill, and avoiding unnecessarily building a graph seems like a good idea due to performance concerns.

I do not think this approach will work with nested modules, but I don't think the original fix does either, so we can leave that until someone complains.

### DIFF
--- a/checkov/common/output/graph_record.py
+++ b/checkov/common/output/graph_record.py
@@ -8,6 +8,6 @@ class GraphRecord(Record):
         super().__init__(record.check_id, record.check_name, record.check_result, record.code_block, record.file_path,
                          record.file_line_range, record.resource, record.evaluations, record.check_class,
                          record.file_abs_path, record.entity_tags, record.caller_file_path,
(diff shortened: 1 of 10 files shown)


pascal-hofmann:
I also needed the resource addresses in #2044. It may make sense to tackle both problems at once.

gruebel (approved):
🍻

mikeurbanski1:
@pascal-hofmann I think we need to look at whether your PR will impact the output that gets used in the Bridgecrew platform, so I am going to merge this as is.

PR #2049, opened 2021-12-02
Fixed SSE validation

Saarett:
In some cases the SSE is not a dict and thus we receive an exception for get method that doesn't exist


checkov/terraform/checks/resource/aws/DynamoDBTablesEncrypted.py
```diff
@@ -15,10 +15,11 @@ def __init__(self):
     def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
         if 'server_side_encryption' in conf.keys():
             sse = conf['server_side_encryption'][0]
-            enabled = sse.get("enabled")
-            kms_key_arn = sse.get("kms_key_arn")
-            if enabled == [True] and kms_key_arn is not None:
-                return CheckResult.PASSED
+            if isinstance(sse, Dict):
```
gruebel, line 18: I'm actually surprised it works like this
```suggestion
            if isinstance(sse, dict):
```

gruebel (approved):
👍

PR #2075, opened 2021-12-09
1st digital Ocean check

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/digitalocean/SpacesBucketVersioning.py
+++ b/checkov/terraform/checks/resource/digitalocean/SpacesBucketVersioning.py
@@ -1,0 +1,19 @@
+
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class SpaceBucketVersioning(BaseResourceValueCheck):
+    def __init__(self):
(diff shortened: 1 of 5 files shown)


checkov/terraform/checks/resource/digitalocean/SpacesBucketVersioning.py
```diff
@@ -0,0 +1,19 @@
+
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class SpaceBucketVersioning(BaseResourceValueCheck):
+    def __init__(self):
+        name = "Ensure the Spaces bucket has versioning enabled"
+        id = "CKV_DIO_1"
+        supported_resources = ['digitalocean_spaces_bucket', 'digitalocean_spaces_bucket_object']
```
gruebel, line 10:
```suggestion
        supported_resources = ['digitalocean_spaces_bucket']
```
I think you just meant to check the bucket for versioning 😉

JamesWoolfenden: I did, id planned to do both which is what tfsec does but the object doesnt have a separate versioning attribute so their check is in error.

gruebel (approved):
🍻

PR #2084, opened 2021-12-10
adjust CKV_AZURE_43 to ignore names, if they could not be evaluated

### LINKED ISSUE #741: False positive for CKV_AZURE_43: check storage account name
I'm building my Storage Account names like this
```
name                = "${local.saname_prefix}diagnostics${module.tf-var-project.random_id}
```

With https://github.com/bridgecrewio/checkov/pull/429 merged I now get a Check failure on the SA name:

```
Check: CKV_AZURE_43: "Ensure the Storage Account naming rules"
	FAILED for resource: azurerm_storage_account.diagnostics
	File: /az_diag_sa.tf:8-22

		8  | resource "azurerm_storage_account" "diagnostics" {
		9  |   #checkov:skip=CKV_AZURE_35:Public access is allowed
		10 |   name                = "${local.saname_prefix}diagnostics${module.tf-var-project.random_id}"

````

gruebel:

Fixes #741 

If we can't evaluate the typical references, we should just ignore it and flag it as `UNKNOWN`.

### DIFF
--- a/checkov/terraform/checks/resource/azure/StorageAccountName.py
+++ b/checkov/terraform/checks/resource/azure/StorageAccountName.py
@@ -1,31 +1,41 @@
+import re
+from typing import List, Dict, Any
+
 from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
 from checkov.common.models.enums import CheckResult, CheckCategories
-import re
-from typing import List
 
-STO_NAME_REGEX = re.compile('^[a-z0-9]{3,24}$')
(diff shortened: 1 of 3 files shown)


schosterbarak (approved):

Ketan-Singrole:
@gruebel As per my understanding in the following case it should have been an unknown case but it seems as it is still identifying it as failed 

image

It also seems like the Guide URL is pointing at the wrong documentation

Issue #2046, opened 2021-12-01
Scan of tfplan with only content being outputs causes parsing error


jkesterson:
**Describe the bug**
When a tfplan consisting of only output values is scanned, it throws a parsing error.

**To Reproduce**
Scan the attached json file.
[plan.txt](https://github.com/bridgecrewio/checkov/files/7637224/plan.txt)

**Expected behavior**
The scan should pass with assumedly no failed checks.

**Desktop (please complete the following information):**
 - OS: Win10/Ubuntu (WSL) 
 - Checkov Version  2.0.546/2.0.620

**Additional Context**
Tried replacing the null value with a string to confirm its not an issue with null - still get a parsing failure.

Issue #1927, opened 2021-11-11
Parsing error with Helm values file


brettcurtis:
**Describe the bug**
When running checkov against a few helm installed applications we use (GoCD & Hazelcast) we get a parsing error back.

**To Reproduce**

**GoCD**

1. `helm repo add gocd https://gocd.github.io/helm-chart`
2. `helm inspect values gocd/gocd > gocd.yaml`
3. `checkov -f gocd.yaml`

**Hazelcast**

1. `helm repo add hazelcast https://hazelcast-charts.s3.amazonaws.com`
2. `helm inspect values hazelcast/hazelcast-enterprise > hazelcast.yaml`
3. `checkov -f hazelcast.yaml`

**Expected behavior**
I'd expect the following to not report back the parsing error:

**GoCD**

```none
By bridgecrew.io | version: 2.0.555 
Update available 2.0.555 -> 2.0.566
Run pip3 install -U checkov to update 


cloudformation scan results:

Passed checks: 0, Failed checks: 0, Skipped checks: 0, Parsing errors: 1

Error parsing file gocd.yaml
secrets scan results:

Passed checks: 0, Failed checks: 2, Skipped checks: 0

Check: CKV_SECRET_6: "Base64 High Entropy String"
        FAILED for resource: 73726940fbbe4f60d623a48c0cf51ba15ececae6
        File: /gocd.yaml:238-239
        Guide: https://docs.bridgecrew.io/docs/git_secrets_6

                238 |       secretName: gocd-server-ssh


Check: CKV_SECRET_6: "Base64 High Entropy String"
        FAILED for resource: 363213b0ce319aa2eaa9bf606f726c2617bcc259
        File: /gocd.yaml:398-399
        Guide: https://docs.bridgecrew.io/docs/git_secrets_6

                398 |       secretName: gocd-agent-ssh
```

**Hazelcast**

```none
❯ 
By bridgecrew.io | version: 2.0.555
Update available 2.0.555 -> 2.0.566
Run pip3 install -U checkov to update


cloudformation scan results:

Passed checks: 0, Failed checks: 0, Skipped checks: 0, Parsing errors: 1

Error parsing file hazelcast.yaml
```

**Additional context**
The failed checks I'm getting back are also a false positive, I can do a new issue there if ya want.

metahertz:
Taking a look at this @brettcurtis 
Thanks for the detailed issue report!

metahertz:
Interesting!

[...]

Issue #1903, opened 2021-11-04
Check of Json file including null values results in parsing error


Pellovic26:
**Describe the bug**
Checking a Terraform Plan output saved as json results in parsing error if there is at least one parameter set to null, even if null is an accepted value. Checking directly the main.tf file is working without parsing error. 

**To Reproduce**
1. Create a Terraform configuration, in which the corresponding plan includes at least one null value. Simple example: Use "azurerm" provider and create a resource "azurerm_resource_group" with the required values for name and location. In the resulting final output the field for tag would be set to "null", which is correct.  
![TFCode](https://user-images.githubusercontent.com/93720518/140331877-56ec11e2-88b7-4990-b4a4-88829698ad19.png)

2. Run CLI command: terraform init
3. Run CLI command: terraform plan -out tf.plan
4. Run CLI command: terraform show -json tf.plan > tf.json (the attached file was pretty-printed to see better the corresponding row, but even with pretty-printing this issue comes up)
![TFPlan-json](https://user-images.githubusercontent.com/93720518/140331895-7b30c08a-60fa-4f95-9835-b880cf902aac.png)
5. LOG_LEVEL=DEBUG checkov -f tf.json
6. Parsing Error, with LOG_LEVEL=DEBUG it is visible that the correct null value is the problem.
![ParsingError](https://user-images.githubusercontent.com/93720518/140332728-24e8c03a-30a8-4753-812f-ae903cce0919.png)

**Expected behavior**
Null values are not resulting in general in parsing errors for json files. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Checkov Version 2.0.545
 - Python Version 3.8.10

**Additional Context**
I think from list of issues, the issues #1414 and #1887 are similar and could have the same root cause.

antoinert:
I also ran into this and for documentation's sake I proceeded reproduce the original case in a notebook with latest Ubuntu.

[https://www.tinnable.com/tins/2250abc7-04b9-4ac3-8840-a2b705619122](https://www.tinnable.com/tins/2250abc7-04b9-4ac3-8840-a2b705619122?utm_source=github&utm_campaign=1ST2021)

BrewToR:
I encounter the same issue, when do you plan a fix ?

Issue #1887, opened 2021-11-02
Parsing error scanning terraform plan output


benglerHIG:
**Describe the bug**
Receiving a parsing error when scanning terraform plan output. parsing error is in the cloudformation scan results. the file being scanned contains no cloudformation code. the terraform_plan scan results are reporting well

**To Reproduce**
1. Go to a folder which contains the output of a terraform plan piped into a tf.json file
2. Run cli command  checkov -f .\tf.json
3. See error in attached scanlog.txt

**Expected behavior**
Expect to be able to scan terraform plan output without encountering a parsing error

**Screenshots**
scanlog.txt
[scanlog.txt](https://github.com/bridgecrewio/checkov/files/7464351/scanlog.txt)

ts.json - terraform plan output being scanned
[tf.json.txt](https://github.com/bridgecrewio/checkov/files/7464354/tf.json.txt)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Checkov Version 2.0.536

**Additional context**
Add any other context about the problem here (e.g. code snippets).

jkesterson:
I have had this issue quite a bit as well.

PR #2085, opened 2021-12-10
CF Parsing Issues for TF Plans and Helm YAMLs

njgibbon:
Hello,

This took hours even though it's just a few lines!

When you pass in a JSON TF Plan file then we currently get this type of error:
```
...
cloudformation scan results:

Passed checks: 0, Failed checks: 0, Skipped checks: 0, Parsing errors: 1

Error parsing file tfplan.json
```

This is because all `.json` file are considered to be potential fodder for the CloudFormation Runner unless explicitly filtered out via CLI config:

https://github.com/bridgecrewio/checkov/blob/master/checkov/cloudformation/cfn_utils.py#L18

CloudFormation templates are identified and returned when it find that the `"Resources"` field exists. If, for **all** input files, there is not a single one which is of the right format I think that it is probabilistically safe to guess the user is not trying to scan CF. Of course there _can_ be cases where the user has accidentally passed in all CF files badly formatted but I think this small change would mark an improvement and we could seek to better improve sub-type checking of input JSON and YAML over time because the same issue is currently occurring when a Helm YAML is passed in.

The `exit()` does not exit the entire program only the Cloud Formation Runner thread(?) as i've tested to ensure that if there is a valid `tf-plan.json` it will still output results as expected.

I've added the simple case unit test for _some_ testing because I can't really unit test well for an `exit` aha.

## Issues that this could resolve:
* https://github.com/bridgecrewio/checkov/issues/2046
* https://github.com/bridgecrewio/checkov/issues/1927
* https://github.com/bridgecrewio/checkov/issues/1903
* https://github.com/bridgecrewio/checkov/issues/1887
* https://github.com/bridgecrewio/checkov/issues/1768
* https://github.com/bridgecrewio/checkov/issues/1752 - partly.
* https://github.com/bridgecrewio/checkov/issues/1257 - partly.

## Additional
I figure lots can be done here but I think this is a viable thing shorter term to fix so many issues.

* Perhaps some helm, cf and tf-plan `-f` integration tests can be added. 
* Perhaps we could look into not starting a runner if the none of the input files are of the right sub-format. For example, tf-plans have `"terraform_version":"0.13.4"`, CF have `"AWSTemplateFormatVersion" : "2010-09-09"` and Helm has, well, nothing ;) (@metahertz - https://github.com/bridgecrewio/checkov/issues/1927) but you can infer something might be helm by knowing it's not CF and not a k8s yaml (api version).

There are other things I will look at with plan parsing where it can fail silently unexpectedly but I think I can look at that separately.

**License**

### DIFF
--- a/checkov/cloudformation/runner.py
+++ b/checkov/cloudformation/runner.py
@@ -56,6 +56,9 @@ class Runner(BaseRunner):
 
         if self.context is None or self.definitions is None or self.breadcrumbs is None:
             self.definitions, self.definitions_raw = create_definitions(root_folder, files, runner_filter, parsing_errors)
+            if self.definitions == {} and self.definitions_raw == {}:
+                logging.debug("Stopping CF Runner. Not 1 CF Template found.")
+                exit()
             if external_checks_dir:
                 for directory in external_checks_dir:
                     cfn_registry.load_external_checks(directory)
--- a/tests/cloudformation/test_cfn_utils.py
+++ b/tests/cloudformation/test_cfn_utils.py
@@ -1,0 +1,21 @@
+import os
+import unittest
+
+from checkov.cloudformation import cfn_utils
+
+class TestCFNUtils(unittest.TestCase):
+    def test_get_files_definitions_where_none_cf(self):
+        # Check for return of {}, {} when there is not a single valid CF Template provided.
+        # This will be used to stop the CF Runner up the stack.
+
+        # Simplest case: no files means no valid templates.
+        # Templates are found based on the file containing CF 'Resources'.
+        definitions, definitions_raw = cfn_utils.get_files_definitions([], {})
+
+        # Empty Dict will resolve to false.
+        self.assertFalse(definitions)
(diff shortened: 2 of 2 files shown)


checkov/cloudformation/runner.py
```diff
@@ -56,6 +56,9 @@ def run(
 
         if self.context is None or self.definitions is None or self.breadcrumbs is None:
             self.definitions, self.definitions_raw = create_definitions(root_folder, files, runner_filter, parsing_errors)
+            if self.definitions == {} and self.definitions_raw == {}:
+                logging.debug("Stopping CF Runner. Not 1 CF Template found.")
+                exit()
```
gruebel, line 61: this will fail on windows environments, because we don't multi process them, just multi thread. This will result in ending the complete process prematurely.

gruebel (commented):
I created a new PR #2087, which is tackling the Terraform plan error during the error creation. And you are right, we should some extra work on identifying the correct framework for the given file 💪

PR #2089, opened 2021-12-11
docker: use py3.10 for base image

chenrui333:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-alpine
+FROM python:3.10-alpine
 
 RUN apk add --no-cache git util-linux bash openssl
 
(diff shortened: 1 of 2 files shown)


gruebel:
Hi, did you try to build them locally and can verify it works?

chenrui333:
@gruebel yes, build and tested out on my machine. This output the same results as the [homebrew test](https://github.com/Homebrew/homebrew-core/blob/master/Formula/checkov.rb#L354-L387).

gruebel (approved):
I double checked it and since the first patch version of Python 3.10 came out yesterday, happy to merge 💪

chenrui333:
Thanks @gruebel!

PR #2091, opened 2021-12-12
add pre-commit + flake8

gruebel:

- added `pre-commit` + config + GitHub Action + docs
- added `flake8` + config + included to `pre-commit`
- fixed couple of issues, which were flagged by `flake8`

Step-by-step I will remove the ignored rules in the `flake8` config. Even for single rules I will split them up, because there too many small fixes and that way it is easier to review them and not get too overwhelmed 😱


checkov/common/checks_infra/checks_parser.py
```diff
@@ -1,17 +1,39 @@
...
+    LessThanAttributeSolver,
+    LessThanOrEqualAttributeSolver,
+)
 from checkov.common.graph.checks_infra.base_check import BaseGraphCheck
 from checkov.common.graph.checks_infra.base_parser import BaseGraphCheckParser
 from checkov.common.graph.checks_infra.enums import SolverType
 from checkov.common.graph.checks_infra.solvers.base_solver import BaseSolver
-from checkov.common.checks_infra.solvers import *
+
+# from checkov.common.checks_infra.solvers import *
```
schosterbarak, line 36: is this comment intended?

checkov/common/output/record.py
```diff
@@ -89,7 +89,7 @@ def _code_line_string(code_block):
 
     def to_string(self, compact=False, use_bc_ids=False):
         status = ''
-        evaluation_message = f''
```
schosterbarak, line 92: is this breaking something?

checkov/common/checks_infra/checks_parser.py
gruebel: 🙈 no, forgot to remove it

checkov/common/output/record.py
gruebel: nope, the `f` before the string is not needed, because there is nothing to format 🙂

nimrodkor (approved):
Nice! @gruebel

PR #2097, opened 2021-12-13
use prisma auth token for fetching data

mikeurbanski1:

Fixes 2 platform integrations that did not use the Prisma auth token when fetching data.


gruebel (approved):
👍

PR #2100, opened 2021-12-13
Add instance_template and instance_from_template to supported resources

kartikp10:

Adds support for `google_compute_instance_template` and `google_compute_instance_from_template` to checks for `google_compute_instance`.

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccount.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccount.py
@@ -9,7 +9,8 @@ class GoogleComputeDefaultServiceAccount(BaseResourceCheck):
     def __init__(self):
         name = "Ensure that instances are not configured to use the default service account"
         id = "CKV_GCP_30"
-        supported_resources = ['google_compute_instance']
+        supported_resources = ['google_compute_instance', 'google_compute_instance_from_template',
+                               'google_compute_instance_template']
         categories = [CheckCategories.NETWORKING]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccount.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccount.py
@@ -29,6 +29,10 @@ class GoogleComputeDefaultServiceAccount(BaseResourceCheck):
             self.evaluated_keys = ['name']
             return CheckResult.PASSED
         self.evaluated_keys = ['service_account/[0]/email', 'name']
+        if 'source_instance_template' in conf.keys() and 'service_account' not in conf.keys():
+            # the absence of service_account from a 'google_compute_instance_from_template' definition does not imply
+            # usage of the default service account.
+            return CheckResult.UNKNOWN
         return CheckResult.FAILED
 
 
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeInstanceOSLogin.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeInstanceOSLogin.py
@@ -1,5 +1,5 @@
 from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
-from checkov.common.models.enums import CheckCategories
+from checkov.common.models.enums import CheckCategories, CheckResult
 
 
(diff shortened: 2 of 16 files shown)


gruebel:
as far as I understand the documentation, when we support `google_compute_instance_template` for a check then it should also work for `google_compute_instance_from_template`, because it is basically an override mechanism, right? But why didn't you add both to all 8 checks?

mikeurbanski1:
> as far as I understand the documentation, when we support `google_compute_instance_template` for a check then it should also work for `google_compute_instance_from_template`, because it is basically an override mechanism, right? But why didn't you add both to all 8 checks?

`from_template` inherits from a template, and so the absence of an attribute does not necessarily mean that it fails. But some checks will fail based off the missing attribute, and so this would be likely to result in false positives for `from_template`.

We'll probably need to use graph checks for the `from_template` to get full coverage, but then it gets awkward because we have both python and yaml for the same functional check. So this is a short term fix for someone who is waiting for it.

kartikp10:
@gruebel Yes, this needs to be ported over to a graph check to inherit values from the `source_instance_template`

checkov/terraform/checks/resource/gcp/GoogleComputeInstanceOSLogin.py
```diff
@@ -7,7 +7,7 @@ def __init__(self):
         name = "Ensure that no instance in the project overrides the project setting for enabling OSLogin" \
                "(OSLogin needs to be enabled in project metadata for all instances)"
         id = "CKV_GCP_34"
-        supported_resources = ['google_compute_instance']
+        supported_resources = ['google_compute_instance', 'google_compute_instance_template']
```
gruebel, line 10:
```suggestion
        supported_resources = ['google_compute_instance', 'google_compute_instance_template', 'google_compute_instance_from_template']
```
because it is a negative check

checkov/terraform/checks/resource/gcp/GoogleComputeSerialPorts.py
```diff
@@ -6,7 +6,7 @@ class GoogleComputeSerialPorts(BaseResourceNegativeValueCheck):
     def __init__(self):
         name = "Ensure 'Enable connecting to serial ports' is not enabled for VM Instance"
         id = "CKV_GCP_35"
-        supported_resources = ['google_compute_instance']
+        supported_resources = ['google_compute_instance', 'google_compute_instance_template']
```
gruebel, line 9:
```suggestion
        supported_resources = ['google_compute_instance', 'google_compute_instance_template', 'google_compute_instance_from_template']
```
also negative check

gruebel (commented):
ok, got it, but then I think those two can also support `google_compute_instance_from_template`, because they are negative checks, right?

checkov/terraform/checks/resource/gcp/GoogleComputeSerialPorts.py
kartikp10: Makes sense

kartikp10: I suppose, if a `from_template` instance passes a negative check condition, that doesn't mean that the underlying `template` will also pass the check. But that won't be solved until we have a graph check for it 🤷🏽‍♂️

gruebel: that's also true, only a graph check could help here

checkov/terraform/checks/resource/gcp/GoogleComputeDefaultServiceAccount.py
```diff
@@ -28,6 +29,10 @@ def scan_resource_conf(self, conf):
             self.evaluated_keys = ['name']
             return CheckResult.PASSED
         self.evaluated_keys = ['service_account/[0]/email', 'name']
+        if 'source_instance_template' in conf.keys() and 'service_account' not in conf.keys():
+            # the absence of service_account from a 'google_compute_instance_from_template' definition does not imply
```
kartikp10, line 33: @mikeurbanski1 Let me know what you think about the usage of `CheckResult.UNKNOWN` in this case

mikeurbanski1: Yeah, this makes sense to me.

kartikp10:
New updates to this PR include:

1.  support for `google_compute_instance_from_template` in all checks
2. when an attribute is missing in `from_template` blocks, the check result is unknown because the underlying template is  not scanned

mikeurbanski1 (approved):

PR #2112, opened 2021-12-15
Clarify log message when module download is skipped and download-external-modules is false

mikeurbanski1:
The log warning is confusing - for external modules, it says "Failed to download module" even though it never tries.

[While some may count this as a failure](https://hips.hearstapps.com/hmg-prod.s3.amazonaws.com/images/baby-yoda-old-yoda-1574103229.jpg?crop=0.486xw:0.973xh;0.514xw,0&resize=480:*), we should be more clear.


gruebel (approved):
👍

PR #2114, opened 2021-12-15
Fixes - use the record absolute file path instead of recalculating it

mikeurbanski1:
This fixes an issue where the absolute path to read the file to send to the fixes API was being incorrectly calculated on Windows, which made the fixes logic not work

Example scan file path: `d:\Dev\bridgecrew\terragoat\terraform\aws\s3.tf`

File path that the fixes logic tries to read: `d:\\Dev\\bridgecrew\\terragoat\\Dev\\bridgecrew\\terragoat\\terraform\\aws\\s3.tf`

Calculating abs path isn't necessary here anyways, as each record already has it.


kartikp10 (approved):
Good catch 👌🏽

Issue #2098, opened 2021-12-13
checkov fails to scan modules hosted in private repos


carloscruzadomatrix:
**Describe the bug**
I cant check external repos, the logs print  "Failed to download module ". I try this in my pc and with github actions (uses: bridgecrewio/checkov-action@master). The error is the same.

**To Reproduce**

1. Run cli command
checkov -d . --download-external-modules true --download-external-modules .external_modules

2. See error
2021-12-13 12:22:15,601 [ThreadPoolEx] [WARNI]  Failed to download module terraform-aws-modules/vpc/aws:3.11.0

![01](https://user-images.githubusercontent.com/95088398/145860130-06310f4a-0203-47a0-acf7-ffcd0cece50d.PNG)
![02](https://user-images.githubusercontent.com/95088398/145860132-3b0809a5-3125-4167-9aaa-6b8fdbbc4ea3.PNG)

gruebel:
Hi @carloscruzadomatrix can you post the `module` usage in your Terraform template, then I can better recreate the issue.

carloscruzadomatrix:
I have 2 modules:

source = "terraform-aws-modules/eks/aws"
source = "terraform-aws-modules/vpc/aws"

2021-12-14 16:36:48,043 [MainThread  ] [WARNI]  Failed to download module terraform-aws-modules/eks/aws:17.24.0
2021-12-14 16:36:48,043 [MainThread  ] [WARNI]  Failed to download module terraform-aws-modules/vpc/aws:3.11.0

gruebel:
I tested it and it works fine, my best guess is, that there is some issue with Windows. As I can see you are running the check on your `G` drive on Windows right? Can you try it on the `C` drive?

carloscruzadomatrix:
but the same problem happend in github actions,
![01](https://user-images.githubusercontent.com/95088398/146264689-f21573aa-3af6-419c-bbb8-4fe2c3e73bcb.PNG)

nimrodkor:
Hey @carloscruzadomatrix , did #2118 get this resolved for you?

tspearconquest:
This is an issue I just ran into

tspearconquest:
```
~/test ❯ checkov -d . -o junitxml
2022-04-27 18:43:18,300 [MainThread  ] [WARNI]  Failed to download module gitlab.com/private/project1/azure:0.0.14 (for external modules, the --download-external-modules flag is required)
[...]

PR #2118, opened 2021-12-16
Fix - External TF Registry Downloads on Windows

njgibbon:
Hello 🙂,

Often you need to do `os.path.join` so that paths work regardless of underlying Operating System.

However, URL / FQDN are platform agnostic and always use `/`. So, just a little oversight here.

**Fixes**
* https://github.com/bridgecrewio/checkov/issues/2098

**License**

### DIFF
--- a/checkov/terraform/module_loading/loaders/registry_loader.py
+++ b/checkov/terraform/module_loading/loaders/registry_loader.py
@@ -31,7 +31,7 @@ class RegistryLoader(ModuleLoader):
         if os.path.exists(self.dest_dir):
             return True
 
-        self.module_version_url = os.path.join(self.REGISTRY_URL_PREFIX, self.module_source, "versions")
+        self.module_version_url = self.REGISTRY_URL_PREFIX + "/" + self.module_source + "/versions"
         if not self.module_version_url.startswith(self.REGISTRY_URL_PREFIX):
             # Local paths don't get the prefix appended
             return False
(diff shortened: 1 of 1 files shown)


checkov/terraform/module_loading/loaders/registry_loader.py
```diff
@@ -31,7 +31,7 @@ def _is_matching_loader(self) -> bool:
         if os.path.exists(self.dest_dir):
             return True
 
-        self.module_version_url = os.path.join(self.REGISTRY_URL_PREFIX, self.module_source, "versions")
+        self.module_version_url = self.REGISTRY_URL_PREFIX + "/" + self.module_source + "/versions"
```
gruebel, line 34:
```suggestion
        self.module_version_url = "/".join((self.REGISTRY_URL_PREFIX, self.module_source, "versions"))
```
a concatenation with "+" is most of the time discouraged, because it creates each time a new string, therefore using `.join()` is most of the time the better idea 🙂 
beware the double brackets are needed, because you need something to iterate through!

checkov/terraform/module_loading/loaders/registry_loader.py
```diff
@@ -54,7 +54,7 @@ def _load_module(self) -> ModuleContent:
 
         best_version = self._find_best_version()
 
-        request_download_url = os.path.join(self.REGISTRY_URL_PREFIX, self.module_source, best_version, "download")
+        request_download_url = self.REGISTRY_URL_PREFIX + "/" + self.module_source + "/" + best_version + "/download"
```
gruebel, line 57: same here

gruebel (commented):
nice catch!

checkov/terraform/module_loading/loaders/registry_loader.py
njgibbon: I did know this I just forgot! Linked to immutability of python strings right. I will update today :)

njgibbon:
> nice catch!

Was another one where it took me a few hours to find even though it's just ending up being two lines. Takes time testing scenario across 2 computers aha.

gruebel (approved):
thanks, nice job 🍻

PR #2129, opened 2021-12-20
Cache parsed module definitions

nimrodkor:

Should save a little bit of overhead

### DIFF
--- a/checkov/terraform/parser.py
+++ b/checkov/terraform/parser.py
@@ -38,6 +38,7 @@ def _filter_ignored_paths(root, paths, excluded_paths):
 
 class Parser:
     def __init__(self, module_class: Type[Module] = Module):
+        self._module_definitions_cache: dict = {}
         self.module_class = module_class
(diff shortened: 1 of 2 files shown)


checkov/terraform/parser.py
```diff
@@ -425,18 +427,21 @@ def _load_modules(self, root_dir: str, module_loader_registry: ModuleLoaderRegis
                             logging.info(f'Got no content for {source}:{version}')
                             continue
 
-                        self._internal_dir_load(directory=content.path(),
-                                                module_loader_registry=module_loader_registry,
-                                                dir_filter=dir_filter, specified_vars=specified_vars,
-                                                module_load_context=module_load_context,
-                                                keys_referenced_as_modules=keys_referenced_as_modules)
+                        if not self._module_definitions_cache.get(content.path()):
```
gruebel, line 430:
```suggestion
                        if content.path() not in self._module_definitions_cache:
```
this is a bit faster

PR #2135, opened 2021-12-21
Initial panos terraform support

jamesholland-uk:

Per the title, this is an initial support for PAN-OS Terraform provider, starting only with the provider check for API keys. More to be added in future. I have built locally per the contributing guides and all checks passed. I have tested against local Terraform code with and without API keys in the provider block.


tests/terraform/checks/provider/panos/test_credentials.py
```diff
@@ -0,0 +1,20 @@
...
+
+class TestCredentials(unittest.TestCase):
+
+    def test_success(self):
+        provider_conf = {}
+        scan_result = check.scan_provider_conf(conf=provider_conf)
+        self.assertEqual(CheckResult.PASSED, scan_result)
+
+    def test_failure(self):
+        provider_conf = {'api_key' :['LUFRPT1yWFdyMFg5NlZxZ1ViU2ZhMTh6aGVEbDJ1UFU9ck9vc2tGcmlHV0tDbWRFa2cxcGUxSU8wMlVjaE9ReU0yYWN5SU1rL2pEOGhDcE50WEt5ABlHQWZoTm8xNG1SQQ==']}
```
mikeurbanski1, line 15: Please parse the resource config specified as a string, like here: https://github.com/bridgecrewio/checkov/blob/master/docs/6.Contribution/Contribute%20New%20Terraform%20Provider.md

The reason is that we have occasionally had issues where the results of parsing the string do not much the hardcoded resource conf, resulting in passing tests but failures in practice.

Although this is a very simple conf, and your definition is correct, it also protects against future parser changes.

jamesholland-uk: Thanks for reviewing, Mike. In the provider testing I created, I did copy the Linode `test_credentials.py` file, which is on the page of the link you provided, so they are almost identical, see below. Are you able to clarify what I need to change here? Many thanks!

Linode example:
```
    def test_success(self):
        provider_conf = {}

        scan_result = check.scan_provider_conf(conf=provider_conf)

    def test_failure(self):
        provider_conf = {'token' :['AKIAIOSFODNN7EXAMPLE']}
        scan_result = check.scan_provider_conf(conf=provider_conf)
        self.assertEqual(CheckResult.FAILED, scan_result)
```

PAN-OS:
```
    def test_success(self):
        provider_conf = {}
        scan_result = check.scan_provider_conf(conf=provider_conf)

    def test_failure(self):
        provider_conf = {'api_key' :['LUFRPT1yWFdyMFg5NlZxZ1ViU2ZhMTh6aGVEbDJ1UFU9ck9vc2tGcmlHV0tDbWRFa2cxcGUxSU8wMlVjaE9ReU0yYWN5SU1rL2pEOGhDcE50WEt5ABlHQWZoTm8xNG1SQQ==']}
        scan_result = check.scan_provider_conf(conf=provider_conf)
        self.assertEqual(CheckResult.FAILED, scan_result)
```

gruebel: @jamesholland-uk I created a new PR #2166 with updated docs on how to write the mentioned `hcl2` syntax. Other than that your changes look great and I'm happy to merge them after that change 🍾

jamesholland-uk:
@gruebel Thanks very much for the update, and for the new docs PR which I saw went live. I (think I've) amended the changes to follow that syntax, let me know what you think...

gruebel (approved):
nice job 🏅

Issue #2136, opened 2021-12-21
git fatal: Could not read from remote repository when using private git repository for terraform modules


pandrel:
**Describe the bug**
git fatal: Could not read from remote repository when using private git repository for terraform modules. 

**To Reproduce**

```terraform
module "sample_bucket" {
    source = "git::ssh://git@github.com/<<org>>>/<<repo>>//modules/s3-encrypted"

    bucket_name           = "sample-bucket"
    enable_lifecycle_rule = true
    enable_versioning     = true
    tags                  = local.default_tags

}
```
```
checkov -d . --download_external_modules True
```

```
2021-12-21 21:53:31,839 [MainThread  ] [WARNI]  Failed to download module git::ssh://git@github.com/<<org>>>/<<repo>>.git//modules/s3-encrypted:None
2021-12-21 21:53:31,916 [MainThread  ] [ERROR]  failed to get git::ssh://github.com/<<org>>>/<<repo>> because of Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --depth=1 ssh://github.com/<<org>>>/<<repo>> /home/ec2-user/environment/<<org>>>/<<repo>>/.external_modules/github.com/<<org>>>/<<repo>>/HEAD
  stderr: 'Cloning into '/home/ec2-user/environment/<<repo>>/.external_modules/github.com/<<org>>>/<<repo>>/HEAD'...
Permission denied (publickey).
fatal: Could not read from remote repository.
```

Error is due to the incorrect url being used in the git command.

git clone -v --depth=1 ssh://github.com/<<org>>>/<<repo>>

should be

git clone -v --depth=1 ssh://git@github.com/<<org>>>/<<repo>>

for some reason git@ is missing in newer releases (2.0.682). This issue was not identified in 2.0.540 release.

gruebel:
the bug is even a bit more specific, it only affects git URLs with sub directories in the source path.

rotemavni:
Hello @pandrel, thank you for bringing this issue to our attention.
I believe I found the cause to this issue, and it will be fixed in the following PR: https://github.com/bridgecrewio/checkov/pull/2138
How do you feel about this solution?

PR #2138, opened 2021-12-22
Fix loading git modules with git@github.com

rotemavni:
Resolves https://github.com/bridgecrewio/checkov/issues/2136
This fix prevents the module source `git::ssh://git@github.com/<<org>>>/<<repo>>...`
To become the following command: `git clone -v --depth=1 ssh://github.com/<<org>>>/<<repo>>` 
and enforce it to be the following: `git clone -v --depth=1 ssh://git@github.com/<<org>>>/<<repo>>`

### DIFF
--- a/checkov/terraform/module_loading/loaders/git_loader.py
+++ b/checkov/terraform/module_loading/loaders/git_loader.py
@@ -74,8 +74,8 @@ class GenericGitLoader(ModuleLoader):
             raise Exception("invalid git url")
 
         username = re.match(re.compile(r"^(.*?@).*"), root_module)
-        if username:
-            root_module = root_module.replace(username[1], "") if username[1] != "git@" else root_module
+        if username and username[1] != "git@":
+            root_module = root_module.replace(username[1], "")
 
         if root_module.endswith(".git"):
             root_module = root_module[:-4]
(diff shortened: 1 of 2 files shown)


checkov/terraform/module_loading/loaders/git_loader.py
```diff
@@ -75,7 +75,7 @@ def _parse_module_source(self) -> ModuleSource:
 
         username = re.match(re.compile(r"^(.*?@).*"), root_module)
         if username:
```
gruebel, line 77:
```suggestion
        if username and username[1] != "git@":
```
why not use the if block here 😉

rotemavni: You're right, as always

gruebel (approved):
🎄

schosterbarak (approved):

pandrel:
Hello @rotemavni, confirmed this fix solved the issue. Thank you for your quick response in resolving this. 
QQ: i am using checkov in codebuild which always installs latest version - probably untested. Are there stable and latest versions that i can point pip to instead of always pointing to a specific version ? like pip install checkov==2.0.stable

PR #2150, opened 2021-12-26
add Terraform attribute null case behaviour

gruebel:

Originally I wanted to migrate a check to  use `BaseResourceNegativeValueCheck`, but then I realized that the current `BaseResourceValueCheck` and `BaseResourceNegativeValueCheck` handle the usage of `null` values incorrectly, which is quite common in Terraform plans, but can also be used in normal template.

### DIFF
--- a/checkov/terraform/checks/resource/aws/IAMPolicyAttachedToGroupOrRoles.py
+++ b/checkov/terraform/checks/resource/aws/IAMPolicyAttachedToGroupOrRoles.py
@@ -1,28 +1,29 @@
-from checkov.common.models.enums import CheckResult, CheckCategories
-from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
-from typing import List
+from typing import List, Any
+
+from checkov.common.models.consts import ANY_VALUE
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
 
 
-class IAMPolicyAttachedToGroupOrRoles(BaseResourceCheck):
-    def __init__(self):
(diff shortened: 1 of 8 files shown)


checkov/terraform/checks/resource/aws/IAMPolicyAttachedToGroupOrRoles.py
```diff
@@ -1,28 +1,29 @@
...
+
+class IAMPolicyAttachedToGroupOrRoles(BaseResourceNegativeValueCheck):
+    def __init__(self) -> None:
         name = "Ensure IAM policies are attached only to groups or roles (Reducing access management complexity may " \
                "in-turn reduce opportunity for a principal to inadvertently receive or retain excessive privileges.)"
         id = "CKV_AWS_40"
-        supported_resources = ['aws_iam_user_policy_attachment', 'aws_iam_user_policy', 'aws_iam_policy_attachment']
-        categories = [CheckCategories.IAM]
+        supported_resources = ('aws_iam_user_policy_attachment', 'aws_iam_user_policy', 'aws_iam_policy_attachment')
+        categories = (CheckCategories.IAM,)
```
nimrodkor, line 14: @gruebel Why not keep as a list?

nimrodkor (approved):
Very nice!!

checkov/terraform/checks/resource/aws/IAMPolicyAttachedToGroupOrRoles.py
gruebel: in general creating and iterating over a static `tuple` is faster and requires less memory compared to a `list`. In the past I checked, if we do more than just iterating over both attributes, but we don't so it is recommended to use `tuple` here 🙂 of course `set` is even faster to iterate over, but they take more time to create and are much more memory hungry 😄

PR #2153, opened 2021-12-27
enable autocompletion

bosmak:
Adds autocomplete, it will look something like this

image

Let me know if I need to fix something or something wrong.


gruebel (approved):
Thanks for the great addition 🍾

PR #2154, opened 2021-12-27
Fix issue 2063 - CKV_AWS_174 incorrect reporting

### LINKED ISSUE #2063: CKV_AWS_174 incorrect reporting
CKV_AWS_174 is being triggered in our terraform code even though we have the viewer certificate set to use TLSv.1.2. Snippet of our code here:

viewer_certificate {
   acm_certificate_arn = aws_acm_certificate.cert.arn
   ssl_support_method       = "sni-only"
   minimum_protocol_version = "TLSv1.2_2019" 
}

Running checkov on our terraform code

**Expected behavior**
This check should be passed

**Additional context**
It looks to me like the issue is in the code between lines 17 and 19. I dont think based on the terraform documentation and the if statements that it would ever pass if using an acm certificate

https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/CloudfrontTLS12.py

rahulmlokurte:

Fixes #2063

This will resolve Issue 2063 CKV_AWS_174 incorrect reporting. @nimrodkor: Can you please check the code. I also have a doubt, if **cloudfront_default_certificate** condition is required at all?

### DIFF
--- a/checkov/terraform/checks/resource/aws/CloudfrontTLS12.py
+++ b/checkov/terraform/checks/resource/aws/CloudfrontTLS12.py
@@ -12,23 +12,18 @@ class CloudFrontTLS12(BaseResourceValueCheck):
 
     def scan_resource_conf(self, conf):
         if "viewer_certificate" in conf.keys():
-            # check if cloudfront_default_certificate is true then this could use less than tls 1.2
             viewer_certificate = conf["viewer_certificate"][0]
-            if 'cloudfront_default_certificate' not in viewer_certificate:
-                if "minimum_protocol_version" in viewer_certificate:
+            # check if cloudfront_default_certificate is true then this could use less than tls 1.2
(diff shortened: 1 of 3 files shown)


gruebel:
@rahulmlokurte thanks for trying out to fix this issue 💪 
could also add a test [here](https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/aws/example_CloudfrontTLS12/main.tf) and then you also need to adjust the test case [here](https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/aws/test_CloudFrontTLS12.py).

And a small challenge for you, you can adjust the check in a way that you don't have to double the `minimum_protocol_version` block.

rahulmlokurte:
@gruebel Removed duplication of `minimum_protocol_version` block and added the test case

gruebel (approved):
Thanks looks good 🍾

PR #2155, opened 2021-12-27
[AUTO-PR] Update pipenv packages

schosterbarak:
bump pipenv packages
- Auto-generated by [pipenv-update github action](https://github.com/bridgecrewio/checkov/blob/master/.github/workflows/pipenv-update.yml)


gruebel:
Dependencies were already updated via a different PR

PR #2160, opened 2021-12-30
Added packages to Twistcli report

Saarett:

### DIFF
--- a/checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
+++ b/checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
@@ -38,7 +38,7 @@ class PackageScanningIntegration(TwistcliIntegration):
                 "type": package.get("type"),
                 "name": package.get("name"),
                 "version": package.get("version"),
-                "licenses":  package.get("licenses", [])
+                "licenses":  package.get("licenses") or []
             }
             for package in twistcli_scan_result.get("packages") or []
         ]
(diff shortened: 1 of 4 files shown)


checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
```diff
@@ -33,6 +33,15 @@ def _create_report(
...
             for vul in twistcli_scan_result.get("vulnerabilities") or []
         ]
+        packages = [
+            {
+                "type": package.get("type"),
+                "name": package.get("name"),
+                "version": package.get("version"),
+                "licenses":  package.get("licenses", [])
+            }
+            for package in twistcli_scan_result.get("packages") or []
```
gruebel, line 43: 👍

checkov/common/bridgecrew/vulnerability_scanning/integrations/package_scanning.py
```diff
@@ -33,6 +33,15 @@ def _create_report(
             }
             for vul in twistcli_scan_result.get("vulnerabilities") or []
         ]
+        packages = [
+            {
+                "type": package.get("type"),
+                "name": package.get("name"),
+                "version": package.get("version"),
+                "licenses":  package.get("licenses", [])
```
gruebel, line 41: does use here ` or []`?

Saarett: @gruebel thanks for the input, `package["licenses"]` might not exist at all so I need a default value.. I could put `None` instead of `[]`, but I'm not sure `or []` is the right approach here. WDYT?

gruebel: it depends what you want to achieve, but I think you should either do
```python
"licenses":  package.get("licenses")
```
or
```python
"licenses":  package.get("licenses") or []
```
then you will send `None` or an empty list to the endpoint.

gruebel (approved):
🍾

PR #2166, opened 2021-12-31
adjust provider tests and docs

gruebel:

When I checked PR #2135 I realized that our docs are quite old for adding a new Provider check, so I adjusted them and migrated all the provider tests to use the `hcl2` syntax.

### DIFF
--- a/checkov/terraform/checks/provider/linode/__init__.py
+++ b/checkov/terraform/checks/provider/linode/__init__.py
@@ -1,5 +1,4 @@
-from os.path import dirname, basename, isfile, join
-import glob
+from pathlib import Path
 
-modules = glob.glob(join(dirname(__file__), "*.py"))
-__all__ = [basename(f)[:-3] for f in modules if isfile(f) and not f.endswith("__init__.py")]
+modules = Path(__file__).parent.glob("*.py")
+__all__ = [f.stem for f in modules if f.is_file() and not f.stem == "__init__"]
--- a/docs/6.Contribution/Contribute New Terraform Provider.md
+++ b/docs/6.Contribution/Contribute New Terraform Provider.md
@@ -21,32 +21,36 @@ First create a new folder `tests/terraform/checks/resource/linode/` and add `tes
 import unittest
 
 import hcl2
+
 from checkov.terraform.checks.resource.linode.authorized_keys import check
 from checkov.common.models.enums import CheckResult
 
 
-class Testauthorized_keys(unittest.TestCase):
-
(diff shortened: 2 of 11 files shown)


nimrodkor (approved):

checkov/terraform/checks/provider/linode/__init__.py
```diff
@@ -1,5 +1,4 @@
-from os.path import dirname, basename, isfile, join
-import glob
+from pathlib import Path
 
-modules = glob.glob(join(dirname(__file__), "*.py"))
-__all__ = [basename(f)[:-3] for f in modules if isfile(f) and not f.endswith("__init__.py")]
+modules = Path(__file__).parent.glob("*.py")
```
schosterbarak, line 3: what does this change?

docs/6.Contribution/Contribute New Terraform Provider.md
```diff
@@ -57,34 +61,37 @@ Add a placeholder file at `tests/terraform/checks/resource/linode/__init__.py`
...
 ```
 
 And also add `checkov/terraform/checks/resource/linode/__init__.py`:
 
 ```python
-from os.path import dirname, basename, isfile, join
-import glob
-modules = glob.glob(join(dirname(__file__), "*.py"))
-__all__ = [ basename(f)[:-3] for f in modules if isfile(f) and not f.endswith('__init__.py')]
+from pathlib import Path
```
schosterbarak, line 91: does this change break backward comptability?

checkov/terraform/checks/provider/linode/__init__.py
gruebel: the functionality is the same, it just uses the "newer" builtin module `pathlib` which is object oriented and supports Windows file paths much better.

docs/6.Contribution/Contribute New Terraform Provider.md
gruebel: nope, no backwards compatibility issues. It does the same as before, but is the modern way to do it 🙂

schosterbarak:
Interesting. What does the modern way improve?

gruebel:
> Interesting. What does the modern way improve?

There are a couple of nice features, which `pathlib` brings to the table, one of it is here used.
Instead of  writing `basename(f)[:-3]` to get the filename and then slice the string to get the name without the file extension, you can just use `f.stem` which returns the same 🙂 there are many more great features, which `pathlib` supports and definitely worth checking out 🚀  https://docs.python.org/3/library/pathlib.html

PR #2168, opened 2022-01-02
fix incorrect handling of empty lists in TF plan

### LINKED ISSUE #1339: CKV_GCP_26 is not working as expected when using terraform plan output.
**Describe the bug**
CKV_GCP_26 check is not providing expected output when it is ran against terraform output json file. Check is gives the result as PASSED even if we don't have the flow logs enabled for the subnet. However if we run it against the directory where we have the .tf files then the check returns the FAILED status which is expected. 

checkov version used: `2.0.236`
Terraform version used: `1.0.1`

**To Reproduce**
1. Create a sample subnet resource in a network.tf file. 
```
resource "google_compute_network" "vpc" {
  name                    = "test-vpc"
  auto_create_subnetworks = "false"
  routing_mode            = "REGIONAL"
}

resource "google_compute_subnetwork" "private-subnet" {
  name                     = "test-sub"
  ip_cidr_range            = "10.1.0.0/16"
  network                  = google_compute_network.vpc.self_link
  private_ip_google_access = true
}
```
2. terraform plan -out tf.plan
3. terraform show -json tf.plan  > tf.json 
4. checkov -f tf.json
5. Now run checkov -f network.tf

Both the commands will give conflicting results

```
Check: CKV_GCP_26: "Ensure that VPC Flow Logs is enabled for every subnet in a VPC Network"
        PASSED for resource: google_compute_subnetwork.private-subnet
        File: /tf/tfplan.json:0-0
        Guide: https://docs.bridgecrew.io/docs/bc_gcp_logging_1
```
```
Check: CKV_GCP_26: "Ensure that VPC Flow Logs is enabled for every subnet in a VPC Network"
        FAILED for resource: google_compute_subnetwork.private-subnet
        File: /network.tf:6-11
        Guide: https://docs.bridgecrew.io/docs/bc_gcp_logging_1

                6  | resource "google_compute_subnetwork" "private-subnet" {
                7  |   name                     = var.private_subnet_name
                8  |   ip_cidr_range            = var.private_subnet_cidr
                9  |   network                  = google_compute_network.vpc.self_link
                10 |   private_ip_google_access = true
                11 | }
```
**Expected behavior**
The check should return the status as FAILED when running it against terraform plan output json

gruebel:

Fixes #1339

- adjusted the evaluation of empty lists to be a missing block, which is very typical in TF plan <- will result in much more checks to flag resources to be not complaint, had to adjust one TF plan test, which flagged 7 resources not being complaint against a check
- adjusted `CKV_GCP_26` to ignore `INTERNAL_HTTPS_LOAD_BALANCER` subnets
- adjusted `CKV_AZURE_6` tests to better syntax and one test was incorrect, because you can pass an empty list for `api_server_authorized_ip_ranges`, which is the same as removing the property

### DIFF
--- a/checkov/terraform/checks/resource/base_resource_value_check.py
+++ b/checkov/terraform/checks/resource/base_resource_value_check.py
@@ -65,7 +65,7 @@ class BaseResourceValueCheck(BaseResourceCheck):
             value = dpath.get(conf, inspected_key)
             if isinstance(value, list) and len(value) == 1:
                 value = value[0]
-            if value is None:
+            if value is None or (isinstance(value, list) and not value):
                 return self.missing_block_result
             if ANY_VALUE in expected_values and value is not None and (not isinstance(value, str) or value):
                 # Key is found on the configuration - if it accepts any value, the check is PASSED
--- a/checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
+++ b/checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
@@ -1,21 +1,23 @@
-from checkov.common.models.enums import CheckCategories
+from typing import List, Any
+
 from checkov.common.models.consts import ANY_VALUE
+from checkov.common.models.enums import CheckCategories
 from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
 class AKSApiServerAuthorizedIpRanges(BaseResourceValueCheck):
-    def __init__(self):
+    def __init__(self) -> None:
         name = "Ensure AKS has an API Server Authorized IP Ranges enabled"
         id = "CKV_AZURE_6"
-        supported_resources = ['azurerm_kubernetes_cluster']
-        categories = [CheckCategories.KUBERNETES]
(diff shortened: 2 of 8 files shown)


checkov/terraform/checks/resource/base_resource_value_check.py
```diff
@@ -65,7 +65,7 @@ def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
             value = dpath.get(conf, inspected_key)
             if isinstance(value, list) and len(value) == 1:
                 value = value[0]
-            if value is None:
+            if value is None or (isinstance(value, list) and not value):
```
YaaraVerner, line 68:
```suggestion
            if not value:
```

YaaraVerner (approved):
Nice work!

checkov/terraform/checks/resource/base_resource_value_check.py
gruebel: I didn't want to do it, because it would also flag `0` or `False` as a missing block case, which would be incorrect. But still good point in general to leverage the truthiness of objects 👍

PR #2172, opened 2022-01-04
Docs update, small reordering change

jamesholland-uk:

Whilst using the documentation, I thought it would be easier for future contributors if the example code blocks were directly adjacent to their descriptions, so this change just reorders the code blocks and brings the provider init next to its description, leaving the consts code block next to its description.

Before:
Screenshot 2022-01-04 at 12 47 26

After:
Screenshot 2022-01-04 at 12 47 10


gruebel (approved):
makes sense 😄

PR #2178, opened 2022-01-05
Additional panos credentials check

jamesholland-uk:

The PAN-OS Terraform provider allows credentials to be passed as an API key, or a username/password combination ([ref](https://registry.terraform.io/providers/PaloAltoNetworks/panos/latest/docs#argument-reference)).
Initially there was only a check for API keys in the provider code, with this change there is now a check for passwords as well.


checkov/terraform/checks/provider/panos/credentials.py
```diff
@@ -17,6 +17,8 @@ def __init__(self) -> None:
     def scan_provider_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
         if self.secret_found(conf, "api_key", panos_api_key_pattern):
             return CheckResult.FAILED
+        if self.secret_found(conf, "password", panos_password_pattern):
```
gruebel, line 20: no need for the pattern check, you can just check, if something is written for the `password` value
```suggestion
        if conf.get("password"):
```
and please add a test to the test file 🙂

checkov/common/models/consts.py
```diff
@@ -13,4 +13,5 @@
 linode_token_pattern = re.compile("(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{64}(?![A-Za-z0-9/+=])") # nosec
 bridgecrew_token_pattern = re.compile(r"^[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}\Z") # nosec
 panos_api_key_pattern = re.compile(r"^LUFRPT1[a-zA-Z0-9]+==\Z") # nosec
+panos_password_pattern = re.compile(".+") # nosec
```
gruebel, line 16: last thing and then it is good to go 💪 please remove the regex here

checkov/terraform/checks/provider/panos/credentials.py
```diff
@@ -3,12 +3,12 @@
 
 from checkov.common.models.enums import CheckResult, CheckCategories
 from checkov.terraform.checks.provider.base_check import BaseProviderCheck
-from checkov.common.models.consts import panos_api_key_pattern
+from checkov.common.models.consts import panos_api_key_pattern, panos_password_pattern
```
gruebel, line 6: and the `panos_password_pattern` import here

gruebel (commented):
after that change I will ready to merge it 👍

jamesholland-uk (commented):
LGTM

gruebel (approved):
Thanks, looks nice 🍻

Issue #1941, opened 2021-11-15
COMMENT_REGEX does not skip custom checks with non-numerical endings


jmeredith16:
**Describe the bug**
The COMMENT_REGEX for skipping checks does not account for custom checks with a non-numerical naming schema.

```
COMMENT_REGEX = re.compile(r'(checkov:skip=|bridgecrew:skip=) *([A-Z_\d]+)(:[^\n]+)?')
```

The above regex matches checks like what Checkov uses internally `CKV_AWS_20`, but does not match something like `ABC_AWS_MyCheckName`. Specifically, the `MyCheckName` section which is not captured by the `\d` in the regex.

I believe changing the regex to something like the following would allow for custom Check skips for those of us wanting a more descriptive naming schema:
```
(checkov:skip=|bridgecrew:skip=) *([A-Z_\S]+)(:)([^\n]+)?
```
(note the `\d` changing to `\S` and the addition of the `(:)` group)

**To Reproduce**
1. Create a check with a string instead of a number for the last part of the Check ID
2. Run Checkov against it with a skip in the terraform
```
$ checkov -f /path/to/custom/AWSKeyPairDisallowed/AWSKeyPairDisallowed.tf --external-checks-dir /path/to/custom/checks
...
terraform scan results:

Passed checks: 0, Failed checks: 1, Skipped checks: 0

Check: ABC_AWS_AWSKeyPairDisallowed: "Ensure aws_key_pair is not used"
	FAILED for resource: aws_key_pair.abc_unittest_fail
	File: /AWSKeyPairDisallowed.tf:3-7

		3 | resource "aws_key_pair" "abc_unittest_fail" {
		4 |   # checkov:skip=ABC_AWS_AWSKeyPairDisallowed:Blah blah blah
		5 |   key_name   = "aws_key_pair_name"
		6 |   public_key = "aws_key_pair_public_key"
		7 | }

```

**Expected behavior**
I expect Checkov to skip my custom check

**Screenshots**
Current:
![image](https://user-images.githubusercontent.com/17982571/141831781-4b4faf0a-10c0-4163-93c2-074446b22c98.png)

Proposed:
![image](https://user-images.githubusercontent.com/17982571/141831872-f2991756-9ed8-464e-9d45-a0c8d5961fa6.png)

**Desktop (please complete the following information):**
 - OS: OSX
 - Checkov Version 2.0.556

**Additional context**
[...]

PR #2181, opened 2022-01-05
Update COMMENT_REGEX to collect non-numerical Check IDs

jmeredith16:

Fixes https://github.com/bridgecrewio/checkov/issues/1941

This edit chances the regex group to match on lower case as well as the other characters/numbers. The screenshot below shows the new regex and how it groups both traditional and non-numerical Check IDs correctly.

image

### DIFF
--- a/checkov/common/comment/enum.py
+++ b/checkov/common/comment/enum.py
@@ -1,3 +1,3 @@
 import re
 
-COMMENT_REGEX = re.compile(r'(checkov:skip=|bridgecrew:skip=) *([A-Z_\d]+)(:[^\n]+)?')
+COMMENT_REGEX = re.compile(r'(checkov:skip=|bridgecrew:skip=) *([a-zA-Z\d_]+)(:[^\n]+)?')


gruebel (approved):
looks good for me, thinking about, if we should add a test for this change 🤔  will add someone else to review this change

jmeredith16:
Thanks @gruebel - I looked around for and didn't see any places where there were tests for skipping that I could just add on to. Could you suggest a location for those tests and a good way of testing skipped vs not skipped?

gruebel:
@jmeredith16 sorry for the delay, we are happy to merge it after you add a test for the new change, you can find skip tests here in the file, but you should probably use this [test](https://github.com/bridgecrewio/checkov/blob/e54fd705676477ab006f62d229cc37d633a3c2fb/tests/terraform/runner/test_runner.py#L105) as a base, because you need to create a custom check and then skip it.

mikeurbanski1:
I missed this and ended up doing it myself, with tests https://github.com/bridgecrewio/checkov/pull/2335

So we'll merge that. But thanks, Joe!

PR #2183, opened 2022-01-06
To eliminate child duplicates on scan locally

eurogig:
Previously the job/cronjob for checkov extracted all resources and relied on checkov to filter out child duplicates in resources based on "ownerReferences" in the yaml.  Now it installs yq (in the Dockerfile) and uses it to filter our child resources locally.  The result should align with previous behaviour so should be compatible with previous runs when users upgrade.  Tested on my local cluster.

### DIFF
--- a/kubernetes/Dockerfile
+++ b/kubernetes/Dockerfile
@@ -12,6 +12,8 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s 
 RUN groupadd -g 12000 -r checkov && useradd -u 12000 --no-log-init -r -g checkov checkov
 RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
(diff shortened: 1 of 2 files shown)


kubernetes/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
 RUN groupadd -g 12000 -r checkov && useradd -u 12000 --no-log-init -r -g checkov checkov
 RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
+RUN curl -L -o /usr/bin/yq https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 \
```
schosterbarak, line 15:
```suggestion
RUN curl -L -o /usr/bin/yq https://github.com/bridgecrewio/yq/releases/download/v4.16.2/yq_linux_amd64 \
```

I've forked it to our org for stability (in case original is deleted) and security (prevent injection of unexpected new code)

schosterbarak:
duplicate of #2184 closing.

PR #2184, opened 2022-01-06
Remove ownerReferences filter from scan

eurogig:
From what I can gather the ownerReferences filter was there only (needs double checking) for the cronjob implementation of checkov.   I've submitted a different PR to this one to locally remove k8s objects with this metadata as part of the job as opposed to being baked into checkov.  

The positive by-product of removing this filter will be that the new checkov admission controller can then properly scan child components during a UPDATE of k8s resources.

### DIFF
--- a/kubernetes/Dockerfile
+++ b/kubernetes/Dockerfile
@@ -12,7 +12,7 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s 
 RUN groupadd -g 12000 -r checkov && useradd -u 12000 --no-log-init -r -g checkov checkov
 RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
-RUN curl -L -o /usr/bin/yq https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 \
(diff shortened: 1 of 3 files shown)


gruebel:
@eurogig as far as I understand the changes, we remove the the objects with `ownerReferences` before running `checkov` and therefore we also don't need the little extra check inside `checkov`, right?

kubernetes/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s
 RUN groupadd -g 12000 -r checkov && useradd -u 12000 --no-log-init -r -g checkov checkov
 RUN mkdir /data && mkdir /app && mkdir /home/checkov
 RUN chown checkov:checkov /data /app /home/checkov
+RUN curl -L -o /usr/bin/yq https://github.com/mikefarah/yq/releases/download/v4.16.2/yq_linux_amd64 \
```
schosterbarak, line 15:
```suggestion
RUN curl -L -o /usr/bin/yq https://github.com/bridgecrewio/yq/releases/download/v4.16.2/yq_linux_amd64 \
```

I've forked it to our org for stability (in case original is deleted) and security (prevent injection of unexpected new code)

schosterbarak:
> `ferences` before running `checkov` and therefore we also don't need the little extra check inside `c`

correct

gruebel (approved):
look good enough 😄

PR #2187, opened 2022-01-07
Improve API error handling

mikeurbanski1:

Catches an issue that can occur if the backend config for assuming the S3 upload role fails, indicating an issue with the Bridgecrew backend itself.

The error message in the S3 role response is:
```
User: arn:aws:sts::___:assumed-role/bc-integrations-api-prod/bc-integrations-api-prod is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::___:role/upload-files-to-scanners-bucket-prod-___
```

This is not caught in the error handling, and results in an infinite loop.

This PR fixes that, and also adds a little extra logging to help catch it and instruct the user on next steps (for any auth error). And, while we're at it, it explicitly handles SSL errors.


gruebel (approved):
🏆

PR #2188, opened 2022-01-07
fix flake8 issues E202,E225,E226

gruebel:

Fixed the next batch of flake8 errors and added type hints, where I passed by 😄


nimrodkor (approved):
Feels like we should have started out with a good flake8 configuration in the first place 😅 
GJ @gruebel

PR #2192, opened 2022-01-10
Increase K8s job's memory

electrocucaracha:
The current values used by the Kubernetes Job causes `OOMKilled` error. Doubling the size request and limit memory avoids it.


schosterbarak:
thanks @electrocucaracha

schosterbarak (approved):

PR #2216, opened 2022-01-16
support check suppression via CFN Metadata

### LINKED ISSUE #1931: Allow suppression to be done in Metadata properties of the AWS CloudFormation template and/or resources
Allow suppression to be done in Metadata properties of the template and/or resources

### LINKED ISSUE #2194: CKV_AWS_116: DLQ applies only to async invocations
**Describe the issue**
CKV_AWS_116: "Ensure that AWS Lambda function is configured for a Dead Letter Queue(DLQ)"
DLQ applies only to synchronous Lambda invocations, so if you have a REST API for example it won't have any effect and thus is unnecessary. The rule fires up for all Lambdas.

I understand that unfortunately in a general case it is impossible to figure out from a CloudFormation template with a Lambda function whether it is synchronous or asynchronous, because you can call any function in either way with AWS CLI or API, without any event mapping.

gruebel:

Fixes #2194
Fixes #1931

Extended the graph to also store the CloudFormation `Metadata` and added the possibility to actually add the skip comment that section. Added an example to the docs.

### DIFF
--- a/checkov/cloudformation/context_parser.py
+++ b/checkov/cloudformation/context_parser.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import operator
 import re
@@ -111,7 +112,7 @@ class ContextParser(object):
                 yield node[kv]
 
     @staticmethod
-    def collect_skip_comments(entity_code_lines: List[Tuple[int, str]]) -> List[_SkippedCheck]:
+    def collect_skip_comments(entity_code_lines: List[Tuple[int, str]], resource_config: Optional[DictNode] = None) -> List[_SkippedCheck]:
         skipped_checks = []
         bc_id_mapping = bc_integration.get_id_mapping()
         ckv_to_bc_id_mapping = bc_integration.get_ckv_to_bc_id_mapping()
@@ -130,6 +131,28 @@ class ContextParser(object):
                     skipped_check["bc_id"] = ckv_to_bc_id_mapping.get(skipped_check["id"])
 
                 skipped_checks.append(skipped_check)
+        if resource_config:
+            metadata = resource_config.get("Metadata")
+            if metadata:
+                ckv_skip = metadata.get("checkov", {}).get("skip", [])
+                bc_skip = metadata.get("bridgecrew", {}).get("skip", [])
+                if ckv_skip or bc_skip:
+                    for skip in itertools.chain(ckv_skip, bc_skip):
+                        skip_id = skip.get("id")
+                        skip_comment = skip.get("comment", "No comment provided")
+                        if skip_id is None:
+                            logging.debug("Check suppression is missing key 'id'")
+                            continue
+
+                        skipped_check = {"id": skip_id, "suppress_comment": skip_comment}
+                        if bc_id_mapping and skipped_check["id"] in bc_id_mapping:
+                            skipped_check["bc_id"] = skipped_check["id"]
+                            skipped_check["id"] = bc_id_mapping[skipped_check["id"]]
+                        elif ckv_to_bc_id_mapping:
+                            skipped_check["bc_id"] = ckv_to_bc_id_mapping.get(skipped_check["id"])
+
+                        skipped_checks.append(skipped_check)
+
         return skipped_checks
 
     @staticmethod
--- a/checkov/cloudformation/context_parser.py
+++ b/checkov/cloudformation/context_parser.py
@@ -141,7 +141,7 @@ class ContextParser(object):
                         skip_id = skip.get("id")
                         skip_comment = skip.get("comment", "No comment provided")
                         if skip_id is None:
-                            logging.debug("Check suppression is missing key 'id'")
+                            logging.warning("Check suppression is missing key 'id'")
                             continue
 
                         skipped_check = {"id": skip_id, "suppress_comment": skip_comment}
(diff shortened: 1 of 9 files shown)


checkov/cloudformation/runner.py
```diff
@@ -111,7 +111,14 @@ def check_definitions(self, root_folder, runner_filter, report):
                         if entity_lines_range and entity_code_lines:
                             # TODO - Variable Eval Message!
                             variable_evaluations = {}
-                            skipped_checks = ContextParser.collect_skip_comments(entity_code_lines)
+
+                            skipped_checks =  resource_context.get("skipped_checks")
```
Saarett, line 115: We have we added this step?

Saarett (approved):
Very nice addition! Looks good to me, added 1 comment in the code.
Please add a test for metadata in json files.
Please wait for another approval :)

docs/2.Basics/Suppressing and Skipping Policies.md
```diff
@@ -94,3 +94,59 @@ Resources:
...
+    'skip': [
+      {
+        'id': 'CKV_AWS_18',
+        'comment': 'Ensure the S3 bucket has access logging enabled'
+      }
+    ]
+  }
+}
+```
+Using `cdk synth` this results in a following CloudFormation template
```
schosterbarak, line 136: show example bash run

docs/2.Basics/Suppressing and Skipping Policies.md
```diff
@@ -94,3 +94,59 @@ Resources:
...
+        Status: Enabled
+    UpdateReplacePolicy: Retain
+    DeletionPolicy: Retain
+    Metadata:
+      checkov:
+        skip:
+          - id: CKV_AWS_18
+            comment: Ensure the S3 bucket has access logging enabled
+
+```
```
schosterbarak, line 152: show the printed cli output when scanning this manifest to show that a policy is being skipped

docs/2.Basics/Suppressing and Skipping Policies.md
```diff
@@ -94,3 +94,59 @@ Resources:
...
+    Properties:
+      DBName: "mydb"
+      DBInstanceClass: "db.t3.micro"
+      Engine: "mysql"
+      MasterUsername: "master"
+      MasterUserPassword: "password"
+```
+
+### CDK Example
+The `Metadata` section of a CDK construct can only be adjusted via the L1 resource.
```
schosterbarak, line 118: what is L1?

checkov/cloudformation/context_parser.py
```diff
@@ -130,6 +131,28 @@ def collect_skip_comments(entity_code_lines: List[Tuple[int, str]]) -> List[_Ski
...
+            metadata = resource_config.get("Metadata")
+            if metadata:
+                ckv_skip = metadata.get("checkov", {}).get("skip", [])
+                bc_skip = metadata.get("bridgecrew", {}).get("skip", [])
+                if ckv_skip or bc_skip:
+                    for skip in itertools.chain(ckv_skip, bc_skip):
+                        skip_id = skip.get("id")
+                        skip_comment = skip.get("comment", "No comment provided")
+                        if skip_id is None:
+                            logging.debug("Check suppression is missing key 'id'")
```
nimrodkor, line 144: @gruebel Let's elevate to `info` / `warning` please, this is highly unexpected! Also add more identifiable data to the log to help the suppressor identify the problematic suppression

checkov/cloudformation/graph_builder/graph_to_definitions.py
```diff
@@ -20,9 +20,13 @@ def convert_graph_vertices_to_definitions(
...
 
         definition = {
-            'Type': vertex.attributes['resource_type'] if vertex.block_type == BlockType.RESOURCE else vertex.block_type,
-            'Properties': vertex.config
+            "Type": vertex.attributes["resource_type"] if vertex.block_type == BlockType.RESOURCE else vertex.block_type,
+            "Properties": vertex.config,
         }
+
+        if vertex.metadata:
+            definition["Metadata"] = vertex.metadata
```
nimrodkor, line 28: @gruebel Why not set it inside the `definition = {...`?

nimrodkor (commented):
Very nice!

checkov/cloudformation/graph_builder/graph_to_definitions.py
gruebel: because it is optional, not every resource has a `Metadata` section. It would be misleading, if I add it with an empty `dict` or `None` value

checkov/cloudformation/context_parser.py
gruebel: makes sense, let's go with `warning` and if users complain being to noisy, then we can reduce it later 😄

docs/2.Basics/Suppressing and Skipping Policies.md
gruebel: this a term used in CDK docs, it stands for Layer 1 or also called Level 1. These are the low level CloudFormation resources. There also L2 and even L3 constructs, which are abstractions of the low level CloudFormation resources.

checkov/cloudformation/runner.py
gruebel: removed

docs/2.Basics/Suppressing and Skipping Policies.md
gruebel: please have a look again

schosterbarak (approved):

Issue #1764, opened 2021-10-15
CKV_GCP_32 not working with terraform plan output


bgoareguer:
**Describe the bug**
When running Checkov on the `terraform plan` output, the `CKV_GCP_32` check does not work.

**To Reproduce**
1. Create a Terraform script that creates a `google_compute_engine` resource with `block-project-ssh-keys` set to `true`
2. Run cli command  'terraform init'
3. Run cli command  'terraform plan --out review.tfplan.binary'
4. Run cli command  'terraform show -json review.tfplan.binary > review.tfplan.json'
5. Run cli command  'checkov -f review.tfplan.json'
6. See error

**Expected behavior**
The CKV_GCP_32 check should be passed

**Screenshots**
![image](https://user-images.githubusercontent.com/43874676/137483848-14aa5679-5d42-405c-b77a-ba4546807014.png)

**Desktop (please complete the following information):**
 - OS: Docker
 - Checkov Version: 2.0.489

**Additional context**
The same Terraform script passes all checks when Checkov is run on the Terraform files instead of on the `terraform plan` output.

jbrule:
I submitted a fix for this #6663 . The policy is coded to only accept a boolean value but Terraform accepts strings as valid input too.

PR #2224, opened 2022-01-17
Fix - Issue 1764 - Terraform Plan Boolean Treatment

njgibbon:
Hello,

Happy new year all!

## Issue
I have one way to resolve this issue: https://github.com/bridgecrewio/checkov/issues/1764

## Description
After investigating what could be going wrong with the issue I found that Terraform Plan treatment does not have a way of consistently representing booleans like is done with the normal parser: https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/parser.py#L535

When performing non-plan scans then booleans whether input as `true` or `True` or `"true"` or "True" are converted so that they are always represented as `True` by the time it is in the dictionary for check evaluation. This difference of representation is what causes the differences in results whether terraform plan or normal terraform scan.

## Details
The Terraform plan - in this particular issue - is representing the GCP resource metadata map as a string regardless of whether a boolean or string is input (this being under `planned_values.root_module.resources[0].values` when a plan is produced for `google_compute_instance` with input:
**Input**
```
  metadata = {
    block-project-ssh-keys = "true"
  }
```
**Output**
```
            "metadata": {
              "block-project-ssh-keys": "true"
            },
```

And I will note that Terraform does not always represent map value booleans as strings. But, at least in this case it does. I feel it has something to do with whether one needs to set the map like this `map = {` or `map {` having been looking at patterns and corresponding terraform plan output.

Where https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#example-usage this must be as it is a single map not a list of maps.

It could also be because other inputs have well-defined types whereas this map of KVPs does not but regardless with the normal terraform parser it is converted to a bool whether terraform its self would have done so or not.

## Test
Test data is a GCP TF plan scan which has had data replaced so that more tests can be exercised against the single plan input.

## Thanks
Please let me know if there is any other way you want me to go with this 🙂 !

## License


checkov/terraform/plan_parser.py
```diff
@@ -46,6 +46,7 @@ def _hclify(obj: DictNode, conf: Optional[DictNode] = None, parent_key: Optional
                 ret_dict[key] = value
             else:
                 ret_dict[key] = [value]
+                ret_dict[key] = _clean_simple_type_list(ret_dict[key])
```
nimrodkor, line 49:
```suggestion
                ret_dict[key] = _clean_simple_type_list([value])
```

nimrodkor (commented):
Very nice @njgibbon ! Just one cleanup comment

njgibbon:
@gruebel @nimrodkor TY Done!

gruebel (approved):
🍻

PR #2231, opened 2022-01-18
check to ensure no aws_elasticache_security_group exist

JamesWoolfenden:
No such groups should not exist

### DIFF
--- a/checkov/terraform/checks/resource/aws/ElasticacheHasSecurityGroup.py
+++ b/checkov/terraform/checks/resource/aws/ElasticacheHasSecurityGroup.py
@@ -7,7 +7,7 @@ class ElasticacheHasSecurityGroup(BaseResourceCheck):
         name = "Ensure no aws_elasticache_security_group resources exist"
         id = "CKV_AWS_196"
         supported_resources = ['aws_elasticache_security_group']
-        categories = [CheckCategories.APPLICATION_SECURITY]
+        categories = [CheckCategories.NETWORKING]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf):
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/aws/ElasticacheHasSecurityGroup.py
```diff
@@ -0,0 +1,19 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class ElasticacheHasSecurityGroup(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure no aws_elasticache_security_group resources exist"
+        id = "CKV_AWS_196"
+        supported_resources = ['aws_elasticache_security_group']
+        categories = [CheckCategories.APPLICATION_SECURITY]
```
gruebel, line 10:
```suggestion
        categories = [CheckCategories.NETWORKING]
```
fits better, we also use it for other similar checks

tests/terraform/checks/resource/aws/test_ElasticCacheHasSecurityGroup.py
```diff
@@ -0,0 +1,32 @@
...
+
+        test_files_dir = current_dir + "/example_ElasticacheHasSecurityGroup"
+        report = runner.run(root_folder=test_files_dir, runner_filter=RunnerFilter(checks=[check.id]))
+        summary = report.get_summary()
+
+        failing_resources = {
+            "aws_elasticache_security_group.exists",
+        }
+
+        failed_check_resources = set([c.resource for c in report.failed_checks])
```
gruebel, line 22:
```suggestion
        failed_check_resources = {c.resource for c in report.failed_checks}
```
it does the same, but is faster, because it doesn't create a list first

gruebel (approved):
🚀 after the tests passed, I will merge it

PR #2234, opened 2022-01-19
add check for mq audit logging

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/aws/MQBrokerAuditLogging.py
+++ b/checkov/terraform/checks/resource/aws/MQBrokerAuditLogging.py
@@ -5,7 +5,7 @@ from checkov.common.models.enums import CheckCategories
 class MQBrokerAuditLogging(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure MQ Broker Audit logging is enabled"
-        id = "CKV_AWS_198"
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/aws/MQBrokerAuditLogging.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class MQBrokerAuditLogging(BaseResourceValueCheck):
+    def __init__(self):
+        name = "Ensure MQ Broker Audit logging is enabled"
+        id = "CKV_AWS_198"
+        supported_resources = ['aws_mq_broker']
+        categories = [CheckCategories.LOGGING]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
```
gruebel, line 12: can you additionally extend the `scan_resource_conf` method to check for the `engine_type`, something like that
```python
        # https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/security-logging-monitoring-rabbitmq.html
        # Audit logging is not supported for RabbitMQ brokers.
        if conf.get("engine_type") == ["RabbitMQ"]:
            return CheckResult.UNKNOWN
```
and of course add a test 😄

PR #2235, opened 2022-01-19
Don't create db security groups outside of a VPC

JamesWoolfenden:

This resource is for legacy BDs created outside of a VPC


gruebel (approved):
💪 let's finish the other check first, so the number test can succeed

PR #2236, opened 2022-01-19
SQSPolicy did not work - reduce false positives

JamesWoolfenden:
False positives were reported.

### DIFF
--- a/checkov/terraform/checks/resource/aws/SQSPolicy.py
+++ b/checkov/terraform/checks/resource/aws/SQSPolicy.py
@@ -19,9 +19,13 @@ class SQSPolicy(BaseResourceCheck):
         :param conf: aws_sqs_queue_policy configuration
         :return: <CheckResult>
         """
-        if "policy" in conf.keys() and type(conf["policy"][0]) is dict:
-            if (conf["policy"][0])['Statement'][0]['Action'] == '*':
-                return CheckResult.FAILED
+        if "policy" in conf.keys():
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/aws/SQSPolicy.py
```diff
@@ -21,9 +19,9 @@ def scan_resource_conf(self, conf):
         :param conf: aws_sqs_queue_policy configuration
         :return: <CheckResult>
         """
-        if "policy" in conf.keys() and is_json(conf["policy"][0]) \
-                and json.loads(conf["policy"][0])['Statement'][0]['Action'] == '*':
-            return CheckResult.FAILED
+        if "policy" in conf.keys() and type(conf["policy"][0]) is dict:
+            if (conf["policy"][0])['Statement'][0]['Action'] == '*':
```
nimrodkor, line 23: @JamesWoolfenden Can we protect this type-wise? Also - to reduce noise, let's pass if the policy has `conditions` 🙏

JamesWoolfenden: sure but im just fixed a policy thats plainly broken here?

nimrodkor: True, but it wouldn't crash when always passes, nor would it have a lot of failures... I want to reduce the pain here

nimrodkor (approved):

PR #2237, opened 2022-01-19
add generic check for unrendered variable and update some checks to return UNKNOWN

mikeurbanski1:
A lot of checks return failures if they encounter an unrendered variable. This also means that our fix engine suggests fixes for them (e.g., change `enabled = var.versioning` to `enabled = true`), which is rarely actually correct.

This PR puts some basic logic in place to check for obvious cases of unrendered variables, and updates some checks to use it. This will be the basis for future more comprehensive changes, and is based on specific feedback from one user who raised it.

### DIFF
--- a/checkov/terraform/checks/resource/base_resource_check.py
+++ b/checkov/terraform/checks/resource/base_resource_check.py
@@ -6,6 +6,7 @@ from checkov.common.checks.base_check import BaseCheck
 from checkov.common.models.enums import CheckResult, CheckCategories
 from checkov.terraform.checks.resource.registry import resource_registry
 from checkov.terraform.parser_functions import handle_dynamic_values
+from checkov.terraform.parser_utils import find_var_blocks
 
 
 class BaseResourceCheck(BaseCheck):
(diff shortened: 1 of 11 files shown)


checkov/terraform/checks/resource/base_resource_check.py
```diff
@@ -28,6 +29,21 @@ def __init__(
...
+        if not isinstance(value, str):
+            return False
+
+        if value.startswith('var.') or value.startswith('local.') or value.startswith('module.'):
+            return True
+
+        if "${" not in value:
+            return False
+
+        if find_var_blocks(value):
```
gruebel, line 43: if I'm not mistaken, then this should cover the `value.startswith('var.')` case

checkov/terraform/checks/resource/base_resource_check.py
```diff
@@ -28,6 +29,21 @@ def __init__(
         self.supported_resources = supported_resources
         resource_registry.register(self)
 
+    @staticmethod
+    def _is_variable_dependant(value: Any) -> bool:
+        if not isinstance(value, str):
+            return False
+
+        if value.startswith('var.') or value.startswith('local.') or value.startswith('module.'):
```
gruebel, line 37:
```suggestion
        if value.startswith(('var.', 'local.', 'module.')):
```
if my other comment is correct, then you can remove the `var.` case there

tests/terraform/checks/test_base_resource_check.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class TestWildcardEntities(unittest.TestCase):
+
+    def test_contains_unrendered_variable(self):
+        self.assertTrue(BaseResourceCheck._is_variable_dependant('var.xyz'))
+        self.assertTrue(BaseResourceCheck._is_variable_dependant('local.xyz'))
+        self.assertTrue(BaseResourceCheck._is_variable_dependant('${var.xyz}'))
+        self.assertTrue(BaseResourceCheck._is_variable_dependant('${local.xyz}'))
+
```
gruebel, line 18: can you add a `module.` test too?

checkov/terraform/checks/resource/base_resource_check.py
mikeurbanski1: We need to check for `var.` before filtering out strings with no interpolation

mikeurbanski1: We need to check for `var.` before filtering out strings that do not have interpolation

gruebel: 👍

gruebel (approved):
💪

Issue #2238, opened 2022-01-20
checkov crashed when checking for missing `kms_key_id`


frek818:
**Describe the issue**
I produced a terraform plan file and ran checkov against it. I expected the checkov command to complete without a traceback. However, this traceback was produced:

**Examples**
```hcl
resource "aws_secretsmanager_secret" "site" {
  name = "the-site-secret"
}
```

**Exception Trace**
Please share the trace for the exception and all relevant output by checkov.
To maximize the understanding, please run checkov with LOG_LEVEL set to debug
as follows:
```sh
2022-01-20 04:15:51,404 [MainThread  ] [ERROR]  Failed to run check: Ensure that Secrets Manager secret is encrypted using KMS CMK for configuration: {'description': [None], 'force_overwrite_replica_secret': [False], 'kms_key_id': [None], 'name': ['ebean-example-kitchen-hagerty-com-secret'], 'recovery_window_in_days': [30], 'tags': [{'Name': 'ebean-example-kitchen-hagerty-com-secret', 'start_line': 753, 'end_line': 753}], 'tags_all': [{'Name': ['ebean-example-kitchen-hagerty-com-secret'], 'start_line': [756], 'end_line': [757]}], 'start_line': [747], 'end_line': [757], '__address__': 'module.elastic_beanstalk.aws_secretsmanager_secret.site'} at file: /../../../_temp/tfplan_KITCHEN.json
Process ForkProcess-6:
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/vsts/.local/lib/python3.8/site-packages/checkov/common/parallelizer/parallel_runner.py", line 29, in func_wrapper
    result = original_func(item)
  File "/home/vsts/.local/lib/python3.8/site-packages/checkov/common/runners/runner_registry.py", line 63, in <lambda>
    lambda runner: runner.run(root_folder, external_checks_dir=external_checks_dir, files=files,
  File "/home/vsts/.local/lib/python3.8/site-packages/checkov/terraform/plan_runner.py", line 67, in run
    self.check_tf_definition(report, runner_filter)
  File "/home/vsts/.local/lib/python3.8/site-packages/checkov/terraform/plan_runner.py", line 93, in check_tf_definition
[...]

PR #2239, opened 2022-01-20
Fix CKV_AWS_149 for terraform_plan

rotemavni:
Fixes issue https://github.com/bridgecrewio/checkov/issues/2238
When creating a plan file with a `aws_secretsmanager_secret` resource that does not specify `kms_key_id`, the run would get an execption.


gruebel (approved):
🍵

PR #2240, opened 2022-01-20
refactoring code with List Comprehension

zjzh:
refactoring code with List Comprehension which is more pythonic, concise and efficient; how do you think this change which has practical value？

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -389,9 +389,10 @@ class Report:
                 )
 
             test_cases[check_name].append(test_case)
-        test_suites = [TestSuite(name=key, 
-                                 test_cases=test_cases[key], 
-                                 package=test_cases[key][0].classname) for key in test_cases.keys()]
+        test_suites = [
+            TestSuite(name=key, test_cases=value, package=value[0].classname)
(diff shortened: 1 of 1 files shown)


checkov/common/output/report.py
```diff
@@ -389,14 +389,10 @@ def get_test_suites(self, use_bc_ids=False) -> List[TestSuite]:
...
-            test_suites.append(
-                TestSuite(
-                    name=key,
-                    test_cases=test_cases[key],
-                    package=test_cases[key][0].classname,
-                )
-            )
+        test_suites = [TestSuite(name=key, 
+                                 test_cases=test_cases[key], 
+                                 package=test_cases[key][0].classname) for key in test_cases.keys()]
```
gruebel, line 394:
```suggestion
        test_suites = [
            TestSuite(name=key, test_cases=value, package=value[0].classname)
            for key, value in test_cases.items()
        ]
```
I think you can even make it better by leveraging the `items()` method, then you don't need to access the `dict` to retrieve `value` for the related `key`

gruebel (commented):
nice find 💪

gruebel (approved):
🚀

PR #2241, opened 2022-01-20
explicitly check TLS types

mikeurbanski1:
A user reported an issue that this check was giving a false negative (i.e., passing) when they had a var with a default value of `null`, i.e.:

```
variable "tls" {
  type = string
  default = null
}

resource "azurerm_storage_account" "example" {
  min_tls_version = var.tls
}
```

This is because our variable rendering logic does not handle default values of `null`, so the `scan_resource_conf` method was getting `min_tls_version = 'var.tls'` and passing the check. This is a quick fix for that issue while we work on the actual variable rendering issue.

Note that there was a UT checking for TLS 1.3 as a future proofing thing, so I just added that in here for good measure even though it's not supported yet.


kartikp10 (approved):
👍🏽

PR #2244, opened 2022-01-21
GoogleComputeExternalIP - This check looked for the wrong place

JamesWoolfenden:
Access_config lies inside a network_config block, it may have worked like this at one time- it doesnt now.

### DIFF
--- a/checkov/terraform/checks/resource/gcp/GoogleComputeExternalIP.py
+++ b/checkov/terraform/checks/resource/gcp/GoogleComputeExternalIP.py
@@ -13,7 +13,7 @@ class GoogleComputeExternalIP(BaseResourceNegativeValueCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf) -> CheckResult:
-        if 'source_instance_template' in conf.keys() and 'network_config' not in conf.keys():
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/gcp/GoogleComputeExternalIP.py
```diff
@@ -13,7 +13,7 @@ def __init__(self):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def scan_resource_conf(self, conf) -> CheckResult:
-        if 'source_instance_template' in conf.keys() and 'access_config' not in conf.keys():
+        if 'source_instance_template' in conf.keys() and 'network_config' not in conf.keys():
```
gruebel, line 16:
```suggestion
        if 'source_instance_template' in conf.keys() and 'network_interface' not in conf.keys():
```
shouldn't be then `network_interface` too?

checkov/terraform/checks/resource/gcp/GoogleComputeExternalIP.py
```diff
@@ -13,7 +13,7 @@ def __init__(self):
...
             # if the source_instance_template value is there (indicating a google_compute_instance_from_template),
             # and the access_config block is not present, then this check cannot PASS, since we don't know what the
```
gruebel, line 18: please adjust the comment here

gruebel (approved):
🍸

PR #2249, opened 2022-01-21
sql server shouldnt be public

JamesWoolfenden:
This check was supposed to only target SQLServer which makes no sense. Can target all dbs.


nimrodkor (approved):
Agreed

PR #2251, opened 2022-01-23
fix source_module assignment for multi module references

gruebel:

Adjusted the `block_dirs_to_modules` creation to make `source_module` less ambiguous, optimally there should be only one reference now.

### DIFF
--- a/checkov/common/graph/graph_builder/graph_components/blocks.py
+++ b/checkov/common/graph/graph_builder/graph_components/blocks.py
@@ -64,8 +64,8 @@ class Block:
         self.get_origin_attributes(base_attributes)
 
         if hasattr(self, "module_dependency") and hasattr(self, "module_dependency_num"):
-            base_attributes["module_dependency_"] = self.module_dependency
-            base_attributes["module_dependency_num_"] = self.module_dependency_num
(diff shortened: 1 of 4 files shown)


checkov/common/graph/graph_builder/graph_components/blocks.py
```diff
@@ -63,6 +63,10 @@ def get_attribute_dict(self, add_hash=True) -> Dict[str, Any]:
         base_attributes = self.get_base_attributes()
         self.get_origin_attributes(base_attributes)
 
+        if hasattr(self, "module_dependency") and hasattr(self, "module_dependency_num"):
+            base_attributes["module_dependency_"] = self.module_dependency
+            base_attributes["module_dependency_num_"] = self.module_dependency_num
```
nimrodkor, line 68: @gruebel Couldn't find it, but seems like `module_dependency_` and `module_dependency_num_` should be existing consts

nimrodkor (approved):

checkov/common/graph/graph_builder/graph_components/blocks.py
gruebel: moved the attributes to the dataclass

PR #2254, opened 2022-01-23
CKV_AWS_92 : Update Failed Status

sumedhrana21:
In case , 'enabled' is not found ['access_logs'][0], the status should be FAILED instead of PASSED.

### DIFF
--- a/checkov/terraform/checks/resource/aws/ELBAccessLogs.py
+++ b/checkov/terraform/checks/resource/aws/ELBAccessLogs.py
@@ -16,7 +16,7 @@ class ELBAccessLogs(BaseResourceCheck):
         if 'access_logs' not in conf:
             return CheckResult.FAILED
         if 'enabled' not in conf['access_logs'][0]:
-            return CheckResult.PASSED
+            return CheckResult.FAILED
         if conf['access_logs'][0]['enabled'] == [True]:
             return CheckResult.PASSED
         return CheckResult.FAILED


gruebel:
@sumedhrana21 thanks for creating the PR, but why is this change needed? The default for the `enabled` property is `true`, so not setting it will enable logging, as long as you define the destination bucket.

sumedhrana21:
@gruebel Yes the default for enabled is indeed true. However, the check should FAIL ,if it doesnt find the 'enabled' key at all in conf['access_logs'][0]. 
As of now, its getting passed.

gruebel:
so, if the default is `true`, why should it fail, if you don't set it.

sumedhrana21:
No matter what the value is for `enabled` flag, if we dont find this key at all in the output of conf['access_logs'][0], I see no reason it should pass.

```
if 'enabled' not in conf['access_logs'][0]:
            return CheckResult.PASSED
```

gruebel:
let's assume we have the following example configuration
```hcl
resource "aws_elb" "example" {
  name = "example"

  subnets = [
    "subnet-123",
    "subnet-456"
  ]

  access_logs {
    bucket        = "example"
    bucket_prefix = "elb"
    interval      = 60
  }

  listener {
    instance_port     = 8000
    instance_protocol = "http"
    lb_port           = 80
    lb_protocol       = "http"
  }
}
```
currently it would pass the check, because access logging is enabled by default, so why do you want it to fail? Because that's what would happen, when we add your code change.

gruebel:
as described I don't see any reason to change the behaviour here, because access logging is enabled by default, if you add the destination bucket config.

PR #2262, opened 2022-01-24
fix CKV_GCP_56 - dont pass for all non postgres dbs

JamesWoolfenden:
also fixes CKV_GCP_53 in the same way


gruebel (approved):
thanks for cleaning up and migrating to the better test syntax :beers

PR #2279, opened 2022-01-25
remove null entries in K8s List

gruebel:

seems like you can have null values in a K8s List object 🤷


Saarett (approved):
[GJ, None, GJ]

PR #2291, opened 2022-01-27
Imagebuilder - ensure AMI disks are encrypted with KMS

JamesWoolfenden:


checkov/terraform/checks/resource/aws/ImagebuilderImageRecipeEBSEncrypted.py
```diff
@@ -0,0 +1,33 @@
...
+        name = "Ensure that Image Recipe EBS Disk are encrypted with CMK"
+        id = "CKV_AWS_200"
+        supported_resources = ["aws_imagebuilder_image_recipe"]
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
+        if conf.get('block_device_mapping'):
+            mappings = conf.get('block_device_mapping')
+            for mapping in mappings:
```
gruebel, line 19: I think you current code doesn't consider that the first block device is encrypted, but the second or any follow is not. Cann you add this test case and check, if you need to adjust your code.

gruebel (commented):
probably you also need to merge master into your branch to make the ID check pass. And congrats for creating the 200th check 🎉 🍾

checkov/terraform/checks/resource/aws/ImagebuilderImageRecipeEBSEncrypted.py
JamesWoolfenden: good catch. have it now

gruebel (approved):
🍻

PR #2294, opened 2022-01-27
CKV_DOCKER_3: Additionally check if user is created by gosu

adberger:


checkov/version.py
```diff
@@ -1 +1 @@
-version = '2.0.762'
+version = '2.0.767'
```
gruebel, line 1: can you revert your change here, this is done automatically when we release a new version

tests/dockerfile/checks/test_UserExists.py
```diff
@@ -15,14 +15,20 @@ def test(self):
...
 
-        passing_resources = {"/success/Dockerfile.USER"}
-        failing_resources = {"/failure/Dockerfile."}
+        passing_resources = {
+            "/success_user/Dockerfile.",
+            "/success_gosu/Dockerfile."
+        }
+        failing_resources = {
+            "/failure_no_user/Dockerfile.FROM",
+            "/failure_no_gosu/Dockerfile.FROM",
```
gruebel, line 24: can you also add a test case for the `gosu` command being inside the `CMD`?

checkov/dockerfile/checks/UserExists.py
```diff
@@ -6,15 +6,20 @@ class UserExists(BaseDockerfileCheck):
...
         super().__init__(name=name, id=id, categories=categories, supported_instructions=supported_instructions)
 
     def scan_entity_conf(self, conf):
-        for instruction, content in conf.items():
-            if instruction == "USER":
-                return CheckResult.PASSED, conf[instruction][0]
-        return CheckResult.FAILED, None
+        for instruction in conf:
+            value = instruction['value']
+            if instruction['instruction'] in self.supported_instructions:
```
gruebel, line 16: I'm surprised you need to check here, that the `instruction` is one of the supported ones, but maybe my memories are blurry 😄

checkov/version.py
adberger: done

tests/dockerfile/checks/test_UserExists.py
adberger: done

checkov/dockerfile/checks/UserExists.py
adberger: I copied this from another test case (`UpdateNotAlone.py`)

adberger:
@gruebel  why was this closed?

PR #2308, opened 2022-01-30
Avoiding crashing when using Pycharm's-debugger on multiprocessing.

itai1357:
Avoiding crashing when using Pycharm's-debugger on multiprocessing by using the env-variable "PYCHARM_HOSTED".
when "PYCHARM_HOSTED" is set and equal to "1", the scans are run in sequence


Saarett (approved):
Good catch! LGTM

PR #2312, opened 2022-01-31
Memorydb snapshots are encrypted

JamesWoolfenden:

### DIFF
--- a/tests/terraform/checks/resource/aws/example_MemoryDBSnapshotEncryptionWithCMK/main.tf
+++ b/tests/terraform/checks/resource/aws/example_MemoryDBSnapshotEncryptionWithCMK/main.tf
@@ -1,0 +1,24 @@
+resource "aws_memorydb_snapshot" "pass" {
+  cluster_name = aws_memorydb_cluster.example.name
+  name_prefix  = aws_memorydb_cluster.example.name
+  kms_key_arn  = aws_kms_key.example.arn
+}
+
+resource "aws_memorydb_cluster" "example" {
+  acl_name                 = "open-access"
+  name                     = "my-cluster"
+  node_type                = "db.t4g.small"
+  num_shards               = 2
+  security_group_ids       = [aws_security_group.example.id]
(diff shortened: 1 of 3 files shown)


tests/terraform/checks/resource/aws/example_MemoryDBSnapshotEncryptionWithCMK/main.tf
```diff
@@ -0,0 +1,24 @@
...
+  security_group_ids       = [aws_security_group.example.id]
+  snapshot_retention_limit = 7
+  subnet_group_name        = aws_memorydb_subnet_group.example.id
+  kms_key_arn              = aws_kms_key.example.arn
+}
+
+resource "aws_memorydb_snapshot" "fail" {
+  cluster_name = aws_memorydb_cluster.example.name
+  name_prefix  = aws_memorydb_cluster.example.name
+}
```
gruebel, line 21: I couldn't find any clear info regarding the encryption of a snapshot, if the DB is encrypted. Is it really using the default key, even you encrypted the DB with your own?

gruebel (approved):
just a clarification question

tests/terraform/checks/resource/aws/example_MemoryDBSnapshotEncryptionWithCMK/main.tf
JamesWoolfenden: uses the cluster key 
image
So I dont think we need this check?

gruebel: either remove it or how about changing it to a YAML check and check if there is a connection between a cluster and if yes then the cluster needs to have a KMS key otherwise the snapshot needs one. I'm good with both options 😄

JamesWoolfenden: ah thinks its way to niche now. Killing it with fire

PR #2316, opened 2022-02-01
Support cloned checkov checks 

rotemavni:
In this PR I've modified the `CustomPoliciesIntegration` class to support cloned checkov python checks.
These cloned checks have the same logic as their original check, but they differ in the metadata - guidelines, severity and title.

To support those check, the cloned checks are stored in the new class member `bc_cloned_checks`, and adding records of those checks to the reports being sent to the `post_runner`.


checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -72,5 +81,25 @@ def _get_policies_from_platform(self):
         policies = [p for p in policies if p['isCustom']]
         return policies
 
+    def post_runner(self, scan_reports):
+        if self.bc_cloned_checks.keys():
```
gruebel, line 85:
```suggestion
        if self.bc_cloned_checks:
```
not sure, if it makes any difference in performance, but there is no need to especially check the `keys`

checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -72,5 +81,25 @@ def _get_policies_from_platform(self):
...
+            scan_reports.failed_checks = self.extend_records_with_cloned_policies(scan_reports.failed_checks)
+            scan_reports.passed_checks = self.extend_records_with_cloned_policies(scan_reports.passed_checks)
+            scan_reports.skipped_checks = self.extend_records_with_cloned_policies(scan_reports.skipped_checks)
+
+    def extend_records_with_cloned_policies(self, records):
+        new_records = deepcopy(records)
```
gruebel, line 91: I don't think you need this line and the `deepcopy` a few lines lower is already doing what you want to achieve. Always trying to avoid `deepcopy` as much as possible 😄

gruebel: i like idempotent methods, but the performance implication can be too harsh and we overwrite the attribute anyway

rotemavni: @gruebel You're right about the implications of the deep copy, but in this case we'll run in an infinite loop if we run without copying it - 
We're adding a new record to the records list, and if we don't deep copy `records`, when adding a record to `new_records` it adds it to the  `records` list as well, which results in an infinite loop.

checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -72,5 +81,25 @@ def _get_policies_from_platform(self):
...
+            cloned_policies = self.bc_cloned_checks.get(record.bc_check_id, [])
+            for cloned_policy in cloned_policies:
+                new_record = deepcopy(record)
+                new_record.check_id = cloned_policy['id']
+                new_record.bc_check_id = cloned_policy['id']
+                new_record.guideline = cloned_policy['guideline']
+                new_record.severity = cloned_policy['severity']
+                new_record.check_name = cloned_policy['title']
+                new_records.append(new_record)
+        return new_records
```
gruebel, line 102:
```suggestion
        bc_check_ids = [record.bc_check_id for records in records]
        for idx, bc_check_id in enumerate(bc_check_ids):
            cloned_policies = self.bc_cloned_checks.get(bc_check_id, [])
            for cloned_policy in cloned_policies:
                new_record = deepcopy(records[idx])
                new_record.check_id = cloned_policy['id']
                new_record.bc_check_id = cloned_policy['id']
                new_record.guideline = cloned_policy['guideline']
                new_record.severity = cloned_policy['severity']
                new_record.check_name = cloned_policy['title']
                records.append(new_record)
        return records
```

how about something like this?

tests/common/integration_features/test_custom_policies_integration.py
```diff
@@ -242,26 +250,111 @@ def test_policy_load(self):
...
+        custom_policies_integration = CustomPoliciesIntegration(instance)
+
+        # mock _get_policies_from_platform method
+        custom_policies_integration._get_policies_from_platform = types.MethodType(_get_policies_from_platform,
+                                                                                   custom_policies_integration)
+
+        custom_policies_integration.pre_scan()
+        self.assertEqual(1, len(custom_policies_integration.policies))
+        self.assertEqual(1, len(custom_policies_integration.bc_cloned_checks))
+
```
gruebel, line 291:
```python
from checkov.common.checks_infra.registry import Registry, get_graph_checks_registry  # adjust the import in th top

        ...

        get_graph_checks_registry("cloudformation").checks = []. # this you don't actually need, because there are no CFN checks, but better to be safe than sorry
        get_graph_checks_registry("terraform").checks = []
```

ad this also at the bottom of the test under this one

gruebel (approved):
🥇

PR #2319, opened 2022-02-01
add empty check to the negative value check

### LINKED ISSUE #2318: CKV_AWS_40: failure even when not setting users
**Describe the issue**
The check fails when executing checkov on Terraform plan's JSON output.

**Examples**
```
resource "aws_iam_policy_attachment" "attachment" {
  ...
  roles = [...]
  # no users
}
```
JSON:
```
{
  "address": "aws_iam_policy_attachment.attachment",
  ...
  "values": {
    ...
    "roles": [
      "data-analytics@eng-0"
    ],
    "users": []
  },
...
```
The `users` field is set to `[]` in JSON, and the [check implementation](https://github.com/bridgecrewio/checkov/blob/e2538c48df14363d6ed46e5b838e19cc71ba6cbf/checkov/terraform/checks/resource/base_resource_negative_value_check.py#L39-L53) doesn't handle this scenario correctly:

https://github.com/bridgecrewio/checkov/blob/e2538c48df14363d6ed46e5b838e19cc71ba6cbf/checkov/terraform/checks/resource/base_resource_negative_value_check.py#L39-L53

```
> checkov --version
2.0.780
```

I saw there are tests for the check implementation, but only targeting Terraform source files. Are there tests targeting JSON output from Terraform plan?

gruebel:

Fixes #2318 

I added this in the past to the `BaseResourceValueCheck` and now it is time to treat its friend 😄

### DIFF
--- a/checkov/terraform/checks/resource/base_resource_negative_value_check.py
+++ b/checkov/terraform/checks/resource/base_resource_negative_value_check.py
@@ -42,11 +42,11 @@ class BaseResourceNegativeValueCheck(BaseResourceCheck):
             value = dpath.get(conf, inspected_key)
             if isinstance(value, list) and len(value) == 1:
                 value = value[0]
+            if value is None or (isinstance(value, list) and not value):
+                return self.missing_attribute_result
             if get_referenced_vertices_in_value(value=value, aliases={}, resources_types=[]):
                 # we don't provide resources_types as we want to stay provider agnostic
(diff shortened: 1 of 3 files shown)


paulolieuthier (approved):
Just tested locally and it fixed my issue.

rotemavni (approved):
Nice one!

PR #2320, opened 2022-02-01
Draft to showcase the new suppression feature

Muhammada3178:

This PR is meant to be a POC for a feature that we are interested in adding. This is only meant as a demonstration and will be closed soon. 

**Feature description:**
We would like to suppress by resource id, this is a useful feature that is present in other tools similar to Checkov such as Regula https://regula.dev/examples/waive-and-disable.html#writing-the-configuration-file . The advantage here is the granularity this feature provides. It allows us to be more specific with our check suppression. 

**Solution should look like:**
The solution that we came up with involves adding a key to the config YAML file. The user would add the following key to the YAML file `skip-check-by-resource-id:` and then specify the resources that they would like to suppress along with the suppression id and comment. Here is an example: 

```
skip-check-by-resource-id:
    aws_s3_bucket.data:
        - id: CKV2_AWS_6
          comment: suppress
```

Once the user has added this argument, when Checkov scans the files, if it finds a resource with the id specified then it should suppress the checks that were specified 

So in the above example, if Checkov finds a resource with the id aws_s3_bucket.data then it should suppress the CKV2_AWS_6 check for it. POC in video below

https://user-images.githubusercontent.com/56619540/152048653-58a5b16a-1599-44d4-8f97-776fdb74b347.mp4

### DIFF
--- a/checkov/common/runners/runner_registry.py
+++ b/checkov/common/runners/runner_registry.py
@@ -53,11 +53,12 @@ class RunnerRegistry:
         guidelines: Optional[Dict[str, str]] = None,
         collect_skip_comments: bool = True,
         repo_root_for_plan_enrichment: Optional[List[Union[str, os.PathLike]]] = None,
+        resource_suppressions = None,
     ) -> List[Report]:
         integration_feature_registry.run_pre_runner()
         if len(self.runners) == 1:
             reports = [self.runners[0].run(root_folder, external_checks_dir=external_checks_dir, files=files,
-                                           runner_filter=self.runner_filter, collect_skip_comments=collect_skip_comments)]
+                                           runner_filter=self.runner_filter, collect_skip_comments=collect_skip_comments, resource_suppressions=resource_suppressions)]
         else:
             reports = parallel_runner.run_function(
                 lambda runner: runner.run(root_folder, external_checks_dir=external_checks_dir, files=files,
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -58,10 +58,7 @@ logging_init()
 logger = logging.getLogger(__name__)
 checkov_runners = [value for attr, value in CheckType.__dict__.items() if not attr.startswith("__")]
 
-DEFAULT_RUNNERS = (tf_graph_runner(), cfn_runner(), k8_runner(),
-                   sls_runner(), arm_runner(), tf_plan_runner(), helm_runner(),
-                   dockerfile_runner(), secrets_runner(), json_runner(), github_configuration_runner(),
-                   gitlab_configuration_runner(), kustomize_runner(), sca_package_runner())
+DEFAULT_RUNNERS = [(tf_graph_runner())]
 
 
 def run(banner=checkov_banner, argv=sys.argv[1:]):
@@ -244,7 +241,8 @@ def run(banner=checkov_banner, argv=sys.argv[1:]):
     elif config.file:
         scan_reports = runner_registry.run(external_checks_dir=external_checks_dir, files=config.file,
                                            guidelines=guidelines,
-                                           repo_root_for_plan_enrichment=config.repo_root_for_plan_enrichment)
+                                           repo_root_for_plan_enrichment=config.repo_root_for_plan_enrichment,
+                                           resource_suppressions = config.skip_check_by_resource_id)
         if baseline:
             baseline.compare_and_reduce_reports(scan_reports)
         if config.create_baseline:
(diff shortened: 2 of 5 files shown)


gruebel:
thanks for creating the POC.

A few things to consider
- how do you want to differentiate between resource IDs, which are in different files, ex modules or folders?
- how would you pass this config as a CLI flag?
- as far as I understand you want to leverage it to skip resources in a `for_each` or `count` block, which wouldn't be possible with your current implementation.

Muhammada3178:
- To address point 1 and 3, I'd be happy to add the filepath as an identifier as well. So all together the resource would be identified with the filepath + the resource id. This would allow for much more granularity than is currently offered by Checkov. Not including the file path would still limit the check suppress to a named resource, rather than skipping a check all together

- I got the inspiration for all of these ideas, including the filepath from Regula https://regula.dev/configuration.html#waiving-rule-results . Their docs are a good reference point to show how the idea works in practice.  

- To the second point, that's one that I do need to ponder more upon. We would need to find a more concise flag for it, because I feel that specifying a filepath + resource id + the check id + the suppression comment all in the CLI is not user friendly as it would lead to very long CLI commands. 

- So unfortunately I don't have a good solution for that yet but I'm happy to think about that some more and definitely open to feedback on that as well if anyone else has any ideas

gruebel:
the first part you can already achieve today with the help of the `baseline` feature https://bridgecrew.io/blog/checkov-prioritize-new-misconfigurations-infrastructure-as-code/ so I currently don't see the benefit of your implementation. I would rather recommend to take a look, how it is implemented and extend that functionality instead of creating a new way.

Muhammada3178:
@gruebel Looking into that functionality right now, does that also work with the `--repo-root-for-plan-enrichment` flag? 

In that, if a resource suppressor is present in my baseline file and then I reference the baseline file along with the root folder for the HCL files, will the suppression still work? 

As an example, currently if we hardcode a suppression comment in the tf file and then run the following command 
`checkov -f ${plan file} --repo-root-for-plan-enrichment ${root path of tf files} `then the suppression works for the plan file as well 

So I'm wondering if a command like 
`checkov -f ${plan file} --baseline ${baseline file} --repo-root-for-plan-enrichment ${root path of tf files} ` would still suppress the checks outlined in the baseline file for the plan 

I was trying to test this on my local machine and was running into an issue so I wanted to double check

Muhammada3178:
To further specify on our specific use case after experimenting with the baseline feature: 

In the case where we have a for_each like this: 

```
 resource "aws_iam_policy" "test" {

  for_each = fileset("policies/", "*.yaml")
  name = replace(basename(each.value), ".yaml", "")
  policy = jsonencode(yamldecode(file("policies/${each.value}")))
}
```

The plan output will have an identifier like this: 

```
"resources": [
                {
                    "address": "aws_iam_policy.test[\"test.yaml\"]",
                    "type": "aws_iam_policy",
                    "name": "test",
                    "index": "test.yaml",
```

Where the resource id will be the type + name + index, so in the above example the resource id should be `aws_iam_policy.test["test.yaml"]`

However when Checkov is scanning the resource, the results of the scan show that only the resource `aws_iam_policy.test` was scanned. I was able to use the baseline feature to filter out all these resources, but not a specific resource. 

Is there any current feature that ensures that Checkov scans the full resource id and which allows us to configure checks that we would like to skip by the full resource id? 

As an example of the difference between a resource that was created with a `for_each` and without: 

non for_each resource:
```
  {
          "address": "aws_iam_role_policy_attachment.admin",
          "mode": "managed",
          "type": "aws_iam_role_policy_attachment",
          "name": "admin",
          "provider_name": "registry.terraform.io/hashicorp/aws",
          "schema_version": 0,
          "values": {
            "id": "admin-arn:aws:iam::aws:policy/AdministratorAccess",
            "policy_arn": "arn:aws:iam::aws:policy/AdministratorAccess",
            "role": "admin"
          }
        }
```

for_each resource: 

```
{
          "address": "aws_iam_role_policy_attachment.policy-attach[\"test-role.yaml=arn:aws:iam::aws:policy/AdministratorAccess\"]",
          "mode": "managed",
          "type": "aws_iam_role_policy_attachment",
          "name": "policy-attach",
          "index": "test-role.yaml=arn:aws:iam::aws:policy/AdministratorAccess",
          "provider_name": "registry.terraform.io/hashicorp/aws",
          "schema_version": 0,
          "values": {
            "policy_arn": "arn:aws:iam::aws:policy/AdministratorAccess",
            "role": "test-role"
          }
       }
```

The baseline feature is pretty awesome so thank you for letting us know about it. The only concern that comes to mind is that feature seems to have been created to specifically show new misconfigs rather than as a tool where you can specify suppressors by resource (though it is able to do that as well). So while it works to a certain extent, as it currently stands, it would have to be used in a roundabout way which isn't ideal.

Perhaps the best path forward is to extend the baseline feature to make it better suited for the task of resource specific suppression and to accept the full resource id 

Curious to hear your thoughts about this 

The video below further specifies the example in a visual way incase anything above was a little vague 

https://user-images.githubusercontent.com/56619540/152258447-c705058d-b04b-4746-be54-8cfe53e72143.mp4

PR #2322, opened 2022-02-02
Add user agent to HTTP requests headers

rotemavni:
Extracted the `User-Agent` from the `SIGNUP_HEADER` and created a method in `http_utils` that retrieve this header.

Then we can merge existing headers with this `User-Agent` header.

### DIFF
--- a/checkov/common/util/http_utils.py
+++ b/checkov/common/util/http_utils.py
@@ -43,7 +43,7 @@ def get_version_headers(client, client_version):
 
 
 def get_user_agent_header():
-    return {'User-Agent': 'Mozilla/5.0 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36'}
+    return {'User-Agent': f'checkov/{checkov_version}'}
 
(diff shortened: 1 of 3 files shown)


checkov/common/util/http_utils.py
```diff
@@ -42,9 +42,13 @@ def get_version_headers(client, client_version):
     }
 
 
+def get_user_agent_header():
+    return {'User-Agent': 'Mozilla/5.0 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36'}
```
gruebel, line 46: just thinking, if we should change it to something, which doesn't look like a fake agent 😄 something like 
```suggestion
    return {'User-Agent': f'checkov/{checkov_version}'}
```

gruebel (approved):
🏆

checkov/common/util/http_utils.py
rotemavni: If it works - let's do it

PR #2323, opened 2022-02-02
Cloudformation graph encryption checks

achiar99:

In this PR we want to add encryption checks for CloudFormation resources inside the graph


checkov/cloudformation/graph_builder/local_graph.py
```diff
@@ -335,6 +339,22 @@ def _create_edge(self, origin_vertex_index: int, dest_vertex_index: int, label:
             self.out_edges[origin_vertex_index].append(edge)
             self.in_edges[dest_vertex_index].append(edge)
 
+
+    def calculate_encryption_attribute(self) -> None:
+        for vertex_index in self.vertices_by_block_type.get(BlockType.RESOURCE, []):
+            vertex = self.vertices[vertex_index]
+            resource_type =  vertex.id.split(".")[0]
```
Saarett, line 345: If I'm not mistaken we could fetch the resource_type from `vertex.attributes.resource_type`, it would be a more stable approach, because if I recall correctly, the resource type might not always be in `vertex.id.split(".")[0]`

checkov/cloudformation/graph_builder/local_graph.py
```diff
@@ -335,6 +339,22 @@ def _create_edge(self, origin_vertex_index: int, dest_vertex_index: int, label:
...
+            encryption_conf = ENCRYPTION_BY_RESOURCE_TYPE.get(resource_type)
+            if encryption_conf:
+                attributes = vertex.get_attribute_dict()
+                is_encrypted, reason = encryption_conf.is_encrypted(attributes)
```
Saarett, line 349: Can we turn the `ENCRYPTION_BY_RESOURCE_TYPE` map into a manager class that would handle this logic?

checkov/terraform/graph_builder/graph_components/generic_resource_encryption.py
```diff
@@ -1,56 +1,22 @@
-from typing import Dict, List, Union, Tuple, Any
-
+from typing import Dict, List, Union
 from checkov.common.graph.graph_builder import EncryptionTypes
+from checkov.common.graph.graph_builder.graph_components.generic_resource_encryption_base import GenericResourceEncryptionBase
 
 
-class GenericResourceEncryption:
+class GenericResourceEncryption(GenericResourceEncryptionBase):
```
Saarett, line 6: Nice

checkov/cloudformation/graph_builder/graph_components/generic_resource_encryption.py
```diff
@@ -0,0 +1,118 @@
...
+                         attribute_values_map,
+                         enabled_by_default,
+                         node_to_node_encryption="NodeToNodeEncryptionOptions")
+        if self.resource_type.startswith("AWS::"):
+            self.default_description = EncryptionTypes.DEFAULT_KMS.value
+
+
+# This map allows dynamically creating the check for each resource type based on GenericResourceEncryption.
+# Please check out the constructor to understand all the edge cases.
+ENCRYPTION_BY_RESOURCE_TYPE = {
```
Saarett, line 23: Should we turn all the resources described below into constants/Enum values?

Saarett (commented):
Great job! A few comments in the code :)
I think we better create a test for yaml templates too, it might give different results from json templates in some cases

checkov/cloudformation/graph_builder/local_graph.py
```diff
@@ -335,6 +339,21 @@ def _create_edge(self, origin_vertex_index: int, dest_vertex_index: int, label:
             self.out_edges[origin_vertex_index].append(edge)
             self.in_edges[dest_vertex_index].append(edge)
 
+    def calculate_encryption_attribute(self) -> None:
+        for vertex_index in self.vertices_by_block_type.get(BlockType.RESOURCE, []):
```
rotemavni, line 343: @achiar99 I think we would benefit from refactoring this method to the base `LocalGraph` class, since its logic is the same in terraform and cloudformation.

rotemavni (commented):
@achiar99 You did an amazing work in identifying the common logic and extracting it to a base class.
I also loved your UT. 👏 👏 👏 👏 

Had only one comment on moving the `calculate_encryption_attribute` method to the base `LocalGraph`

checkov/cloudformation/graph_builder/local_graph.py
achiar99: @Saarett  the problem with that is vertex.attributes.get('resource_type') is a list. so im not sure how can i use it here

gruebel:
@achiar99 please merge the master branch into yours to get the workflow adjustments

checkov/cloudformation/graph_builder/local_graph.py
achiar99: @rotemavni pls review the next commits

rotemavni (approved):

PR #2324, opened 2022-02-02
remove pipenv version

gruebel:

replace the outdated pipenv action by just installing it manually, looks like that installing `pipenv` changed with new Win Server version 2022. Other actions in the marketplace are doing the same, so I don't see any need to use them at the moment.


rotemavni (approved):
🍾 🍾 🍾 🍾

PR #2342, opened 2022-02-05
Add tf plan ref

### LINKED ISSUE #2327: CKV2_AWS_6: invalid result when executed over a Terraform plan
**Describe the issue**
The check works when executed over the Terraform source file. If executed over a Terraform plan, the `aws_s3_bucket_public_access_block`  is not detected.

**Examples**
```
resource "aws_s3_bucket" "example" {
  bucket = "example"
  ...
}

resource "aws_s3_bucket_public_access_block" "example" {
  bucket = aws_s3_bucket.example.id

  block_public_acls       = true
  block_public_policy     = true
}
```

**Version**
```
> checkov --version
2.0.782
```

gruebel:

Fixes #2327

I added an extra attribute for the references, which are only available in a TF plan. This is then leveraged for the graph checks to make the connection. The previous approach only added missing attributes and now it will add a reference if available.

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -3,6 +3,7 @@ import itertools
 import json
 import sys
 from collections import defaultdict
+from collections.abc import Iterable
 from dataclasses import dataclass
 from typing import List, Dict, Union, Any, Optional, Set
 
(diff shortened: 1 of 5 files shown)


schosterbarak:
@gruebel interesting. does that mean we support connection queries now?

gruebel:
we supported them already before, but now I added the missing ones. If I'm not wrong, then these should be all connections, which are possible in a TF plan file.

schosterbarak:
sweet!

PR #2349, opened 2022-02-06
Using new endpoint to report all sca findings at once

itai1357:
* Using new endpoint to report all sca findings at once
* creating the method 'report_results_to_bridgecrew' for supporting commiting news-sca-scanner's results out 
* refactoring sca-package-runner's run method for enabling getting directly the scan-results as dict (rather than Report object) and adding test for the relevant new method 'prepare_and_scan'


checkov/sca_package/runner.py
```diff
@@ -16,55 +16,72 @@
...
 
         # skip complete run, if flag '--check' was used without a CVE check ID
         if runner_filter.checks and all(not check.startswith("CKV_CVE") for check in runner_filter.checks):
-            return report
+            return None
 
         if not bc_integration.bc_api_key:
             logging.info("The --bc-api-key flag needs to be set to run SCA package scanning")
-            return report
+            return None
```
Saarett, line 39: Why have we changed those to `None`?

Saarett, line 39: Why have we changed those to `None`?

nimrodkor (approved):
Nice, check out @Saarett 's comments :)

checkov/sca_package/runner.py
itai1357: I moved this logic from the method "run" into inner method called "prepare_and_scan"
all those cases represents cases where the scan can't be executed, so we return empty report finally.
when it was in "run", we should return "Report" object, so this is the old behvior.
now, I use the internal method "prepare_and_scan" which do the preparation and the scan itself, but the convertion to reports still happens in "run". returning None letting the external method that scan wasn't executed, and then "run" will return empty "Report"

itai1357: same as above

tests/sca_package/test_runner.py
```diff
@@ -113,6 +113,36 @@ def test_run_with_skip(mocker: MockerFixture, scan_result):
     assert record.check_id == "CKV_CVE_2020_29652"
 
 
+def test_prepare_and_scan(mocker: MockerFixture, scan_result):
```
gruebel, line 116: can you separate the test and make two out of it

itai1357: sure!

PR #2354, opened 2022-02-07
check is mq broker is encrypted

JamesWoolfenden:
Check undocumented encryption parameters are set


gruebel (approved):
🚀 let's merge the other one first and then we can do this one 😄

PR #2356, opened 2022-02-07
Batch job does not describe a privileged container

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/aws/BatchJobIsNotPrivileged.py
+++ b/checkov/terraform/checks/resource/aws/BatchJobIsNotPrivileged.py
@@ -17,9 +17,8 @@ class BatchJobIsNotPrivileged(BaseResourceCheck):
                 container = json.loads(conf.get("container_properties")[0])
             else:
                 container = conf.get("container_properties")[0]
-            if 'privileged' in container:
-                if container['privileged']:
-                    return CheckResult.FAILED
+            if container.get("privileged"):
+                return CheckResult.FAILED
             return CheckResult.PASSED
         return CheckResult.UNKNOWN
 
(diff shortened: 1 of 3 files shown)


tests/terraform/checks/resource/aws/example_BatchJobIsNotPrivileged/main.tf
```diff
@@ -0,0 +1,129 @@
...
+
+
+
+resource "aws_batch_job_definition" "unknown" {
+  name = "tf_test_batch_job_definition"
+  type = "container"
+}
+
+
+resource "aws_batch_job_definition" "pass2" {
```
gruebel, line 92: could you add one passing test which leverages `jsonencode()` instead of heredoc to define the container definition, just want to be sure, that the check handle it 🙂

checkov/terraform/checks/resource/aws/BatchJobIsNotPrivileged.py
```diff
@@ -0,0 +1,24 @@
...
+        supported_resources = ['aws_batch_job_definition']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if conf.get("container_properties"):
+            container = json.loads(conf.get("container_properties")[0])
+            if 'privileged' in container:
+                if container['privileged']:
+                    return CheckResult.FAILED
```
gruebel, line 19:
```suggestion
            if container.get("privileged"):
                return CheckResult.FAILED
```
this will do the same job 😄

gruebel (commented):
looks great, especially the extra test would be great

tests/terraform/checks/resource/aws/example_BatchJobIsNotPrivileged/main.tf
JamesWoolfenden: fixed it. good call.

checkov/terraform/checks/resource/aws/BatchJobIsNotPrivileged.py
JamesWoolfenden: ta, resolved

gruebel (approved):
nice, thanks for adding the `jsonencode` test 💪

PR #2362, opened 2022-02-08
Removing the old packages-scanner

itai1357:
* Removing the old sca-packages scanner and its tests
* creating tests to the new sca-packages scanner, based on the old-scanner's tests


gruebel (approved):
noice 🍻

PR #2363, opened 2022-02-08
Fix colored example

duncan485:
The colored example does not generate a junit xml file. Uploading the artifact will fail for this pipeline.


gruebel (approved):
thanks, your are right!

PR #2368, opened 2022-02-08
Adds support for aws_wafv2_web_acl_association in CKV2_AWS_29

### LINKED ISSUE #2367: CKV2_AWS_29 should support aws_wafv2_web_acl_association
The check "Ensure public API gateway are protected by WAF" only supports the association if it is done using the `aws_wafregional_web_acl_association`. Considering that the rule description in bridgecrew mentions WAFv2, it should support `aws_wafv2_web_acl_association`.

https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/graph_checks/aws/APIProtectedByWAF.yaml#L36

BenjaDiaz:
Fixes #2367

### DIFF
--- a/tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
+++ b/tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
@@ -62,7 +62,18 @@ resource "aws_api_gateway_stage" "pass" {
   stage_name    = "example"
 }
 
+resource "aws_api_gateway_stage_wafv2" "pass" {
+  deployment_id = aws_api_gateway_deployment.example.id
+  rest_api_id   = aws_api_gateway_rest_api.pass.id
+  stage_name    = "example"
+}
+
 resource "aws_wafregional_web_acl_association" "pass" {
   resource_arn = aws_api_gateway_stage.pass.arn
   web_acl_id   = aws_wafregional_web_acl.foo.id
+}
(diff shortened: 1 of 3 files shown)


tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
```diff
@@ -62,7 +62,18 @@ resource "aws_api_gateway_stage" "pass" {
   stage_name    = "example"
 }
 
+resource "aws_api_gateway_stage_wafv2" "pass" {
```
gruebel, line 65:
```suggestion
resource "aws_api_gateway_stage" "wafv2_pass" {
```
that's what you probably meant, right?

gruebel (commented):
and then you need to adjust the name for the references

tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
gruebel: and don't forget to change it in `expected.yml` file 😄

BenjaDiaz: Hahaha, you are totally right. Sorry! Fixing it now.

gruebel (approved):
Thanks for the contribution 🎉

Issue #2298, opened 2022-01-28
Checkov showing non-compliance of certain parameters even with right expected values.


charanreddypruthvi:
**Describe the issue**
CKV_AZURE_41/Ensure that the expiration date is set on all secrets.
This check should pass if we provide the expiration date of Key Vault secret which is not happening .

**Examples**
Below is the default setting of the expiration date on secrets that are being created in Azure Kay Vault.
expiration_date = timeadd(formatdate("YYYY-MM-DD'T'00:00:00Z", timestamp()), "2191h")

**Version (please complete the following information):**
 - Checkov Version - 2.0.679

**Additional context**

nimrodkor:
I believe the issue is we don't support these operators yet (`timeadd`, `formatdate` etc) - hence checkov doesn't know how to handle it

PR #2375, opened 2022-02-09
2298/add terraform date functions

MooreDerek:

This a part solution for #2298. Providing the TF timeadd() and timestamp() date functions. Relatively new to Pyhton, so looking for feedback on whether this is the best approach.


nimrodkor (approved):
Brilliant @MooreDerek , thank you!

elvis2:
@MooreDerek Thanks for the contribution.

Do you have any docs on writing Checkov rules for your contribution?

I have a need to check a tag that has a timestamp in a string format (from TF's timestamp() function). I would like to ask Check if the date is older than 90 days (now() - tag.datetime_string) > 90 days.

I

PR #2376, opened 2022-02-09
Only validate api_server_authorized_ip_ranges if cluster is not private

duncan485:
The api_server_authorized_ip_ranges can and should only be set if private_cluster_enabled is false. Azure won't allow setting the api_server_authorized_ip_ranges if the cluster is private.

*This is my first experience with writing a policy, any notes or approvements are welcome!*

### DIFF
--- a/checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
+++ b/checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
@@ -1,7 +1,7 @@
 from typing import List, Any
 
 from checkov.common.models.consts import ANY_VALUE
-from checkov.common.models.enums import CheckCategories
+from checkov.common.models.enums import CheckCategories, CheckResult
 from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
```diff
@@ -1,23 +1,23 @@
...
-        return [ANY_VALUE]
-
+    def scan_resource_conf(self, conf):
+        private_cluster_enabled = conf.get('private_cluster_enabled',[False])[0]
+        api_server_authorized_ip_ranges =  conf.get('api_server_authorized_ip_ranges',[[]])[0]
+     
+        if not private_cluster_enabled:
+            if api_server_authorized_ip_ranges == []:
+                return CheckResult.FAILED
+        return CheckResult.PASSED
```
gruebel, line 21:
```suggestion
    def scan_resource_conf(self, conf):
        # can't be set for private cluster
        private_cluster_enabled = conf.get('private_cluster_enabled',[False])[0]
        if private_cluster_enabled:
            return CheckResult.PASSED
        return super().scan_resource_conf(conf)
```
you can something like this and leave the old code like it was 🙂

gruebel (commented):
nice job 💪 and please add a test config for a private cluster

gruebel:
Tests for this check go here:
https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/azure/example_AKSApiServerAuthorizedIpRanges/main.tf
https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/checks/resource/azure/test_AKSApiServerAuthorizedIpRanges.py

checkov/terraform/checks/resource/azure/AKSApiServerAuthorizedIpRanges.py
duncan485: Thanks, added the scan_resource_conf() as you suggested!

Example and test are also updated

gruebel (approved):
thanks, great work 🥇

PR #2378, opened 2022-02-09
add extra info to the report and add flag to enable package.json scan

gruebel:

- add extra info to the SCA report
- add flag to enable scanning of `package.json` even, if the lock file exists


YaaraVerner (approved):
Good job!

PR #2390, opened 2022-02-10
Set Github org using env variable

rotemavni:
Set the github org name by setting the `GITHUB_ORG` environment variable.


checkov/github/dal.py
```diff
@@ -72,7 +73,7 @@ def get_organization_security(self):
                         }
                     }
                 }
-                """, variables={'org': 'bridgecrewio'})
+                """, variables={'org': self.org})
```
gruebel, line 76: does anything bad happen, when this is an empty string?

rotemavni: Nope, it just doesn't fetch any data

PR #2402, opened 2022-02-11
prevent pointless destroy of Key materials

JamesWoolfenden:
# "From the Provider Documentation"
        # CryptoKeys cannot be deleted from Google Cloud Platform.Destroying a Terraform - managed CryptoKey will remove
        # it from state and delete all CryptoKeyVersions, rendering the key unusable, but will not delete the resource
        # from the project.When Terraform destroys these keys, any data previously encrypted with these keys will be
        # irrecoverable.For this reason, it is strongly recommended that you add lifecycle hooks to the resource to
        # prevent accidental destruction.


gruebel (approved):
🍺

PR #2406, opened 2022-02-11
add tests covering Prisma integrations

mikeurbanski1:

I became worried that our logic for matching --repo-id with platform suppressions did not properly account for Prisma org names (all numeric). There is no issue, but I added some tests to cover it and did a small refactor so that we do not recompile a regex repeatedly.


kartikp10 (approved):
LGTM 👑

nimrodkor (approved):

PR #2410, opened 2022-02-13
Change git config files result to relative path

achiar99:

In this pr we set the git configuration resource (resources that coming from github api) to be relative


checkov/json_doc/runner.py
```diff
@@ -85,3 +85,9 @@ def import_registry(self):
 
     def require_external_checks(self):
         return True
+
+    @staticmethod
+    def _change_files_path_to_relative(report: Report):
+        for record in report.get_all_records():
+            record.file_path = record.file_path.replace(os.getcwd(), '')
+            record.resource = record.resource.replace(os.getcwd(), '')
```
Saarett, line 93: Is it OS agnostic?

Saarett (commented):
Good job! A question in the code :)

checkov/json_doc/runner.py
achiar99: @Saarett  what do u mean?

Saarett: Would it yield the same result for different operating systems? Windows, Linux....

Saarett: And what if the user would run it locally? Or is it not possible?

achiar99: it is possible. when the user run checkov cli. but in both cases `os.getcwd()` will be the same

Saarett (approved):

PR #2411, opened 2022-02-13
add tf azure check for ACR admin account disabled

arock95:

Custom Policy id: CKV_AZURE_137

Custom Policy name: Ensure ACR admin account is disabled

Custom Policy IaC type: Terraform

Custom Policy type and provider: AzureRM

IaC configuration documentation (If available): https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/container_registry#admin_enabled

Any additional information that would help other members to better understand the check:
By default the setting is disabled but can be explicitly enabled.  The security ramifications of enabling this are outlined in https://docs.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#admin-account

### DIFF
--- a/checkov/terraform/checks/resource/azure/ACRAdminAccountDisabled.py
+++ b/checkov/terraform/checks/resource/azure/ACRAdminAccountDisabled.py
@@ -1,0 +1,21 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class ACRAdminAccountDisabled(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure ACR admin account is disabled"
+        id = "CKV_AZURE_137"
+        supported_resources = ['azurerm_container_registry']
+        categories = [CheckCategories.IAM]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/azure/ACRAdminAccountDisabled.py
```diff
@@ -0,0 +1,21 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class ACRAdminAccountDisabled(BaseResourceCheck):
```
gruebel, line 5: I recommend to use the `BaseResourceNegativeValueCheck` as base class, then `checkov` can leverage additional logic to make the `check` more resilient, here is an example https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/aws/MemoryDBClusterIntransitEncryption.py

gruebel (commented):
thanks, nice addition, I added a recommendation

checkov/terraform/checks/resource/azure/ACRAdminAccountDisabled.py
arock95: Thanks for the recommendation and example, I can make another commit if that's the best way forward.

gruebel (approved):
Thanks and great work with the new check 🥇

PR #2418, opened 2022-02-14
Pyright GitHub action

achiar99:

In this pr i added pyright github action to avoid typing and runtime errors.
We need to remove the folders in the `exclude` parameter in pyproject.tomb and make all the folders to pass


gruebel:
any specific goal you want to achieve by adding this action to the workflow?

PR #2427, opened 2022-02-14
PAN-OS resource check for http in mgmt profile

jamesholland-uk:

First resource check for PAN-OS Terraform provider. Ensures that plain-text HTTP is not enabled for managing PAN-OS firewalls, we only want to see HTTPS being used for firewall management.


gruebel (approved):
nice job 🥇

PR #2428, opened 2022-02-14
Panos telnet mgmt service check

jamesholland-uk:

Another resource check for PAN-OS Terraform provider for management, almost identical to the previous one for HTTP. This one ensures that plain-text Telnet is not enabled for managing PAN-OS firewalls, we only want to see SSH being used for firewall CLI management.


gruebel (approved):
nice 🚀

PR #2434, opened 2022-02-14
CKV_GCP_89 - Public vertex ai notebook check

jdyke:

This PR adds a new GCP check for public Vertex AI instances. These instances are the underlining compute that powers the notebooks and are __by default public__. 

The check logic and tests accounts for three scenarios:

- If the `no_public_ip` setting does not exist - Fail
- If the `no_public_ip` setting exists but is set to `false` - Fail
- If the `no_public_ip` setting exists and is set to `true` - Pass


gruebel (approved):
🍻

PR #2437, opened 2022-02-15
fix CKV_AWS_8

gruebel:

Fixed an issue in CKV_AWS_8, when using the old way of defining a block and it is empty by default


YaaraVerner (approved):
Looks great!

PR #2448, opened 2022-02-15
CKV_GCP_95 - Ensure Memorystore for Redis has AUTH enabled

jdyke:

This PR adds a GCP terraform check for "Ensure Memorystore for Redis has AUTH enabled". The AUTH configuration for Memorystore for redis is not enabled by default, but when enabled, "incoming client connections must authenticate in order to connect". We generally recommend this feature be enabled to limit who/what can connect to redis databases.

This check accounts for 3 scenarios:

- If `auth_enabled` does not exist - FAIL
- If `auth_enabled` exists but is set to `false` - FAIL
- If `auth_enabled` exists but is set to `true` - PASS

You can read more about the AUTH feature [here](https://cloud.google.com/memorystore/docs/redis/auth-overview)


gruebel (approved):
thanks, nice check 🍸

PR #2472, opened 2022-02-17
New yaml operator jsonpath_equals

matansha:

add new yaml operator that search attribute via jsonpath.
for example: 
```
- cond_type: "attribute"
  resource_types:
    - "aws_security_group"
  attribute: "ingress[?(@.to_port == 22 & @.from_port == 22)].cidr_blocks"
  operator: "jsonpath_equals"
  value: 0.0.0.0/0
```


checkov/common/checks_infra/solvers/attribute_solvers/jsonpath_equals_attribute_solver.py
```diff
@@ -0,0 +1,35 @@
...
+            for match in parsed_attr.find(vertex):
+                full_path = str(match.full_path)
+                if full_path not in vertex:
+                    vertex[full_path] = match.value
+
+                attribute_matches.append(full_path)
+
+            return self.resource_type_pred(vertex, self.resource_types) and any(
+                self._get_operation(vertex=vertex, attribute=attr) for attr in attribute_matches
+            )
```
gruebel, line 31: since we can get several possible matches, this could lead into a misinterpretation. You only match one of the possible matches and don't care about the rest, right?

tests/terraform/graph/checks_infra/attribute_solvers/jsonpath_equals_solver/test_solver.py
```diff
@@ -0,0 +1,29 @@
...
+
+from tests.terraform.graph.checks_infra.test_base import TestBaseSolver
+
+TEST_DIRNAME = os.path.dirname(os.path.realpath(__file__))
+
+
+class TesJsonpathEqualsSolver(TestBaseSolver):
+    def setUp(self):
+        self.checks_dir = TEST_DIRNAME
+        super(TesJsonpathEqualsSolver, self).setUp()
```
gruebel, line 11:
```suggestion
        super().setUp()
```
no need to set it here, it will automatically the parent classes `setUp`

tests/terraform/graph/checks_infra/attribute_solvers/jsonpath_equals_solver/test_solver.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+from tests.terraform.graph.checks_infra.test_base import TestBaseSolver
+
+TEST_DIRNAME = os.path.dirname(os.path.realpath(__file__))
+
+
+class TesJsonpathEqualsSolver(TestBaseSolver):
```
gruebel, line 8:
```suggestion
class TestJsonpathEqualsSolver(TestBaseSolver):
```
missed a t 😉

tests/terraform/graph/checks_infra/attribute_solvers/jsonpath_equals_solver/test_solver.py
```diff
@@ -0,0 +1,29 @@
+import os
+
+from tests.terraform.graph.checks_infra.test_base import TestBaseSolver
```
gruebel, line 3: I'm not 100% sure, if the checks you add are also cleaned up, so they don't trigger extra results. in other tests, need to double check

checkov/common/checks_infra/solvers/attribute_solvers/jsonpath_equals_attribute_solver.py
GiladMark: @gruebel I think the right thing to look is to fail a policy if at least one ingress rule is not compliant, no?

gruebel: i would also think so, that's why it should be `all` not `any` in the code. so whatever the `jsonpath` syntax find, all of them should be equal to value

tests/terraform/graph/checks_infra/attribute_solvers/jsonpath_equals_solver/test_solver.py
matansha: fixed

tests/terraform/graph/checks_infra/attribute_solvers/jsonpath_equals_solver/test_solver.py
matansha: fixed

checkov/common/checks_infra/solvers/attribute_solvers/jsonpath_equals_attribute_solver.py
matansha: i checked with Gilad, and it should be all, fixed

gruebel (approved):
great job 🚀

PR #2473, opened 2022-02-17
fix in cfn variable renderring

YaaraVerner:
go over a copy of the vertex attribute for not getting runtime error of changing the dictionary size during iteration.


gruebel (approved):
this will make the scans slower 😢 but can't be changed

PR #2481, opened 2022-02-18
Add Subset operator

mikeurbanski1:

Adds a pair of operators: "subset" and "not_subset". These will be useful for things like IP address allowlists (or any allowlist).


checkov/common/checks_infra/solvers/attribute_solvers/subset_attribute_solver.py
```diff
@@ -0,0 +1,28 @@
...
+        super().__init__(resource_types=resource_types,
+                         attribute=attribute, value=SubsetAttributeSolver.to_set(value))
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:
+        attr_val = SubsetAttributeSolver.to_set(vertex.get(attribute))
+        return attr_val.issubset(self.value)
+
+    @staticmethod
+    def to_set(value: Any) -> Set[Any]:
+        if isinstance(value, Set):
```
gruebel, line 21:
```suggestion
        if isinstance(value, set):
```

checkov/common/checks_infra/solvers/attribute_solvers/subset_attribute_solver.py
```diff
@@ -0,0 +1,28 @@
...
+            return value
+        elif isinstance(value, (list, dict)):
+            return set(value)
+        else:
+            s = set()
+            s.add(value)
+            return s
```
gruebel, line 28:
```suggestion
            return {value}
```

tests/terraform/graph/resources/arrays/main.tf
```diff
@@ -0,0 +1,46 @@
+resource "x" "pass1" {
+  arr = [
+    "allowed1"
+  ]
+}
+
+resource "x" "pass2" {
+  arr = [
+  ]
+}
```
gruebel, line 10: not sure about this one, is no value a subset of a list of values? mathematically it is correct, but does it make sense for the user? @JamesWoolfenden any thoughts?

gruebel (commented):
nice, great addition

tests/terraform/graph/resources/arrays/main.tf
mikeurbanski1: Well, it's either this or return a mathematically incorrect value

gruebel (approved):
🍻

PR #2483, opened 2022-02-18
Postgres would like logs retained

JamesWoolfenden:


gruebel:
I'm thinking, if it would make more sense to create a YAML connection check for making sure, that to each PostgreSQL server a log configuration is attached, wdyt?

gruebel (approved):
🍺

PR #2486, opened 2022-02-18
Panos policy service any check

jamesholland-uk:

A resource check for PAN-OS Terraform provider for security policy rules. This one ensures that 'services' is not set to 'any' which is overly permissive. The 'services' attribute can be set in two different resource types, both are covered by this check.


gruebel (approved):
thanks, nice job 🥇

PR #2479, opened 2022-02-17
Add aws_apigatewayv2_domain_name as supported resource for CKV2_AWS_23

moosamus:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2488, opened 2022-02-18
Fix Route53 filter to include aws_alb as well

rafaljanicki:
Currently, the filter only checks for aws_lb and aws_elb. But aws_alb is a correct resource name as well as it's an alias for aws_lb

### DIFF
--- a/tests/terraform/graph/checks/resources/Route53ARecordAttachedResource/main.tf
+++ b/tests/terraform/graph/checks/resources/Route53ARecordAttachedResource/main.tf
@@ -97,14 +97,22 @@ resource "aws_route53_record" "pass4" {
   }
 }
 
+resource "aws_alb" "example" {
+  name               = "example"
+  internal           = false
+  load_balancer_type = "application"
+  security_groups    = [aws_security_group.lb_sg.id]
+  subnets            = [for subnet in aws_subnet.public : subnet.id]
+}
+
 resource "aws_route53_record" "pass_alb" {
   zone_id = data.aws_route53_zone.example.zone_id
   name    = "example"
   type    = "A"
 
   alias {
-    name                   = data.aws_alb.example.dns_name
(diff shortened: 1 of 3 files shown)


gruebel:
nice catch, can you also add a test. here is an example https://github.com/bridgecrewio/checkov/pull/2479

rafaljanicki:
Added @gruebel , I hope I've done it correctly. If not, I'll adjust

tests/terraform/graph/checks/resources/Route53ARecordAttachedResource/main.tf
```diff
@@ -97,6 +97,18 @@ resource "aws_route53_record" "pass4" {
...
+  zone_id = data.aws_route53_zone.example.zone_id
+  name    = "example"
+  type    = "A"
+
+  alias {
+    name                   = data.aws_alb.example.dns_name
+    zone_id                = data.aws_alb.example.zone_id
+    evaluate_target_health = true
+  }
+}
```
gruebel, line 110:
```suggestion
resource "aws_alb" "example" {
  name               = "example"
  internal           = false
  load_balancer_type = "application"
  security_groups    = [aws_security_group.lb_sg.id]
  subnets            = [for subnet in aws_subnet.public : subnet.id]
}

resource "aws_route53_record" "pass_alb" {
  zone_id = data.aws_route53_zone.example.zone_id
  name    = "example"
  type    = "A"

  alias {
    name                   = aws_alb.example.dns_name
    zone_id                = aws_alb.example.zone_id
    evaluate_target_health = true
  }
}
```
almost, something like this 😄

rafaljanicki:
Thanks @gruebel , I wasn't sure where the data comes from. Just pushed the change :)

gruebel (approved):
great job and thanks for the adjustments 🥇

PR #2492, opened 2022-02-18
CKV2_GCP_9 - Ensure that Container Registry repositories are not anonymously or publicly accessible

jdyke:

This PR is a new graph check for GCP [Container Registry repositories](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_registry).

This is a good scenario for a graph check in checkov. Container registry terraform resources themselves only support two keys `location` and `project`. The way to make them public is via a GCS bucket IAM resource like `google_storage_bucket_iam_member`, but if we only check for that resource, it would technically be a GCS public bucket check.

I think it's important to keep them separate so that the remediation guidance walks through the risks of public container registries. With a graph check, we can specifically hit the public repo check if `google_container_registry` AND `google_storage_bucket_iam_member` OR `google_storage_bucket_iam_binding` exist.

Let me know what you think. Happy to adjust however needed.

Thanks!


nimrodkor (approved):
TY @jdyke !

PR #2495, opened 2022-02-19
docker run ephemeral container automatic cleanup

new23d:
The `--rm` switch to docker run cleans up the ephemeral container after its run. Otherwise _spent_ containers are left behind -- which are not reused. This does not impact the container image download/cache.


gruebel (approved):
thanks, good point 👍

PR #2500, opened 2022-02-20
Add support for outputting results to file

lirshindalman:

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -267,13 +267,11 @@ class Report:
                     f"Created a checkov baseline file at {created_baseline_path}",
                     "blue",
                 )
-
         if baseline:
             output_data += colored(
                     f"Baseline analysis report using {baseline.path} - only new failed checks with respect to the baseline are reported",
                     "blue",
                 )
-
         print(output_data)
         # Remove colors from the output
         ansi_escape = re.compile(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
@@ -425,7 +423,7 @@ class Report:
 
     def get_test_suites(self, use_bc_ids=False) -> List[TestSuite]:
         test_cases = defaultdict(list)
-        
+
         records = self.passed_checks + self.failed_checks + self.skipped_checks
         for record in records:
             check_name = f"{record.get_output_id(use_bc_ids)}/{record.check_name}"
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -264,13 +264,11 @@ class Report:
         if created_baseline_path:
             output_data += colored(
                     f"Created a checkov baseline file at {created_baseline_path}",
-                    "blue",
-                )
+                    "blue",)
         if baseline:
             output_data += colored(
                     f"Baseline analysis report using {baseline.path} - only new failed checks with respect to the baseline are reported",
-                    "blue",
-                )
+                    "blue",)
         print(output_data)
         # Remove colors from the output
         ansi_escape = re.compile(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
@@ -415,8 +413,7 @@ class Report:
                 result,
                 headers=["check_id", "file", "resource", "check_name", "guideline"],
                 tablefmt="github",
-                showindex=True,
-            ) + "\n\n---\n\n"
+                showindex=True,) + "\n\n---\n\n"
         print(output_data)
         return output_data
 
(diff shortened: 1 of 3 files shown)


checkov/common/output/report.py
```diff
@@ -242,24 +244,29 @@ def print_console(
             else:
                 message = f"\nPassed checks: {summary['passed']}, Failed checks: {summary['failed']}, Skipped checks: {summary['skipped']}\n"
         print(colored(message, "cyan"))
-
+        output_data += message
```
YaaraVerner, line 247: can you print the `output_data` in the end? this way you won't need to print and add to this variable each line

checkov/common/output/report.py
```diff
@@ -420,6 +433,12 @@ def print_failed_github_md(self, use_bc_ids=False) -> None:
             )
         )
         print("\n\n---\n\n")
+        return tabulate(
```
YaaraVerner, line 436: please save it to a variable instead of duplicating it

checkov/common/runners/runner_registry.py
```diff
@@ -90,12 +100,14 @@ def print_reports(
...
         if "cli" in config.output and not config.quiet:
             print(f"{self.banner}\n")
+        file_output_format = config.output[0]
         exit_codes = []
         cli_reports = []
         report_jsons = []
         sarif_reports = []
         junit_reports = []
         cyclonedx_reports = []
+        data_outputs = {'cli': '', 'github_failed_only': ''}
```
YaaraVerner, line 110: why not use an empty dict?

YaaraVerner (commented):
Looks good! had some comments 😊

checkov/common/runners/runner_registry.py
```diff
@@ -78,6 +79,15 @@ def _handle_report(self, scan_report, guidelines, repo_root_for_plan_enrichment)
...
         self.scan_reports.append(scan_report)
 
+    def save_output_to_file(self, file_name, data, data_format):
+        try:
+            with open(file_name, 'w') as f:
+                f.write(data)
+            print(f"\nWrote output in {data_format} format to the file '{file_name}')")
+        except EnvironmentError as e:
+            print(f"\nAn error occurred while writing {data_format} results to file: {file_name}")
+            print(f"More details: \n {e}")
```
gruebel, line 89: please use `logging.info` and `logging.error`

gruebel (commented):
general question, do we really want to print the output to the terminal and to a file at the same time? Independent of this question, I would recommend to change the methods to return the output data and the caller function can decide to print it or do something else with it. like `print_console(...)` -> `create_console_output(...)` and move the print statement to the `runner_registry` class.

checkov/common/runners/runner_registry.py
lirshindalman: because the firs use is += so i have to initialize the values before.
but i change to empty dict and check the keys before using '+='

nimrodkor: @lirshindalman Try `defaultdict` 🙏

nimrodkor:
@gruebel Regarding your general comment - what would you print if an output file is specified, nothing? I'm not sure what should be the correct answer here, especially since checkov supports multiple output formats.

checkov/main.py
```diff
@@ -311,6 +311,8 @@ def add_parser_args(parser: ArgumentParser) -> None:
     parser.add('-o', '--output', action='append', choices=OUTPUT_CHOICES,
                default=None,
                help='Report output format. Add multiple outputs by using the flag multiple times (-o sarif -o cli)')
+    parser.add('--output-file-path', default=None,
+               help='Path for output file. The first output you use in the output flag will be the serve to the file (default output is cli)')
```
gruebel, line 315:
```suggestion
               help='Path for output file. The first selected output via output flag will be saved to the file (default output is cli)')
```

checkov/common/output/report.py
```diff
@@ -226,65 +227,65 @@ def print_console(
...
+                output_data += record.to_string(compact=is_compact, use_bc_ids=use_bc_ids)
             if not is_quiet:
                 for record in self.skipped_checks:
-                    print(record.to_string(compact=is_compact, use_bc_ids=use_bc_ids))
+                    output_data += record.to_string(compact=is_compact, use_bc_ids=use_bc_ids)
 
         if not is_quiet:
             for file in self.parsing_errors:
                 Report._print_parsing_error_console(file)
+                output_data += colored(f"Error parsing file {file}", "red")
```
YaaraVerner, line 263: please notice that you will print it twice

YaaraVerner (approved):
LGTM!

PR #2511, opened 2022-02-21
ECS:Ensure cluster logging with CMK

JamesWoolfenden:


checkov/terraform/checks/resource/aws/ECSClusterLoggingEncryptedWithCMK.py
```diff
@@ -0,0 +1,30 @@
...
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf):
+        if conf.get("configuration"):
+            if conf.get("configuration")[0].get('execute_command_configuration'):
+                command_conf = conf.get("configuration")[0].get('execute_command_configuration')[0]
+                if not command_conf.get('logging') == ['NONE']:
+                    if command_conf.get('kms_key_id'):
+                        log_conf = command_conf.get('log_configuration')[0]
```
gruebel, line 19: I think it is possible to not set the `log_configuration` block at all, then this would break. please add an extra if condition or add to the previous one to check for the existence of `log_configuration` -> 
```python
if command_conf.get('log_configuration'):
```

JamesWoolfenden: You're not wrong

gruebel (approved):
👍

PR #2513, opened 2022-02-21
PAN-OS resource check for rules with no desc

jamesholland-uk:

A resource check for PAN-OS Terraform provider for security policy rules. This one ensures that the description attribute is populated. This attribute can be set in two different resource types, both are covered by this check.


checkov/terraform/checks/resource/panos/PolicyDescription.py
```diff
@@ -0,0 +1,46 @@
...
+            # Iterate over each rule
+            for secrule in rules:
+
+                # Check if a description is defined in the rule
+                if 'description' in secrule:
+
+                    # If a description is defined, get the value
+                    desc = secrule.get('description')
+
+                    if desc[0] == "":
```
gruebel, line 33:
```suggestion
                    if desc[0].strip() == "":
```
how about adding `strip` here, then we can even trick the sneaky ones, who just add a space or multiple 😄

gruebel (commented):
nice, looks good, let me know, what you think about my comment, otherwise it is good to merge 🚀

checkov/terraform/checks/resource/panos/PolicyDescription.py
jamesholland-uk: Yes, I like your suggestion 👍

jamesholland-uk:
⬆️ And some tests for the strings composed of spaces too
Good to go if you're happy Anton?

gruebel (approved):
great thanks 🏆

PR #2515, opened 2022-02-21
Exclude raw HELM templates from Kubernetes runner

metahertz:
Prevents HELM template files which have not gone through the HELM runner being picked up directly by the kubernetes runner.

Previously this would cause the kubernetes runner to fail at a yaml.loads due to {{- HELM variable/markup syntax within the files.


gruebel (approved):
looks good, less runtime errors is always great 😄 👍

PR #2516, opened 2022-02-21
Linux VM no password authentication

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/azure/VMDisablePasswordAuthentication.py
+++ b/checkov/terraform/checks/resource/azure/VMDisablePasswordAuthentication.py
@@ -1,0 +1,20 @@
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_negative_value_check import BaseResourceNegativeValueCheck
+
+
+class VMDisablePasswordAuthentication(BaseResourceNegativeValueCheck):
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/azure/VMDisablePasswordAuthentication.py
```diff
@@ -0,0 +1,20 @@
...
+class VMDisablePasswordAuthentication(BaseResourceNegativeValueCheck):
+    def __init__(self):
+        name = "Ensure that Virtual machine does not enable password authentication"
+        id = "CKV_AZURE_149"
+        supported_resources = ['azurerm_linux_virtual_machine_scale_set', 'azurerm_linux_virtual_machine']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_forbidden_values(self) -> str:
+        return [True]
```
gruebel, line 14:
```suggestion
        return [False]
```
I think you meant to use `False` here, because if it is set to false, then it is actually enabled 😄

JamesWoolfenden: must have been a late night with the double negatives

gruebel:
@JamesWoolfenden can you also adjust the tests to the new logic?

PR #2523, opened 2022-02-22
PAN-OS resource check for log forwarding profiles

jamesholland-uk:

A resource check for PAN-OS Terraform provider for security policy rules. This one ensures that security policy rules have a log forwarding profile enabled, such that logs are collected at Panorama and/or an external logging server. Log forwarding can be set in two different resource types, both are covered by this check.


gruebel (approved):
nice, looks good 🍻

PR #2528, opened 2022-02-22
CKV_GCP_102 - Ensure that GCP Cloud Run services are not anonymously or publicly accessible

jdyke:

This PR adds a new checkov check for public Cloud Run services. Ideally Cloud Run services would not be exposed for unauthenticated access and instead a loud balancer would be used.


gruebel (approved):
thanks again 🍻

PR #2529, opened 2022-02-22
Ensure NACL restrict sensitive ports from 0.0.0.0/0

JamesWoolfenden:


gruebel:
in general code wise it is all good, but this would deviate from our current port checks. I would prefer to add for each port a separate check, because then the user can easier decide, which of it is worth checking or needs to be skipped. @nimrodkor or @schosterbarak any thoughts? I will accept, if I'm overruled 😄

JamesWoolfenden:
happy to split it out to 4 checks, like you say it gives the user the opportunity to allow some and not others

checkov/terraform/checks/resource/aws/AbsNACLUnrestrictedIngress.py
```diff
@@ -0,0 +1,58 @@
...
+                    return CheckResult.FAILED
+            return CheckResult.PASSED
+
+        return CheckResult.UNKNOWN
+
+    def check_rule(self, rule):
+        if rule.get('cidr_block'):
+            if rule.get('cidr_block') == ["0.0.0.0/0"]:
+                if rule.get('action') == ["allow"] or rule.get('rule_action') == ["allow"]:
+                    if rule.get('from_port')[0] >= self.port and rule.get('to_port')[0] <= self.port:
```
gruebel, line 51:
```suggestion
                    if rule.get('from_port')[0] <= self.port and rule.get('to_port')[0] >= self.port:
```
I think that's how it should be, can you either change one of the test cases or add a new one, which will test a range like, 10 - 100, then it should fail of course 😄

JamesWoolfenden: wont that be invalid terraform to start with or am i not getting it?

JamesWoolfenden: i should put some ranges in though

gruebel: sorry, what I meant was to add a test case like that
```hcl
resource "aws_network_acl_rule" "pass" {
  network_acl_id = aws_network_acl.pass.id
  rule_number    = 200
  egress         = false
  protocol       = "tcp"
  rule_action    = "allow"
  cidr_block     = "10.0.0.0/32"
  from_port      = 3300
  to_port        = 3400
}
```
and this should fail for the RDP check

JamesWoolfenden: yeah i understand now, i put the ranges in. good catch

gruebel (approved):
nice, looks great 💯

PR #2530, opened 2022-02-23
add tf azure check for ensuring ml cluster min nodes set to 0

arock95:

Custom Policy id - CKV_AZURE_150
Custom Policy name - "Ensure Machine Learning Compute Cluster Minimum Nodes Set To 0"
Custom Policy IaC type - terraform
Custom Policy type and provider - azurerm
IaC configuration documentation (If available) - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/machine_learning_compute_cluster#min_node_count

Any additional information that would help other members to better understand the check - https://docs.microsoft.com/en-us/azure/machine-learning/concept-vulnerability-management#compute-clusters
Per MS doc, if min nodes is set to 0 then not only is it a cost-saving configuration but it also allows the VMs in the cluster to get the latest VM images (with security patches).

### DIFF
--- a/checkov/terraform/checks/resource/azure/MLComputeClusterMinNodes.py
+++ b/checkov/terraform/checks/resource/azure/MLComputeClusterMinNodes.py
@@ -6,7 +6,7 @@ class MLComputeClusterMinNodes(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure Machine Learning Compute Cluster Minimum Nodes Set To 0"
         id = "CKV_AZURE_150"
-        supported_resources = ['machine_learning_compute_cluster']
+        supported_resources = ['azurerm_machine_learning_compute_cluster']
         categories = [CheckCategories.GENERAL_SECURITY]
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/azure/MLComputeClusterMinNodes.py
```diff
@@ -0,0 +1,20 @@
+from checkov.common.models.enums import CheckCategories
+from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+
+
+class MLComputeClusterMinNodes(BaseResourceValueCheck):
+    def __init__(self):
+        name = "Ensure Machine Learning Compute Cluster Minimum Nodes Set To 0"
+        id = "CKV_AZURE_150"
+        supported_resources = ['machine_learning_compute_cluster']
```
gruebel, line 9:
```suggestion
        supported_resources = ['azurerm_machine_learning_compute_cluster']
```
I think you meant this resource name 🙂

gruebel (commented):
nice addition and good to know about this specific behaviour 💪 please have a look at my comments

arock95:
Ah geez sorry about that @gruebel, latest commit fixes the resource type!

gruebel (approved):
thanks for the contribution 🚀 no problem can happen 😄

Issue #2433, opened 2022-02-14
Custom policy URLs for YAML defined policies


celik0311:
I would like to define custom policy URLs for YAML-defined policies. I expected the following to display a guideline URL when a check fails however no guideline URL is displayed:

```
---
metadata:
 name: "Test policy"
 id: "CKV2_AWS_1"
 category: "IAM"
 guideline: "www.thisisatest.com"
scope:
 provider: aws
definition:
       cond_type: "attribute"
...
```

Expected:
```
        FAILED for resource: some_resource
        File: somefile.tf:109-118
        Guide: www.thisisatest.com
```

Actual:
```
        FAILED for resource: some_resource
        File: somefile.tf:109-118
```

```
$ checkov --version
2.0.780
```

This MR looks to have added this ability https://github.com/bridgecrewio/checkov/pull/1682 but I see no reference in the docs on how to do this for YAML defined policies.

tkishel:
PR for this: https://github.com/bridgecrewio/checkov/pull/2532

Note that in the policy above, the key in YAML should be `guidelines` not `guideline` (even though the key in code is `guideline`).

```
--- 
metadata: 
  name: "TJK S3 Force Destroy Must Be False (Code Editor)"
  category: "storage"
  severity: "low"
  guidelines: "https://www.example.com/tjk"
scope: 
  provider: "aws"
definition:
     cond_type: "attribute"
     resource_types:
     - "aws_s3_bucket"
     attribute: "force_destroy"
     operator: "equals"
     value: false
```

```
# checkov -d example --bc-api-key xxx

Check: tkishel73_AWS_1644533850571: "TJK S3 Force Destroy Must Be False (Code Editor)"
	FAILED for resource: aws_s3_bucket.foo
	Severity: LOW
	File: /main.tf:9-12
	Guide: https://www.example.com/tjk

		9  | resource "aws_s3_bucket" "foo" {
		10 |   bucket        = "bar-bc-iac"
		11 |   force_destroy = true
		12 | }
```

celik0311:
Thank you! This was very helpful.

PR #2532, opened 2022-02-23
display guidelines for custom policies

tkishel:
In `_handle_customer_run_config()`, custom policies are stored in `run_config['customPolicies']` instead of in `run_config['policyMetadata']`.

This PR adds guidelines for custom policies to `self.check_metadata` in the same way that `_handle_public_metadata()` adds guidelines to `self.check_metadata` resulting in those guidelines being displayed in the output of the `checkov` command.

Addresses: https://github.com/bridgecrewio/checkov/issues/2433


mikeurbanski1 (approved):
LGTM - thanks John!

PR #2534, opened 2022-02-24
Support empty resource types for YAML queries

rotemavni:
In case we have an empty list of resource types, we will apply attribute queries on all the `resource` vertices in the graph.


gruebel (approved):
🥇

PR #2535, opened 2022-02-24
Yaml runner

schosterbarak:
create a generic yaml runner. It is similar to the json runner that enables loading generic external yaml checks. 

This will be used later on for cicd parsing.


checkov/common/parsers/yaml/parser.py
```diff
@@ -0,0 +1,48 @@
...
+        if filename.endswith(".yaml") or filename.endswith(".yml"):
+            (template, template_lines) = loader.load(filename)
+
+        if template:
+            if isinstance(template, list):
+                for t in template:
+                    if t and isinstance(t, dict):
+                        return t, template_lines
+                    if t and isinstance(t, list):
+                        return t, template_lines
```
achiar99, line 23: There is a case that t is not dict or list? if so we can just return the first element
Also what about the content in the yaml that we do not return? maybe the parse needs to return list of parsed yamls

achiar99 (approved):
Very cool! 😎  
one small question

checkov/common/parsers/yaml/parser.py
schosterbarak: changed the syntax. honestly, i don't remember why I did that.

PR #2536, opened 2022-02-24
PAN-OS Policy Logging Check

jamesholland-uk:

A resource check for PAN-OS Terraform provider for security policy rules. This one ensures that security policy rules have logging enabled, such that logs are written to the firewall storage (this is separate to "log forwarding", that ensures logs are then subsequently sent to Panorama and/or an external logging server). Logging should be set for "end of session" via the log_end attribute, and can be set in two different resource types, both are covered by this check.


gruebel (approved):
thanks, great work 💯

PR #2537, opened 2022-02-24
Fix types comparison in AbsNACLUnrestrictedIngress check

rotemavni:
Support the case when ports are written as strings instead of integers, which causes exception when trying to compare the port ranges.


Saarett (approved):
🍪

PR #2541, opened 2022-02-25
change external check loading to also consider the file path

gruebel:

Changed the logic for loading external checks, this makes it possible to have the same check file name, but in different folders. This is needed for #2535
Also added some missing type hints.


tests/terraform/checks/module/registry/test_ModuleCheck.py
```diff
@@ -39,6 +41,8 @@ def test_immutable_module(self):
         self.assertEqual(summary["skipped"], 0)
         self.assertEqual(summary["parsing_errors"], 0)
 
+        check = next(c for c in module_registry.checks["module"] if c.id == "CKV_TF_MODULE_1")
+        module_registry.checks["module"].remove(check)
```
itai1357, line 45: same

mypy.ini
```diff
@@ -3,6 +3,7 @@ files = checkov
 exclude = checkov/(arm|cloudformation|common|dockerfile|github|bitbucket|gitlab|helm|json_doc|kubernetes|kustomize|sca_package|secrets|serverless|terraform)
 strict = True
 ignore_missing_imports = True
+show_error_codes = True
```
itai1357, line 6: nice! I checked it on google.
just out of curiosity, regarding the addition if the "error_code", do you have some concrete example in checkov where it help you to understand better the error?

itai1357 (approved):
Good! just several questions =]

mypy.ini
gruebel: this is actually used for adding `# type: ignore` comments with a specific error code, so you only ignore a very specific error and not just turn type checking off for the whole line

tests/terraform/checks/module/registry/test_ModuleCheck.py
```diff
@@ -24,6 +24,8 @@ def test_module_version(self):
 
         check = next(c for c in module_registry.checks["module"] if c.id == "CKV_TF_MODULE_1")
         module_registry.checks["module"].remove(check)
+        check = next(c for c in module_registry.checks["module"] if c.id == "CKV_TF_MODULE_2")
+        module_registry.checks["module"].remove(check)
```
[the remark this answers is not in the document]
gruebel: 1. sure, let me add a comment. It is purely for cleaning without removing any other possible checks, which don't exist for `module` checks currently, but if they are added, then it makes more sense 😄 
2. As far as I know, we don't have, because there was never a real need for it. This kind of issue only happens during unit testing, because we check registry is a singleton

PR #2544, opened 2022-02-27
Fix AbsNACLUnrestrictedIngress check

rotemavni:
Handled cases where input for AbsNACLUnrestrictedIngress check can be different - 
1. No ports exist
2. ingress is an array.


gruebel (approved):
nice 🚀 that's probably old Terraform code what caused the issue 🙈

PR #2549, opened 2022-02-28
CFN - Added check of Properties as a dict

Saarett:

Some checkov checks would crash the whole scan if the Properties is malformed. In this PR I make sure the Properties is indeed a Dict so it won't crash the scan

### DIFF
--- a/checkov/cloudformation/cfn_utils.py
+++ b/checkov/cloudformation/cfn_utils.py
@@ -91,7 +91,7 @@ def get_entity_value_as_string(value: Any) -> str:
 
 
 def get_folder_definitions(
-    root_folder: str, excluded_paths: Optional[List[str]], out_parsing_errors: Dict[str, str] = {}
+        root_folder: str, excluded_paths: Optional[List[str]], out_parsing_errors: Dict[str, str] = {}
 ) -> Tuple[Dict[str, DictNode], Dict[str, List[Tuple[int, str]]]]:
     files_list = []
     for root, d_names, f_names in os.walk(root_folder):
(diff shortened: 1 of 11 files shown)


mikeurbanski1:
I suggest adding a specific test for the case that we know failed it

checkov/cloudformation/cfn_utils.py
```diff
@@ -200,8 +201,16 @@ def _parse_file(file):
         try:
             template, template_lines = parse_result
             if isinstance(template, DictNode) and isinstance(template.get("Resources"), DictNode):
-                definitions[path] = template
-                definitions_raw[path] = template_lines
+                template_resources = template.get("Resources")
+                is_valid = True
+                for resource in template_resources.values():
+                    if 'Properties' not in resource or (not isinstance(resource['Properties'], DictNode)):
+                        out_parsing_errors.update({file: 'Resource Properties is not a dictionary'})
```
gruebel, line 208: additionally it would be good to add a log message here with details about which file and the resource name.

Saarett:
> I suggest adding a specific test for the case that we know failed it

@mikeurbanski1 Some (or all) of the current failing tests are due to templates with malformed resources, tomorrow I'm going to talk with @rotemavni about how to handle such cases - Mark the whole file as malformed, or perhaps some other solution. WDYT?

rotemavni (commented):
@Saarett Great work!
Can you add a test before merging for testing a template like 
```
Resources:
  Function:
    Type: 'AWS::Serverless::Function'
    Properties: not dict
```

And verifying we get a parsing error?

PR #2553, opened 2022-02-28
PAN-OS resource check for IPsec algorithms

jamesholland-uk:

A resource check for PAN-OS Terraform provider to ensure use of secure algorithms for IPsec VPNs. It can be configured in two different resource types, both are covered by this check.


gruebel (approved):
thanks, looks good 🚀

PR #2556, opened 2022-03-01
CKV_GCP_103 - Ensure Dataproc Clusters do not have public IPs

jdyke:

This PR adds a new Checkov check for Public IP Dataproc clusters. By default Dataproc clusters have public IPs.

### DIFF
--- a/checkov/terraform/checks/resource/gcp/DataprocPublicIpCluster.py
+++ b/checkov/terraform/checks/resource/gcp/DataprocPublicIpCluster.py
@@ -13,8 +13,4 @@ class DataprocPublicIpCluster(BaseResourceValueCheck):
     def get_inspected_key(self):
         return 'cluster_config/[0]/gce_cluster_config/[0]/internal_ip_only'
 
-    # Accounts for if key is present but is set to False
-    def get_expected_value(self):
-        return True
-
 check = DataprocPublicIpCluster()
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/gcp/DataprocPublicIpCluster.py
```diff
@@ -0,0 +1,20 @@
...
+        supported_resources = ['google_dataproc_cluster']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self):
+        return 'cluster_config/[0]/gce_cluster_config/[0]/internal_ip_only'
+
+    # Accounts for if key is present but is set to False
+    def get_expected_value(self):
+        return True
```
nimrodkor, line 18: @jdyke This isn't needed as it is the default of the parent

nimrodkor (approved):
Thanks @jdyke !

checkov/terraform/checks/resource/gcp/DataprocPublicIpCluster.py
jdyke: Oh weird I thought I had to use that in other examples and the same applied here. I removed it, thanks for catching that

gruebel (approved):
nice 👍

PR #2558, opened 2022-03-02
add automatic pipeline to update admission controller checkov version

schosterbarak:


.github/workflows/build.yml
```diff
@@ -275,6 +276,29 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: update checkov-k8s version
+        run: |
+          checkov_version=${{ needs.publish-package.outputs.version }}
+          checkov_major_version=$(echo "${checkov_version}" | head -c1)
+          echo ::set-env name=RELEASE_VERSION::$(echo $checkov_version)
+          echo ::set-env name=RELEASE_MAJOR_VERSION::$(echo $checkov_major_version)
+          echo $RELEASE_VERSION
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
```
nimrodkor, line 292: @schosterbarak needed here? You just wrote a check against it 😆

admissioncontroller/k8s/deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
             drop:
               - ALL
               - NET_RAW
-        image: eurogig/whorf@sha256:5e8fbbda58cce84226c4ddcdff50a28e3806252b5c5af1de9d82565e7be44dc0
+        image: bridgecrew/whorf
```
nimrodkor, line 34: @schosterbarak Do we want to add a tag here? I'd vote yes

.github/workflows/build.yml
schosterbarak: it is required for the set-env

admissioncontroller/k8s/deployment.yaml
schosterbarak: how would i bump it?

.github/workflows/build.yml
```diff
@@ -170,8 +170,9 @@ jobs:
           ## update python version
           echo "version = '$new_tag'" > 'checkov/version.py'
           echo "checkov==$new_tag" > 'kubernetes/requirements.txt'
+          echo "checkov==$new_tag" > 'admissioncontroller/checkov-requirements.txt'
```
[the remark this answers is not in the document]
schosterbarak: same effect :)

admissioncontroller/k8s/deployment.yaml
nimrodkor: sed...

.github/workflows/build.yml
nimrodkor: Aha! If so, let's use the `set-output` [mechanism](https://github.com/bridgecrewio/checkov/blob/e4a42f0fcc9c0301d7337e55cfbaef8a749e0000/.github/workflows/build.yml#L179). Please make sure you add an id to this step, and then reference it where you need. [Example reference](https://github.com/bridgecrewio/checkov/blob/e4a42f0fcc9c0301d7337e55cfbaef8a749e0000/.github/workflows/build.yml#L187)

admissioncontroller/k8s/deployment.yaml
schosterbarak: can you assis?

eurogig: Did we solve this?  I do a bump to the latest in another repo like this.  Might be hackish.  
```
        name: Build and push
        id: docker_build
        uses: docker/build-push-action@v2
        with:
          file: clust3rf8ck/Dockerfile
          push: true
          tags: eurogig/clust3rf8ck:${{ steps.get-latest-tag.outputs.tag }}, eurogig/clust3rf8ck:latest
      -
        name: Image digest
        run: echo ${{ steps.docker_build.outputs.digest }}
      - name: Update deployment
        uses: jacobtomlinson/gha-find-replace@master
        with:
          find: "- image: eurogig/clust3rf8ck@sha256:[a-f0-9]{64}"
          replace: "- image: eurogig/clust3rf8ck@${{ steps.docker_build.outputs.digest }}"
          include: "clust3rf8ck/clust3rf8ck.yaml"`

nimrodkor: Looks really nice, we did not solve this yet @eurogig

nimrodkor (approved):

PR #2559, opened 2022-03-02
Fix Route53 filter to include vars as well

rafaljanicki:
Currently, the filter checks for module and data, but it should be possible to provide the DNS names via variables in case Route53 records are handled in a different module


nimrodkor (approved):

nimrodkor:
Thanks @rafaljanicki !

PR #2563, opened 2022-03-03
bump hcl2 version

gruebel:

- bumped the `hcl2` version to latest one and adjusted some tests, the new output looks better I would say 😄 
- I also limited the `networkx` version, so the newest release won't be picked up, because they officially don't support Python 3.7 anymore
- some other minor adjustments, which I stumbled upon

### DIFF
--- a/tests/terraform/graph/variable_rendering/expected_data.py
+++ b/tests/terraform/graph/variable_rendering/expected_data.py
@@ -31,7 +31,7 @@ expected_eks = {
             "version": ["1.19"],
             "vpc_config": {
                 "security_group_ids": ["aws_security_group.master.id"],
-                "subnet_ids": "Tree('full_splat_expr_term', ['aws_subnet.eks', 'id'])"
+                "subnet_ids": "aws_subnet.eks[*].id"
             },
         }
     }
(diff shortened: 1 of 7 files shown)


tests/terraform/graph/variable_rendering/expected_data.py
```diff
@@ -31,7 +31,7 @@
             "version": ["1.19"],
             "vpc_config": {
                 "security_group_ids": ["aws_security_group.master.id"],
-                "subnet_ids": "Tree('full_splat_expr_term', ['aws_subnet.eks', 'id'])"
+                "subnet_ids": "aws_subnet.eks[*].id"
```
nimrodkor, line 34: @gruebel Very cool that we now support it. How do we parse it as graph node attributes? As a string?

gruebel: our Transformer in the `hcl2` lib was missing a method, that's why it looked so weird. let me check, how it looks like in the graph node 🧐

gruebel: @nimrodkor currently it is just a string

Saarett:
Nice!

Saarett (approved):

PR #2574, opened 2022-03-07
Workdir should be absolute path

JamesWoolfenden:

### DIFF
--- a/checkov/dockerfile/checks/WorkdirIsAbsolute.py
+++ b/checkov/dockerfile/checks/WorkdirIsAbsolute.py
@@ -3,6 +3,8 @@ import re
 from checkov.common.models.enums import CheckCategories, CheckResult
 from checkov.dockerfile.base_dockerfile_check import BaseDockerfileCheck
 
+ISABSOLUTE = re.compile("(^/[A-z0-9-_+]*)|(^[A-z0-9-_+]:\\\\.*)|(^\\$[{}A-z0-9-_+].*)")
+
 
 class WorkdirIsAbsolute(BaseDockerfileCheck):
     def __init__(self):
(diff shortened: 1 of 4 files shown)


checkov/dockerfile/checks/WorkdirIsAbsolute.py
```diff
@@ -0,0 +1,26 @@
...
+        name = "Ensure that WORKDIR values are absolute paths"
+        id = "CKV_DOCKER_10"
+        supported_instructions = ["WORKDIR"]
+        categories = [CheckCategories.CONVENTION]
+        super().__init__(name=name, id=id, categories=categories, supported_instructions=supported_instructions)
+
+    def scan_entity_conf(self, conf):
+        for mydir in conf:
+            mypath = mydir["value"]
+            if not re.match("(^/[A-z0-9-_+]*)|(^[A-z0-9-_+]:\\\\.*)|(^\\$[{}A-z0-9-_+].*)", mypath):
```
gruebel, line 21: please use `re.compile(...)` to create the regex as a "static" variable and then use it here, like here https://github.com/bridgecrewio/checkov/blob/f43ef11c3bc489cb1a28c24826ce4d778cf35813/checkov/arm/checks/resource/NSGRulePortAccessRestricted.py#L11

JamesWoolfenden: have done it. Whats does that achieve?

gruebel: then the `flake8` plugin `dlint` is able to check, if we have some issue inside the regex

gruebel (approved):
thanks 🚀

PR #2581, opened 2022-03-08
fix single item list attribute flattening

gruebel:

I did a mistake in my last PR, which will be fixed with this one, also added a test to prevent someone else for doing the same mistake 🙈


Saarett (approved):
🍪 🥛

PR #2592, opened 2022-03-09
Panos zone protection

jamesholland-uk:

A resource check for PAN-OS Terraform provider to ensure Zone Protection Profiles are configured for each zone. This can be configured in two different resource types, both are covered by this check.


gruebel (approved):
🥇

PR #2596, opened 2022-03-10
PAN-OS check for User-ID IncludeACL in Zones

jamesholland-uk:

Another check for PAN-OS. If User-ID is enabled for a zone (it is optional), an "include ACL" should be defined to provide the scope for source IP addresses used within User-ID. This can be configured in "panos_zone" or "panos_panorama_zone" resources. Both resource types are covered by this check.


gruebel (approved):
Thanks, as always great job 🥇

PR #2597, opened 2022-03-11
k8s:tf:Do not admit containers wishing to share the host IPC namespace

JamesWoolfenden:
Also contains a fix for CKV_K8S_3 as it can relate to a pos and a PSP


gruebel (approved):
💯

PR #2603, opened 2022-03-12
skip SLS sub resource if it is a file reference

gruebel:

- added a bit of code to skip `file` reference in a resource block
```yaml
resources:
  Resources: ${file(./serverless.${opt:stage}.yml)} 
```
- also moved the `DEFAULT_VAR_PATTERN` to the `common` module, because I somehow had circular dependency

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/ECRPolicy.py
+++ b/checkov/cloudformation/checks/resource/aws/ECRPolicy.py
@@ -1,22 +1,25 @@
+from __future__ import annotations
+
 import json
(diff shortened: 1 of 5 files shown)


schosterbarak:
would this change our output for: https://github.com/OWASP/DVSA?

gruebel:
@schosterbarak nope, since it currently results in a runtime error for the whole file 😄  we just gain more resilient code.

nimrodkor (approved):

PR #2609, opened 2022-03-13
moving the method "is_docker_file" to be in utils file

itai1357:
BCE-5356 moving the method "is_docker_file" to be in common/utils for making it exposed


schosterbarak (approved):
UT please.

PR #2614, opened 2022-03-14
Implement CKV_K8S_8 for tf

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/kubernetes/LivenessProbe.py
+++ b/checkov/terraform/checks/resource/kubernetes/LivenessProbe.py
@@ -1,0 +1,38 @@
+from typing import Any
+from checkov.common.models.consts import ANY_VALUE
+from checkov.common.models.enums import CheckCategories, CheckResult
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/kubernetes/LivenessProbe.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def get_inspected_key(self) -> str:
+        return "spec/[0]/container/[0]/liveness_probe/[0]"
+
+    def scan_resource_conf(self, conf) -> CheckResult:
+        spec = conf.get('spec')[0]
+        if spec:
+            containers = spec.get("container")
+            for idx, container in enumerate(containers):
+                if type(container) != dict:
```
gruebel, line 25:
```suggestion
                if not isinstance(container, dict):
```
most of the times it is better to use `isinstance` instead of `type` to check what kind of type a variable is, because it also includes subtypes.

gruebel (commented):
other than that, nice 👍

PR #2624, opened 2022-03-15
implement CKV_K8S_14 for terraform - image tags are fixed

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/kubernetes/ImageTagFixed.py
+++ b/checkov/terraform/checks/resource/kubernetes/ImageTagFixed.py
@@ -1,0 +1,52 @@
+from typing import Any
+from checkov.common.models.consts import ANY_VALUE
+from checkov.common.models.enums import CheckCategories, CheckResult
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/kubernetes/ImageTagFixed.py
```diff
@@ -0,0 +1,44 @@
...
+                    return CheckResult.UNKNOWN
+                if container.get("image"):
+                    name = container.get("image")[0]
+                    if ":" in name:
+                        if name.split(":")[1] == "latest" or "":
+                            self.evaluated_keys = [f'spec/[0]/container/[{idx}]/image']
+                            return CheckResult.FAILED
+                        break
+                    if "@" in name:
+                        break
```
gruebel, line 35: I think this both should be `continue` not `break`. Can you change one of the failed examples to have the first container with a version or hash and second one being latest.

gruebel (approved):
👍

PR #2627, opened 2022-03-15
Trust openssl to be on PATH not hardcoded

metahertz:
Had issues in some environments where openssl was at a different path, we also use openssl without hard path later on via $(), so think this is acceptable! 


eurogig:
The reason it was hardcoded to local was because the default version of OpenSSL on MacOS doesn't work with these options.  You need to do a brew install OpenSSL of the "real" version for it to work.  Perhaps this needs some kind of OS check as well?

gruebel:
@metahertz and @eurogig do we want to merge it or should we close it?

PR #2635, opened 2022-03-16
AWS Kinesis firehose - Delivery streams are encrypted

JamesWoolfenden:


gruebel (commented):
Should we also put this exception in the check

> Server-side encryption should not be enabled when a kinesis stream is configured as the source of the firehose delivery stream.

JamesWoolfenden:
your not wrong. fixed it.

gruebel (approved):
thanks 🚀

PR #2639, opened 2022-03-16
Fix -f support for SCA scanning

nimrodkor:

If -f was passed, leverage the original file path as the path of the results, no relative paths


gruebel (approved):
💯

PR #2642, opened 2022-03-17
Enforce client certificates for api management ckv azure 152

JamesWoolfenden:


checkov/terraform/checks/resource/kubernetes/ReadonlyRootFilesystem.py
```diff
@@ -0,0 +1,31 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class ReadonlyRootFilesystem(BaseResourceCheck):
```
gruebel, line 5: I'm not sure, but is it correct that this check is also added with the PR? 😄

JamesWoolfenden: no that was not my intention

gruebel (approved):
thanks 🚀

PR #2669, opened 2022-03-23
check for drop of NET_RAW or ALL - CKV_k8s_28 for terraform

JamesWoolfenden:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2670, opened 2022-03-23
check for security context CKV_K8S_29 for terraform k8s

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/kubernetes/PodSecurityContext.py
+++ b/checkov/terraform/checks/resource/kubernetes/PodSecurityContext.py
@@ -1,0 +1,52 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class PodSecurityContext(BaseResourceCheck):
+
+    def __init__(self):
+        # CIS-1.5 5.7.3
+        name = "Apply security context to your pods and containers"
+        # Security context can be set at pod or container level.
+        id = "CKV_K8S_29"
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/kubernetes/PodSecurityContext.py
```diff
@@ -0,0 +1,52 @@
...
+class PodSecurityContext(BaseResourceCheck):
+
+    def __init__(self):
+        # CIS-1.5 5.7.3
+        name = "Apply security context to your pods and containers"
+        # Security context can be set at pod or container level.
+        id = "CKV_K8S_29"
+
+        supported_resources = ['kubernetes_pod', 'kubernetes_deployment', 'kubernetes_daemonset']
+        categories = [CheckCategories.GENERAL_SECURITY]
```
rotemavni, line 14: why not `CheckCategories.KUBERNETES`?

rotemavni (commented):
Nice 😄 
Please refer to the comments in this PR - https://github.com/bridgecrewio/checkov/pull/2669
We need to be more careful when accessing attributes without validation (or catching possible errors)

checkov/terraform/checks/resource/kubernetes/PodSecurityContext.py
JamesWoolfenden: because every check for the provider would be that.

JamesWoolfenden: Ive added type checking

PR #2671, opened 2022-03-24
Prisma policy metadata

mikeurbanski1:

Use prisma policy metadata (check title, severity), if available, when running with a Prisma API key.

Also fixes an unrelated issue where CKV_SECRETS checks were not using BC policy IDs with the --list arg (I thought it was related, so I fixed it before realizing)


nimrodkor (approved):
⚡

PR #2694, opened 2022-03-27
Create a cache for referred modules

rotemavni:


checkov/terraform/runner.py
```diff
@@ -414,17 +416,38 @@ def _strip_module_referrer(file_path: str) -> Tuple[str, Optional[str]]:
...
+
+            for modules in file_content["module"]:
+                cached = False
+                for module_name, module_content in modules.items():
+                    if "__resolved__" not in module_content:
+                        continue
+
+                    self.definitions_with_modules[file] = file_content
+                    cached = True
+                    continue
```
gruebel, line 451:
```suggestion
                    break
```
it feels like `break` fits here better, because there is no reason to iterate further.

checkov/terraform/runner.py
```diff
@@ -414,17 +416,38 @@ def _strip_module_referrer(file_path: str) -> Tuple[str, Optional[str]]:
...
         return None
+
+    def _prepare_definitions_with_modules(self):
+        for file, file_content in self.definitions.items():
+            if "module" not in file_content:
+                continue
+
+            for modules in file_content["module"]:
+                cached = False
+                for module_name, module_content in modules.items():
```
gruebel, line 445:
```suggestion
                for module_content in modules.values():
```

gruebel (approved):
nice find 💪 this should speed up the scans for module heavy user.

checkov/terraform/runner.py
gruebel, line 445: thought about this double `for` loop with two `break`, you could rewrite it in two ways.
1. move it a separate function, remove both `break` and instead just use ` return file_content`
2. leverage `for` ... `else` ... this one looks always strange to me, but would be correct Python code 😄

PR #2695, opened 2022-03-27
fix mypy issues

gruebel:

I think that `parsing_errors: dict[str, Exception]` should be `parsing_errors: list[str]` but need to verify in a later PR, because this would result in some code changes in the other runners code.


nimrodkor (approved):
TY!

PR #2697, opened 2022-03-28
fix save graph for tf graph scanner

achiar99:


gruebel:
@achiar99 could change one thing for me, then I don't have to create a separate PR for it 😄 thanks!
https://github.com/bridgecrewio/checkov/blob/5f4fefbc0d5584efbddd32d0f32ed702445a591a/checkov/bicep/runner.py#L84
`None` -> `{}`

gruebel (approved):
👍

PR #2701, opened 2022-03-28
Fix - Logging Init

njgibbon:
Hey all, hope you're well!

Just looking at some issues to pick something up and found that there are some `INFO` logs leaking into CLI output by default.

I've found that this is due to how the logger is initialised and we need to ensure that it is done before anything else has an opportunity to run where we would like it to inherit the logging.

I am not an expert with any of this stuff so please guide me if you'd like another implementation to get the job done! For example, could import it as before and just run `init()` before some of the other imports.

## Value
Better User Experience with the cleaner CLI output and also users are less likely to raise an issue because they are being confused with DEBUG logs in the CLI output where they are not representing a functional problem - meaning less operational overhead.

## Details
**Before (default):**
```
INFO:root:Leveraging the IAM definition at /path/iam-definition.json
INFO:checkov.common.checks_infra.registry:Loading external checks from /path/graph_checks
INFO:checkov.common.checks_infra.registry:Searching through [] and ['SQLServerAuditingEnabled.yaml']
INFO:root:Resultant set of frameworks (removing skipped frameworks): all
```

**After (default):**
```
Nothing output!
```

**After (LOG_LEVEL=INFO):**
You can see the central logger is being used by how the output looks.
```
2022-03-28 12:35:59,363 [MainThread  ] [INFO ]  Leveraging the IAM definition at /path/iam-definition.json
2022-03-28 12:35:59,517 [MainThread  ] [INFO ]  Loading external checks from /path/graph_checks
2022-03-28 12:35:59,518 [MainThread  ] [INFO ]  Searching through [] and ['SQLServerAuditingEnabled.yaml']
2022-03-28 12:35:59,564 [MainThread  ] [INFO ]  Resultant set of frameworks (removing skipped frameworks): all
```

**License**

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -15,7 +15,7 @@ from configargparse import ArgumentParser
 from configargparse import Namespace
 from urllib3.exceptions import MaxRetryError
 
-import checkov.logging_init # should be imported before the others to ensure correct logging setup
+import checkov.logging_init  # should be imported before the others to ensure correct logging setup
 from checkov.arm.runner import Runner as arm_runner
 from checkov.bitbucket.runner import Runner as bitbucket_configuration_runner
 from checkov.cloudformation.runner import Runner as cfn_runner
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -15,6 +15,7 @@ from configargparse import ArgumentParser
 from configargparse import Namespace
 from urllib3.exceptions import MaxRetryError
 
+import checkov.logging_init
 from checkov.arm.runner import Runner as arm_runner
 from checkov.bitbucket.runner import Runner as bitbucket_configuration_runner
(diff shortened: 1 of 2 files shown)


checkov/main.py
```diff
@@ -15,6 +15,7 @@
 from configargparse import Namespace
 from urllib3.exceptions import MaxRetryError
 
+import checkov.logging_init
```
gruebel, line 18:
```suggestion
import checkov.logging_init  # should be imported before the others to ensure correct logging setup
```
please add some extra info, so the next developer doesn't move it 😄

gruebel (commented):
@njgibbon thanks! nice observation. I first thought, that's a weird behaviour, but after seeing the order of imports it makes sense.

gruebel (approved):
thanks 🥇

PR #2706, opened 2022-03-29
Fix non-dict labels in TillerService

rotemavni:
Handle the case where the labels are not dictionary, to avoid throwing exceptions


tronxd (approved):
LGTM :)

PR #2712, opened 2022-03-29
Checkov breaks with nested modules

jdyke:

Checkov will fail on some of the "check public" resource checks in GCP due to us not performing a safety check before diving into our evaluation logic. We ran into a specific scenario where nested modules broke some of checkov's checks due to:

```
TypeError: 'NoneType' object is not subscriptable
```

aka we were not checking to see if the key exists before assigning it to a variable. I didn't think this was possible since these resource keys are required, but there are scenarios where it could happen.


nimrodkor (approved):
TY @jdyke !

PR #2714, opened 2022-03-30
Missing Job from CKV_K8S_31

Kostavro:
RIght now, if the block:
```
securityContext:
  seccompProfile:
    type:RunTimeDefault
```
is added to a Job, and checkov runs against this job, it will still report CKV_K8S_31 as FAILED.

In this file it is obvious that if instead, we add the annotation `seccomp.security.alpha.kubernetes.io/pod`, then checkov will not report CKV_K8S_31 as FAILED. However, this annotation is deprecated as per Kubernetes documentation: https://kubernetes.io/docs/tutorials/security/seccomp/#create-a-pod-with-a-seccomp-profile-for-syscall-auditing

So, since Job follows the spec order of Deployment, I have added it next to Deployment and StatefulSet. I have also added DaemonSet and ReplicaSet.


nimrodkor (approved):
TY @Kostavro !

PR #2715, opened 2022-03-30
Support nested rules for AbsSecurityListUnrestrictedIngress OCI check

rotemavni:
Handle a case where the `ingress_security_rules` are parsed into double nested array.


ayajbara (commented):
Add the test to 3389 too.

ayajbara (approved):

PR #2722, opened 2022-03-30
Cache non-existing referrers

rotemavni:

### DIFF
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -420,10 +420,12 @@ class Runner(BaseRunner):
         cached_referrer = self.referrer_cache.get(full_file_path)
         if cached_referrer:
             return cached_referrer
+        if full_file_path in self.non_referred_cache:
+            return None
+
         if not self.definitions_with_modules:
             self._prepare_definitions_with_modules()
         for file, file_content in self.definitions_with_modules.items():
-
             for modules in file_content["module"]:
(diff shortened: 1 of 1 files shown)


checkov/terraform/runner.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         self.graph_registry = get_graph_checks_registry(self.check_type)
         self.definitions_with_modules: Dict[str, Dict] = {}
         self.referrer_cache: Dict[str, str] = {}
+        self.non_referred_cache: List[str] = []
```
gruebel, line 68:
```suggestion
        self.non_referred_cache: Set[str] = set()
```
of course `Set` needs to be imported

checkov/terraform/runner.py
```diff
@@ -433,6 +436,8 @@ def _find_id_for_referrer(self, full_file_path) -> Optional[str]:
                         id_referrer = f"module.{module_name}"
                         self.referrer_cache[full_file_path] = id_referrer
                         return id_referrer
+
+        self.non_referred_cache.append(full_file_path)
```
gruebel, line 440:
```suggestion
        self.non_referred_cache.add(full_file_path)
```

gruebel (approved):
how about using a set instead of a list, look ups are faster, but the memory footprint is abit higher, which shouldn't be a problem for this kind of list. I don't expect 1 million entries here 😄

PR #2723, opened 2022-03-30
handle boolean values in graph equals check

mikeurbanski1:

Fixes an issue where custom policies with a condition like `== "false"` would fail against a Terraform attribute like `= false` due to case sensitivity in Python booleans.


kartikp10 (approved):
Good catch 💡

PR #2736, opened 2022-03-31
Feature/platform only policies

mikeurbanski1:

With this change, by default, if using an API key, checkov will only run policies that exist in the platform (currently in the BC platform, regardless of API key type - because the results should match what they see in the platform). Platform policies are identified by the existence if `bc_id` on the check (in some cases, runners are responsible for setting this on their own checks, if they load them in a nonstandard way, like the secrets runner).

Also adds the flag `--include-all-checkov-policies` to include OOTB policies that only exist in checkov. Custom policies loaded locally will still run without this flag.

So, specifically:

* if no API key, run all checkov policies and all local custom policies
* with API key and without the `--include...` flag, run all policies that exist in the platform, all platform custom policies, and all local custom policies
* you can also `--check` specific check IDs to include them if they are checkov-only, without including all checkov policies
* with an API key and `--include...`, then runs all policies (same as the current behavior before this change)

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -128,7 +128,7 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
         config.var_file = [os.path.abspath(f) for f in config.var_file]
 
     runner_filter = RunnerFilter(framework=config.framework, skip_framework=config.skip_framework, checks=config.check,
-                                 skip_checks=config.skip_check,
+                                 skip_checks=config.skip_check, include_all_checkov_policies=config.include_all_checkov_policies,
                                  download_external_modules=convert_str_to_bool(config.download_external_modules),
                                  external_modules_download_path=config.external_modules_download_path,
                                  evaluate_variables=convert_str_to_bool(config.evaluate_variables),
@@ -203,6 +203,7 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
             return None
     else:
         logger.debug('No API key found. Scanning locally only.')
+        config.include_all_checkov_policies = True
 
     if config.check and config.skip_check:
         if any(item in runner_filter.checks for item in runner_filter.skip_checks):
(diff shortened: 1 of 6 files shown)


kartikp10 (approved):
Noice 🚀

checkov/main.py
```diff
@@ -496,6 +506,15 @@ def normalize_config(config: Namespace) -> None:
         logger.warning('--skip-policy-download is deprecated and will be removed in a future release. Use --skip-download instead')
         config.skip_download = True
 
+    if config.bc_api_key and not config.include_all_checkov_policies:
+        # info because we expect this to be the standard usage
+        logger.info('You are using an API key and did not set the --include-all-checkov-policies flag, so policies '
```
gruebel, line 511: nice 👍

checkov/main.py
```diff
@@ -337,6 +338,13 @@ def add_parser_args(parser: ArgumentParser) -> None:
                help='Name for output file. The first selected output via output flag will be saved to the file (default output is cli)')
     parser.add('--output-bc-ids', action='store_true',
                help='Print Bridgecrew platform IDs (BC...) instead of Checkov IDs (CKV...), if the check exists in the platform')
+    parser.add('--include-all-checkov-policies', action='store_true',
+               help='When running with an API key, the Checkov will omit any policies that do not exist '
```
gruebel, line 342:
```suggestion
               help='When running with an API key, Checkov will omit any policies that do not exist '
```
can be wrong, but sounds better without it 😄

gruebel (approved):
🍻

checkov/main.py
mikeurbanski1: You are absolutely not wrong, lol

mikeurbanski1:
Wait until Sunday / Monday to merge, with a release announcement

mikeurbanski1:
Merge on Sunday / Monday in conjunction with a release announcement

nimrodkor (commented):
@mikeurbanski1 I don't think graph checks are affected by your changes... Changes should be around this method:
https://github.com/bridgecrewio/checkov/blob/d8ac7646a5d02173af3437bf1f31333b6cec6f2e/checkov/common/checks_infra/registry.py#L31

mikeurbanski1:
> @mikeurbanski1 I don't think graph checks are affected by your changes... Changes should be around this method:
> 
> https://github.com/bridgecrewio/checkov/blob/d8ac7646a5d02173af3437bf1f31333b6cec6f2e/checkov/common/checks_infra/registry.py#L31

The metadata integration loads the graph checks before setting the metadata. I do see that it still only loads the TF graph check registry. It's fine up until now, because we have no OOTB graph checks for the others. Updated.

PR #2740, opened 2022-04-01
Handle nested WAF

### LINKED ISSUE #2617: Check Azure Front Door WAF enabled fails even when a WAF is correctly assigned
**Describe the issue**
[`CKV_AZURE_121`](https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py) fails despite a Web Application Firewall policy being correctly applied. 

WAF policies are applied by specifying a value for `web_application_firewall_policy_link_id` inside a `frontend_endpoint` block within the `azurerm_frontdoor` resource itself.

The [documentation](https://docs.bridgecrew.io/docs/ensure-that-azure-front-door-enables-waf) seems to expect that the `web_application_firewall_policy_link_id` attribute is defined in the resource block itself, rather than in a sub-block (`frontend_endpoint`).

- [`azurerm_frontdoor` resource documentation reference](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/frontdoor#web_application_firewall_policy_link_id)

**Examples**
```terraform
resource "azurerm_frontdoor" "test" {
  name                                         = "test-front-door"
  resource_group_name                          = var.resource_group_name
  enforce_backend_pools_certificate_name_check = false
  tags                                         = var.tags

  frontend_endpoint {
    name                                    = "DefaultFrontend"
    host_name                               = "test-front-door.azurefd.net"
    web_application_firewall_policy_link_id = azurerm_frontdoor_firewall_policy.test.id
  }

  # ... 
```

 - Checkov Version: 2.0.930

**Additional context**

tsmithv11:

Fixes #2617

### DIFF
--- a/checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
+++ b/checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
@@ -12,7 +12,7 @@ class AzureFrontDoorEnablesWAF(BaseResourceValueCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
-        return "web_application_firewall_policy_link_id" or "frontend_endpoint/[0]/web_application_firewall_policy_link_id"
+        return "frontend_endpoint/[0]/web_application_firewall_policy_link_id"
 
     def get_expected_value(self):
         return ANY_VALUE
--- a/checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
+++ b/checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
@@ -12,7 +12,7 @@ class AzureFrontDoorEnablesWAF(BaseResourceValueCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
-        return "web_application_firewall_policy_link_id"
+        return "web_application_firewall_policy_link_id" or "frontend_endpoint/[0]/web_application_firewall_policy_link_id"
 
     def get_expected_value(self):
         return ANY_VALUE
--- a/tests/terraform/checks/resource/azure/test_AzureFrontDoorEnablesWAF.py
+++ b/tests/terraform/checks/resource/azure/test_AzureFrontDoorEnablesWAF.py
@@ -111,6 +111,55 @@ class TestAzureFrontDoorEnablesWAF(unittest.TestCase):
         scan_result = check.scan_resource_conf(conf=resource_conf)
         self.assertEqual(CheckResult.PASSED, scan_result)
 
+        def test_success(self):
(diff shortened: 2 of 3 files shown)


checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
```diff
@@ -12,7 +12,7 @@ def __init__(self):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
-        return "web_application_firewall_policy_link_id"
+        return "web_application_firewall_policy_link_id" or "frontend_endpoint/[0]/web_application_firewall_policy_link_id"
```
gruebel, line 15:
```suggestion
        return "frontend_endpoint/[0]/web_application_firewall_policy_link_id"
```
the documentation doesn't even suggest, that it is possible to add a `web_application_firewall_policy_link_id` argument in the root level, so lets remove it and then of course remove the test, which expects the wrong way to be correct.

tests/terraform/checks/resource/azure/test_AzureFrontDoorEnablesWAF.py
```diff
@@ -111,6 +111,55 @@ def test_success(self):
         scan_result = check.scan_resource_conf(conf=resource_conf)
         self.assertEqual(CheckResult.PASSED, scan_result)
 
+        def test_success(self):
```
gruebel, line 114: it would be great, if you can migrate the tests here to the better syntax with a separate TF file, like one of James PR #2730

tsmithv11: Updated. Let me know if this works.

checkov/terraform/checks/resource/azure/AzureFrontDoorEnablesWAF.py
tsmithv11: Good catch! Updated.

tests/terraform/checks/resource/azure/test_AzureFrontDoorEnablesWAF.py
```diff
@@ -1,116 +1,42 @@
...
-import hcl2
-
+from checkov.runner_filter import RunnerFilter
+from checkov.terraform.runner import Runner
 from checkov.terraform.checks.resource.azure.AzureFrontDoorEnablesWAF import check
-from checkov.common.models.enums import CheckResult
 
 
-class TestAzureFrontDoorEnablesWAF(unittest.TestCase):
+class TestOSSBucketAccessLogs(unittest.TestCase):
```
gruebel, line 9:
```suggestion
class TestAzureFrontDoorEnablesWAF(unittest.TestCase):
```
just adjust the test name and then it is good to merge 😉

gruebel (approved):
👍

PR #2747, opened 2022-04-03
removed code blocks in compact json output

### LINKED ISSUE #2713: Add --compact behaviour to json output
**Describe the issue**

For the json output, it would be nice to have the possibility of using the '--compact' option as for the CLI.

As for the CLI, maybe we only want to know the "file_path" and the "file_line_range" without having the code itself.
The "code_block" in the json is really hard to read because it creates a field per line with the line number.
Checkov will copy that block for each check and makes the json really long if the block is long (e.g. kubernetes cluster description) and it has several vulnerabilities in it.
So it would be nice to be able to not include it in the json output.

e.g. :
"code_block": [
	[
		24,
		"resource \"google_compute_subnetwork\" \"dummy\" {\n"
	],
	[
		25,
		"  name          = \"${var.name_root}-${random_pet.subnet.id}\"\n"
	],
	[
		26,
		"  region        = var.region\n"
	],
	[
		27,
		"  network       = google_compute_network.dummy.name\n"
	],
	[
		28,
		"  ip_cidr_range = \"10.40.1.0/24\"\n"
	],
	[
		29,
		"}\n"
	]
]
**Additional context**

Eliran-Turgeman:

resolves #2713


checkov/common/runners/runner_registry.py
```diff
@@ -283,3 +285,13 @@ def get_enriched_resources(repo_roots: List[Union[str, os.PathLike]]) -> Dict[st
                                 "skipped_checks": skipped_checks,
                             }
         return enriched_resources
+
+    @staticmethod
+    def strip_code_blocks_from_json(report_jsons):
+        for report in report_jsons:
+            results = report.get('results', [])
```
itai1357, line 292: it seems that at the next line you expect "results" us dict.
maybe you meant: 
`results = report.get('results', dict())`
?

checkov/common/runners/runner_registry.py
```diff
@@ -283,3 +285,13 @@ def get_enriched_resources(repo_roots: List[Union[str, os.PathLike]]) -> Dict[st
                                 "skipped_checks": skipped_checks,
                             }
         return enriched_resources
+
+    @staticmethod
+    def strip_code_blocks_from_json(report_jsons):
```
itai1357, line 290: consider to add a test for checking this method

itai1357 (commented):
nice! just short comment please have a look =]

checkov/common/runners/runner_registry.py
gruebel: I think Itai is right 👍 but better to use
```python
results = report.get('results', {})
```

checkov/common/runners/runner_registry.py
gruebel, line 290:
```suggestion
    def strip_code_blocks_from_json(report_jsons: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
```
this method is actually non-idempotent, therefore it doesn't need to return the `report_jsons` they will be changed in-place, because `dict` in Python is a mutable object.
If you decide to remove the return statement then the type hints will be
```python
def strip_code_blocks_from_json(report_jsons: List[Dict[str, Any]]) -> None:
```

gruebel (commented):
good job, please also add 1-2 UTs for this change

checkov/common/runners/runner_registry.py
Eliran-Turgeman: Thanks for clarifying this point

gruebel (approved):
💯

PR #2752, opened 2022-04-03
Added Openapi registry

ayajbara:


checkov/common/checks/object_registry.py
```diff
@@ -0,0 +1,179 @@
+from checkov.common.checks.base_check_registry import BaseCheckRegistry
+from checkov.common.models.enums import CheckResult
+from checkov.yaml_doc.enums import BlockType
+
+
+class Registry(BaseCheckRegistry):
```
schosterbarak, line 6: can that be resolved using inheritance/ usage of the yaml or json registry?

checkov/openapi/checks/resource/Version.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict, Any
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.json_doc.enums import BlockType
+from checkov.openapi.checks.base_openapi_check import BaseOpenapiCheck
+
+
+class Version(BaseOpenapiCheck):
```
schosterbarak, line 7: why a k8 check and not a real check?

checkov/openapi/runner.py
```diff
@@ -13,11 +18,22 @@ def import_registry(self) -> BaseCheckRegistry:
         from checkov.openapi.checks.registry import openapi_registry
         return openapi_registry
 
-    def _parse_file(self, f: str) -> None:
-        raise Exception("parser should be implemented")
+    def _parse_file(self, f: str) -> tuple[dict[str, Any] | list[dict[str, Any]], list[tuple[int, str]]] | tuple[None, None]:
+        if f.endswith(".json"):
+            return JsonRunner._parse_file(self, f)
```
schosterbarak, line 23: before returning the json or yaml file results we need to check it's indeed an open API file and not any yaml (for example not a cfn yaml). for that you'll need to either validate the file schema or look for an element that is always there. if it's not an open API spec you should return None,None.

in github actions it is solved by inspecting path: https://github.com/bridgecrewio/checkov/blob/master/checkov/github_actions/runner.py#L24

in cloudformation by inspecting an element 
https://github.com/bridgecrewio/checkov/blob/master/checkov/cloudformation/parser/cfn_yaml.py#L218

in github vcs by inspecting a schema:
https://github.com/bridgecrewio/checkov/blob/0c1e05c6ac94d4fd8957b1b5a1cb4334b6e1fa0f/checkov/github/base_github_org_security.py#L23

schema is the most accurate. Does openapi have a json schema you can validate or an element you check for existence?

schosterbarak: maybe this can be used to validate a file is indeed a json schema: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.json

checkov/openapi/checks/resource/Version.py
ayajbara: its a dummy check.

checkov/common/checks/object_registry.py
ayajbara: why not creating a new common enums for both json and yaml?

checkov/openapi/checks/resource/Version.py
schosterbarak: please use a real one so it will have context for future maintenance or use "foo/goo" and not k8.

checkov/openapi/runner.py
schosterbarak: sounds good.

Saarett (approved):

PR #2755, opened 2022-04-04
fix in arm runner and registry

YaaraVerner:
handle the case when `type` is not in arm resource
fix edge-case in PasswordPolicyMaxLogin check


gruebel (approved):
do you also have an example, which can be used for a UT?

PR #2760, opened 2022-04-04
Leverage line numbers from hcl2 parser

### LINKED ISSUE #864: bc-python-hcl2: Multi-line ternary parse error
**Describe the bug**
From #855, multi-line ternary operations do not parse correctly. For example, the following:
```
multiline = (local.local_true) ?
  "correct" : "wrong"
```
Results in this error:

> Unexpected token Token(__ANON_0, '\\n') at line 10, column 35.
> Expected one of: 
>  * LSQB
>  * LBRACE
>  * LPAR
>  * NULL
>  * __ANON_11
>  * DECIMAL
>  * FALSE
>  * __ANON_3
>  * __ANON_10
>  * STRING_LIT
>  * MINUS
>  * TRUE
>  * BANG

**Expected behavior**
This should be legal terraform and I would expect a parsed result equivalent to:
```
multiline = (local.local_true) ? "correct" : "wrong"
```
(that is: `${local.local_true ? 'correct' : 'wrong'`)

### LINKED ISSUE #1871: Error parsing multiline expression
checkov (2.0.510 and 2.0.528) fails to parse terraform multiline expressions when a line starts with a new operator.  The following expressions all parses ok:
```
locals {
  ok1 = (false || true)
  ok2 = (false ||
         true)
  ok3 = (false ?
         true
         : false)
}
```
These expressions all fails to parse:
```
locals {
  fail1 = (false
           || true)
  faill2 = (false
            ? true
            : false)
}
```

gruebel:

Fixes #1871 🚀 
Fixes #864

- most work was done in the `hcl2` parser
- because each TF block is a bit different, needed to change `enrich_definition_block` to an abstract method and implement the different flavours
- some tests had duplicate resources, which messed up the result -> removed
- added a small performance improvement in `checkov/terraform/graph_builder/local_graph.py`, resulted in 3 seconds faster builds for my test repo, which sounds a lot, but the whole run took 2:48min -> 2:45min 😄


checkov/terraform/context_parsers/parsers/resource_context_parser.py
```diff
@@ -13,5 +15,19 @@ def get_entity_context_path(self, entity_block: Dict[str, Dict[str, Any]]) -> Li
...
+    def enrich_definition_block(self, definition_blocks: List[Dict[str, Any]]) -> Dict[str, Any]:
+        for entity_block in definition_blocks:
+            entity_type, entity_value = next(iter(entity_block.items()))
+            entity_name, entity_config = next(iter(entity_value.items()))
+
+            self.context[entity_type][entity_name] = {
+                "start_line": entity_config[START_LINE],
+                "end_line": entity_config[END_LINE],
+                "code_lines": self.file_lines[entity_config[START_LINE] - 1: entity_config[END_LINE]],
+            }
```
nimrodkor, line 27: @gruebel Is this duplicated from [data_context_parser.py](https://github.com/bridgecrewio/checkov/pull/2760/files#diff-c9156a98f78acf8cd94baf468cc169483044003c4f8ccc81a673d84836db3c96)?

checkov/terraform/context_parsers/parsers/variable_context_parser.py
```diff
@@ -26,10 +27,16 @@ def get_entity_context_path(self, entity_block: Dict[str, Dict[str, Any]]) -> Li
...
-        for variable_block in definition_blocks:
-            if isinstance(variable_block, dict):
-                self._collect_default_variables_values(variable_block)
+        for entity_block in definition_blocks:
+            entity_name, entity_config = next(iter(entity_block.items()))
+            self.context[entity_name] = {
+                "start_line": entity_config[START_LINE],
+                "end_line": entity_config[END_LINE],
+                "code_lines": self.file_lines[entity_config[START_LINE] - 1: entity_config[END_LINE]],
+            }
```
nimrodkor, line 36: @gruebel Is this duplicated from `checkov/terraform/context_parsers/parsers/module_context_parser.py`?

nimrodkor (commented):
Very, very good for performance and for simplicity!

rotemavni (approved):
Loved it!

checkov/terraform/context_parsers/parsers/variable_context_parser.py
gruebel: they are slightly different

checkov/terraform/context_parsers/parsers/locals_context_parser.py
```diff
@@ -19,10 +24,16 @@ def get_entity_context_path(self, entity_block: Dict[str, Dict[str, Any]]) -> Li
...
     def enrich_definition_block(self, definition_blocks: List[Dict[str, Any]]) -> Dict[str, Any]:
-        self.context = super().enrich_definition_block(definition_blocks)
-        for locals_block in definition_blocks:
-            if isinstance(locals_block, dict):
-                self._collect_local_values(locals_block)
+        for entity_block in definition_blocks:
+            if START_LINE in entity_block.keys():
+                self.context["start_line"] = entity_block[START_LINE]
+            if END_LINE in entity_block.keys():
+                self.context["end_line"] = entity_block[END_LINE]
```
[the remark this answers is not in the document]
gruebel: they will always be there, but the `locals` is processed differently than the others. We iterate through a list of dicts single entries and two of them are start and end line.

PR #2763, opened 2022-04-05
Fix in PasswordPolicyMaxLogin check

YaaraVerner:
fix edge-case in PasswordPolicyMaxLogin check return UNKNOWN if max_login_attempts is variable.
add a resource to tests for unknown result


checkov/terraform/checks/resource/alicloud/PasswordPolicyMaxLogin.py
```diff
@@ -28,6 +28,8 @@ def scan_resource_conf(self, conf):
 
         if conf.get('max_login_attempts'):
             length = force_int(conf.get('max_login_attempts')[0])
+            if not length:
```
gruebel, line 31:
```suggestion
            if length is None:
```
because `0` is a valid value 🙂

gruebel (commented):
nice 👍

PR #2766, opened 2022-04-05
Fix relative path in openapi runner result records 

nimrodkor:


gruebel (commented):
even you didn't ask me 😄 I think there will be an issue with Windows and different drives

nimrodkor:
> even you didn't ask me 😄 I think there will be an issue with Windows and different drives

@gruebel It works for terraform runner, 🤷 
https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/runner.py#L184

gruebel (approved):
👍

PR #2779, opened 2022-04-07
add possibility to adjust parallelization type

gruebel:

- Changed the default multiprocess context method to `spawn`
- It can be adjusted via env `CHECKOV_PARALLELIZATION_TYPE` I'm still open for a better name
  - valid configs `fork`, `spawn`, `thread` or `none`
  - invalid config results in `none`
- also added a test to run a file parsing with each type and added an exclusion for `fork` + `Windows` = 💔 
- added/adjusted type hints
- added a decorator, which is meant for development to time function calls

### DIFF
--- a/checkov/arm/runner.py
+++ b/checkov/arm/runner.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import logging
 import os
-from typing import Optional, List, Dict, Tuple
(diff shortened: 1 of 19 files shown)


nimrodkor (approved):
Just a tiny bit too many changes @gruebel

PR #2786, opened 2022-04-08
remove unneeded move operation in helm error handler

mikeurbanski1:

If the k8s runner raises an exception for a helm chart, then the exception handler is guaranteed to crash the overall checkov run:

* `shutil.move(target_dir, save_error_dir)` moves the `target_dir` `TemporaryDirectory` to another place
* When `target_dir` gets automatically cleaned up, Checkov crashes due to the resulting `No such file or directory` error

It seems like (based on the TODO on line 226) that there is more work that might happen to do something with this failed result, so for now I just commented out the whole second temp dir.


gruebel (approved):
👍

nimrodkor (approved):

PR #2790, opened 2022-04-10
Connection_one_exists solver

### LINKED ISSUE #2196: OneOf conditional checking based on the configuration of multiple like-resources
**Describe the issue**
I have more of a question and am looking for guidance, potentially on contributing back to checkov.  My team and I are trying to implement a graph-based check that fails if "one of" a specifically configured resource does not exist.   The current concrete example is:  

> If a VPC is present, then at least one VPC Flow Log must be present that is configured with a LogDestination of s3, and the destination path must be a specific centralized bucket. 

As far as I can tell there isn't a way I can write this sort of check.  We have explored a yml graph-based approach as well as pondered how we could write this in python proper.  It seems to me that the yml logic diverges from the available to-be-subclassed python classes for writing checks.  In the yaml case we have access to the graph and can write checks based on connected resources, but (again afaict) don't have the ability to say "one of _n_ like resource types must be configured a certain way".  On the python side we could accomplish the latter, but only have access to one resource at a time, and don't know how many we might encounter.

Is my interpretation of current state accurate?

I can think of two possible ways checkov might support this, but would greatly appreciate some input. 

1. implement a solver (e.g. `OneOfSolver` that operates like https://github.com/bridgecrewio/checkov/blob/master/checkov/common/checks_infra/solvers/complex_solvers/and_solver.py but expects `any()` to pass.

2. somehow offer another implementation or variation on `scan_resource_conf` that provides greater context than it does today.  for example maybe all instances of configured `resource_types` for a check could be passed into the scan function.  or the rest of the template could be provided as context.  something like that. 

Absolutely welcome alternate thoughts and approaches.  If there is a potential desired path here details would be great as my team and I would consider PRing the functionality.

Thanks!

**Examples**
An example template that illustrates the situation I described above.  1 vpc, 2 flow logs, one of which meet a "check passed" criteria.  However, the `team-flow-log` would cause a failure without the ability to define `OneOf` conditional criteria.

```
{
    "Resources": {
      "VPCB9E5F0B4": {
        "Type": "AWS::EC2::VPC"
      },
      "team-flow-log": {
        "Type": "AWS::EC2::FlowLog",
        "Properties": {
          "ResourceId": {
            "Ref": "VPCB9E5F0B4"
          },
          "ResourceType": "VPC",
          "TrafficType": "ALL",
          "LogDestination": "arn:aws:s3:::my-teams-bucket",
          "LogDestinationType": "s3"
        }
      },
      "company-required-flow-log": {
        "Type": "AWS::EC2::FlowLog",
        "Properties": {
          "ResourceId": {
            "Ref": "VPCB9E5F0B4"
          },
          "ResourceType": "VPC",
          "TrafficType": "ALL",
          "LogDestination": "arn:aws:s3:::company-required-bucket",
          "LogDestinationType": "s3"
        }
      }
    }
  }
```

 - Checkov Version - latest

**Additional context**
N/A

NoaAzoulay:

Fixes #2196

new solver : Connection_one_exists 
this solver allows to define specific requirements in connection between resources, and it recognize the resource which meets the requirements even if there are resources which not

new example of graph check : Ensure a specific VPC flow log exists in VPC's flow logs  (demonstrates the new solver action)

### DIFF
--- a/checkov/common/checks_infra/checks_parser.py
+++ b/checkov/common/checks_infra/checks_parser.py
@@ -31,6 +31,8 @@ from checkov.common.checks_infra.solvers import (
 )
 from checkov.common.checks_infra.solvers.attribute_solvers.not_subset_attribute_solver import NotSubsetAttributeSolver
 from checkov.common.checks_infra.solvers.attribute_solvers.subset_attribute_solver import SubsetAttributeSolver
+from checkov.common.checks_infra.solvers.connections_solvers.connection_one_exists_solver import \
+    ConnectionOneExistsSolver
 from checkov.common.graph.checks_infra.base_check import BaseGraphCheck
 from checkov.common.graph.checks_infra.base_parser import BaseGraphCheckParser
 from checkov.common.graph.checks_infra.enums import SolverType
@@ -67,7 +69,11 @@ operators_to_complex_solver_classes = {
     "or": OrSolver,
 }
 
-operator_to_connection_solver_classes = {"exists": ConnectionExistsSolver, "not_exists": ConnectionNotExistsSolver}
+operator_to_connection_solver_classes = {
+    "exists": ConnectionExistsSolver,
+    "one_exists": ConnectionOneExistsSolver,
+    "not_exists": ConnectionNotExistsSolver
+}
 
 operator_to_complex_connection_solver_classes = {"and": AndConnectionSolver, "or": OrConnectionSolver}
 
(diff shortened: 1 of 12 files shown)


tronxd (approved):
Very nice! 💯

gruebel (approved):
nice, great job 🥇

gruebel:
@NoaAzoulay can you add the new operator to the docs https://github.com/bridgecrewio/checkov/blob/master/docs/3.Custom%20Policies/YAML%20Custom%20Policies.md

Issue #2399, opened 2022-02-11
Support AWS provider version 4.0.0


HorizonNet:
**Describe the issue**
The latest AWS provider version 4.0.0 introduces multiple breaking changes, especially regarding the `aws_s3_bucket` resource

**Examples**
The following snippet is based on version 4.0.0 and will make the following checks fail

* aws-s3-enable-bucket-encryption
* aws-s3-enable-versioning
* aws-s3-encryption-customer-key

```terraform
resource "aws_s3_bucket" "example" {
  bucket = "example"
}

resource "aws_s3_bucket_acl" "example" {
  bucket = aws_s3_bucket.example.id
  acl    = "private"
}

resource "aws_s3_bucket_server_side_encryption_configuration" "example" {
  bucket = aws_s3_bucket.example.id

  rule {
    apply_server_side_encryption_by_default {
      kms_master_key_id = aws_kms_key.example.arn
      sse_algorithm     = "aws:kms"
    }
  }
}

resource "aws_s3_bucket_versioning" "example" {
  bucket = aws_s3_bucket.example.id

  versioning_configuration {
    status = "Enabled"
  }
}

resource "aws_kms_key" "example" {
  description         = "example"
  enable_key_rotation = true
  multi_region        = false
}
```

**Version (please complete the following information):**
 - Checkov Version 2.0.823 (probably more versions)

**Additional context**
See the [migration guide](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade).

gruebel:
Hi @HorizonNet thanks for creating the issue 💪 

We will use it to track our effort to complain with the recent changes in the new provider version.

babyhuey:
Does closing this issue mean that the issue is fixed? I'm still seeing failures due to my changes to support 4.x branch

MrHash:
+1 specifically it's not recognizing the CMK encryption config on some buckets for me

gruebel:
@MrHash I checked and you are right the CMK check was not ported to the new syntax will add a fix. @babyhuey any specific use cases, which don't work for you?

babyhuey:
So the biggest one I've found is CKV_AWS_19. It is saying to "Ensure all data stored in the S3 bucket is securely encrypted at rest", but I definitely have a aws_s3_bucket_server_side_encryption_configuration defined for the buckets

Also Check: CKV_AWS_145: "Ensure that S3 buckets are encrypted with KMS by default"
[...]

PR #2798, opened 2022-04-11
add grant cases for S3 ACL

gruebel:

Fixes https://github.com/bridgecrewio/checkov/issues/2399#issuecomment-1095191013

Added use cases for S3 Bucket ACL in combination with `access_control_policy` and `grant`

### DIFF
--- a/checkov/terraform/checks/graph_checks/aws/S3PublicACLRead.yaml
+++ b/checkov/terraform/checks/graph_checks/aws/S3PublicACLRead.yaml
@@ -137,6 +137,24 @@ definition:
                   attribute: acl
                   operator: not_equals
                   value: authenticated-read
+            - and:
+                - cond_type: attribute
+                  resource_types:
+                    - aws_s3_bucket_acl
+                  attribute: access_control_policy
+                  operator: exists
+                - or:
+                    - cond_type: attribute
+                      resource_types:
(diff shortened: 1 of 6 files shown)


Saarett (approved):
Nice new addition! Quick thought, what about `aws-exec-read, bucket-owner-read, bucket-owner-full-control, log-delivery-write`?

gruebel:
> Nice new addition! Quick thought, what about `aws-exec-read, bucket-owner-read, bucket-owner-full-control, log-delivery-write`?

@Saarett good question. Those are canned ACLs and have nothing to do with `access_control_policy` and `grant` 😄 
But they are also not so harmful in a sense of making your bucket/objects public for everyone.
- `aws-exec-read` gives AWS EC2 read permissions to retrieve an AMI, which was stored in the bucket
- `bucket-owner-read` and `bucket-owner-full-control` those define the permissions, which the bucket owner has and the one who uploaded them, so usually it is not a concern, but depending on the situation you may not want the object uploader to be able to retrieve object again. So, you usually protect yourself with other means from a malicious uploader, who don't don't trust completely, but I think this check is already complex enough and we have others, which give recommendations on how to protect yourself from an evil uploader.
- `log-delivery-write` this used by AWS to write logs to the bucket

PR #2800, opened 2022-04-12
Added output limitation

ayajbara:

limit the lines of the printed output

### DIFF
--- a/checkov/common/output/record.py
+++ b/checkov/common/output/record.py
@@ -16,7 +16,7 @@ init(autoreset=True)
 
 DEFAULT_SEVERITY = "none"  # equivalent to a score of 0.0 in the CVSS v3.0 Ratings
 
-OUTPUT_VIOLATION_LEN = force_int(os.getenv('CHECKOV_OUTPUT_VIOLATION_LENGTH')) or 50
+OUTPUT_CODE_LINE_LIMIT = force_int(os.getenv('CHECKOV_OUTPUT_CODE_LINE_LIMIT')) or 50
 
 class Record:
     check_id = ""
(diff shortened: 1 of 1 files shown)


checkov/common/output/record.py
```diff
@@ -15,6 +15,7 @@
 
 DEFAULT_SEVERITY = "none"  # equivalent to a score of 0.0 in the CVSS v3.0 Ratings
 
+output_violation_len = os.environ.get('CHECKOV_OUTPUT_VIOLATION_LENGTH', 50)
```
gruebel, line 18:
```suggestion
OUTPUT_VIOLATION_LEN = force_int(os.getenv('CHECKOV_OUTPUT_VIOLATION_LENGTH') or 50
```
since it is static, you usually use upper case for the name.
if an env var is used, then it is always a string, we have a function which forces an int cast, if it fails then `None` is returned. In that case the `or` part will change it to 50 🙂 and don't forget to import the function 😄

gruebel (commented):
nice job 💪

checkov/common/output/record.py
```diff
@@ -10,11 +10,13 @@
 from checkov.common.models.enums import CheckResult
 from checkov.common.typing import _CheckResult
 from checkov.common.util.file_utils import convert_to_unix_path
+from checkov.common.util.type_forcers import force_int
 
 init(autoreset=True)
 
 DEFAULT_SEVERITY = "none"  # equivalent to a score of 0.0 in the CVSS v3.0 Ratings
 
+OUTPUT_VIOLATION_LEN = force_int(os.getenv('CHECKOV_OUTPUT_VIOLATION_LENGTH')) or 50
```
nimrodkor, line 19:
```suggestion
OUTPUT_VIOLATION_LEN = force_int(os.getenv('CHECKOV_OUTPUT_CODE_LINE_LIMIT')) or 50
```

checkov/common/output/record.py
```diff
@@ -118,6 +120,9 @@ def _code_line_string(code_block: List[Tuple[int, str]], colorized: bool = True)
         color_codes = (Fore.WHITE if colorized else "", Fore.YELLOW if colorized else "")
         last_line_number_len = len(str(code_block[-1][0]))
 
+        if len(code_block) >= OUTPUT_VIOLATION_LEN:
+            return f'\t\t{color_codes[1]}Text is too long.'
```
nimrodkor, line 124:
```suggestion
            return f'\t\t{color_codes[1]}Code lines for this resource are too many. Please use IDE of your choice to review the file.'
```

nimrodkor (approved):
Very nice!

PR #2801, opened 2022-04-12
Alicloud k8s Ensure that network plugin for k8s policy support is installed

JamesWoolfenden:

### DIFF
--- a/checkov/terraform/checks/resource/alicloud/K8sEnableNetworkPolicies.py
+++ b/checkov/terraform/checks/resource/alicloud/K8sEnableNetworkPolicies.py
@@ -32,31 +32,26 @@ class K8sEnableNetworkPolicies(BaseResourceCheck):
                 if addon.get("name")
             ]
 
-            # either addon will do
-            if not any(name in names for name in ("flannel", "terway-eniip")):
-                self.evaluated_keys = ["addons/[0]/config"]
-                return CheckResult.FAILED
-            else:
-                if "terway-eniip" in names:
-                    if conf.get("pod_vswitch_ids") and isinstance(conf.get("pod_vswitch_ids"), list):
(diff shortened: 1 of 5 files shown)


checkov/terraform/checks/resource/alicloud/K8sEnableNetworkPolicies.py
```diff
@@ -0,0 +1,68 @@
...
+        # addons
+        if conf.get("addons") and isinstance(conf.get("addons"), list):
+            names = []
+            if len(conf.get("addons")) > 1:
+                addons = conf.get("addons")
+                for addon in addons:
+                    names.append(addon.get("name")[0])
+            else:
+                addons = conf.get("addons")[0]
+                names = addons.get("name")
```
gruebel, line 36:
```suggestion
            names = [
                addon["name"][0]
                for addon in conf["addons"]
                if addon.get("name")
            ]
```
I think this should do the same

checkov/terraform/checks/resource/alicloud/K8sEnableNetworkPolicies.py
```diff
@@ -0,0 +1,68 @@
...
+
+            # either addon will do
+            if not any(name in names for name in ("flannel", "terway-eniip")):
+                self.evaluated_keys = ["addons/[0]/config"]
+                return CheckResult.FAILED
+            else:
```
gruebel, line 42: I think you don't need this `if` block, because in the `else` block you check for each of the names and if they don't exist, you fail it at the end anyway.

JamesWoolfenden: you can guess which bit i wrote first!

gruebel (approved):
🍻

PR #2810, opened 2022-04-13
upgrade hcl2

### LINKED ISSUE #2808: HCL2 parser cannot parse functions with comments interleaved in the arguments.
**Describe the issue**
The HCL2 parser fails to parse a file that contains an expression with a Terraform function call that contains comments interleaved within the arguments.

**Example Value**
A file that contains the following exaple variable will fail to parse.
```hcl
variable "example" {
  default = function(
    # this comment is fine
    argument1,
    # this comment causes a parsing error
    argument2
    # this comment is fine
  )
}
```

This seems to be a replicated issue in the downstream as well > https://github.com/amplify-education/python-hcl2/issues/95.

I have opened a PR to fix this in the bridgecrewio specific parser > https://github.com/bridgecrewio/python-hcl2/pull/29.

**Question**
Is the bridgecrewio HCL2 Parser intened to be merged upstream?
If not, I will implement the change in Amplify's codebase separately.

**An aside**
Checkov is an awesome tool, it makes the jobs of myself and the rest of the Platform/DevOps Engineers on my team so much easier!

gruebel:

Fixes #2808 

@domroutley thanks for the great work on the parser 🚀

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -33,7 +33,7 @@ dlint = "*"
 #
 # REMINDER: Update "install_requires" deps on setup.py when changing
 #
-bc-python-hcl2 = "==0.3.38"
+bc-python-hcl2 = "==0.3.39"
 deep_merge = "*"
 tabulate = "*"
 colorama="*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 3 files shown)


schosterbarak (changes requested):
@gruebel we should also update setup.py

gruebel:
@schosterbarak good catch 🙈 added

PR #2820, opened 2022-04-14
cache twistcli at .bridgecrew directory

NoaAzoulay:

Improve checkov with api token startup time by caching twistcli download locally 
The twistcli binary is cached at the .bridgecrew directory (where API tokens are stored). 
If twistcli binary is older than a week - purge and download new. otherwise use the existing binary.


gruebel (commented):
looks good, I think writing a test for this (what happens inside `setup_twictcli`) would be great, creating the subfolder, downloading it, if it downloaded and 1 week old, delete it and download it again. This could become quite tricky, because you need to mock the download, put something a temp dir, change time invoke it again, etc.

schosterbarak:
> looks good, I think writing a test for this (what happens inside `setup_twictcli`) would be great, creating the subfolder, downloading it, if it downloaded and 1 week old, delete it and download it again. This could become quite tricky, because you need to mock the download, put something a temp dir, change time invoke it again, etc.

@gruebel maybe an integration test with API token with timeout for deletion of 10 seconds

checkov/sca_package/scanner.py
```diff
@@ -2,25 +2,35 @@
...
 from typing import Tuple, Dict, Any
 
 from aiomultiprocess import Pool
 
+from checkov.common.bridgecrew.platform_key import bridgecrew_dir
 from checkov.common.bridgecrew.vulnerability_scanning.integrations.package_scanning import package_scanning_integration
 
 TWISTCLI_FILE_NAME = 'twistcli'
-
+SEC_IN_WEEK = 604800
```
schosterbarak, line 16: maybe put in a configurable environment variable. will also help with testing

checkov/sca_package/scanner.py
```diff
@@ -2,35 +2,55 @@
...
+from checkov.common.bridgecrew.platform_key import bridgecrew_dir
 from checkov.common.bridgecrew.vulnerability_scanning.integrations.package_scanning import package_scanning_integration
 
+
 TWISTCLI_FILE_NAME = 'twistcli'
+SEC_IN_WEEK = 604800
+
+
+def get_expiration_time_in_sec():
+    return int(os.getenv("EXPIRATION_TIME_IN_SEC", SEC_IN_WEEK))
```
nimrodkor, line 21: @NoaAzoulay 2 things:
1. Please prefix all env vars with `CHECKOV_`
2. do we need to wrap this as a function? Can it be written in-line in the caller function?

tests/sca_package/conftest.py
```diff
@@ -6,6 +7,21 @@
 from checkov.common.bridgecrew.platform_integration import BcPlatformIntegration, bc_integration
 
 
+@pytest.fixture(scope="module", autouse=True)
+def init_env_var():
```
nimrodkor, line 11: @NoaAzoulay required?

nimrodkor (commented):
Very nice @NoaAzoulay !

checkov/sca_package/scanner.py
```diff
@@ -2,35 +2,51 @@
...
             raise
 
+    def should_download(self) -> bool:
+        if not self.twistcli_path.exists():
+            return True
+        last_modification = os.stat(self.twistcli_path)
+        file_age = (time.time() - last_modification.st_mtime)
+        if file_age >= int(os.getenv("CHECKOV_EXPIRATION_TIME_IN_SEC", CHECKOV_SEC_IN_WEEK)):
+            return True
+        return False
```
nimrodkor, line 42:
```suggestion
        return file_age >= int(os.getenv("CHECKOV_EXPIRATION_TIME_IN_SEC", CHECKOV_SEC_IN_WEEK))
```

nimrodkor (approved):

tests/sca_package/test_scanner.py
```diff
@@ -1,60 +1,50 @@
...
+    twistcli_path = tmp_path / "twistcli"
+    twistcli_path.touch()
+    scanner.twistcli_path = twistcli_path
 
-    def download_twistcli(cli_file_name: Path):
-        cli_file_name.touch()
+    # then
+    assert not scanner.should_download()
+    # cleaning mock file
+    twistcli_path.unlink()
```
gruebel, line 33: no need to clean up, since you are using `tmp_path` from `pytest`, it will be cleaned up automatically after each test.

tests/sca_package/conftest.py
```diff
@@ -1,3 +1,4 @@
+import os
```
gruebel, line 1: if this import is not used, please remove it 🙂

gruebel (commented):
nice job, 2 small things

gruebel (approved):
🥇

PR #2822, opened 2022-04-14
Whorfnsignore

eurogig:
Made changes to ensure the admission controller doesn't block manifests from critical namespaces but also allows for customisation of what is a critical namespace via a new configMap.


admissioncontroller/k8s/deployment.yaml
```diff
@@ -59,7 +60,12 @@ spec:
         - name: BC_SOURCE
           value: admissionController
         - name: CKV_GITHUB_CONFIG_FETCH_DATA
-          value: "False"
+          value: "False"     
+        - name: DEBUG
+          valueFrom:
+            configMapKeyRef:
+              name: whorfconfig # The ConfigMap this value comes from.
+              key: debug        # Are we in debug mode?
```
schosterbarak, line 68: do we want to keep it in debug?

admissioncontroller/whorf.py
```diff
@@ -31,10 +40,16 @@ def validating_webhook():
         webhook.logger.error('K8s UID failed security checks. Request rejected!')
         return admission_response(False, uid, response)
 
+    # check we're not in the kube-system namespace
+    namespace = request_info["request"].get("namespace")
+    if namespace in ignore_list:
+        print("Namespace in ignore list. Ignoring validation")
```
schosterbarak, line 46: use the logger over print

eurogig: Both are present.  Shall I delete print anyway?

admissioncontroller/k8s/deployment.yaml
eurogig: Debug status is dictated by the config map which is by default 'no'

nimrodkor (approved):

PR #2823, opened 2022-04-14
handling not valid text file that are formatted as accepted dockerfile

itai1357:
handling not valid text file that are formatted as accepted dockerfile


tests/dockerfile/test_runner.py
```diff
@@ -266,6 +266,16 @@ def test_wrong_check_imports(self):
 
         assert len(check_imports) == 0, f"Wrong imports were added: {check_imports}"
 
+    def test_get_files_definitions(self):
+        current_dir = os.path.dirname(os.path.realpath(__file__))
```
gruebel, line 270:
```suggestion
        current_dir = os.path.dirname(__file__)
```
please try to avoid the usage of `os.path.realpath`, it is also not needed here 🙂

gruebel (approved):
easy peasy lemon squeezy 🍋

PR #2831, opened 2022-04-17
Leverage twistlock api in checkov

Eliran-Turgeman:

Changed checkov to leverage the twistlock wrapper API.

Package scanning won't require installing twistcli client and will store results in the cache.
Image scanning will still require to install the twistcli client but will utilize the cache as well.


checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
```diff
@@ -109,5 +114,29 @@ def scan(self, docker_image_id: str, dockerfile_path: str, skip_extract_image_na
...
+            logging.error(response_json["outputData"])
+            return {}
+
+        elif output_type == "Result":
+            return json.loads(
+                decompress_file_gzip_base64(
+                    response.json()["outputData"]
+                )
+            )
+        return {}
```
itai1357, line 139: i suggest one of the next here:
1. writing a comment here for describing that output_type == "Empty" in case it reaches to line 139, 
2.  adding third 'elif' (output_type == "Empty") statement and another else
in the else segment, just raise an error, so we are monitored in case there is new output_type which are not known here. 
basically, it is up to you
@nimrodkor what I suggest as option 2 is good practice or too much according to your opinion

checkov/sca_package/scanner.py
```diff
@@ -2,103 +2,110 @@
...
-        command: str,
-        input_path: Path,
-        output_path: Path,
-    ) -> Dict[str, Any]:
+    def run_scan(self, input_path: Path) -> dict:
         logging.info(f"Start to scan package file {input_path}")
-        process = await asyncio.create_subprocess_shell(
-            command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+
+        headers = merge_dicts(
```
itai1357, line 55: maybe create a method "get_header" in this class. we have the same logic in line 22.
if it isn't to hard, the best is create some "global" function "get_header" (which can be part of 'bc_integration') and than we can use also it instead the header-creation in image_scanner.py:118 (and apparently in a lot of other places), but it may be more complicated, so let's start do it in this class

checkov/sca_package/scanner.py
```diff
@@ -2,103 +2,110 @@
...
+        current_state = "Empty"
+        desired_state = "Result"
+
+        headers = merge_dicts(
+            get_default_get_headers(bc_integration.bc_source, bc_integration.bc_source_version),
+            {"Authorization": bc_integration.get_auth_token()},
+        )
+        response = requests.Response()
+
+        while current_state != desired_state:
```
itai1357, line 93: add some limitation. (e.g: after 2 minutes, abort and log it)

tests/sca_package/test_scanner.py
```diff
@@ -1,137 +1,102 @@
...
-def test_setup_twistcli_exists(mocker: MockerFixture, tmp_path: Path):
-    # given
-    scanner = Scanner()
 
-    integration_mock = MagicMock()
-    mocker.patch(
-        "checkov.common.bridgecrew.vulnerability_scanning.integrations.package_scanning.package_scanning_integration.download_twistcli",
-        side_effect=integration_mock,
-    )
+def assert_equal_dicts(a, b):
```
itai1357, line 9: if we use a={'a': 1}, b= {'a': 2} we got False-Positive
I guess that you have any assumption regarding the dicts 'a' and 'b'. please write them in the documentation (or as a comment) of the function

tests/sca_package/test_scanner.py
```diff
@@ -1,137 +1,102 @@
...
+                return False
+    return True
 
-    # then
-    assert twistcli_path.exists()
-    integration_mock.assert_not_called()
 
-
-def test_setup_twistcli_not_exists(mocker: MockerFixture, tmp_path: Path):
+@responses.activate
```
itai1357, line 23: nice decorator! Big Like

tests/sca_package/test_scanner.py
```diff
@@ -1,137 +1,102 @@
...
-    assert not twistcli_path.exists()
+    assert assert_equal_dicts(result, scan_result2)
+    responses.assert_call_count(mock_bc_integration.bc_api_url + "/api/v1/vulnerabilities/scan", 1)
+    assert len(responses.calls) >= 2
 
 
-@pytest.mark.asyncio
-async def test_run_scan(mocker: MockerFixture, tmp_path: Path, mock_bc_integration, scan_result):
+@responses.activate
+def test_run_scan_fail_on_scan(mock_bc_integration):
```
itai1357, line 74: nice test!!

itai1357 (commented):
Great work and awesome tests! some small comments

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -21,6 +21,7 @@
 from tqdm import trange
 from urllib3.exceptions import HTTPError, MaxRetryError
 
+from checkov.common.bridgecrew.bc_source import get_source_type
```
nimrodkor, line 24: @Eliran-Turgeman necessary import?

checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
nimrodkor: Would log an info log in this case, something like `Got an empty result for file {file_name}`.
1 - doesn't help the average user
2 - we can't monitor the cli runs on user computers... So we'll have to let them know, gently, so they can reach out. raising an error will crash the scan, and we don't want that if we did something wrong...

checkov/sca_package/scanner.py
```diff
@@ -2,103 +2,110 @@
...
-            self.twistcli_path.unlink()
-        logging.info('twistcli file removed')
-
-    def scan(self, input_output_paths: "Iterable[Tuple[Path, Path]]", cleanup_twictcli: bool = True) \
-            -> "Sequence[Dict[str, Any]]":
-        self.setup_twictcli()
+        self.base_url = bc_integration.api_url
+        self.headers = merge_dicts(
+            get_default_get_headers(bc_integration.bc_source, bc_integration.bc_source_version),
+            {"Authorization": bc_integration.get_auth_token()},
```
nimrodkor, line 24: @Eliran-Turgeman This might be a time-limited token, and so should be computed as close to the request as possible. It contains a cache by itself, so don't worry calling it too many times :)

checkov/sca_package/scanner.py
nimrodkor: Agree with @itai1357 , Let's limit it to 60 seconds please

nimrodkor (commented):
Very nice work @Eliran-Turgeman !
Need some cleaning up around the requests to reuse code as much as possible

Eliran-Turgeman:
Thanks for your review @itai1357 @nimrodkor!
I addressed all your comments, the PR is ready for a second review :)

There's a small addition i'll have to make regarding the input validation on docker image id (make it in a sha256 format) which I am testing at the moment

checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
```diff
@@ -109,5 +113,28 @@ def scan(self, docker_image_id: str, dockerfile_path: str, skip_extract_image_na
             logging.error("Failed to scan docker image", exc_info=True)
             return 1
 
+    def get_scan_results_from_cache(self, image_id: str) -> Dict[str, Any]:
+        image_id_encode = quote_plus(':'.join(['sha256', image_id]))
```
gruebel, line 117:
```suggestion
        image_id_encode = quote_plus(f"sha256:{image_id}")
```
better to use a f-string literal here, also looks nicer 😉

checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
```diff
@@ -109,5 +113,28 @@ def scan(self, docker_image_id: str, dockerfile_path: str, skip_extract_image_na
...
+        output_type = response_json["outputType"]
+
+        if output_type == "Error":
+            logging.error(response_json["outputData"])
+            return {}
+
+        elif output_type == "Result":
+            return json.loads(
+                decompress_file_gzip_base64(
+                    response.json()["outputData"]
```
gruebel, line 133:
```suggestion
                    response_json["outputData"]
```
no need to do it again

checkov/common/util/file_utils.py
```diff
@@ -1,3 +1,36 @@
...
+
+
+def decompress_file_gzip_base64(compressed_file_body: str) -> str:
+    try:
+        # 1. convert string to bytes
+        # 2. decode base64 data
+        # 3. wrap decoded binary data with BytesIO to enable reading
+        # 4. gunzip compressed data
+        decoded_utf8 = bytes(compressed_file_body, "utf-8")
+        decoded_base64 = base64.b64decode(decoded_utf8)
```
gruebel, line 31:
```suggestion
        base64_bytes = compressed_file_body.encode("utf-8")
        decoded_base64 = base64.b64decode(base64_bytes)
```
to keep similar to the compress function code

checkov/sca_package/scanner.py
```diff
@@ -5,116 +5,102 @@
...
-    def setup_twictcli(self) -> None:
-        try:
-            if self.should_download():
-                if not os.path.exists(bridgecrew_dir):
-                    os.makedirs(bridgecrew_dir)
-                self.cleanup_twictcli()
-                package_scanning_integration.download_twistcli(self.twistcli_path)
-        except Exception:
-            logging.error("Failed to setup twictcli for package scanning", exc_info=True)
-            raise
```
gruebel, line 33: just to make sure, do we remove already what Noa implemented today?

gruebel (commented):
nice work 💪

checkov/sca_package/scanner.py
nimrodkor: We spoke, it needs to be implemented on the sca_image runner - but on the sca_package runner it will become obsolete soon

checkov/sca_package/scanner.py
```diff
@@ -5,116 +5,102 @@
...
-        logging.info(f"Successfully scanned package file {input_path}")
+            if current_state == "Error":
+                logging.error(response_json["outputData"])
+                return {}
 
-        # read and delete the report file
-        scan_result = json.loads(output_path.read_text())
-        output_path.unlink()
+            time.sleep(SLEEP_DURATION)
+            total_sleeping_time += SLEEP_DURATION
```
nimrodkor, line 96: @Eliran-Turgeman let's put this line after the `if total_sleeping_time > MAX_SLEEP_DURATION:` to avoid waiting without making a call

tests/sca_package/test_scanner.py
nimrodkor: Need to fix this still @Eliran-Turgeman

checkov/sca_package/scanner.py
itai1357: @Eliran-Turgeman just making sure you fixed it

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -641,5 +641,16 @@ def repo_matches(self, repo_name):
         # matches xyz_org/repo or org/repo (where xyz is the BC org name and the CLI repo prefix from the platform)
         return re.match(re.compile(f'^(\\w+_)?{self.repo_id}$'), repo_name) is not None
 
+    def get_default_headers(self, request_type: str) -> dict:
```
itai1357, line 644: nice!

checkov/sca_package/scanner.py
Eliran-Turgeman: Yeah done, refer to lines 95 - 100

checkov/sca_image/runner.py
```diff
@@ -52,8 +62,8 @@ def scan(
             image_scanner.cleanup_scan()
             raise
 
-    @staticmethod
     async def execute_scan(
```
itai1357, line 65: why did you choose to put the method "execute_scan" in the runner rather than in "image_scanner.py"?

checkov/sca_image/runner.py
```diff
@@ -73,8 +83,26 @@ async def execute_scan(
             logging.error(stderr.decode())
             return {}
 
-        # read and delete the report file
+        # read the report file
         scan_result: Dict[str, Any] = json.loads(output_path.read_text())
+
+        # upload results to cache
+        image_id_sha = ':'.join(['sha256', image_id])
```
itai1357, line 90: I suggest to use here also 'quote_plus'
`image_id_sha = quote_plus(f"sha256:{image_id}")`
(as you did in the image_scanner.py)

tests/sca_package/test_runner.py
```diff
@@ -134,59 +134,39 @@ def test_prepare_and_scan(mocker: MockerFixture, scan_result):
     assert runner._code_repo_path == EXAMPLES_DIR
 
 
-def test_prepare_and_scan_sca_package_scan_disabled(mocker: MockerFixture, scan_result):
```
itai1357, line 137: why did you delete it?

itai1357 (commented):
Awesome. some small questions. if the reasons are good, tell me and I will apptove

checkov/sca_image/runner.py
Eliran-Turgeman: It was originally in the runner, I haven't moved it.
Should I move it to the image_scanner?

checkov/sca_image/runner.py
Eliran-Turgeman: Will modify to f-string, but I think the usage of `quote_plus` isn't needed and might lead to undesired behavior since the id is passed in the body and not the url itself so it won't be decoded

tests/sca_package/test_runner.py
Eliran-Turgeman: This test is verifying that the feature flag around sca package scanning is working and returns an empty result.
Since we wanted to delete the feature flag around this feature, I deleted the test, otherwise it would fail.

itai1357 (approved):
Great work!

PR #2837, opened 2022-04-19
Document and enable monchrome console output

### LINKED ISSUE #2805: No colour option
**Describe the issue**
Checkov cli output is heavily targeted for a dark background. Using a light theme makes it very hard to read.
Is there a way to remove color or control the theme used?

**Version:**
 - Checkov Version 2.0.1050

**Additional context**
Small search I found you use termcolor: https://github.com/bridgecrewio/checkov/blob/d55c68ff909c65e341e91938c13a155c99601f5f/checkov/common/output/report.py#L16
but I couldn't find a way to "mute" it. Maybe simply overwriting that function "colored" with a dummy lambda would do it.

MooreDerek:
Resolve #2805 by using the `termcolor.py` existing setting of  `ANSI_COLORS_DISABLED`

### DIFF
--- a/checkov/common/output/record.py
+++ b/checkov/common/output/record.py
@@ -18,6 +18,8 @@ DEFAULT_SEVERITY = "none"  # equivalent to a score of 0.0 in the CVSS v3.0 Ratin
 
 OUTPUT_CODE_LINE_LIMIT = force_int(os.getenv('CHECKOV_OUTPUT_CODE_LINE_LIMIT')) or 50
 
+ANSI_COLORS_DISABLED  = bool(os.getenv('ANSI_COLORS_DISABLED'))
+
 class Record:
     check_id = ""
     check_name = ""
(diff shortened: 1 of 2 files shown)


checkov/common/output/record.py
```diff
@@ -165,7 +167,8 @@ def to_string(self, compact: bool = False, use_bc_ids: bool = False) -> str:
         )
         code_lines = ""
         if self.code_block:
-            code_lines = "\n{}\n".format("".join([self._code_line_string(self.code_block)]))
+            ANSI_COLORS_DISABLED
```
nimrodkor, line 170: @MooreDerek is this line necessary?

MooreDerek: @nimrodkor The code in `_code_line_string` actually handles the adding of text colour itself - it doesn't use the `colored` function. Happy to refactor to use the same functionality as elsewhere.
https://github.com/bridgecrewio/checkov/blob/master/checkov/common/output/record.py#L124

checkov/common/output/record.py
```diff
@@ -165,7 +167,8 @@ def to_string(self, compact: bool = False, use_bc_ids: bool = False) -> str:
...
+            code_lines = "\n{}\n".format("".join([self._code_line_string(self.code_block, not(ANSI_COLORS_DISABLED))]))
```
nimrodkor, line 171: @MooreDerek This is what I meant - not sure why we need the first line:
```suggestion
            code_lines = "\n{}\n".format("".join([self._code_line_string(self.code_block, not(ANSI_COLORS_DISABLED))]))
```

nimrodkor (approved):
TY!

PR #2846, opened 2022-04-20
Flake8 - All Rules followed Outside of Checkov and Tests

njgibbon:
Hello,

I have saw @gruebel slowly making way through the linting and this is to try to help with that.

This change handles all pep8 rules for any python outside of the main `./checkov` and `./tests` directory.

1 unused import, mainly whitespace and newlines.

I only did a minor refactor for the line-length limits. I have been a bit more DRY with one output message and otherwise have used a nice way for creating f strings in a pythonic multi-line way which I learned about through doing this.

e.g
```
foo = "foo"
bar = "bar"
new_string = (
  f'{foo}/'
  f'{bar}'
)
print(new_string)
```

**License**

### DIFF
--- a/admissioncontroller/whorf.py
+++ b/admissioncontroller/whorf.py
@@ -1,5 +1,5 @@
 from flask import Flask, request, jsonify
-from os import environ, remove, getenv 
+from os import remove, getenv
 import logging
 import json
 import subprocess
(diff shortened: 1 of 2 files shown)


gruebel (approved):
nice, looks good 👍

PR #2854, opened 2022-04-21
Ensure API Gateway EDGE is protected by WAF

pdiroot:
Currently only the `REGIONAL` type of API Gateway can pass `CKV2_AWS_29`. I added the `EDGE` type as well.

### DIFF
--- a/tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
+++ b/tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
@@ -1,4 +1,4 @@
-resource "aws_api_gateway_rest_api" "pass" {
+resource "aws_api_gateway_rest_api" "regional" {
   name = var.name
 
   policy = ""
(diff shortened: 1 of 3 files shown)


gruebel:
I thought the edge optimized API Gateway endpoints are protected through the WAF attached to CloudFront or can you really attach one to edge one?

pdiroot:
You actually can attach WAF Web ACL to a Stage of the API Gateway (EDGE). We have it in our terraform infrastructure, and checkov triggers with `CKV2_AWS_29` on it. So this PR is the fix.

tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
```diff
@@ -1,4 +1,14 @@
...
+  name = var.name
+
+  policy = ""
+
+  endpoint_configuration {
+    types = ["REGIONAL"]
+  }
+}
+
+resource "aws_api_gateway_rest_api" "edge" {
```
gruebel, line 11: I think you forgot to change the endpoint type here 🙂

gruebel (changes requested):

gruebel:
> You actually can attach WAF Web ACL to a Stage of the API Gateway (EDGE). We have it in our terraform infrastructure, and checkov triggers with CKV2_AWS_29 on it. So this PR is the fix.

gotcha!

tests/terraform/graph/checks/resources/APIProtectedByWAF/main.tf
pdiroot: True! That's embarrassing) Fixed

gruebel (approved):
nice, great job 🚀

Issue #2853, opened 2022-04-21
Plain text db password in terraform files


Roqche:
Hi,
I'm new to infrastructure security. I know exactly that I have several databases in terraform files such as SQL Server or PostgreSQL with admin passwords in plain text. I don't understand why checkov can't tell me this is some kind of issue? In my IT experience, this is quite a critical issue as anyone with access to the GIT repository can see the passwords. Am I wrong?

I'm attaching two examples.
```
resource "azurerm_sql_server" "sql_server" {
  provider                     = azurerm.providername
  name                         = "${var.sql_server_name}"
  resource_group_name          = azurerm_resource_group.sql_rg.name
  location                     = var.location
  version                      = "12.0"
  administrator_login          = "adminuser"
  administrator_login_password = "My_STRONG_password(!)"
  tags                         = var.tags
}

resource "azurerm_postgresql_flexible_server" "psql" {
  name                   = "${var.postgresql_name}"
  resource_group_name    = azurerm_resource_group.postrgressql_rg.name
  location               = azurerm_resource_group.postrgressql_rg.location
  version                = "13"
  delegated_subnet_id    = var.delegated_subnet_id
  private_dns_zone_id    = azurerm_private_dns_zone.private_dns.id
  administrator_login    = "psqladmin"
  administrator_password = "My_STRONG_password(!)"
  zone                   = "1"

  sku_name   = "B_Standard_B1ms"
}
```

MooreDerek:
@Roqche There two ways of addressing this behaviour that I'm aware of 
1. Use a `var.password` in the tf file and passing in the password as a variable on the command line or
2. Using a `data` to retrieve a password from the azurem key vault. 
So a simple check could be to verify that it's simply not a string, would that work for you?

Roqche:
@MooreDerek Thank you for the solution on how to handle secrets.
But my question was rather, why didn't checkov inform me there are plain text passwords in files?

nimrodkor:
We have some detections for secrets and tokens, but usually they are more random (base64 / hex entropy).
We have room for improvement on the `secrets` runner, are you up for helping out? @Roqche

MooreDerek:
@Roqche The secrets detection is failing on the example above due to the complexity of the password :) The check only looks for alphanumeric characters the `(!)` causes the the regex to fail and the password is not detected. If you test the following examples:

``` hcl
resource "azurerm_postgresql_flexible_server" "psql" {
  name                   = "${var.postgresql_name}"
[...]

PR #2855, opened 2022-04-22
Add test for hard coded DB password

MooreDerek:
Address issue raised in #2853


nimrodkor:
The issue I have here @MooreDerek is that the `conf` dict is after variable resolution - so using `var.password`, if resolved, will still be flagged. I'd much rather leverage the `secrets` runner for this type of things

MooreDerek:
@nimrodkor Understand your point. Hadn't been aware of the `secrets` code. Will have a look and see if I can use that instead. Will close this PR as it's not the best solution.

PR #2856, opened 2022-04-23
Detect relevant frameworks with --file

mikeurbanski1:

When scanning files (not directories), only run the frameworks that are relevant for the file type.

Note: In most cases, runner declare their supported files in the `super().__init__` call to `BaseRunner`. In some cases, runners have to override what gets set by their parent class. For example, Helm uses the K8s runner, but only looks for `Chart.yaml`.

### DIFF
--- a/checkov/cloudformation/runner.py
+++ b/checkov/cloudformation/runner.py
@@ -35,8 +35,9 @@ class Runner(BaseRunner):
         source: str = "CloudFormation",
         graph_class: Type[LocalGraph] = CloudformationLocalGraph,
         graph_manager: Optional[GraphManager] = None,
-        external_registries: Optional[List[BaseRegistry]] = None,
+        external_registries: Optional[List[BaseRegistry]] = None
     ) -> None:
+        super().__init__(file_extensions=['.json', '.yml', '.yaml'])
         self.external_registries = [] if external_registries is None else external_registries
         self.graph_class = graph_class
         self.graph_manager = (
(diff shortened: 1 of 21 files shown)


checkov/sca_image/runner.py
```diff
@@ -20,11 +20,13 @@ class Runner(PackageRunner):
     check_type = CheckType.SCA_IMAGE
 
     def __init__(self) -> None:
+        super().__init__()
         self._check_class: Optional[str] = None
         self._code_repo_path: Optional[Path] = None
         self._check_class = f"{image_scanner.__module__}.{image_scanner.__class__.__qualname__}"
         self.raw_report: Optional[Dict[str, Any]] = None
         self.image_referencers: Optional[ImageReferencer] = None
+        self.file_names = ['Dockerfile']
```
nimrodkor, line 29: @mikeurbanski1 dockerfile has a mask / regex to identify all file names

checkov/cloudformation/runner.py
```diff
@@ -35,8 +35,9 @@ def __init__(
         source: str = "CloudFormation",
         graph_class: Type[LocalGraph] = CloudformationLocalGraph,
         graph_manager: Optional[GraphManager] = None,
-        external_registries: Optional[List[BaseRegistry]] = None,
+        external_registries: Optional[List[BaseRegistry]] = None
     ) -> None:
+        super().__init__(file_extensions=['.json', '.yml', '.yaml'])
```
nimrodkor, line 40: @mikeurbanski1 Should add `.template` at least, I believe cfn doesn't require a specific extension...

nimrodkor (commented):
Nice @mikeurbanski1 !

checkov/sca_image/runner.py
mikeurbanski1: Ah, yeah, I did it correctly in one place but not in this one :|

nimrodkor (approved):

Issue #2844, opened 2022-04-20
False positive for CKV_AZURE_120


SebastianBalle:
**Describe the issue**
If it is related to an existing check, please note the relevant check ID.
Also, explain the logic for this addition/change.

The check CKV_AZURE_120 returns an error when using the configuration `policy_settings { enabled  = true }` inside the resource `azurerm_web_application_firewall_policy` instead of the `waf_configuration { enabled = true }` configuration inside the resource `azurerm_application_gateway`.

The code should however check for which Application Gateway the `azurerm_web_application_firewall_policy` is attached to.

**Examples**
Please share an example code sample (in the IaC of your choice) + the expected outcomes.

The check should return successful if any of the following are true:
- `policy_settings { enabled = true }` for resource `azurerm_web_application_firewall_policy`
-  `waf_configuration { enabled = true }` for resource `azurerm_application_gateway`

**Version (please complete the following information):**
 - Checkov Version 2.0.990

**Additional context**

gruebel:
@SebastianBalle you are right. To support this logic, the current check has to be converted to a graph check and make sure, that there is a policy attached or the WAF is directly enabled.

PR #2857, opened 2022-04-24
Add graph check for firewall policy

MooreDerek:
Fix for #2844 add a graph check for `azurerm_web_application_firewall_policy`


nimrodkor (approved):
TY @MooreDerek !

gruebel (approved):
thanks again 🚀

PR #2394, opened 2022-02-10
Use new runconfig endpoint; add severities to check, skipcheck, hard-fail-on, soft-fail-on

mikeurbanski1:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

- Uses the new endpoint in the platform if you are using an API key
  - This prevents having separate API calls per integration feature (suppressions, custom policies, etc  - but fixes are still separate, because they occur later)
  - The integration features just consume what they need from the endpoint response
  - The API key endpoint returns check metadata (severities, BC ID mapping, guidelines, benchmarks, categories, prisma title, prisma severity, and more), custom policies, suppressions, repo configs)
  - The anonymous public endpoint that returns check ID mapping and guidelines still exists
- Consolidates check metadata logic into a new integration feature class (another step towards splitting out the behemoth of platform_integration)
- Adds severity logic to --check and --skip-check
  - If you put a severity for --check, it will run that severity and higher
  - If you use a severity for --skip-check, it will skip that severity and lower
  - You can combine --check and --skjp-check, in which case it will first apply the --check filter (remove anything not matching), and then substract anything matching --skip-check. It's covered in the docs
- Adds severity logic to --hard-fail and --soft-fail
  - Using a severity with soft-fail will soft-fail for that severity or lower
  - Using a severity with hard-fail will hard fail that severity or higher (effectively the same as the code review settings in the platform)
  - Specific check IDs take precedence over severities - so you can hard-fail HIGH (and above) but soft-fail CKV_123 (assume this is a high severity)
  - Covered more in docs
- Added logic to every runner to handle this
- Eliminated the min-cve-severity-flag and incorporated into the logic described above (this was decided in a thread a couple weeks ago)
- Eliminated the skip-suppressions, skip-custom-policies, and no-guide flags in favor of a single skip-download flag (but the other flags will trigger a warning)
- There is also the start of logic to fetch Prisma policy metadata (severity, title) - not used yet
- It also fetches other policy metadata - benchmarks, category
- This is a pre-req for the following future features:
  - Use Prisma policy metadata when a prisma API key is used
  - Add check / skip / etc logic for benchmarks, categories
  - Inherit more repo config settings to use in the run, mainly code review thresholds

PR #2858, opened 2022-04-24
Flake8 - Unused imports removed under tests directory

njgibbon:
Hello,

Continuing with linting. Focus on Flake8 F class errors - which are mainly unused imports.

This changes all code under `./tests` to restrict the amount of file changes per PR.

**Care**
When doing this you need to be careful that fixing lint errors doesn't accidentally break something.

In 2 files I have removed an unused import but there was a note in the import section saying not to remove some imports because they prevent a circular dependency. I couldn't see how this specific unused import could be required in the test but just need to be careful on that.

There were added in: https://github.com/bridgecrewio/checkov/pull/2394
e.g https://github.com/bridgecrewio/checkov/blob/master/tests/terraform/runner/test_runner.py - not clear how the unused import prevents a cycle or which import lines are in scope.

My hope is that regression checks are ok and my assumption is alright.

**Extra**
I have just noticed that @gruebel has also been doing this specific lint class - https://github.com/bridgecrewio/checkov/pull/2842 - but hopefully this still helps.

**License**

### DIFF
--- a/tests/common/integration_features/test_custom_policies_integration.py
+++ b/tests/common/integration_features/test_custom_policies_integration.py
@@ -1,6 +1,5 @@
 import json
 import os
-import types
 import unittest
 
 from checkov.common.bridgecrew.integration_features.features.custom_policies_integration import \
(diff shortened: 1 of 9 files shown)


nimrodkor:
If it's in the tests directory and the tests pass - it's good enough :)

nimrodkor (approved):
TY!

Issue #2341, opened 2022-02-05
Shielded GKE nodes is default as of provider version 4.0.0


brettcurtis:
**Describe the issue**
I wouldn't expect to get the alert if not defined explicitly. 

**Examples**
https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.0.0

**Version (please complete the following information):**
 - Checkov Version 2.0.780

gruebel:
hi @brettcurtis thanks for creating the issue.

looks like we should adjust our check to respect the new default behaviour. Do you want to do the change and contribute back to `checkov`?

brettcurtis:
Let me look at the guidelines and see if I can figure it out!

PR #2896, opened 2022-04-28
GKE nodes shielded by default in v4.0.0

MooreDerek:
Fix for #2341 
Not sure if there should be a check that the correct GCP provider is being used, or the assumption is that the latest is being used?


gruebel (approved):
thanks for adjustment 🚀 the v4 provider is out for almost 6 months, so lets accept the new default 🙂

PR #2898, opened 2022-05-01
ut to check image reference flow is triggering image scan flow calls

Eliran-Turgeman:

Added ut that checks that the image reference flow is triggering the image scanning flow calls.
Moved uploading to cache to a separate function.

### DIFF
--- a/checkov/sca_image/runner.py
+++ b/checkov/sca_image/runner.py
@@ -91,12 +91,20 @@ class Runner(PackageRunner):
         scan_result: Dict[str, Any] = json.loads(output_path.read_text())
 
         # upload results to cache
+        self.upload_results_to_cache(output_path, image_id)
+
+        # delete the report file
+        output_path.unlink()
+
+        return scan_result
+
+    def upload_results_to_cache(self, output_path, image_id):
         image_id_sha = f"sha256:{image_id}" if not image_id.startswith("sha256:") else image_id
 
         request_body = {
-                "compressedResult": compress_file_gzip_base64(str(output_path)),
-                "compressionMethod": "gzip",
-                "id": image_id_sha
+            "compressedResult": compress_file_gzip_base64(str(output_path)),
+            "compressionMethod": "gzip",
+            "id": image_id_sha
         }
         response = requests.request(
             "POST", f"{self.base_url}/api/v1/vulnerabilities/scan-results",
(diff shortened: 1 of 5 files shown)


checkov/sca_image/runner.py
```diff
@@ -91,12 +91,20 @@ async def execute_scan(
...
 
         # upload results to cache
+        self.upload_results_to_cache(output_path, image_id)
+
+        # delete the report file
+        output_path.unlink()
+
+        return scan_result
+
+    def upload_results_to_cache(self, output_path: Path, image_id: str) -> None:
```
itai1357, line 101: nice method!

tests/sca_image/examples/.github/workflows/vulnerable_container.yaml
```diff
@@ -0,0 +1,29 @@
+on: pull_request
```
itai1357, line 1: what is this file? when do you use it?

tests/sca_image/test_runner.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    image_runner = Runner()
+    image_runner.image_referencers = [GHA_Runner()]
+    image_runner.run(root_folder=EXAMPLES_DIR)
+
+    responses.assert_call_count(
+        mock_bc_integration.bc_api_url + f"/api/v1/vulnerabilities/scan-results/{image_id_encoded}", 1)
+
+    assert len(responses.calls) >= 1
```
itai1357, line 52: is it possible to check that "/api/v1/vulnerabilities/scan-results" (the post request) also was called exactly once?
in addition, if you have a way for test a case where the results exist, I think it may be helpful for the future

tests/sca_image/examples/.github/workflows/vulnerable_container.yaml
Eliran-Turgeman: It's an example of a github actions file which is referencing an image (that we want to scan)

tests/sca_image/test_runner.py
Eliran-Turgeman: Since I mock `execute_scan` this call isn't triggered - will try to think of a workaround.

I'll try to add the test you described

tests/sca_image/examples/.github/workflows/vulnerable_container.yaml
itai1357: got it

itai1357 (approved):
LGTM!

Eliran-Turgeman:
@itai1357 I made a few changes since you last reviewed.
* moved the uploading to cache function to be called outside of execute_scan so we will be able to test that it does get uploaded
* limited the test to run on environments that have python3.8+ since on python3.7 the mocking mechanism broke on the async function (on 3.8+ it's working OOTB)

PR #2905, opened 2022-05-02
add bitbucketpipelines runner

schosterbarak:
add image scanning of bitbucket pipelines


checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+from checkov.common.images.image_referencer import ImageReferencer, Image
+from checkov.common.output.report import CheckType
+from checkov.yaml_doc.runner import Runner as YamlRunner
+
+
+class Runner(YamlRunner, ImageReferencer):
+    check_type = CheckType.BITBUCKET_PIPELINES
+
+    def __init__(self):
+        super().__init__()
```
gruebel, line 13: I don't think you need that, when you don't do anything extra inside `__init__`, it will call the parent method anyway

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+    def import_registry(self):
+        return registry
+
+    def _parse_file(self, f):
+        if self.is_workflow_file(f):
+            return super()._parse_file(f)
+
+    def is_workflow_file(self, file_path):
+        """
+        :return: True if the file mentioned is in a is named bitbucket-pipelines.yml. Otherwise: False
```
gruebel, line 27:
```suggestion
        :return: True if the file mentioned is named bitbucket-pipelines.yml. Otherwise: False
```

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        abspath = os.path.abspath(file_path)
+        return abspath.endswith(("bitbucket-pipelines.yml", "bitbucket-pipelines.yaml"))
```
gruebel, line 30:
```suggestion
        return file_path.endswith(("bitbucket-pipelines.yml", "bitbucket-pipelines.yaml"))
```
no need to evaluate the `abspath`, because you are not using it

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+
+        pipelines = workflow.get("pipelines", {})
+        default_pipeline = pipelines.pop("default", {})
+        self.add_default_pipeline_images(file_path, images, default_pipeline)
+
+        self.add_pipeline_images(file_path, images, pipelines)
+
+        return images
+
+    def add_root_image(self, file_path: str, images: [], root_image: str, workflow_line_numbers: dict) -> None:
```
gruebel, line 90:
```suggestion
    def add_root_image(self, file_path: str, images: list, root_image: str, workflow_line_numbers: tuple[int, str]) -> None:
```
I also recommend to add this line to the top of the file, then you can use the built-in types for type hints
```python
from __future__ import annotations
```

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+                if "image" in line_txt and not line_txt.startswith(' '):
+                    start_line = line_number
+                    end_line = line_number
+                    image_id = self.pull_image(root_image)
+                    image_obj = Image(file_path=file_path, name=root_image, image_id=image_id,
+                                      start_line=start_line,
+                                      end_line=end_line)
+                    images.add(image_obj)
+
+    def add_pipeline_images(self, file_path: str, images: [], pipelines: dict):
```
gruebel, line 102:
```suggestion
    def add_pipeline_images(self, file_path: str, images: list, pipelines: dict):
```

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+        if root_image:
+            for line_number, line_txt in workflow_line_numbers:
+                if "image" in line_txt and not line_txt.startswith(' '):
+                    start_line = line_number
+                    end_line = line_number
+                    image_id = self.pull_image(root_image)
+                    image_obj = Image(file_path=file_path, name=root_image, image_id=image_id,
+                                      start_line=start_line,
+                                      end_line=end_line)
+                    images.add(image_obj)
```
gruebel, line 100:
```suggestion
                    image_id = self.pull_image(root_image)
                    image_obj = Image(
                        file_path=file_path,
                        name=root_image,
                        image_id=image_id,
                        start_line=line_number,
                        end_line=line_number,
                    )
                    images.add(image_obj)
```

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+                                if image_id:
+                                    start_line = step_obj_def['__startline__']
+                                    end_line = step_obj_def['__endline__']
+
+                                    image_obj = Image(file_path=file_path, name=image, image_id=image_id,
+                                                      start_line=start_line,
+                                                      end_line=end_line)
+                                    images.add(image_obj)
+
+    def add_default_pipeline_images(self, file_path: str, images: [], default_pipeline: dict):
```
gruebel, line 121:
```suggestion
    def add_default_pipeline_images(self, file_path: str, images: list, default_pipeline: dict):
```

checkov/bitbucket_pipelines/runner.py
```diff
@@ -0,0 +1,135 @@
...
+                if image:
+                    image_id = self.pull_image(image)
+                    if image_id:
+                        start_line = step_obj['__startline__']
+                        end_line = step_obj['__endline__']
+
+                        image_obj = Image(file_path=file_path, name=image, image_id=image_id,
+                                          start_line=start_line,
+                                          end_line=end_line)
+                        images.add(image_obj)
```
gruebel, line 135: this looks very similar to the previous method, probably you can abstract some parts

Saarett (approved):
💯

PR #2907, opened 2022-05-02
adding request_wrapper for overcoming ConnectionErrors

itai1357:
adding request_wrapper for overcoming ConnectionErrors (we encountered 2:
`'Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')`
`('Connection aborted.', OSError(107, 'Socket not connected'))` 
they appear from time to time)

resources I got assisted with:
https://localcoder.org/python-error-104-connection-reset-by-peer
https://stackoverflow.com/questions/52051989/requests-exceptions-connectionerror-connection-aborted-connectionreseterro


checkov/sca_package/scanner.py
```diff
@@ -14,6 +14,8 @@
 
 SLEEP_DURATION = 2
 MAX_SLEEP_DURATION = 60
+REQUEST_MAX_TRIES = os.getenv('REQUEST_MAX_TRIES', 2)
+REQUEST_SLEEP_BETWEEN_TRIES = os.getenv('REQUEST_SLEEP_BETWEEN_TRIES', 0.1)
```
nimrodkor, line 18: @itai1357 Please use `CHECKOV_` as the prefix of every env var we use, and set base retry to 3
Also, shouldn't these be `int`s?

nimrodkor (commented):
Very nice @itai1357 !

checkov/sca_package/scanner.py
itai1357: yep you are right!

itai1357:
> Very nice @itai1357 !

thanks! I will work on UT

checkov/sca_package/scanner.py
```diff
@@ -98,3 +100,16 @@ def parse_api_result(self, origin_file_path: Path, response: str) -> dict:
...
+        # from time to time ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')).
+        remaining_tries = CHECKOV_REQUEST_MAX_TRIES
+        try:
+            remaining_tries -= 1
+            return requests.request(method, url, headers=headers, data=data)
+        except requests.exceptions.ConnectionError as err:
+            if remaining_tries == 0:
+                raise err
+            else:
+                time.sleep(CHECKOV_REQUEST_SLEEP_BETWEEN_TRIES)
```
gruebel, line 115:
```suggestion
                time.sleep(CHECKOV_REQUEST_SLEEP_BETWEEN_TRIES * (CHECKOV_REQUEST_MAX_TRIES - remaining_tries))
```
how about something like this. instead of linear, a bit of adaptive backoff 🙂 so it waits 0.1, then 0.2. exponential could be also a good approach.

gruebel (commented):
nice job, just some possible alternatives on the backoff strategy 😄

checkov/sca_package/scanner.py
itai1357: interesting. OK let's do it =]

checkov/sca_package/scanner.py
```diff
@@ -98,3 +100,17 @@ def parse_api_result(self, origin_file_path: Path, response: str) -> dict:
...
+        remaining_tries = self._request_max_tries
+        while True:
+            try:
+                return requests.request(method, url, headers=headers, data=data)
+            except requests.exceptions.ConnectionError as err:
+                remaining_tries -= 1
+                if remaining_tries == 0:
+                    raise err
+                else:
+                    time.sleep(self._sleep_between_request_tries * (self._request_max_tries - remaining_tries))
```
nimrodkor, line 116: @itai1357 Let's also catch and retry 504 HTTP errors:
```
HTTPError - 504 Server Error: Gateway Time-out for url: https://www.bridgecrew.cloud/api/v1/vulnerabilities/scan
```
Also - let's log an info saying we got the error ..., retrying

itai1357: @nimrodkor resolved

gruebel (approved):
🥇

PR #2911, opened 2022-05-03
Pass module to platform for policy filtering

mikeurbanski1:

Uses the `module` parameter of the run config endpoint to get the appropriate policy set based on token type.

The platform side of this is also live.


kartikp10 (approved):
Looks good 👌🏽

PR #2912, opened 2022-05-03
Kustomize and bicep custom properties

PelegLi:

- inject runner-specific data to bicep and kustomize runners
- added definitions context to bicep runner

### DIFF
--- a/checkov/bicep/runner.py
+++ b/checkov/bicep/runner.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import cast, Type, TYPE_CHECKING, Any, List, Tuple
+from typing import cast, Type, TYPE_CHECKING, Any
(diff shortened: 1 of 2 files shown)


checkov/bicep/runner.py
```diff
@@ -108,6 +109,9 @@ def run(
 
         return report
 
+    def set_definitions_raw(self, definitions_raw: dict[Path, List[Tuple[int, str]]]) -> None:
```
gruebel, line 112:
```suggestion
    def set_definitions_raw(self, definitions_raw: dict[Path, list[tuple[int, str]]]) -> None:
```
and please remove the added imports for `List, Tuple` you can use the built-in types as long as the file has
```python
from __future__ import annotations
```
at the beginning 🙂

gruebel (commented):
good catch 👍

PR #2914, opened 2022-05-03
Scan external private modules

kartikp10:


checkov/terraform/module_loading/loader.py
```diff
@@ -69,7 +76,9 @@ def load(
         if os.path.exists(module_path):
             return ModuleContent(dir=module_path)
 
-        self.logger.debug(f"getting module {self.module_source} version: {self.version}")
+        self.logger.debug(f"Using {self.__class__.__name__} attempting to get module "
+                          f"{self.module_source if '@' not in self.module_source else self.module_source.split('@')[1]} "
+                          f"version: {self.version}")
```
rotemavni, line 81: What happens if the `module_source` has no '@'? The access to `[1]` would fail, is there any way to avoid it?

checkov/terraform/module_loading/loaders/bitbucket_access_token_loader.py
```diff
@@ -0,0 +1,27 @@
...
+    def discover(self):
+        self.username = os.getenv('BITBUCKET_USERNAME', '')
+        self.app_password = os.getenv('BITBUCKET_APP_PASSWORD', '')
+        self.token = os.getenv('BITBUCKET_TOKEN', '')
+
+    def _is_matching_loader(self) -> bool:
+        # https://www.terraform.io/docs/modules/sources.html#bitbucket
+        if self.token:
+            # https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/
+            if self.module_source.startswith("bitbucket.org"):
```
rotemavni, line 16: From the looks of it I feel like there's a lot of code duplication between Bitbucket, Github and Gitlab.

Isn't it possible to extract this logic to `GenericGitLoader._is_matching_loader` (to the parent class) and add the prefix 
(for example `bitbucket.org`) as the class property, and use it instead of the hardcoded string?

rotemavni (commented):
Incredible work!

nimrodkor:
Please take a look at CodeQL findings @kartikp10 🙏

checkov/common/util/file_utils.py
```diff
@@ -8,6 +9,12 @@ def convert_to_unix_path(path: str) -> str:
     return path.replace('\\', '/')
 
 
+def extract_tar_archive(source_path: str, dest_path: str) -> None:
+    tar = tarfile.open(source_path)
+    tar.extractall(path=dest_path)
+    tar.close()
```
gruebel, line 15:
```suggestion
    with tarfile.open(source_path) as tar:
        tar.extractall(path=dest_path)
```
always better to use a context manager 😉

gruebel (commented):
nice job 🍻 , probably it would be good to have an integration test to really test it for one of the flavours, wdyt?

kartikp10:
> nice job 🍻 , probably it would be good to have an integration test to really test it for one of the flavours, wdyt?

good call. I'll add tests for Terraform cloud and Github (private repo)

docs/7.Scan Examples/Terraform Plan Scanning.md
```diff
@@ -71,3 +71,23 @@ checkov -d .terraform # Module TF files.
...
 
 It is worth noting however, that when scanning the `.terraform` directory, Checkov cannot differentiate between third-party and internally written modules. That said, you will benefit from scanning coverage across all of them.
+
+### Scanning Private Terraform Modules
+
+In case third-party modules are stored in a private repository or a private Terraform Cloud registry, you can provide access tokens as environment variables for checkov to attempt to clone those modules. 
+
+| Variable Name          | Description                                                                |
+|------------------------|----------------------------------------------------------------------------|
+| GITHUB_TOKEN           | Github personal access token with read access to the private repository    |
```
mikeurbanski1, line 81: GITHUB_PAT?

kartikp10: Fixed!

PR #2918, opened 2022-05-04
change request body for scan endpoint

gruebel:

since the content encoding was set to `json`, but no json was sent, it results in an error.

### DIFF
--- a/checkov/common/util/http_utils.py
+++ b/checkov/common/util/http_utils.py
@@ -1,14 +1,20 @@
+from __future__ import annotations
+
 import json
 import requests
 import logging
 import time
 import os
-from typing import Optional, Any
+from typing import Any, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


checkov/common/util/http_utils.py
```diff
@@ -58,8 +64,14 @@ def get_default_post_headers(client: SourceType, client_version: str):
...
-def request_wrapper(method: str, url: str, headers: Any, data: Optional[Any] = None, json: Optional[Any] = None,
-                    should_call_raise_for_status: bool = False):
+def request_wrapper(
+    method: str,
+    url: str,
+    headers: dict[str, Any],
+    data: Any | None = None,
+    json: dict[str, Any] | None = None,
+    should_call_raise_for_status: bool = False
+) -> Response | None:
```
itai1357, line 74: nice!

Issue #2373, opened 2022-02-08
CKV2_AWS_35 is failing on a plan scan even when a Nat Gateway is used 


jonprice:
**Describe the issue**
Issue relates to.
Check: CKV2_AWS_35: "AWS NAT Gateways should be utilized for the default route"

which is failing on a terraform plan scan that is using a Nat Gateway. 

**Examples**
check: CKV2_AWS_35: "AWS NAT Gateways should be utilized for the default route"
    	FAILED for resource: aws_route.private_nat_gateway
    	File: /************.json:4955-4975
    
    		4956 |                     "carrier_gateway_id": "",
    		4957 |                     "destination_cidr_block": "0.0.0.0/0",
    		4958 |                     "destination_ipv6_cidr_block": "",
    		4959 |                     "destination_prefix_list_id": "",
    		4960 |                     "egress_only_gateway_id": "",
    		4961 |                     "gateway_id": "",
    		4962 |                     "id": "r-rtb-0***************",
    		4963 |                     "instance_id": "",
    		4964 |                     "instance_owner_id": "",
    		4965 |                     "local_gateway_id": "",
    		4966 |                     "nat_gateway_id": "nat-03*************",
    		4967 |                     "network_interface_id": "",
    		4968 |                     "origin": "CreateRoute",
    		4969 |                     "route_table_id": "rtb-0*************",
    		4970 |                     "state": "active",
    		4971 |                     "timeouts": {
    		4972 |                       "create": "5m"
    		4973 |                     },
    		4974 |                     "transit_gateway_id": "",
    		4975 |                     "vpc_endpoint_id": "",

The plan output above shows a Nat Gateway in use, and no instance_id

**Version (please complete the following information):**
checkov-2.0.812

**Additional context**

gruebel:
hi @jonprice thanks for raising this issue. It is indeed an issue with the Terraform plan output.

PR #2926, opened 2022-05-05
2373 add check for blank instance

MooreDerek:
Fix for #2373 The second test condition in the `yaml` fix is really there to allow for testing as I couldn't figure out how to test with
a plan file.


gruebel (approved):
@MooreDerek thanks, no need for a real plan test, I usually also try to adjust the resources to fit the check, so I don't have to create a real plan test 😄

PR #2938, opened 2022-05-08
Remove incorrect RBAC wildcard

yuvalavra:
K8s RBAC doesn't consider `*/*` as a wildcard, only `*` or `*/<subresource>` (source: https://github.com/kubernetes/kubernetes/blob/acb85012ba9c5e9f723971ef706f396d9629e93e/pkg/apis/rbac/types.go#L45)

Don't check for  `*/*`.


gruebel (approved):
thanks 🍻

PR #2939, opened 2022-05-08
Improve type hints in common

gruebel:

Sorry for the big PR, it somehow grew bigger and bigger, because I had to fix different parts after adding some folders under `common`.
There are only few real code changes, because we sometimes didn't check `None` types. I also added a couple of type libraries for the libraries we use to enable better type checking. Additionally added a file named `py.typed` to symbolize that `checkov` is typed, of course it is not complete yet, but we also don't have nothing 😄 

There couple of `type:ignore` some are limitations, which are intended by `mypy` but some are our doings, which I need to tackle an other time, didn't want to blow it up more than it is already 😵


checkov/common/graph/db_connectors/db_connector.py
```diff
@@ -1,12 +1,22 @@
-class DBConnector:
-    def save_graph(self, local_graph):
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, TypeVar, Generic
+
+if TYPE_CHECKING:
+    from checkov.common.graph.graph_builder.local_graph import LocalGraph
+
+T = TypeVar("T")
```
Saarett, line 8: Why do we need it?

checkov/common/graph/graph_builder/variable_rendering/vertex_reference.py
```diff
@@ -1,11 +1,17 @@
...
-from checkov.common.graph.graph_builder.graph_components.block_types import BlockType
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
+from typing import Any, TypeVar, Generic, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from checkov.common.graph.graph_builder.graph_components.block_types import BlockType  # noqa
+
+_BlockT = TypeVar("_BlockT", bound="BlockType")
```
Saarett, line 9: Why do we need it?

checkov/common/multi_signature.py
```diff
@@ -1,16 +1,33 @@
...
+
 import inspect
 from abc import ABCMeta
 from functools import update_wrapper
-from typing import Callable, Any, TypeVar, Dict, List
+from types import CodeType
+from typing import Callable, Any, TypeVar, cast
+from typing_extensions import Protocol
+
+_MultiT = TypeVar("_MultiT")
```
Saarett, line 10: 🤔

Saarett (approved):
Some questions in the code 👁️

checkov/common/graph/db_connectors/db_connector.py
gruebel: Because `DBConnector` is a generic class, so when someone subclasses it, he can just add the type, which will be returned by the `get_reader_endpoint` and `get_writer_endpoint`

checkov/common/graph/graph_builder/variable_rendering/vertex_reference.py
gruebel: also here `VertexReference` can handle different types of `BlockType`, if it is in a context of Terraform it is a Terraform one and for CloudFormation a CloudFormation one. The `bound` limits the possibility to only allow types which subclass `common.BlockType`. This makes type resolution better down the line, because you know, which flavour exactly it is and not just it is a `common.BlockType` or a subclass of it.

checkov/common/multi_signature.py
gruebel: here I'm still not really sure, how to type hint this whole file. It is basically a place holder, which tells the type checker, that whatever you put in will come out in different places as the same type

PR #2940, opened 2022-05-08
fixed file_path for bicep report

PelegLi:
Strip `file_path` attribute in bicep report from root folder

### DIFF
--- a/checkov/bicep/runner.py
+++ b/checkov/bicep/runner.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import logging
 from pathlib import Path
 from typing import cast, Type, TYPE_CHECKING, Any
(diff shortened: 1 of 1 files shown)


rotemavni (approved):

checkov/bicep/runner.py
```diff
@@ -159,6 +162,9 @@ def add_python_check_results(self, report: Report, runner_filter: RunnerFilter)
                                 record.set_guideline(check.guideline)
                                 report.add_record(record=record)
 
+    def extract_file_path_from_abs_path(self, path: Path) -> str:
+        return f"/{os.path.relpath(str(path), self.root_folder)}"
```
gruebel, line 166:
```suggestion
        return f"/{os.path.relpath(path, self.root_folder)}"
```
it should work without making it a string 🙂

gruebel (commented):
nice 👍

PR #2944, opened 2022-05-09
improve image referencer pull performance

schosterbarak:
inspect if image sha exists locally. if not then pull. 

we should add a rest API call to check if the sha exists for the name remotely on the server.


nimrodkor (approved):
Agreed

gruebel (approved):
just a tiny thing 🙂

PR #2956, opened 2022-05-11
Added run_integration_test.sh script

ayajbara:
Added script that runs integration tests locally.

### DIFF
--- a/integration_tests/run_integration_tests.sh
+++ b/integration_tests/run_integration_tests.sh
@@ -24,7 +24,7 @@ prepare_data () {
 
   python checkov/main.py -s -f repositories/terragoat/terraform/aws/s3.tf --bc-api-key $BC_KEY > checkov_report_s3_singlefile_api_key_terragoat.txt
   python checkov/main.py -s -d repositories/terragoat/terraform/azure/ --bc-api-key $BC_KEY > checkov_report_azuredir_api_key_terragoat.txt
-#  export CHECKOV_EXPERIMENTAL_IMAGE_REFERENCING=True #cant run it on M1 mac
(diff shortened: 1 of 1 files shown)


integration_tests/run_integration_tests.sh
```diff
@@ -0,0 +1,104 @@
...
+   exit 1
+fi
+
+echo $BC_API_URL
+if [[ -z "$BC_API_URL" ]]; then
+   echo "BC_API_URL is missing."
+   exit 1
+fi
+
+docker run node:14.16
```
gruebel, line 69: I recommend to just run `docker -v` or `docker info`

integration_tests/run_integration_tests.sh
```diff
@@ -0,0 +1,104 @@
...
+# Create repositories dir
+mkdir repositories
+cd repositories
+
+echo "Cloning repositories"
+clone_repositories
+
+cd ..
+
+deactivate
```
gruebel, line 84: this will fail, if you didn't activate a venv before, I know, why you are adding it, but I would rather be a bit more careful and check first, if you are in a venv, this can be done by checking if the env var `$VIRTUAL_ENV` is not empty

gruebel (commented):
great job 💪

integration_tests/run_integration_tests.sh
ayajbara: no need to run docker at all because I have added it for "CHECKOV_EXPERIMENTAL_IMAGE_REFERENCING" flag needs and cant run it on Mac M1, so docker deleted

gruebel (approved):
🥇

Eliran-Turgeman:
Looking awesome :fire:

PR #2959, opened 2022-05-12
moved definitions logic from parser class to utils

PelegLi:

### DIFF
--- a/checkov/bicep/utils.py
+++ b/checkov/bicep/utils.py
@@ -1,6 +1,17 @@
 from __future__ import annotations
 
+import os
 from pathlib import Path
+from typing import Optional
+import logging
+
+from checkov.common.runners.base_runner import filter_ignored_paths
+from checkov.runner_filter import RunnerFilter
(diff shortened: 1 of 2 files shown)


checkov/bicep/utils.py
```diff
@@ -23,3 +34,42 @@ def clean_file_path(file_path: Path) -> Path:
...
+                full_path = os.path.join(root, file)
+                files_list.add(Path(full_path))
+    parser = Parser()
+
+    return parser.get_files_definitions(files_list)
+
+
+def create_definitions(
+        root_folder: str,
+        files: "Collection[Path] | None" = None,
```
gruebel, line 58: you forgot to import `from collections.abc import Collection` 🙂

achiar99 (commented):
You dont need to update the imports in the files that use these functions?

PelegLi:
> You dont need to update the imports in the files that use these functions?

they are already updated

achiar99 (approved):

PR #2967, opened 2022-05-14
fix ARM checks and add Bicep tests

gruebel:

Fixed couple of ARM checks, which probably never worked and an other one just supported the old api version. Also added Bicep tests for those checks to make sure they work for both ARM and Bicep.

### DIFF
--- a/checkov/arm/checks/resource/AppServiceAuthentication.py
+++ b/checkov/arm/checks/resource/AppServiceAuthentication.py
@@ -1,3 +1,7 @@
+from __future__ import annotations
+
+from typing import Any
+
 from checkov.common.models.enums import CheckResult, CheckCategories
 from checkov.arm.base_resource_check import BaseResourceCheck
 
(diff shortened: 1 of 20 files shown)


checkov/arm/checks/resource/AppServiceAuthentication.py
```diff
@@ -1,36 +1,39 @@
...
-            if conf["type"] == "Microsoft.Web/sites/config":
-                if "name" in conf:
-                    if "authsettings" in conf["name"]:
+    def scan_resource_conf(self, conf: dict[str, Any]) -> CheckResult:
+        if self.entity_type == "Microsoft.Web/sites/config":
+            if "name" in conf and "authsettings" in conf["name"]:
+                if "properties" in conf and "enabled" in conf["properties"]:
+                    if str(conf["properties"]["enabled"]).lower() == "true":
+                        return CheckResult.PASSED
+                return CheckResult.FAILED
```
tronxd, line 25: Probably it's better to indent `return CheckResult.FAILED` to the same indentation level of `if "name" in conf and "authsettings" in conf["name"]:`

tronxd (approved):
🥇

checkov/arm/checks/resource/AppServiceAuthentication.py
gruebel: yeah, you are totally right 😄

PR #2976, opened 2022-05-17
List platform only policies

kartikp10:

Bug fix:

1. Fixes an issue where even though you use a platform key (`--bc-api-key`) CKV2 policies that do not exist in the platform were still being run. 
2. Added an integration test to confirm that if an API key is used with the `--output-bc-ids` flag, no CKV ids will be present in `checkov --list`. 
3. That said, other check id formats - external checks, platform checks will be present. 

New behavior:

1. When you use --list with an API key, only platform checks will be printed.

### DIFF
--- a/integration_tests/test_checkov_platform_only_policies.py
+++ b/integration_tests/test_checkov_platform_only_policies.py
@@ -1,0 +1,33 @@
+import os
+import platform
+import sys
+import unittest
+import re
+
+from checkov.common.models.consts import ckv_check_id_pattern
+
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
+
+class TestCheckovPlatformOnlyPolicies(unittest.TestCase):
+
+    def test_no_ckv_ids(self):
+        checks_list_path = os.path.join(current_dir, '..', 'checkov_checks_list.txt')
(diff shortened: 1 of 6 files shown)


integration_tests/test_checkov_platform_only_policies.py
```diff
@@ -0,0 +1,33 @@
...
+    def test_no_ckv_ids_api_key(self):
+        checks_list_path = os.path.join(current_dir, '..', 'checkov_checks_list.txt')
+        if sys.version_info[1] == 7 and platform.system() == 'Linux':
+            with open(checks_list_path, encoding='utf-8') as f:
+                for i, line in enumerate(f):
+                    if i in [0, 1]:
+                        # skip the header lines
+                        continue
+                    line = "".join(line.split())
+                    if type(line) == str and line:
```
gruebel, line 23:
```suggestion
                    if line and isinstance(line):
```
most of the time `isinstance` usage is preferred compared to `type` for checking the of a variable.

gruebel (approved):
🥇

integration_tests/test_checkov_platform_only_policies.py
kartikp10: thanks for the tip!

PR #2983, opened 2022-05-19
Created class level cache for policies in CFN BaseCloudsplainingIAMCheck

rotemavni:
Cached `PolicyDocument` result to avoid slow calculation begin done repeatedly.


gruebel (approved):
🏅

Issue #2979, opened 2022-05-18
Check: CKV_GCP_76: should check against 3 strings rather than a boolean


parabolic:
Hello everyone! I am having an issue with the check `CKV_GCP_76`. I've put all of the details below.

**Describe the issue**
Check: CKV_GCP_76: "Ensure that Private google access is enabled for IPV6" checks for a boolean but eventually it accepts 3 string values.
Those 3 values are:
```
ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE
DISABLE_GOOGLE_ACCESS
ENABLE_BIDIRECTIONAL_ACCESS_TO_GOOGLE
```
I've used these commands to find the possible values that this api field accepts.
```sh
gcloud compute networks subnets update --private-ipv6-google-access-type=enable-bidirectional-access <self_link>
gcloud compute networks subnets update --private-ipv6-google-access-type=disable <self_link>
gcloud compute networks subnets update --private-ipv6-google-access-type=enable-outbound-vm-access  <self_link>
```

**Examples**
With the following code I am able to satisfy check `CKV_GCP_76` and `terraform apply` works. 
```tf
resource "google_compute_subnetwork" "subnets" {
....
  private_ipv6_google_access = true
....
}
```
I get a cyclical change on the next plan/apply due to the API not checking the input coming from the SDK and disabling `setPrivateIpv6GoogleAccess` silently.
```sh
      ~ private_ipv6_google_access = "DISABLE_GOOGLE_ACCESS" -> "true"
```
If I change the value in terraform to:
```tf
  private_ipv6_google_access = "ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE"
```
Checkov complains with
```sh
Check: CKV_GCP_76: "Ensure that Private google access is enabled for IPV6"
        FAILED for resource: google_compute_subnetwork.subnetwork
....
.....
.....
        24 |   private_ipv6_google_access = "ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE"

```
**Version (please complete the following information):**
 - version: 2.0.1140

**Additional context**
I've two paths here, either I disable the check with the `ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE` or I leave out `true` and get changes everytime I run terraform.

Thank you!

nimrodkor:
Hey @parabolic !

I agree with your recommendation, and would love to have your contribution to the existing check to align it with what you wrote. You can find it here:
[...]

PR #2988, opened 2022-05-19
CKV_GCP_76/update-the-gcp-compute-subnetwork-private-ipv6-google-access-check

parabolic:
This change fixes the issue with the argument `private_ipv6_google_access` in the `google_compute_subnetwork` terraform resource. It accepts true but does no change to the actual setting. Checkov expects true and fails otherwise. You can find more details in the linked issue below.
https://github.com/bridgecrewio/checkov/issues/2979

The tests are passing and I have also created a version and tested it against dummy terraform templates.
**Test Output:**
```sh
$ pipenv run python -m coverage run -m pytest tests/terraform/checks/resource/gcp/test_GoogleSubnetworkIPV6PrivateGoogleEnabled.py 
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.7.10, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/kikola/Documents/work/checkov, configfile: pytest.ini
plugins: mock-3.7.0, xdist-2.5.0, cov-3.0.0, forked-1.4.0, asyncio-0.18.3
asyncio: mode=legacy
gw0 [1] / gw1 [1]
.                                                                                                                                                                                                           [100%]
================================================================================================ warnings summary =================================================================================================
../../../.local/share/virtualenvs/checkov-b6-xzvGD/lib/python3.7/site-packages/pytest_asyncio/plugin.py:191
../../../.local/share/virtualenvs/checkov-b6-xzvGD/lib/python3.7/site-packages/pytest_asyncio/plugin.py:191
../../../.local/share/virtualenvs/checkov-b6-xzvGD/lib/python3.7/site-packages/pytest_asyncio/plugin.py:191
  /home/kikola/.local/share/virtualenvs/checkov-b6-xzvGD/lib/python3.7/site-packages/pytest_asyncio/plugin.py:191: DeprecationWarning: The 'asyncio_mode' default value will change to 'strict' in future, please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.
    config.issue_config_time_warning(LEGACY_MODE, stacklevel=2)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================================================== 1 passed, 3 warnings in 2.29s ==========================================================================================

```

**Changes:**
Adjust the check to match the private_ipv6_google_access argument against the following values:
- ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE
- ENABLE_BIDIRECTIONAL_ACCESS_TO_GOOGLE

Adjust the test to check against those 2 values as well.


gruebel (approved):
thanks, great work 🚀

diablo296:
Why does it only accept two values and not three? For the value true checkov returns failed. Why do not accept value 

`DISABLE_GOOGLE_ACCESS `
But accepts only values
```
ENABLE_BIDIRECTIONAL_ACCESS_TO_GOOGLE
ENABLE_OUTBOUND_VM_ACCESS_TO_GOOGLE
```

gruebel:
@diablo296 because the purpose of the check is `"Ensure that Private google access is enabled for IPV6"` and not disabled.

diablo296:
Ok but what if I want to disable access? Previously if I set true the field was set to `DISABLE_GOOGLE_ACCESS `

PR #2990, opened 2022-05-20
Fix CFN boolean as string checks

mikeurbanski1:

Fixes 3 checks that only look for boolean values as strings, and fail if you provide an actual boolean (e.g., `attribute: true` instead of `attribute: "true"`)


kartikp10 (approved):
Looks good!

Saarett:
@mikeurbanski1 , Good catch!
I wanna suggest a few changes as a new commit, mind waiting before you merge it?

PR #2991, opened 2022-05-20
Enforce OpenApi v2 checks to check first the version of the file

Saarett:


checkov/openapi/checks/resource/v2/BaseOpenapiCheckV2.py
```diff
@@ -0,0 +1,27 @@
...
+        )
+
+    @abstractmethod
+    def scan_openapi_conf(self, conf: Dict[str, Any], entity_type: str) -> Tuple[CheckResult, Dict[str, Any]]:
+        raise NotImplementedError()
+
+    def scan_entity_conf(self, conf: Dict[str, Any], entity_type: str) -> Tuple[CheckResult, Dict[str, Any]]:  # type: ignore
+        if "swagger" in conf and conf.get("swagger") == '2.0':
+            return self.scan_openapi_conf(conf, entity_type)
+        return CheckResult.PASSED, conf
```
gruebel, line 27:
```suggestion
        return CheckResult.UNKNOWN, conf
```
shouldn't it be unknown, because the v3 files don't really pass them, it is just not meant for them? If this is the case, then we just use `CheckResult.UNKNOWN` as in other checks.

checkov/openapi/checks/resource/v2/BaseOpenapiCheckV2.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(
+            name=name,
+            id=id,
+            categories=categories,
+            supported_entities=supported_entities,
+            block_type=block_type,
+        )
+
+    @abstractmethod
+    def scan_openapi_conf(self, conf: Dict[str, Any], entity_type: str) -> Tuple[CheckResult, Dict[str, Any]]:
```
gruebel, line 21:
```suggestion
    def scan_openapi_conf(self, conf: dict[str, Any], entity_type: str) -> tuple[CheckResult, dict[str, Any]]:
```
if add `from __future__ import annotations`, like you did, then you can use the built-in types for typing `Dict`, `List`, `Set` and `Tuple` and then of course you can remove the import of those types, otherwise `flake8` will blame you 😄

checkov/openapi/checks/resource/v2/BaseOpenapiCheckV2.py
```diff
@@ -0,0 +1,27 @@
...
+        raise NotImplementedError()
+
+    def scan_entity_conf(self, conf: Dict[str, Any], entity_type: str) -> Tuple[CheckResult, Dict[str, Any]]:  # type: ignore
```
gruebel, line 24:
```suggestion
    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> tuple[CheckResult, dict[str, Any]]:  # type:ignore[override]
```
please always add the error code, if you use the `type:ignore`

checkov/openapi/checks/resource/v2/Oauth2SecurityRequirement.py
```diff
@@ -16,7 +16,7 @@ def __init__(self) -> None:
         super().__init__(name=name, id=id, categories=categories, supported_entities=supported_resources,
                          block_type=BlockType.DOCUMENT)
 
-    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> tuple[CheckResult, dict[str, Any]]:  # type:ignore[override] # return type is different than the base class
+    def scan_openapi_conf(self, conf: Dict[str, Any], entity_type: str) -> Tuple[CheckResult, dict[str, Any]]:
```
gruebel, line 19: same here for the types, just use them as before 🙂 same goes for the other checks you adjusted

gruebel (approved):
looks good 👍

PR #2993, opened 2022-05-22
support var element rendering in Bicep

gruebel:

Added support for `var` element rendering in Bicep + chain rendering also works (ex. `param` -> `var` -> `resource`). I moved the `update_inner_attribute()` method from Terraform to common, because it perfectly works with Bicep, but somehow it doesn't with `CloudFormation`, need extra investigation, which I didn't want to invest for this PR 😄 Also added some missing type hints.


tronxd (approved):
🥇

PR #3007, opened 2022-05-24
fix CKV_AZURE_116 to reflect changes in v2.97.0 of Terraform Azure provider

### LINKED ISSUE #3006: CKV_AZURE_116 fails with Terraform Azure provider >= v.2.97.0
**Describe the issue**
This issue is regarding CKV_AZURE_116.
Since v2.97.0 of the Azure Terraform provider, the Azure policies add-on is enabled by setting

```terraform
azure_policy_enabled = true
```

as a top-level property in a `azurerm_kubernetes_cluster` resource definition, instead of the following

```terraform
addon_profile {
  azure_policy {
    enabled = true
  }
}
```
The check should be updated to reflect the change.

**Examples**
```terraform
resource "azurerm_kubernetes_cluster" "example" {
  azure_policy_enabled = true
}
```
should pass the check for CKV_AZURE_116.

**Version:**
2.0.1160

**Additional context**
None

I will submit a PR to fix this.

dommaes:
fixes #3006 

Unfortunately, I couldn't find the place to update the docs at https://docs.bridgecrew.io/docs/ensure-that-aks-uses-azure-policies-add-on. If someone can point me to the corresponding place, I will also update the docs.

### DIFF
--- a/checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
+++ b/checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
@@ -11,7 +11,7 @@ class AKSUsesAzurePoliciesAddon(BaseResourceValueCheck):
         super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
 
     def get_inspected_key(self):
-        return "addon_profile/[0]/azure_policy/[0]/enabled"
+        return "azure_policy_enabled"
 
 
 check = AKSUsesAzurePoliciesAddon()
--- a/checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
+++ b/checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
@@ -1,8 +1,10 @@
-from checkov.common.models.enums import CheckCategories
-from checkov.terraform.checks.resource.base_resource_value_check import BaseResourceValueCheck
+from typing import Dict, List, Any
+
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
 
 
-class AKSUsesAzurePoliciesAddon(BaseResourceValueCheck):
+class AKSUsesAzurePoliciesAddon(BaseResourceCheck):
     def __init__(self):
         name = "Ensure that AKS uses Azure Policies Add-on"
         id = "CKV_AZURE_116"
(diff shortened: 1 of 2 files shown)


gruebel:
hi @dommaes thanks for the contribution 🚀 the docs are separate from the repository and you just need to click on **Suggest Edits**
regarding your changes, I'm totally up for only supporting the newer provider versions, but sadly our users need usually a bit longer to adapt to the new changes. I have to consult with my colleagues on how to best approach the transition time.
But thanks a lot for contributing and I will get back to you 🍻

dommaes:
Thanks for the explanation, I totally get that.

In order to support both the old and the new version, what about reading the provider version from the Terraform lock file?
With that it would be possible to implement the check in a version dependent way.

I suppose this would turn out to be a bit more complex than it sounds.
Still, any thoughts on that?

rotemavni:
Hi @dommaes, thank you for your contribution!
I have a suggestion on how to support both versions.

Currently the check extends `BaseResourceValueCheck` and my suggestion is to convert it to `BaseResourceCheck`. The check will check if one of the attributes exists and equals `True`. 
In case one of the attributes exist, set `self.evaluated_keys = [<found attribute>]`.

What do you think about this idea?

dommaes:
I implemented your suggestion, thanks a lot for that @rotemavni! =)

checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
```diff
@@ -1,17 +1,31 @@
...
+        azure_policy_enabled = conf.get("azure_policy_enabled", [None])[0]
+        if azure_policy_enabled:
+            self.evaluated_keys = ["azure_policy_enabled"]
+            return CheckResult.PASSED
+        # up to and including Azure provider v2.96.0
+        addon_profile = conf.get("addon_profile", [None])[0]
+        if addon_profile and isinstance(addon_profile, dict):
+            azure_policy = addon_profile.get("azure_policy", [None])[0]
+            if azure_policy and isinstance(azure_policy, dict) and azure_policy.get("enabled", [None])[0]:
+                self.evaluated_keys = ["addon_profile/[0]/azure_policy/[0]/enabled"]
```
gruebel, line 26:
```suggestion
        self.evaluated_keys = ["addon_profile/[0]/azure_policy/[0]/enabled"]
        addon_profile = conf.get("addon_profile", [None])[0]
        if addon_profile and isinstance(addon_profile, dict):
            azure_policy = addon_profile.get("azure_policy", [None])[0]
            if azure_policy and isinstance(azure_policy, dict) and azure_policy.get("enabled", [None])[0]:
```
just move the `evaluated_keys` to the beginning of the condition check 🙂

gruebel (commented):
other than that, great job 🚀

checkov/terraform/checks/resource/azure/AKSUsesAzurePoliciesAddon.py
dommaes: Done, thanks for the suggestion =)

gruebel (approved):
thanks, great work 🥇

rotemavni (approved):

PR #3009, opened 2022-05-25
relpath fails everywhere for windows paths that different from the calling drive

JamesWoolfenden:
The built in function relpath fails on windows when the working directory is different from where the target code is:
e.g.
checkov runs from c:\python\....
tf files on f:\code
**fails**

if you set your working dir (in a debugger) or use the same drive it works.
or 
checkov runs from f:\python\....
tf files on f:\code
**Works**

I don't really know what intended purpose of relpath is hence my hack around here. I know relpath is supposed to work on windows but it doesn't this way.

Karthik/Mike maybe you can think of a better way?
 

JamesWoolfenden:
Ive updated 3 instances of relpath, this fixes the crash for manulife but its not the only use of relpath which is flawed when using windows if checkov is run from a different drive from where your iac is.

nimrodkor (approved):
Why work with windows at all? 😆

PR #3011, opened 2022-05-25
Support branch protection rules

arielkru:

We want to add a branch protection rule for master branch, that requires a minimal amount of reviewers on every PR and restricts who can push to the branch.
Adding that would break some of the current github actions workflows, since they would try to push to master without the right permissions.
The PR implements checking out to the code with a PAT that has permissions to push to master, in every job that runs `git commit` or `git push`.


gruebel (approved):
💯

nimrodkor (approved):

PR #3022, opened 2022-05-26
IAM Rule added

Jhanvi12:
This rule will be raised when Action is allowed on All the resource i.e Resource = "*"


gruebel:
hi @Jhanvi12 thanks for the contribution 🚀 I understand the reason behind this check, but I think it will give to many false positives, because adding a `*` in the resource block is very often valid, due to the way of how IAM actions are scoped. A lot of the actions can't be restricted to only work for a specific resource.

Jhanvi12:
Hey , thanks for the comment above . I am contributing for the first time , so asking this what can be the next step here ?

PR #3027, opened 2022-05-28
add environment variable to hard fail on parsing errors

mikeurbanski1:

By customer request, adds a mechanism to hard-fail the run if there is a file with a parsing error. It uses an environment variable, as this is a fairly niche requirement.


gruebel (approved):
🥇

nimrodkor (approved):

PR #3032, opened 2022-05-30
refactor terraform plan runner

YaaraVerner:

### Description
refactor terraform_plan runner - split the logic to the runner and plan_utils.
create plan_utils that include calculating the definitions and building definitions context.
add an option to inject the definitions, the context, and the graph manager in order to not calculate it in the run function.
run the checks on the tf_definitions after calculating all of them.
build the graph with all the files data and not only the last.

## Checklist:


checkov/terraform/plan_parser.py
```diff
@@ -125,15 +125,13 @@ def _find_child_modules(child_modules: ListNode) -> List[Dict[str, Dict[str, Any
     return resource_blocks
 
 
-def parse_tf_plan(tf_plan_file: str, out_parsing_errors: Dict[str, str]) -> Tuple[Optional[Dict[str, Dict[str, Any]]], Optional[List[Tuple[int, str]]]]:
+def parse_tf_plan(tf_defintions: Dict[str, Dict[str, Any]], template_lines: Dict[str, List[Tuple[int, str]]], tf_plan_file: str, out_parsing_errors: Dict[str, str]):
```
tronxd, line 128: @YaaraVerner why does this method now accepts `tf_definitions` and not returning it? `create_definitions` anyways return the definitions, so I don't see a reason to add additional arguments, and It's not a best practice to modify in-place a function argument, unless we have a good reason for it

YaaraVerner: until now for each plan file, we parsed it and run the checks on it. the function is parsing only one file and I wanted to save all the results in the same object. but I agree that it isn't the best solution, I will change it and extend the tf_definitions outside the function.

YaaraVerner: @tronxd done :)

rotemavni (approved):
🥇

PR #3050, opened 2022-06-03
Filter by prisma labels

kartikp10:

## Description

Adds a new `--policy-metadata-filter` flag that can only be used with a Prisma Cloud API key. This adds the ability to filter checks based on [policy metadata filters](https://prisma.pan.dev/api/cloud/cspm/policy#operation/get-policy-filters-and-options). 

Filters can be specified in a comma-separated key-value list, for example: 

```
checkov -d . --bc-api-key "xxx::xxx" --policy-metadata-filter policy.label=A,cloud.type=aws,policy.mode=custom
```
When you list checks with a filter, only checks returned by the filter are included.  

Some important behaviors:
- If a filter is invalid, checkov will continue execution with a warning and will run as though no filters were specified.
- `policy.enabled=True` and `policy.subtype=build` are assumed and cannot be supplied by the user.
- Checkov will not run any other checks outside of those returned by the filter. This implies if a check is explicitly supplied by the `--check` flag, but it isn't in the list of filtered policies, it will not be run. 

## Checklist:


checkov/common/util/docs_generator.py
```diff
@@ -46,20 +46,21 @@ def get_compare_key(c: list[str] | tuple[str, ...]) -> list[tuple[str, str, int,
 
 
 def print_checks(frameworks: Optional[List[str]] = None, use_bc_ids: bool = False,
-                 include_all_checkov_policies: bool = True) -> None:
+                 include_all_checkov_policies: bool = True, filtered_policy_ids: List[str] = []) -> None:
```
gruebel, line 49: please don't use `[]` as a default value, because it mutable, you can use `None` and then check, if it is `None`, then set `[]`. Using a mutable object as a default in Python results in surprising behaviour, because it is reused with consecutive calls.

checkov/common/util/docs_generator.py
```diff
@@ -46,20 +46,21 @@ def get_compare_key(c: list[str] | tuple[str, ...]) -> list[tuple[str, str, int,
...
+                                       include_all_checkov_policies=include_all_checkov_policies, filtered_policy_ids=filtered_policy_ids)
     print(
         tabulate(printable_checks_list, headers=["Id", "Type", "Entity", "Policy", "IaC"], tablefmt="github",
                  showindex=True))
     print("\n\n---\n\n")
 
 
-def get_checks(frameworks: Optional[List[str]] = None, use_bc_ids: bool = False, include_all_checkov_policies: bool = True) -> List[Tuple[str, str, int, int, str]]:
+def get_checks(frameworks: Optional[List[str]] = None, use_bc_ids: bool = False,
+               include_all_checkov_policies: bool = True, filtered_policy_ids: List[str] = []) -> List[Tuple[str, str, int, int, str]]:
```
gruebel, line 60: same here

checkov/common/util/docs_generator.py
```diff
@@ -127,9 +128,10 @@ def add_from_repository(registry: Union[BaseCheckRegistry, BaseGraphRegistry], c
         add_from_repository(openapi_registry, "resource", "OpenAPI")
     if any(x in framework_list for x in ("all", "secrets")):
         for check_id, check_type in CHECK_ID_TO_SECRET_TYPE.items():
-            if use_bc_ids:
-                check_id = metadata_integration.get_bc_id(check_id)
-            printable_checks_list.append((check_id, check_type, "secrets", check_type, "secrets"))
+            if check_id in filtered_policy_ids or not filtered_policy_ids:
```
gruebel, line 131:
```suggestion
            if not filtered_policy_ids or check_id in filtered_policy_ids:
```
I think the other way around is better/more performant, because the empty check is much faster done and will probably be the most common one for BC users.

checkov/common/util/type_forcers.py
```diff
@@ -115,3 +115,23 @@ def convert_csv_string_arg_to_list(csv_string_arg: list[str] | str | None) -> li
...
+    {'policy.label': 'label1', 'cloud.type': 'aws'}
+    Note that the API does not accept lists https://prisma.pan.dev/api/cloud/cspm/policy#operation/get-policies-v2
+    This is not allowed: policy.label=label1,label2
+    """
+    filter_params = {}
+    if filter_string:
+        try:
+            for f in filter_string.split(','):
+                f_name = f.split('=')[0]
+                f_value = f.split('=')[1]
```
gruebel, line 133:
```suggestion
                f_name, f_value = f.split('=')
```
if you are sure, that there will be only one `=`, otherwise
```suggestion
                f_name, f_value, *_ = f.split('=')
```
then it will throw away excessive parts

checkov/main.py
```diff
@@ -222,13 +222,19 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
         bc_integration.skip_download = True
 
     bc_integration.get_platform_run_config()
+    bc_integration.get_prisma_build_policies(config.filter)
 
     integration_feature_registry.run_pre_scan()
 
+    if integration_feature_registry.features[0].__dict__.get('filtered_policy_ids'):
```
gruebel, line 229: I think it would be better to check if this feature `isinstance(PolicyMetadataIntegration)` and then directly access the `filtered_policy_ids` without the need of using the `__dict__` method.

checkov/runner_filter.py
```diff
@@ -34,7 +34,8 @@ def __init__(
             excluded_paths: Optional[List[str]] = None,
             all_external: bool = False,
             var_files: Optional[List[str]] = None,
-            skip_cve_package: Optional[List[str]] = None
+            skip_cve_package: Optional[List[str]] = None,
+            filtered_policy_ids: List[str] = []
```
gruebel, line 38: also here, please don't use `[]` as a default, instead use `None` and then check if it is `None` and set `[]` for the variable.

checkov/runner_filter.py
```diff
@@ -82,6 +83,7 @@ def __init__(
         self.all_external = all_external
         self.var_files = var_files
         self.skip_cve_package = skip_cve_package
+        self.filtered_policy_ids: List[str] = filtered_policy_ids
```
gruebel, line 86:
```suggestion
        self.filtered_policy_ids: List[str] = filtered_policy_ids or []
```
that's how you could do easily the `None` check here

gruebel (commented):
nice job, added some possible improvements.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -440,11 +446,85 @@ def get_customer_run_config(self) -> None:
...
+        if self.skip_download is True:
+            logging.debug("Skipping prisma policy API call")
+            return
+        if not self.is_prisma_integration():
+            return
+        if not self.bc_api_key or not self.is_integration_configured():
+            raise Exception(
+                "Tried to get prisma build policy metadata, "
+                "but the API key was missing or the integration was not set up")
+        if not policy_filter:
```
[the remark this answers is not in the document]
kartikp10: Makes sense!

checkov/common/util/docs_generator.py
kartikp10: Got it! Thanks for pointing this out

checkov/common/util/docs_generator.py
kartikp10: Good call 👍🏽

checkov/common/bridgecrew/integration_features/features/policy_metadata_integration.py
```diff
@@ -113,15 +120,25 @@ def _handle_public_metadata(self, check_metadata):
...
             for custom_policy in run_config['customPolicies']:
                 if 'guideline' in custom_policy:
                     self.check_metadata[custom_policy['id']] = {
                         'guideline': custom_policy['guideline']
                     }
+                if custom_policy.get('pcPolicyId'):
+                    self.pc_to_ckv_id_mapping[custom_policy['pcPolicyId']] = custom_policy['id']
+
+    def _handle_customer_prisma_policy_metadata(self, prisma_policy_metadata):
+        if isinstance(prisma_policy_metadata, list):
```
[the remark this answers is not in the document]
kartikp10: The API only returns a list and if it doesn't for some reason, it would indicate there was an error and the result will be filtering is skipped. So I think this should be alright!

checkov/common/util/type_forcers.py
kartikp10: Accepted option 1 since no more `=` will be present

checkov/main.py
kartikp10: Makes sense

checkov/main.py
```diff
@@ -497,6 +504,10 @@ def add_parser_args(parser: ArgumentParser) -> None:
     parser.add('--skip-cve-package',
                help='filter scan to run on all packages but a specific package identifier (denylist), You can '
                     'specify this argument multiple times to skip multiple packages', action='append', default=None)
+    parser.add('--policy-metadata-filter',
```
gruebel, line 507: can you also add it to the docs https://github.com/bridgecrewio/checkov/blob/master/docs/2.Basics/CLI%20Command%20Reference.md

gruebel (approved):
nice work 🏆

tsmithv11 (approved):

checkov/common/util/type_forcers.py
```diff
@@ -115,3 +115,22 @@ def convert_csv_string_arg_to_list(csv_string_arg: list[str] | str | None) -> li
...
+    {'policy.label': 'label1', 'cloud.type': 'aws'}
+    Note that the API does not accept lists https://prisma.pan.dev/api/cloud/cspm/policy#operation/get-policies-v2
+    This is not allowed: policy.label=label1,label2
+    """
+    filter_params = {}
+    if filter_string:
+        try:
+            for f in filter_string.split(','):
+                f_name, f_value = f.split('=')
+                filter_params[f_name] = f_value
```
nimrodkor, line 133: @kartikp10 I'd expect to see some `trim()` here, to avoid stupid user input mistakes... Example bad input:
`'policy.label=label, cloud.type=aws'`

Can be done in another PR

checkov/runner_filter.py
```diff
@@ -153,3 +160,8 @@ def notify_external_check(check_id: str) -> None:
     @staticmethod
     def is_external_check(check_id: str) -> bool:
         return check_id in RunnerFilter.__EXTERNAL_CHECK_IDS
+
+    def is_policy_filtered(self, check_id: str) -> bool:
+        if not self.filtered_policy_ids:
+            return True
+        return check_id in self.filtered_policy_ids
```
nimrodkor, line 167: Can be another PR:
```suggestion
        return self.filtered_policy_ids and check_id in self.filtered_policy_ids
```

nimrodkor (approved):

Issue #1502, opened 2021-08-16
Add option to output checks skipped from baseline


bobziuchkovski:
**Is your feature request related to a problem? Please describe.**
Yes.  We've inherited a terraform repo with a lot of legacy problems.  We've introduced checkov to ensure new config going 
forward meets our standards.  We've also generated baseline files for all of the existing legacy terraform config, so that we only fail checks for new problems.  However, we'd like to keep visibility on all of the baseline/legacy config that we're skipping.  In particular, with junitxml output, we'd like to see the checks that are being skipped due to the baseline.  So we'd effectively be treating all of the legacy config as skipped tests with visibility on those as we update/fix them.

**Describe the solution you'd like**
Add the ability to explicitly output baseline-skipped checks as baseline-skipped.  For junitxml output formatting, render these as skipped tests.

**Describe alternatives you've considered**
We've considered running checkov twice.  Once with the `--baseline` flag and once without, and then comparing the output junitxml files.  Failures unique to the run without the `--baseline` flag would basically be those skipped checks, so we could post-process the result XML and display those as skipped tests.  However, this is pretty ugly.

nimrodkor:
Sounds very interesting. We're open for contributions here, should be implemented somewhere near this:
https://github.com/bridgecrewio/checkov/blob/ca2673633080910d69ae5a6c8133eefdaeb36787/checkov/common/output/baseline.py#L13

oliver-pham:
I'm interested in giving this a try. Has anyone worked on this issue yet?

bobziuchkovski:
Our team at my company had intended to, but haven't been able to prioritize it yet.  Would still love to see this, though.

chtnnh:
@nimrodkor I could give this a shot, if @oliver-pham isn't already working on this

oliver-pham:
@chtnnh you can go ahead 🚀

PR #3052, opened 2022-06-04
output baseline skipped in results

yuvalyacoby:

## Description

added boolean flag `--output-baseline-as-skipped` to output skipped checks due to baseline in the results report.

Fixes [#1502](https://github.com/bridgecrewio/checkov/issues/1502)

### DIFF
--- a/checkov/common/output/baseline.py
+++ b/checkov/common/output/baseline.py
@@ -69,15 +69,15 @@ class Baseline:
             scan_report.skipped_checks = [
                 check for check in scan_report.skipped_checks if self._is_check_in_baseline(check)
             ]
-            failed_checks = []
-            for check in scan_report.failed_checks:
-                if not self._is_check_in_baseline(check):
-                    failed_checks.append(check)
-                elif self.output_skipped:
-                    check.check_result["suppress_comment"] = "baseline-skipped"
-                    check.check_result["result"] = CheckResult.SKIPPED
-                    scan_report.skipped_checks.append(check)
-            scan_report.failed_checks = failed_checks
+            if self.output_skipped:
+                for check in scan_report.failed_checks:
+                    if self._is_check_in_baseline(check):
+                        check.check_result["suppress_comment"] = "baseline-skipped"
+                        check.check_result["result"] = CheckResult.SKIPPED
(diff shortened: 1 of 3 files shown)


checkov/main.py
```diff
@@ -497,6 +497,9 @@ def add_parser_args(parser: ArgumentParser) -> None:
     parser.add('--skip-cve-package',
                help='filter scan to run on all packages but a specific package identifier (denylist), You can '
                     'specify this argument multiple times to skip multiple packages', action='append', default=None)
+    parser.add('--output-baseline-as-skipped',
```
gruebel, line 500: I think it would be better to put it directly after the baseline flag, so three lines higher 🙂

checkov/common/output/baseline.py
```diff
@@ -67,9 +69,15 @@ def compare_and_reduce_reports(self, scan_reports: list[Report]) -> None:
...
-            ]
+            failed_checks = []
+            for check in scan_report.failed_checks:
+                if not self._is_check_in_baseline(check):
+                    failed_checks.append(check)
+                elif self.output_skipped:
+                    check.check_result["suppress_comment"] = "baseline-skipped"
+                    check.check_result["result"] = CheckResult.SKIPPED
+                    scan_report.skipped_checks.append(check)
+            scan_report.failed_checks = failed_checks
```
gruebel, line 80: since this is a very special case, I don't think it is worth to remove the list comprehension and use the slower iteration, just run through the failed checks, if someone wants the output
```suggestion
            if self.output_skipped:
                for check in scan_report.failed_checks:
                    check.check_result["suppress_comment"] = "baseline-skipped"
                    check.check_result["result"] = CheckResult.SKIPPED
                    scan_report.skipped_checks.append(check)
```

checkov/main.py
gruebel: and please add it to the docs 💪 https://github.com/bridgecrewio/checkov/blob/master/docs/2.Basics/CLI%20Command%20Reference.md

gruebel (commented):
nice job, added some suggestions

gruebel (approved):
🎉

tronxd (approved):

PR #3074, opened 2022-06-09
add exception for dicts when evaluating sub in Cloudformation

### LINKED ISSUE #3068: find_all_interpolations crashing cloudformation with UserData
**Describe the issue**
Explain what you expected to happen when checkov crashed.
https://github.com/bridgecrewio/checkov/blob/master/checkov/cloudformation/graph_builder/utils.py#L38
this Line Crashes, when parsing cloudformation templates where the UserData is a dict
the function appears to expect a string

`def find_all_interpolations(str_value: str) -> List[str]:
    return re.findall(FIND_INTERPOLATION_PATTERN, str_value)
`

**Source code screenshot**
source_template
source_referenece

**Examples**
images provided
trace_console
debug
trace

**Exception Trace**

 - OS: MAC
 - Checkov Version [2.0.1204] tried newer and older versions seems to remain

**Additional context**

maxamel:
This PR prevents crashing when processing Cloudformation templates with Sub functions containing dicts (currently similar to lists). 
 

Fixes #3068 

## Checklist:

### DIFF
--- a/checkov/cloudformation/graph_builder/variable_rendering/renderer.py
+++ b/checkov/cloudformation/graph_builder/variable_rendering/renderer.py
@@ -285,8 +285,8 @@ class CloudformationVariableRenderer(VariableRenderer):
 
     def _evaluate_sub_connection(self, value: str, dest_vertex_attributes: Dict[str, Any]) -> (
             Optional[str], Optional[str]):
-        if isinstance(value, list):
-            # TODO: Render values of list type
+        if isinstance(value, list) or isinstance(value, dict):
+            # TODO: Render values of list/dict types
             return None, None
         evaluated_value = None
         attribute_at_dest = None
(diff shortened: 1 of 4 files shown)


nimrodkor (commented):
@maxamel Can you add a test case so we make sure we won't have a regression?

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
```diff
@@ -285,8 +285,8 @@ def _evaluate_getatt_connection(value: List[str], dest_vertex_attributes: Dict[s
 
     def _evaluate_sub_connection(self, value: str, dest_vertex_attributes: Dict[str, Any]) -> (
             Optional[str], Optional[str]):
-        if isinstance(value, list):
-            # TODO: Render values of list type
+        if isinstance(value, list) or isinstance(value, dict):
```
gruebel, line 288:
```suggestion
        if isinstance(value, (list, dict)):
```
more beautiful 😄

maxamel:
Added test containing Cloudformation template with a Sub function referencing a dict. This fix ensures we don't crash in this case.

nimrodkor (approved):
Just fix the UTs :)

PR #3076, opened 2022-06-09
Fix cloudformation EBS encryption check

maxamel:
Fix Cloudformation EBS encryption check to fail fast and only in case EBS is not encypted. Other DeviceBlockMappings are disregarded.

## Checklist:


checkov/cloudformation/checks/resource/aws/LaunchConfigurationEBSEncryption.py
```diff
@@ -17,19 +17,15 @@ def scan_resource_conf(self, conf):
         :param conf: aws_launch_configuration configuration
         :return: <CheckResult>
         """
-        ebs_encryption_confs = []
         if 'Properties' in conf.keys() and conf['Properties'].get('BlockDeviceMappings'):
             for block_device_mapping in conf['Properties']['BlockDeviceMappings']:
                 if not isinstance(block_device_mapping, dict):
                     return CheckResult.UNKNOWN
-                if block_device_mapping.get('Ebs') and not block_device_mapping.get('VirtualName'):
```
gruebel, line 25: I think removing `block_device_mapping.get('VirtualName')` is tricky, only the newer instance types encrypt the instance store by default. hard to say, how we want to handle it.

gruebel: @schosterbarak @nimrodkor thoughts?

maxamel: @gruebel  I removed it since according to this AWS [doc](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-autoscaling-launchconfiguration-blockdevicemapping.html#cfn-autoscaling-launchconfiguration-blockdevicemapping-virtualname:~:text=To%20define%20a%20block%20device%20mapping%2C%20set%20the%20device%20name%20and%20exactly%20one%20of%20the%20following%20properties%3A%20Ebs%2C%20NoDevice%2C%20or%20VirtualName.) , a block device mapping cannot be both VirtualName and Ebs. Do you mean VirtualName should be handled separately?

gruebel: let's remove it, all good. instance store volumes encryption is connected to the instance type. So would be an other check.

gruebel (approved):
🚀

Saarett (approved):

PR #3086, opened 2022-06-14
Minor fix for output format github_failed_only

nascit:

## Description

When using `github_failed_only` format, the table does not say which check the table corresponds to and also it prints the table even though there are no failed checks.

E.g.:
```
checkov -d .       --output github_failed_only --soft-fail                                         
2022-06-14 15:47:33,410 [MainThread  ] [WARNI]  Failed to download module git::https://githubstage.mmm.com/DSC/terraform-aws-sonarqube-dcg-393.git?ref=v1.3.0:None
| check_id   | file   | resource   | check_name   | guideline   |
|------------|--------|------------|--------------|-------------|


---
```

This small PR will first see if the check results are populated before creating a table and will also a title for each table

Fixes # (issue)

## Checklist:

### DIFF
--- a/checkov/common/output/report.py
+++ b/checkov/common/output/report.py
@@ -398,14 +398,18 @@ class Report:
                     record.guideline,
                 ]
             )
-        output_data = tabulate(
-            result,
-            headers=["check_id", "file", "resource", "check_name", "guideline"],
-            tablefmt="github",
-            showindex=True,
-        ) + "\n\n---\n\n"
-        print(output_data)
-        return output_data
+        if result:
+            table = tabulate(
+                result,
+                headers=["check_id", "file", "resource", "check_name", "guideline"],
+                tablefmt="github",
+                showindex=True,
+            )
+            output_data = f"### {self.check_type} scan results:\n\n{table}\n\n---\n"
+            print(output_data)
+            return output_data
+        else:
+            return "\n\n---\n\n"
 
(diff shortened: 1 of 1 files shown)


checkov/common/output/report.py
```diff
@@ -398,14 +398,18 @@ def print_failed_github_md(self, use_bc_ids: bool = False) -> str:
...
-        print(output_data)
-        return output_data
+        if result:
+            table = tabulate(
+                result,
+                headers=["check_id", "file", "resource", "check_name", "guideline"],
+                tablefmt="github",
+                showindex=True,
+            )
+            output_data = f"### {self.check_type} scan results:\n {table}  \n\n---\n\n"
```
gruebel, line 408:
```suggestion
            output_data = f"### {self.check_type} scan results:\n\n{table}\n\n---\n"
```
how about doing it this way?

gruebel (commented):
great work 💪 definitely a good idea to add some indication to which framework the results belong. I'm just not sure about removing the table, because I can't tell, if other users rely on having the table there even when it is empty. @schosterbarak any thoughts?

nascit:
> great work 💪 definitely a good idea to add some indication to which framework the results belong. I'm just not sure about removing the table, because I can't tell, if other users rely on having the table there even when it is empty. @schosterbarak any thoughts?

Good point. IMHO if the table is empty, it's not bringing any value. But I agree it's good to have more input on it.

gruebel (approved):
🚀

schosterbarak (approved):

PR #3088, opened 2022-06-14
Pass (some) graph check attribute conditions with unrendered variables

mikeurbanski1:

**Currently in draft while I finished adding tests for individual operators**

## Description

This PR includes updates to pass graph checks when there are unrendered variables, and when the operator involves some type of equality check.

There are two main updates in this PR.

**Update some attribute solvers to pass (since we can't easily return "unknown") if there is an unrendered variable.**

This has been a common customer request for other Python checks, and also matches the default behavior in `BaseResourceCheck`. It makes the overall policy logic more consistent; users do not want errors for variables with no default values.

I chose to only update operators with some equality check, like equals, within, etc, as well as the > / >= / < / <= operators. This is because other operators, like regex or contains, can still work on partially rendered variables. It does mean that you would still get errors for conditions like `var.xyz contains "abc"`. I am open to changing this logic now, or we can wait and see if we get feedback on it.

**Update some of the variable evaluation logic to make unrendered variables easier to identify in graph checks**

Updates the expression evaluation logic to replace interpolated values after removing them, if:
 - removing the interpolation did not help evaluation, and
 - the expression is more than a simple `var.xyz`

So, for example:
 - This interpolation will still be removed: `${var.xyz}` -> `var.xyz`
 - This will be kept: `abc-${var.xyz}`

This does not have an effect on variable rendering. However, it means that vertex attribute values that have unrendered variables will be more easily identified by the `_is_variable_dependent` methods.

Without the change above, the vertex would have a value like `abc-var.xyz`, which would not be identified as having a variable by the method above.

*This is currently for TF, only. CFN will come later (soon)*.

## Checklist:


checkov/common/checks_infra/solvers/attribute_solvers/less_than_attribute_solver.py
```diff
@@ -14,4 +15,10 @@ def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bo
         if vertex.get(attribute) is None:  # type:ignore[arg-type]  # due to attribute can be None
             return False
 
+        attr_val = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        if BaseAttributeSolver._is_variable_dependant(attr_val, vertex['source_']):
```
nimrodkor, line 21: You can do this and avoid the import
```suggestion
        if self._is_variable_dependant(attr_val, vertex['source_']):
```
(relevant for all similar changes)

nimrodkor (commented):
Looks good! 2 comments

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -88,3 +89,13 @@ def get_attribute_patterns(attribute: str) -> Tuple[Pattern[str], Pattern[str]]:
...
 
         return pattern_with_index, pattern_without_index
+
+    @staticmethod
+    def _is_variable_dependant(value: Any, source: str) -> bool:
+        if source == 'Terraform' and is_terraform_variable_dependent(value):
+            return True
+        # TODO add logic for CloudFormation
+        # elif source == 'CloudFormation' and is_cloudformation_variable_dependent(value):
+        #     return True
```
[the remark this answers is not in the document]
mikeurbanski1: Because the rendering logic for CFN is quite different and requires its own analysis which will take some more time, and the customer is impatiently waiting for the TF behavior change.

checkov/terraform/graph_builder/variable_rendering/evaluate_terraform.py
```diff
@@ -41,7 +52,14 @@ def evaluate_terraform(input_str: Any, keep_interpolations: bool = True) -> Any:
     evaluated_value = evaluate_json_types(evaluated_value)
     second_evaluated_value = _try_evaluate(evaluated_value)
 
-    return evaluated_value if callable(second_evaluated_value) else second_evaluated_value
+    # return evaluated_value if callable(second_evaluated_value) else second_evaluated_value
```
gruebel, line 55:
```suggestion
```
feel free to remove the commented code block 🙂

gruebel (approved):
nice work 🚀 yeah the CFN dependant logic needs some extra work, which can be done in a later stage. I also realized, that it would be great to remove the boilerplate config in the current policies. I think all of the updated S3 policies have extra cases just for handling unrendered variables. I will open a follow up issue to tackle and track it.

checkov/terraform/graph_builder/variable_rendering/evaluate_terraform.py
mikeurbanski1: Oops. thanks

checkov/common/checks_infra/solvers/attribute_solvers/equals_attribute_solver.py
```diff
@@ -13,6 +13,11 @@ def __init__(self, resource_types: List[str], attribute: Optional[str], value: A
 
     def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:
         attr_val = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        # handle edge cases in some policies that explicitly look for blank values
+        if self.value != '' and self._is_variable_dependant(attr_val, vertex['source_']):
```
rotemavni, line 19: @mikeurbanski1 wasn't it possible to implement this once in `get_operation` in `base_attribute_solver`?

rotemavni (approved):
Great work! had only one question ❓

checkov/common/util/var_utils.py
```diff
@@ -0,0 +1,43 @@
+import re
+from typing import Any
+
+from checkov.common.util.parser_utils import find_var_blocks
+
+CFN_VARIABLE_DEPENDANT_REGEX = re.compile(r"(?:Ref)\.[^\s]+")
```
Saarett, line 6: Easy win that would catch most CFN cases would be checking for variables with the keyword `Fn::`, and PROBABLY `AWS::` too. I'm not sure regarding `AWS::` because all CFN types have it as a prefix so I don't really remember or not really sure if we'll break anything if we add it..

Saarett (commented):
Great job! Added 1 important comment regarding CFN 🥇

checkov/common/util/var_utils.py
nimrodkor: @Saarett We're not implementing the CFn logic yet as it is not too straightforward. Can you open n issue detailing all the cases you can think of, that we should take into account?

checkov/common/checks_infra/solvers/attribute_solvers/equals_attribute_solver.py
mikeurbanski1: I think it only applies to `equals`. I added it because a single OOTB policy test was failing. The logic for the policy is `attribute_1 equals ''`. Thus, if the resource has `attribute_1 = aws_instance.i.id`, for example, this check was automatically returning `True`, when of course any value for the variable would make it fail.

There are many other possible little cases like this that could arise for specific policies and operators, so I decided to just solve this one policy for now.

PR #3103, opened 2022-06-20
adding retry when trying to report results to server

itai1357:

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/common/bridgecrew/vulnerability_scanning/integrations/twistcli.py
+++ b/checkov/common/bridgecrew/vulnerability_scanning/integrations/twistcli.py
@@ -9,6 +9,7 @@ from typing import Dict, Any, List
 from datetime import datetime, timedelta
 
 import aiohttp
+import asyncio
 from checkov.common.bridgecrew.platform_integration import bc_integration, BcPlatformIntegration
 from checkov.common.util.data_structures_utils import merge_dicts
 from checkov.common.util.http_utils import get_default_post_headers, request_wrapper
@@ -90,7 +91,7 @@ class TwistcliIntegration(ABC):
                     logging.info(f"Successfully send report for package file {file_path}")
                     return 0
                 elif i != request_max_tries - 1:
-                    time.sleep(sleep_between_request_tries * (i + 1))
+                    await asyncio.sleep(sleep_between_request_tries * (i + 1))
                     continue
                 else:
                     logging.error(f"Failed to send report for package file {file_path}")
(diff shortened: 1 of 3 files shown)


checkov/common/bridgecrew/vulnerability_scanning/integrations/twistcli.py
```diff
@@ -71,19 +73,29 @@ async def report_results_async(
...
-            else:
-                logging.error(f"Failed to send report for package file {file_path}")
-                logging.error(f"Status code: {response.status}, Reason: {response.reason}, Content: {content}")
-                return 1
+        # adding retry mechanism for avoiding repeated unexpected Gateway Timeout from the server
+        logging.info(f"[twistcli](report_results_async) reporting results to the server")
+        request_max_tries = int(os.getenv('REQUEST_MAX_TRIES', 3))
+        sleep_between_request_tries = float(os.getenv('SLEEP_BETWEEN_REQUEST_TRIES', 1))
+
+        for i in range(request_max_tries):
```
gruebel, line 81: better to move the for loop between `async with aiohttp` and `async with session` then the session context doesn't need to be recreated.

checkov/common/bridgecrew/vulnerability_scanning/integrations/twistcli.py
```diff
@@ -71,19 +73,29 @@ async def report_results_async(
...
+                async with session.post(
+                    url=f'{bc_integration.api_url}{self.vulnerabilities_base_path}/results', headers=headers, json=payload
+                ) as response:
+                    content = await response.text()
+
+                if response.ok:
+                    logging.info(f"Successfully send report for package file {file_path}")
+                    return 0
+                elif i != request_max_tries - 1:
+                    time.sleep(sleep_between_request_tries * (i + 1))
```
gruebel, line 93: please use the non blocking version https://docs.python.org/3/library/asyncio-task.html#sleeping

checkov/common/bridgecrew/vulnerability_scanning/integrations/twistcli.py
itai1357: but in the random case we have connection error, shouldn't we restart the connection from zero?

gruebel: so basically it is managing the connection pool, so I wouldn't want to recreate it with each retry 🙂 I don't think it is worth to recreate the pool with each retry, because the typical errors we get are from our endpoint and not that the actual connection is broke.

itai1357: yeah indeed we get 504 so it is from our endpoint. will fix =]

Eliran-Turgeman (approved):

PR #3107, opened 2022-06-21
Add jsonpath_not_exists operator

mikeurbanski1:

## Description

Adds the `jsonpath_not_exists` operator as the inverse to `jsonpath_exists`. This required overriding `JsonpathEqualsSolver.get_operation`, which does not delegate to the `_get_operation` method if it can't find the attribute. This caused every evaluation result to return `False`.

I think that the path to `_get_operation` is not actually going to occur. But I am not sure why the jsonpathequals class has the `if self.attributes` check, so I maintained that logic.

## Checklist:


gruebel (approved):
nice 🥇

nimrodkor (approved):

PR #3124, opened 2022-06-24
explicitly handle lists and dicts in equals operator

mikeurbanski1:

## Description

Explicitly handles comparisons for `list` and `dict` types in the `equals` operator. Previously, it would compare the string values, which generally worked for lists and some dicts, but not in all cases.

## Checklist:


gruebel (approved):
nice 🍺 could you also add one example with `not_equals` just to make sure, it also works as expected

nimrodkor (approved):

PR #3127, opened 2022-06-25
sort baseline output by file and resource

### LINKED ISSUE #3119: baseline output can change resource order for each run
If I generate a baseline file and I have then made some improvements to my Terraform code and I run the baseline again. What I am finding is that the order of the resources for each file can often change which then shows up as a diff against the prevous baseline file - when in reality nothing has change but the order of the resources in the findings array in the baseline output file 

I was wondering could the findings array just be sorted before being output? Then the resource order should be fixed and any actual diffs should be real changes to check_ids (which is sorted already) or new resources being added?

e.g. this is a diff from two runs of generating a baseline file nothing has actually change just resources moved around in the array.

```
@@ -100,13 +100,12 @@
             "file": "/main.tf",
             "findings": [
                 {
-                    "resource": "aws_s3_bucket.canary_artifacts",
+                    "resource": "aws_s3_bucket.backups",
                     "check_ids": [
                         "CKV2_AWS_6",
                         "CKV_AWS_144",
                         "CKV_AWS_145",
-                        "CKV_AWS_18",
-                        "CKV_AWS_21"
+                        "CKV_AWS_18"
                     ]
                 },
                 {
@@ -119,12 +118,13 @@
                     ]
                 },
                 {
-                    "resource": "aws_s3_bucket.lambdas",
+                    "resource": "aws_s3_bucket.canary_artifacts",
                     "check_ids": [
                         "CKV2_AWS_6",
                         "CKV_AWS_144",
                         "CKV_AWS_145",
-                        "CKV_AWS_18"
+                        "CKV_AWS_18",
+                        "CKV_AWS_21"
                     ]
                 },
                 {
@@ -137,7 +137,7 @@
                     ]
                 },
                 {
-                    "resource": "aws_s3_bucket.backups",
+                    "resource": "aws_s3_bucket.lambdas",
                     "check_ids": [
                         "CKV2_AWS_6",
                         "CKV_AWS_144",
```

gruebel:

## Description

The output is now sorted by file path and for each file by resource.

Fixes #3119

## Checklist:

### DIFF
--- a/checkov/common/output/baseline.py
+++ b/checkov/common/output/baseline.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import json
 from collections import defaultdict
+from operator import itemgetter
+
 from checkov.common.models.enums import CheckResult
 from typing import Any, TYPE_CHECKING
 
@@ -56,9 +58,9 @@ class Baseline:
             formatted_findings = []
             for finding in findings:
                 formatted_findings.append({"resource": finding["resource"], "check_ids": finding["check_ids"]})
-            failed_checks_list.append({"file": file, "findings": formatted_findings})
+            failed_checks_list.append({"file": file, "findings": sorted(formatted_findings, key=itemgetter("resource"))})
 
-        resp = {"failed_checks": failed_checks_list}
+        resp = {"failed_checks": sorted(failed_checks_list, key=itemgetter("file"))}
         return resp
 
     def compare_and_reduce_reports(self, scan_reports: list[Report]) -> None:
--- a/tests/common/output/test_baseline.py
+++ b/tests/common/output/test_baseline.py
@@ -31,12 +31,16 @@ def test_to_dict():
                 ],
             },
             {
-                "file": "/main_copy.tf",
+                "file": "/main_2.tf",
                 "findings": [
                     {
                         "resource": "aws_s3_bucket.destination_2",
                         "check_ids": ["CKV2_AWS_6", "CKV_AWS_18"],
-                    }
+                    },
+                    {
+                        "resource": "aws_s3_bucket.destination_3",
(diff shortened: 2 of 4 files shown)


checkov/common/output/baseline.py
```diff
@@ -56,9 +58,9 @@ def to_dict(self) -> dict[str, Any]:
             formatted_findings = []
             for finding in findings:
                 formatted_findings.append({"resource": finding["resource"], "check_ids": finding["check_ids"]})
-            failed_checks_list.append({"file": file, "findings": formatted_findings})
+            failed_checks_list.append({"file": file, "findings": sorted(formatted_findings, key=itemgetter("resource"))})
 
-        resp = {"failed_checks": failed_checks_list}
+        resp = {"failed_checks": sorted(failed_checks_list, key=itemgetter("file"))}
```
nimrodkor, line 63: @gruebel Within the same files, will the resources keep their order?

tests/common/output/test_baseline.py
```diff
@@ -0,0 +1,43 @@
+import argparse
```
itai1357, line 1: as you want to affect also the order of the resources in some specific file, aren't better we that at least one of the files will continue one then 1 resource, while yor new logic will indeed affect the order?

itai1357 (approved):
Nice! approved with one comment regarding the test

schosterbarak (approved):

checkov/common/output/baseline.py
gruebel: yes, I made sure of it, two lines up `sorted(formatted_findings, key=itemgetter("resource"))`

tests/common/output/test_baseline.py
gruebel: yeah good point 😄

PR #3135, opened 2022-06-27
Add missing cfn checks

### LINKED ISSUE #1794: CKV_AWS_117 (require VPC networking for Lambdas) is often counterproductive
Checkov currently flags any Lambda which doesn't have VPC networking configured with CKV_AWS_117 / BC_AWS_GENERAL_65:

https://docs.bridgecrew.io/docs/ensure-that-aws-lambda-function-is-configured-inside-a-vpc-1

I think this check should be more nuanced because a Lambda which only calls AWS services, which is extremely common, would have increased security risk if this check is followed. The AWS-managed Lambda VPC does not give users the chance to configure it insecurely and attaching a Lambda to a VPC which it otherwise does not need would increase the range of resources at risk if it's exploited.

I think it would make more sense to have this check describe those tradeoffs and suggest this as appropriate only when the Lambda has a need to access a VPC resource or use a static IP address.

gruebel:

## Description

- added a couple of missing CFN checks
- adjusted a log message, which was not helpful at all
- enabled flake8 for ARM runner

Fixes #1794

## Checklist:

### DIFF
--- a/checkov/arm/runner.py
+++ b/checkov/arm/runner.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 from typing import Optional, List, Dict, Tuple
@@ -16,7 +18,7 @@ ARM_POSSIBLE_ENDINGS = [".json"]
 
 
 class Runner(BaseRunner):
-    check_type = CheckType.ARM
+    check_type = CheckType.ARM  # noqa: CCE003  # a static attribute
 
     def __init__(self) -> None:
         super().__init__(file_extensions=ARM_POSSIBLE_ENDINGS)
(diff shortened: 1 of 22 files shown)


checkov/arm/runner.py
```diff
@@ -26,9 +28,10 @@ def run(
         root_folder: str,
         external_checks_dir: Optional[List[str]] = None,
         files: Optional[List[str]] = None,
-        runner_filter: RunnerFilter = RunnerFilter(),
+        runner_filter: RunnerFilter | None = None,
```
itai1357, line 31: why? flake8?

gruebel: yes, correct. a function call as an argument default value will be invoked once even when you invoke it again. It is not so critical here, because we only invoke the `run` method once per a whole run, but why take the risk. Instead use `None` and if it is not passed as an argument it will be instantiated a few lines lower.

rotemavni (approved):
🌻

PR #3137, opened 2022-06-28
Fixing pbar on sca scanner

Eliran-Turgeman:

## Description

Fixing progress bar on sca scanner, in case pbar is None, creating a turned off ProgressBar

## Checklist:

### DIFF
--- a/checkov/sca_package/scanner.py
+++ b/checkov/sca_package/scanner.py
@@ -24,7 +24,9 @@ MAX_SLEEP_DURATION = 60
 class Scanner:
     def __init__(self, pbar: ProgressBar = None, root_folder: str | Path | None = None) -> None:
         self._base_url = bc_integration.api_url
-        self.pbar = pbar
+        self.pbar = pbar or ProgressBar('')
+        if not pbar:
+            self.pbar.turn_off_progress_bar()
         self.root_folder = root_folder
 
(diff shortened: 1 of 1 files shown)


checkov/sca_package/scanner.py
```diff
@@ -24,7 +24,9 @@
 class Scanner:
     def __init__(self, pbar: ProgressBar = None, root_folder: str | Path | None = None) -> None:
         self._base_url = bc_integration.api_url
-        self.pbar = pbar
+        self.pbar = pbar or ProgressBar('')
+        if not pbar:
+            self.pbar.turn_off_progress_bar()
```
gruebel, line 29:
```suggestion
        self.pbar = pbar
        if not self.pbar:
            self.pbar = ProgressBar('')
            self.pbar.turn_off_progress_bar()
```
did you mean this? 😄 otherwise it will never be `None`

matansha: actually `self.pbar` will never be but `pbar` will.
and i like more your way

Eliran-Turgeman: Isn't that equal?
if pbar is `None` then self.pbar will be initiated with `ProgressBar('')` and so the condition will pass (the condition is on pbar, not self.pbar)

matansha (approved):

gruebel (approved):
noice 🥇

PR #3140, opened 2022-06-28
add 'not' complex solver

mikeurbanski1:

## Description

Adds `not` as a possible complex solver (can appear at the same place as `and` and `or`). It will invert the sub-solver. It requires exactly one sub-solver. It will accept a list, like `and` and `or` today, but requires it to have exactly 1 element.

Even though we have a `not_` variant of every operator, there is at least one case where this is insufficient to express any policy by applying DeMorgan's law. That is when you want to do operations over an array.

For example, given the following resource:

```
resource "aws_security_group" "sg" {
  ingress {
    cidr_blocks = ["0.0.0.0/0"]
  }
  ingress {
    cidr_blocks = ["192.168.1.0/24"]
  }
}
```

If you write a policy that says `ingress.*.cidr_blocks contains "0.0.0.0/0"`. it will evaluate to true. If you invert the operator to `ingress.*.cidr_blocks not_contains "0.0.0.0/0"`, it will also evaluate to true. In each case, there is one matching item in the wildcard array, so it passes. So, this `not` complex solver will allow you to properly pass a security group that does NOT contain 0.0.0.0 in any of its ingress rules.

I also made a tiny change to the parser to convert a sub-solver from a dict to a list of length 1. This allows you to specify either:

```
not:
  - cond_type: attribute
     ...
```

Or:
```
not:
  cond_type: attribute
  ...
```

The second is more intuitive for `not`. Doing it this way allows us to reuse all the other code related to complex solver types.

## Checklist:

### DIFF
--- a/checkov/common/checks_infra/checks_parser.py
+++ b/checkov/common/checks_infra/checks_parser.py
@@ -19,6 +19,7 @@ from checkov.common.checks_infra.solvers import (
     NotEndingWithAttributeSolver,
     AndSolver,
     OrSolver,
+    NotSolver,
     ConnectionExistsSolver,
     ConnectionNotExistsSolver,
     AndConnectionSolver,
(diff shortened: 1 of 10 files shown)


rotemavni (approved):
Such an elegant implementation! Can you please update the docs as well?
https://github.com/bridgecrewio/checkov/blob/master/docs/3.Custom%20Policies/YAML%20Custom%20Policies.md

gruebel (commented):
nice job 🍻 could you also add an example to the example docs 💪 
https://github.com/bridgecrewio/checkov/blob/master/docs/3.Custom%20Policies/Examples.md

gruebel (approved):

PR #3143, opened 2022-06-29
skip handling irrelevant files for openapi runner

PelegLi:

## Description
skip handling irrelevant for (mostly but not only) openapi runner. added an example file that failed the runner tests before the fix

Fixes # (issue)
https://redlock.atlassian.net/browse/BCE-7158

## Checklist:


schosterbarak:
please add a sample UT

checkov/common/runners/object_runner.py
```diff
@@ -92,8 +92,10 @@ def run(
                     continue
 
                 # result record
-                if result_config:
+                if result_config and isinstance(result_config, (list, dict)):
```
achiar99, line 95: so we will not generate record for that?

achiar99 (approved):

checkov/common/runners/object_runner.py
```diff
@@ -92,8 +92,10 @@ def run(
...
                     end, start = self.get_start_end_lines(end, result_config, start)
+                else:
+                    continue
```
achiar99, line 98: so in this case we will not generate record?

achiar99: @PelegLi maybe it will be better to try catch here instead of checking this specific case

rotemavni (approved):

PR #3144, opened 2022-06-29
fix yaml runner id

schosterbarak:

## Description

* if I have two violating examples it only reports on the last one

### Fix
* fix the id of results in yaml runner list objects

## Checklist:

### DIFF
--- a/checkov/yaml_doc/base_registry.py
+++ b/checkov/yaml_doc/base_registry.py
@@ -227,7 +227,7 @@ class Registry(BaseCheckRegistry):
                        entity_type: str,
                        scanned_file: str,
                        skip_info: _SkippedCheck) -> str:
-        if STARTLINE_MARK and ENDLINE_MARK in entity_configuration:
+        if STARTLINE_MARK in entity_configuration and ENDLINE_MARK in entity_configuration:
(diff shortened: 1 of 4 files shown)


checkov/yaml_doc/base_registry.py
```diff
@@ -225,6 +225,23 @@ def get_result_key(self, check: BaseCheck,
                        skip_info: _SkippedCheck) -> str:
         if "__startline__" and "__endline__" in entity_configuration:
             return f'{entity_type}.{entity_name}.{check.id}[{entity_configuration["__startline__"]}:{entity_configuration["__endline__"]}]'
+        if isinstance(entity_configuration, list):
+            start_line = None
+            end_line = None
+            for sub_conf in entity_configuration:
+                if "__startline__" and "__endline__" in sub_conf:
```
gruebel, line 232: just realizing, this only checks, if `"__endline__" in sub_conf` the left side of the and is always `True`, because it is a not empty string. Please create a a tuple `("__startline__", "__endline__") as a constant in or outside of the class and then use something like
```python
if all(key in sub_conf for key in START_ENDLINE_KEYS):
```
same for line 226

eurogig (changes requested):
Looks solid.  I was just confused by the test file.  If this is just testing the presence of two violations instead of one, why not make the code identical for each.  Confusing otherwise.

eurogig (approved):
Ok with it.

gruebel (approved):

PR #3154, opened 2022-07-02
feat: added is_empty and is_not_empty operators

kadoshms:

## Description

Added support for `is_empty` and `is_not_empty` operators. 
Equivalent way would be to have an `equals/not_equals` operator with a `value: ""` but this one is more straight-forward IMO.

## Checklist:

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] I have made corresponding changes to the documentation
- [ X] I have added tests that prove my feature, policy, or fix is effective and works
- [ X] New and existing tests pass locally with my changes
- [ X] Any dependent changes have been merged and published in downstream modules


gruebel (commented):
great job 🍻

checkov/common/checks_infra/solvers/attribute_solvers/is_empty_attribute_solver.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:
+        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
+
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        if self._is_variable_dependant(attr, vertex["source_"]):
+            return True
+
+        return attr == ""
```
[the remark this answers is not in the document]
kadoshms: good point :)

rotemavni (approved):
🥇

checkov/common/checks_infra/solvers/attribute_solvers/is_empty_attribute_solver.py
```diff
@@ -0,0 +1,24 @@
...
+
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        if self._is_variable_dependant(attr, vertex["source_"]):
+            return True
+
+        if isinstance(attr, (list, dict)):
+            return len(attr) == 0
+
+        return attr == ""
```
gruebel, line 24: need to import this
```python
from collections.abc import Sized
```
then you can use
```suggestion
        if isinstance(attr, Sized):
            return len(attr) == 0

        return False
```
I think default `False` for anything, which can't be checked for being empty sounds reasonable 🙂

kadoshms: @gruebel how would it work for empty string though?  :) 
if `attr = ''` then we will pass this check, no?

kadoshms: Great point!!
Now learned that `Sized` also works for strings!

gruebel (approved):
great job 🚀 looking forward to more contribution from you 💪

PR #3162, opened 2022-07-04
Leverage Pyston-lite 

ayajbara:

Leverage Pyston-lite to speed up checkov

improving performance by leverage pytston-lite which is a JIT for python 3.8. 
for more about pytson-lite [here](https://blog.pyston.org/)

## Checklist:

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -122,8 +122,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.7"]
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        python: ["3.7", "3.8", "3.9", "3.10"]
+        os: [ubuntu-latest]
     env:
       working-directory: ./performance_tests
     runs-on: ${{ matrix.os }}
--- a/Pipfile
+++ b/Pipfile
@@ -77,7 +77,7 @@ jsonschema = "~=3.0"
 prettytable = ">=3.0.0"
 pycep-parser = "==0.3.7"
 charset-normalizer = "*"
-pyston_lite_autoload = ">=2.3.4.1"
+pyston_lite_autoload = {version = ">=2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
 
 [requires]
 python_version = "3.7"
--- a/Pipfile
+++ b/Pipfile
@@ -77,7 +77,7 @@ jsonschema = "~=3.0"
 prettytable = ">=3.0.0"
 pycep-parser = "==0.3.7"
 charset-normalizer = "*"
-pyston_lite_autoload = {version = ">=2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
+pyston_lite_autoload = {version = "==2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
 
 [requires]
 python_version = "3.7"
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -68,6 +68,7 @@
         "prettytable>=3.0.0",
         "pycep-parser==0.3.7",
         "charset-normalizer",
+        "pyston_lite_autoload==2.3.4.1; python_version=='3.8' and (sys_platform=='linux' or sys_platform=='darwin')"
```
nimrodkor, line 71: @ayajbara Why lock it to version 3.8?

itai1357: 1) what will happen in case the user has "windows"? is checkov expected to installed anyway, but without "pyston_lite_autoload"?
2) why do you lock `pyston_lite_autoload==2.3.4.1`, and not `pyston_lite_autoload>=2.3.4.1` as in the Pipfile?
3) in case pytson_lite works only for python 3.8 with mac/linux, shouldn't you add condition in Pipfile?

.github/workflows/pr-test.yml
```diff
@@ -122,8 +122,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.7"]
-        os: [ ubuntu-latest, macos-latest, windows-latest ]
```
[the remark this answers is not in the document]
gruebel: I asked her to change it, to better compare the different Python version. I added the OS comparison to get check how they actually compare between each other, no need to further keep it 🙂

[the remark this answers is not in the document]
ayajbara: @itai1357 its a performance test, related somehow :) . but @gruebel have asked to update it

setup.py
gruebel: 1. yes, it will be just without the package
2. & 3. good idea to keep it consistent between both files. Actually the `Pipfile.lock` is only used for local development. the `setup.py` is the important one, when users install it via `pypi`

gruebel: @ayajbara should be possible to add it https://pipenv-fork.readthedocs.io/en/latest/advanced.html#specifying-basically-anything

ayajbara: @gruebel updated, thanks!!

.github/workflows/pr-test.yml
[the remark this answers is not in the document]
itai1357: cool!

Pipfile
```diff
@@ -77,6 +77,7 @@ jsonschema = "~=3.0"
 prettytable = ">=3.0.0"
 pycep-parser = "==0.3.7"
 charset-normalizer = "*"
+pyston_lite_autoload = {version = ">=2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
```
itai1357, line 80:
```suggestion
pyston_lite_autoload = {version = "==2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
```
no?

setup.py
itai1357: cool!!

itai1357 (approved):

Pipfile
```diff
@@ -77,6 +77,7 @@ jsonschema = "~=3.0"
 prettytable = ">=3.0.0"
 pycep-parser = "==0.3.7"
 charset-normalizer = "*"
+pyston_lite_autoload = {version = "==2.3.4.1", markers="python_version == '3.8' and (sys_platform=='linux' or sys_platform=='darwin')"}
```
itai1357, line 80: nice!!

nimrodkor (approved):

gruebel (approved):
🚀

PR #3174, opened 2022-07-07
feature BCE-6958 add github actions metadata to record

NoaAzoulay:

## Description
add github actions metadata to record, added triggers, jobs and workflow name

Fixes # (issue)
BCE-6958
## Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [] I have added tests that prove my feature, policy, or fix is effective and works
- [ x] New and existing tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/checkov/common/runners/object_runner.py
+++ b/checkov/common/runners/object_runner.py
@@ -8,6 +8,7 @@ from abc import abstractmethod
 from collections.abc import Iterable
 from typing import Any, TYPE_CHECKING, Callable
 
+from checkov.common.output.github_actions_record import GithubActionsRecord
 from checkov.common.output.record import Record
 from checkov.common.output.report import Report
 from checkov.common.parallelizer.parallel_runner import parallel_runner
@@ -18,34 +19,46 @@ from checkov.common.util.suppression import collect_suppressions_for_context
 if TYPE_CHECKING:
     from checkov.common.checks.base_check_registry import BaseCheckRegistry
 
+START_LINE = '__startline__'
+END_LINE = '__endline__'
+GITHUB_ACTIONS = "github_actions"
+
 
 class Runner(BaseRunner):
+    workflow_name = ""
+    jobs = {}
+    triggers = ""
+
     def _load_files(
-        self,
-        files_to_load: list[str],
-        definitions: dict[str, dict[str, Any] | list[dict[str, Any]]],
-        definitions_raw: dict[str, list[tuple[int, str]]],
-        filename_fn: Callable[[str], str] | None = None,
+            self,
+            files_to_load: list[str],
+            definitions: dict[str, dict[str, Any] | list[dict[str, Any]]],
+            definitions_raw: dict[str, list[tuple[int, str]]],
+            filename_fn: Callable[[str], str] | None = None,
     ) -> None:
         files_to_load = [filename_fn(file) if filename_fn else file for file in files_to_load]
         results = parallel_runner.run_function(lambda f: (f, self._parse_file(f)), files_to_load)
         for file, result in results:
             if result:
                 (definitions[file], definitions_raw[file]) = result
+                if self.check_type == GITHUB_ACTIONS:
+                    self.workflow_name = result[0].get('name')
+                    self.jobs = self._get_jobs(result)
+                    self.triggers = self._get_triggers(result)
 
     @abstractmethod
     def _parse_file(
-        self, f: str
+            self, f: str
     ) -> tuple[dict[str, Any] | list[dict[str, Any]], list[tuple[int, str]]] | None:
         raise Exception("parser should be imported by deriving class")
 
     def run(
-        self,
-        root_folder: str | None = None,
-        external_checks_dir: list[str] | None = None,
-        files: list[str] | None = None,
-        runner_filter: RunnerFilter | None = None,
-        collect_skip_comments: bool = True,
+            self,
+            root_folder: str | None = None,
+            external_checks_dir: list[str] | None = None,
+            files: list[str] | None = None,
+            runner_filter: RunnerFilter | None = None,
+            collect_skip_comments: bool = True,
     ) -> Report:
         runner_filter = runner_filter or RunnerFilter()
         if not runner_filter.show_progress_bar:
@@ -82,7 +95,8 @@ class Runner(BaseRunner):
         for file_path in definitions.keys():
             self.pbar.set_additional_data({'Current File Scanned': os.path.relpath(file_path, root_folder)})
             skipped_checks = collect_suppressions_for_context(definitions_raw[file_path])
-            results = registry.scan(file_path, definitions[file_path], skipped_checks, runner_filter)  # type:ignore[arg-type]  # this is overridden in the subclass
+            results = registry.scan(file_path, definitions[file_path], skipped_checks,
+                                    runner_filter)  # type:ignore[arg-type]  # this is overridden in the subclass
             for key, result in results.items():
                 result_config = result["results_configuration"]
                 start = 0
@@ -96,6 +110,7 @@ class Runner(BaseRunner):
                     end, start = self.get_start_end_lines(end, result_config, start)
                 if platform.system() == "Windows":
                     root_folder = os.path.split(file_path)[0]
+
                 record = Record(
                     check_id=check.id,
                     bc_check_id=check.bc_id,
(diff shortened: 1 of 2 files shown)


checkov/common/runners/object_runner.py
```diff
@@ -18,34 +20,46 @@
...
 
+START_LINE = '__startline__'
+END_LINE = '__endline__'
+GITHUB_ACTIONS = "github_actions"
+
 
 class Runner(BaseRunner):
+    workflow_name: str = ""
+    jobs: typing.Dict[str, dict[str, str]] = {}
+    triggers: typing.Set[str] = set()
```
gruebel, line 31: please add them to a `__init__` method, because they are not static.

checkov/common/runners/object_runner.py
```diff
@@ -82,7 +96,8 @@ def run(
         for file_path in definitions.keys():
             self.pbar.set_additional_data({'Current File Scanned': os.path.relpath(file_path, root_folder)})
             skipped_checks = collect_suppressions_for_context(definitions_raw[file_path])
-            results = registry.scan(file_path, definitions[file_path], skipped_checks, runner_filter)  # type:ignore[arg-type]  # this is overridden in the subclass
+            results = registry.scan(file_path, definitions[file_path], skipped_checks,
+                                    runner_filter)    # this is overridden in the subclass
```
gruebel, line 100: `# type:ignore[arg-type]` not needed anymore? nice, then you can also remove the other comment 😄

checkov/common/runners/object_runner.py
```diff
@@ -138,3 +161,32 @@ def _change_files_path_to_relative(report: Report) -> None:
...
             record.resource = record.resource.replace(os.getcwd(), "")
+
+    def _get_triggers(self, result: Tuple[Dict[str, Any], Dict[str, Any]]) -> set[str]:
+        triggers_set = set()
+        triggers = result[0].get(True)
+        try:
+            if triggers:
+                for key in triggers.keys():
+                    if key != START_LINE and key != END_LINE:
+                        triggers_set.add(key)
```
gruebel, line 172: you can write this as a set comprehension `{... for ... if ...}` this has better performance

checkov/common/runners/object_runner.py
```diff
@@ -138,3 +159,34 @@ def _change_files_path_to_relative(report: Report) -> None:
...
+        if jobs:
+            for key, value in jobs.items():
+                if key != START_LINE and key != END_LINE:
+                    jobs_dict[key] = {}
+                    tmp_key = key
+                else:
+                    if key == START_LINE:
+                        jobs_dict[tmp_key][START_LINE] = value
+                    if key == END_LINE:
+                        jobs_dict[tmp_key][END_LINE] = value
```
Eliran-Turgeman, line 187: A question regarding the usage of `tmp_key`.
In case we get to this block first, `tmp_key` is an empty string, also if we get to this block in a later iteration then `tmp_key` holds the last key that didn't get to this block - is that intended? if so, I suggest changing the variable name to `prev_key`

NoaAzoulay: changed it :)

gruebel (approved):
`mypy` is happy, makes me happy 😄

nimrodkor (commented):
@NoaAzoulay Please add a few more UTs that actually parse our own GHA workflow files - they're already in the repo :) - plus some other sample workflow files (can take from the official docs)

checkov/common/runners/object_runner.py
```diff
@@ -138,3 +159,30 @@ def _change_files_path_to_relative(report: Report) -> None:
...
+        triggers_set = set()
+        # it is correct that 'True' can be a key. It is easier to ingore the typing here,
+        # then to support it all the way up.
+        triggers = definition.get(True)  # type:ignore[call-overload]
+        try:
+            if triggers:
+                triggers_set = {key for key in triggers.keys() if key != START_LINE and key != END_LINE}
+
+        except Exception as e:
+            logging.info(f"Error:{str(e)}")
```
nimrodkor, line 173: @NoaAzoulay Please provide a more substantial text which explains where this error is being generated and what it means. i.e.: `failed to parse workflow triggers because of ...`, `failed to find triggers, file: ...` etc

checkov/common/runners/object_runner.py
```diff
@@ -138,3 +159,30 @@ def _change_files_path_to_relative(report: Report) -> None:
...
+        return triggers_set
+
+    def _get_jobs(self, definition: dict[str, Any]) -> dict[str, dict[str, int]]:
+        jobs_dict: dict[str, dict[str, int]] = {}
+        jobs = definition.get('jobs')
+        if jobs:
+            for key, value in jobs.items():
```
nimrodkor, line 180: @NoaAzoulay key, value means nothing to anyone debugging this. Please avoid using generic names such as key, value. In this example:
```suggestion
            for job_name, job_instance in jobs.items():
```

checkov/common/runners/object_runner.py
```diff
@@ -138,3 +159,30 @@ def _change_files_path_to_relative(report: Report) -> None:
...
+                if key != START_LINE and key != END_LINE:
+                    jobs_dict[key] = {}
+                    if value.get(START_LINE):
+                        jobs_dict[key][START_LINE] = value[START_LINE]
+                    if value.get(END_LINE):
+                        jobs_dict[key][END_LINE] = value[END_LINE]
```
nimrodkor, line 186: @NoaAzoulay the `.get()` already gives you the necessary protection...
```suggestion
                    jobs_dict[key][START_LINE] = value.get(START_LINE)
                    jobs_dict[key][END_LINE] = value.get(END_LINE)
```

checkov/common/runners/object_runner.py
```diff
@@ -104,13 +118,20 @@ def run(
...
+                    resource=self.get_resource(file_path, key, check.supported_entities),# type:ignore[arg-type]  # key is str not BaseCheck
                     evaluations=None,
                     check_class=check.__class__.__module__,
                     file_abs_path=os.path.abspath(file_path),
                     entity_tags=None,
                     severity=check.severity,
                 )
+                if self.check_type == CheckType.GITHUB_ACTIONS:
+                    record = GithubActionsRecord(
+                        record=record,
```
nimrodkor, line 130: @NoaAzoulay IMO, it is better and cleaner to not use the class you are extending inside your own class constructor.
If we had mypy on this directory I would feel more confident, but we don't. So my suggestion - check what's your check_type, and if it is GHA create a GHARecord with all the fields **set one by one**. If not - create the regular record.
The constructor doing the spreading for its super is... harder to maintain, and we've broken it a few times already

nimrodkor (commented):
@NoaAzoulay Very nice, had some comments. Also - are these new fields in the GHARecord affecting the output in any way?

tests/common/runners/test_object_runner.py
```diff
@@ -0,0 +1,848 @@
...
+         (58, '        with:\n'), (59, '          languages: python\n'),
+         (60, '          setup-python-dependencies: false\n'), (61, '      - name: Autobuild\n'),
+         (62, '        uses: github/codeql-action/autobuild@v2\n'), (63, '      - name: Perform CodeQL Analysis\n'),
+         (64, '        uses: github/codeql-action/analyze@v2\n')])
+
+    def test_get_jobs_on_result(self):
+        expected_jobs = {"bridgecrew": {"__startline__": 8, "__endline__": 20},
+                         "bridgecrew2": {"__startline__": 21, "__endline__": 33}, }
+
+        jobs = Runner._get_jobs(self, self.result[0])
```
nimrodkor, line 774: @NoaAzoulay This is testing the GHA runner, not the object runner. Why put this test here?

tests/common/runners/test_object_runner.py
```diff
@@ -0,0 +1,848 @@
+import unittest
+
+from checkov.common.runners.object_runner import Runner
+
+
+class TestObjectRunner(unittest.TestCase):
+    result = ({"name": "Supply Chain", True: {"workflow_dispatch": None,
```
nimrodkor, line 7: @NoaAzoulay Let the runner parse the file itself, to verify parser doesn't break please

nimrodkor (approved):

nimrodkor:
@NoaAzoulay Using relative paths is problematic, try to convert it to absolute paths with tricks like [this one](https://github.com/bridgecrewio/checkov/blob/d94f9070a25ac7d012ec5500da216c3bbb21a994/tests/arm/checks/resource/test_AppServiceIdentity.py#L13)

PR #3179, opened 2022-07-08
fix: avoid flattening single-item-list vertex values if the attribute…

kadoshms:

## Description

1. Fixed vertex attribute values serializiation on graph creation  - verteices with single-item values would be flattned, for instance:
```
 cidr_blocks: ["0.0.0.0/0"]
```

becomes:
```
cidr_blocks: "0.0.0.0/0"
```

On graph creations, fixed this to only apply to attributes without a '.' notation.

2. Added two attribute solvers `length_equals` and `length_not_equals`

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
@@ -2,7 +2,7 @@ from typing import List, Optional, Any, Dict
 from collections.abc import Sized
 from checkov.common.checks_infra.solvers.attribute_solvers.base_attribute_solver import BaseAttributeSolver
 from checkov.common.graph.checks_infra.enums import Operators
-from checkov.common.util.type_forcers import force_integer
+from checkov.common.util.type_forcers import force_int
 
 
 class LengthEqualsAttributeSolver(BaseAttributeSolver):
@@ -22,6 +22,6 @@ class LengthEqualsAttributeSolver(BaseAttributeSolver):
             return True
 
         if isinstance(attr, Sized):
-            return len(attr) == force_integer(self.value)
+            return len(attr) == force_int(self.value)
 
         return False
--- a/checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
@@ -1,0 +1,27 @@
+from typing import List, Optional, Any, Dict
+from collections.abc import Sized
+from checkov.common.checks_infra.solvers.attribute_solvers.base_attribute_solver import BaseAttributeSolver
+from checkov.common.graph.checks_infra.enums import Operators
+from checkov.common.util.type_forcers import force_integer
+
+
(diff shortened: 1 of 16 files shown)


checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
```diff
@@ -0,0 +1,27 @@
...
+        if vertex.get(attribute) is None:  # type:ignore[arg-type]  # due to attribute can be None
+            return False
+
+        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        if self._is_variable_dependant(attr, vertex['source_']):
+            return True
+
+        if isinstance(attr, Sized):
```
gruebel, line 24: nice 👍 😄

checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
```diff
@@ -0,0 +1,27 @@
...
+            return len(attr) == force_integer(self.value)
```
gruebel, line 25:
```suggestion
            return len(attr) == force_int(self.value)
```
don't need to add a new function, it already exists in the same file 😄

checkov/common/util/type_forcers.py
```diff
@@ -44,6 +44,15 @@ def force_float(var: Any) -> float | None:
         return None
 
 
+def force_integer(var: Any) -> int | None:
```
gruebel, line 47: you can remove it and just use `force_int` 😉

gruebel (commented):
nice job 🍻 looking forward to the next operators 😄

checkov/common/checks_infra/solvers/attribute_solvers/length_equals_attribute_solver.py
kadoshms: missed that ^^

nimrodkor (changes requested):
@kadoshms Regarding this:
> verteices with single-item values would be flattned

Could be very problematic. Let's say I want to query the attribute `cidr_blocks`. Should I use string operators on it, or list operators? The name of the field is in plural, so I expect a list... Why is this change needed?

gruebel:
@nimrodkor 
> Could be very problematic. Let's say I want to query the attribute cidr_blocks. Should I use string operators on it, or list operators? The name of the field is in plural, so I expect a list... Why is this change needed?

This PR is fixing the behaviour of flattening single value lists 😄

nimrodkor (approved):
Got it.

kadoshms:
yes, in case `cidr_blocks` looks like this `cidr_blocks: ['0.0.0.0/0']` it will evaluate to `cidr_blocks: '0.0.0.0/0'` on the graph (while `cidr_blocks: ['0.0.0.0/0', '0.0.0.0/0']` will remain `cidr_blocks: ['0.0.0.0/0', '0.0.0.0/0']`) so it's in-consistent

rotemavni (approved):
@kadoshms 👏 👏 👏

PR #3183, opened 2022-07-08
ALIcloud ACL should not allow 0.0.0.0/0

JamesWoolfenden:
ACL can define 0.0.0.0/0 so heres a check for that, effectively make the acl pointless


gruebel (approved):
🍻

nimrodkor (approved):

PR #3184, opened 2022-07-08
ALicloud rds allow minor updates

JamesWoolfenden:
Check that we opts into the minor db updates

### DIFF
--- a/checkov/terraform/checks/resource/alicloud/RDSInstanceAutoUpgrade.py
+++ b/checkov/terraform/checks/resource/alicloud/RDSInstanceAutoUpgrade.py
@@ -14,8 +14,8 @@ class RDSInstanceAutoUpgrade(BaseResourceValueCheck):
     def get_inspected_key(self):
         return 'auto_upgrade_minor_version'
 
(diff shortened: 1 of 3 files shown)


checkov/terraform/checks/resource/alicloud/RDSInstanceAutoUpgrade.py
```diff
@@ -0,0 +1,21 @@
...
+        id = "CKV_ALI_30"
+        supported_resources = ['alicloud_db_instance']
+        categories = [CheckCategories.GENERAL_SECURITY]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def get_inspected_key(self):
+        return 'auto_upgrade_minor_version'
+
+    def get_expected_values(self) -> List[Any]:
+        return ["Auto"]
```
nimrodkor, line 18: Better to overwrite the source of truth if it is a single value
```suggestion
    def get_expected_value(self) -> Any:
        return "Auto"
```

rotemavni (approved):

PR #3185, opened 2022-07-08
alicloud ensure nodepools are set to auto repair

JamesWoolfenden:
Ensure that nodepools are set to auto repair (true) - defaults to false


gruebel (approved):
🍹

nimrodkor (approved):

PR #3188, opened 2022-07-08
Fix custom policy wildcard iteration

mikeurbanski1:

## Description

Adds word boundaries to the regexes used to find attributes when wildcards are used in custom policies. Before this, if you had an attribute like `ingress.*.cidr_blocks` in a policy, it would include `ingress.0.cidr_blocks.0`, `ingress.0.cidr_blocks.1`, etc. The result was that the following policy would pass, when it should fail. The reason is that each individual string in the array would get checked with the `contains` operator.

While this policy itself is nonsensical, it is still a bug.

```
definition:
  cond_type: attribute
  resource_types:
    - aws_security_group
  operator: contains
  attribute: "ingress.*.cidr_blocks"
  value: "0.0.0.0"
```

```
resource "aws_security_group" "sg2" {
  ingress {
    cidr_blocks = ["10.0.0.0/8", "192.168.1.0/24"]  # passes, because "10.0.0.0/8" contains "0.0.0.0"
  }
  ingress {
    cidr_blocks = ["192.168.0.124/32"]
  }
}
```

## Checklist:


gruebel (approved):
💯

nimrodkor (approved):

PR #3192, opened 2022-07-10
feat: added new operators - length_less_than & length_greater_than

kadoshms:

## Description

Added two new operator solvers:

1. `length_less_than`
2. `length_greater_than`

Example usage:

```
metadata:
  id: "test"
definition:
  cond_type: "attribute"
  resource_types:
    - aws_security_group
  attribute: egress.cidr_blocks
  operator: length_less_than
  value: 2
```

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/common/checks_infra/solvers/attribute_solvers/length_greater_than_attribute_solver.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class LengthGreaterThanAttributeSolver(BaseAttributeSolver):
+    operator = Operators.LENGTH_GREATER_THAN  # noqa: CCE003  # a static attribute
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:  # type:ignore[override]
+        if vertex.get(attribute) is None:  # type:ignore[arg-type]  # due to attribute can be None
+            return False
+
+        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
```
gruebel, line 15:
```suggestion
        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
        if attr is None:  # type:ignore[arg-type]  # due to attribute can be None
            return False
        
```
you can just move it up and then you don't need to call it twice 🙂

checkov/common/checks_infra/solvers/attribute_solvers/length_greater_than_or_equal_attribute_solver.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .length_less_than_attribute_solver import LengthLessThanAttributeSolver
+from checkov.common.graph.checks_infra.enums import Operators
+
+
+class LengthGreaterThanOrEqualAttributeSolver(LengthLessThanAttributeSolver):
+    operator = Operators.LENGTH_GREATER_THAN  # noqa: CCE003  # a static attribute
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:  # type:ignore[override]
+        return not super()._get_operation(vertex, attribute)
```
gruebel, line 11: sneaky, but smart 😄

checkov/common/checks_infra/solvers/attribute_solvers/length_less_than_attribute_solver.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class LengthLessThanAttributeSolver(BaseAttributeSolver):
+    operator = Operators.LENGTH_LESS_THAN  # noqa: CCE003  # a static attribute
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:  # type:ignore[override]
+        if vertex.get(attribute) is None:  # type:ignore[arg-type]  # due to attribute can be None
+            return False
+
+        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
```
gruebel, line 15: same here, just move it up and then use `attr`

gruebel (commented):
nice job 🍻 please also rename the YAML files, I think you copy pasted it from the `length_equals` solver 😄

gruebel (approved):
🍻

ankon:
I think `length_less_than_or_equal` doesn't really work properly for the "cop-out" case in the base class: `LengthGreaterThanAttributeSolver._get_operation` returns `True` when the value length cannot be determined, so `LengthLessThanOrEqualAttributeSolver` ends up returning `not True`  in that case -- I would expect it to also return `True`.

Is this intentional?

kadoshms:
@ankon I think you are right there, seems like this "shortcut" wasn't a good cool.
I'll open a fix PR for that as I agree with that.

PR #3238, opened 2022-07-20
Fix module version parsing

kartikp10:

## Description
Fixes an issue where the module finder did not pass the operator for a module version to the loaders. 

Additionally, this adds logic so that the "best version" in the case of an inexact operator (like ~>) is used for the destination directory instead of using the raw version string. This will ensure that module versions are not misrepresented in the local cache directory which can cause unexpected behavior.

For example:
If `version = "~> 1.2.1"` is resolving to `1.2.2`, the destination directory for the external module will be `.exteral_modules/github.com/org/repo/v1.2.2` instead of `.exteral_modules/github.com/org/repo/~>\ v1.2.2`

## Checklist:


gruebel (approved):
nice, looks good 🍸

schosterbarak (approved):

PR #3247, opened 2022-07-22
feature  support for is_true and is_false operators 

kadoshms:

## Description

Introducing `is_true` and `is_false` operators, which are basically some sort of syntactic sugar for the `equals` operator and give a more fluent approach.

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/common/checks_infra/solvers/attribute_solvers/is_true_attribute_solver.py
```diff
@@ -0,0 +1,20 @@
...
+    operator = Operators.IS_TRUE  # noqa: CCE003  # a static attribute
+
+    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:  # type:ignore[override]
+        if vertex.get(attribute) is None:  # type:ignore[arg-type]  # due to attribute can be None
+            return False
+
+        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
+        # if this value contains an underendered variable, then we cannot evaluate the check,
+        # so return True (since we cannot return UNKNOWN)
+        if self._is_variable_dependant(attr, vertex['source_']):
```
gruebel, line 17:
```suggestion
    def _get_operation(self, vertex: Dict[str, Any], attribute: Optional[str]) -> bool:  # type:ignore[override]
        attr = vertex.get(attribute)  # type:ignore[arg-type]  # due to attribute can be None
        if attr is None:
            return False

        # if this value contains an underendered variable, then we cannot evaluate the check,
        # so return True (since we cannot return UNKNOWN)
        if self._is_variable_dependant(attr, vertex['source_']):
```

gruebel (approved):
nice job 🍻

gruebel (approved):

tests/terraform/graph/checks_infra/attribute_solvers/is_true_solver/TrueValue.yaml
```diff
@@ -0,0 +1,11 @@
+metadata:
+  id: "BooleanString"
+scope:
+  provider: "Azure"
+definition:
+  cond_type: "attribute"
+  resource_types:
+    - "azurerm_storage_account"
+  attribute: "allow_blob_public_access"
+  operator: "is_true"
```
Saarett, line 10: I suggest creating a test case where the attribute is actually `False` and we are trying to check for `is_true`

tests/terraform/graph/checks_infra/attribute_solvers/is_false_solver/FalseValue.yaml
```diff
@@ -0,0 +1,11 @@
+metadata:
+  id: "BooleanString"
+scope:
+  provider: "Azure"
+definition:
+  cond_type: "attribute"
+  resource_types:
+    - "azurerm_storage_account"
+  attribute: "allow_blob_public_access"
+  operator: "is_false"
```
Saarett, line 10: I suggest creating a test case where the attribute is actually `True` and we are trying to check for `is_false`

Saarett (approved):
Great job! Added 2 comments, and also approved :)

PR #3260, opened 2022-07-25
secret scanning use correct plugins for source code files

lirshindalman:

## Description
secret scanning - Today, we have a custom plugin to reduce false positives on IaC files - first runs the keyword plugin. If it finds a match - it runs the entropy scanners.

add logic for source code files ('.py', '.js', '.properties', '.pem', '.php', '.xml', '.ts', '.env', '.java', '.rb',
                         'go', 'cs', '.txt') to run the 2 original (Base64HighEntropyString + HexHighEntropyString, KeywordDetector) plugins  and for IaC files its remain as was.

## Checklist:

- [ v] My code follows the style guidelines of this project
- [ v] I have performed a self-review of my own code
- [ v] I have commented my code, particularly in hard-to-understand areas
- [ v] I have made corresponding changes to the documentation
- [v ] I have added tests that prove my feature, policy, or fix is effective and works
- [v ] New and existing tests pass locally with my changes
- [v ] Any dependent changes have been merged and published in downstream modules


checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -26,9 +28,18 @@ def analyze_line(
...
+        for source code files run and marge the tow plugin.
         """
+        entropy_matches = set()
         if len(line) <= MAX_LINE_LENGTH:
             keyword_matches = self.keyword_scanner.analyze_line(filename, line, line_number, **kwargs)
+            if f".{filename.split('.')[-1]}" in SOURCE_CODE_EXTENSION:
+                for entropy_scanner in self.high_entropy_scanners:
+                    matches = entropy_scanner.analyze_line(filename, line, line_number, **kwargs)
+                    if matches:
+                        entropy_matches = matches
```
itai1357, line 40: notice the variable entropy_matches is overridden each time that the the function arrive to line 40. is it expected?

lirshindalman: yes it's ok i need one of the high_entropy_scanners

itai1357: got it. I suggest to add 'break'. this way it is clearer that we need only one

itai1357 (approved):
approved with one suggestion

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -26,9 +28,18 @@ def analyze_line(
         """
         This method first runs the keyword plugin. If it finds a match - it runs the entropy scanners, and if
         one of the entropy scanners find a match (on a line which was already matched by keyword plugin) - it is returned.
+        for source code files run and marge the tow plugin.
```
arielkru, line 31:
```suggestion
        for source code files run and merge the two plugins.
```
Little typos

arielkru (approved):

PR #3270, opened 2022-07-27
new AWS resources new CMK checks

JamesWoolfenden:
A number of new AWS resources have been published, a few can use CMKs. these checks ensure that you do.

New checks for:
262 aws_kendra_index (#24920) -encryption block 
263 aws_appflow_flow (#24017) -customer managed kms key 
264 aws_appflow_connector_profile (#23892)  -customer managed kms key
265 aws_keyspaces_table (#24351) encryption_specification and CMK
266 aws_db_snapshot_copy (#9886) kms key id is set

### DIFF
--- a/checkov/terraform/checks/resource/aws/KeyspacesTableUsesCMK.py
+++ b/checkov/terraform/checks/resource/aws/KeyspacesTableUsesCMK.py
@@ -1,0 +1,25 @@
+from checkov.common.models.enums import CheckCategories, CheckResult
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+
+
+class KeyspacesTableUsesCMK(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure Keyspaces Table uses CMK"
+        id = "CKV_AWS_265"
+        supported_resources = ['aws_keyspaces_table']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf) -> CheckResult:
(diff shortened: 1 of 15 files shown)


checkov/terraform/checks/resource/aws/KeyspacesTableUsesCMK.py
```diff
@@ -0,0 +1,25 @@
...
+
+class KeyspacesTableUsesCMK(BaseResourceCheck):
+    def __init__(self):
+        name = "Ensure Keyspaces Table uses CMK"
+        id = "CKV_AWS_265"
+        supported_resources = ['aws_keyspaces_table']
+        categories = [CheckCategories.ENCRYPTION]
+        super().__init__(name=name, id=id, categories=categories, supported_resources=supported_resources)
+
+    def scan_resource_conf(self, conf) -> CheckResult:
```
gruebel, line 13: why not just use a `BaseResourceValueCheck` with an inspected key of `encryption_specification/[0]/type`? or is it possible to set it without actually setting `kms_key_identifier`?

JamesWoolfenden: someone code have written the tf that way, id have preferred what you say - and it was originally that way.

JamesWoolfenden: just went for the safer option

gruebel: ok gotcha

gruebel (approved):
🏖️

lirshindalman (approved):

PR #3273, opened 2022-07-27
Add details to reports

eternalblue:

## Description

Add a new field `details` to record in order to provide more information to the end user about why the policy failed or succeed or any other information that might be useful to the user. These details can be added by calling `self.details.append(msg)` on a python policy.

This field will be present on CLI, Sarif JSON, CSV, and Github Actions.

```python
class S3PCIPrivateACL(BaseResourceCheck):
    def __init__(self):
        name = "Ensure PCI Scope buckets has private ACL (enable public ACL for non-pci buckets)"
        id = "CKV_AWS_999"
        supported_resources = ['aws_s3_bucket']
        # CheckCategories are defined in models/enums.py
        categories = [CheckCategories.BACKUP_AND_RECOVERY]

        """
            Looks for ACL configuration at aws_s3_bucket and Tag values:
            https://www.terraform.io/docs/providers/aws/r/s3_bucket.html
        :param conf: aws_s3_bucket configuration
        :return: <CheckResult>
        """
        if 'tags' in conf.keys():
            environment_tag = Token("IDENTIFIER", "Scope")
            if environment_tag in conf['tags'][0].keys():
                if conf['tags'][0][environment_tag] == "PCI":
                    if 'acl' in conf.keys():
                        acl_block = conf['acl']
                        if acl_block in [["public-read"], ["public-read-write"], ["website"]]:
                            self.details.append("acl not in public-read, public-read-write, website")
                            self.details.append("another one bites the dust")
                            return CheckResult.FAILED
        return CheckResult.PASSED


scanner = S3PCIPrivateACL()
```

image

**On Sarif JSON, the 'attachment' field was used being the closest one but not sure if correct**
> "attachment": {
      "description": "An artifact relevant to a result.",

## Checklist:


tsmithv11:
Cool idea! Can you make the CSV column only show up if there are "traces"? Since none of the OOTB checks have these new "traces" it will be a confusing column to many and the CSV output was meant to be an inventory not a comprehensive set of all information.

eternalblue:
Hey @tsmithv11 ! Sure, didn't know that. Does this way look good to you?

Also, I've used the word 'Traces' because couldn't a better word 😓

tsmithv11:
> Also, I've used 'Traces' because I couldn't come up with a more meaningful word 😓

How about 'Details'?

eternalblue:
> > Also, I've used 'Traces' because I couldn't come up with a more meaningful word 😓
> 
> How about 'Details'?

Sounds good to me. I updated the issue regarding the name in the Serif JSON output. 

Traces was not a valid field on [the specification](https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json) so I chose to use 'attachment' which was the closest but still, not sure if semantically makes sense... Otherwise, I can remove it from the Serif JSON output.

checkov/common/output/csv.py
```diff
@@ -100,16 +101,23 @@ def add_iac_resources(self, resource: Record | ExtraResource, git_org: str, git_
...
+            "Misconfigurations": misconfig,
+            "Severity": severity,
+        }
+
+        if len(resource.details):
+            self.iac_rows_have_details = True
+            row = {
+                **row,
+                "Details": "|".join(resource.details)
             }
```
gruebel, line 118:
```suggestion
        if resource.details:
            self.iac_rows_have_details = True
            row["Details"] = "|".join(resource.details)
```
a non empty check can be done by just checking for truthiness of the value, like `if resource.details`
since the `row` dict is already created, you can just add an other key to it, no need for unpacking it again 🙂

checkov/common/output/github_actions_record.py
```diff
@@ -23,6 +23,7 @@ def __init__(self,
                  entity_tags: Optional[Dict[str, str]] = None,
                  bc_check_id: Optional[str] = None,
                  severity: Optional[Severity] = None,
+                 details: List[str] = []
```
gruebel, line 26:
```suggestion
                 details: Optional[List[str]] = None
```
never use a mutable object, such as `{}` or `[]` as a default value for method parameters, otherwise strange things can happen, because the object is actually only once created, even you call it several times.

checkov/common/output/record.py
```diff
@@ -46,7 +46,8 @@ def __init__(
         description: Optional[str] = None,
         short_description: Optional[str] = None,
         vulnerability_details: Optional[Dict[str, Any]] = None,
-        connected_node: Optional[Dict[str, Any]] = None
+        connected_node: Optional[Dict[str, Any]] = None,
+        details: Optional[List[str]] = []
```
gruebel, line 50:
```suggestion
        details: Optional[List[str]] = None
```
same here

checkov/common/output/record.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         self.vulnerability_details = vulnerability_details  # Stores package vulnerability details
         self.connected_node = connected_node
         self.guideline: str | None = None
+        self.details: List[str] = details
```
gruebel, line 83:
```suggestion
        self.details: List[str] = details or []
```
this will create an empty `list`, if `details` is `None`

checkov/common/output/record.py
```diff
@@ -152,6 +154,18 @@ def to_string(self, compact: bool = False, use_bc_ids: bool = False) -> str:
...
+        detail = ""
+        if len(self.details):
+            detail_buffer = ""
+
+            detail_buffer += colored(f"\tDetails: {self.details[0]}\n", "blue")
+
+            for t in self.details[1:]:
+                detail_buffer += colored(f"\t\t{t}\n", "blue")
+
+            detail = detail_buffer
```
gruebel, line 167:
```suggestion
        if self.details:
            detail_buffer = [colored(f"\tDetails: {self.details[0]}\n", "blue")]

            for t in self.details[1:]:
                detail_buffer.append(colored(f"\t\t{t}\n", "blue"))

            detail = "".join(detail_buffer)
```
using string concatenation with a `+` results in Python creating multiple strings, therefore creating a list of strings and then joining them is most of the time faster and uses less memory.

checkov/common/output/report.py
```diff
@@ -325,6 +325,7 @@ def get_sarif_json(self, tool: str) -> Dict[str, Any]:
                 "ruleId": record.check_id,
                 "ruleIndex": idx,
                 "level": level,
+                "attachments": [{'description': detail} for detail in record.details],
```
gruebel, line 328: lets go with that, even `attachments` doesn't actually fit, but there is nothing else 😄

checkov/common/checks/base_check.py
```diff
@@ -113,3 +114,6 @@ def get_evaluated_keys(self) -> List[str]:
 
     def get_output_id(self, use_bc_ids: bool) -> str:
         return self.bc_id if self.bc_id and use_bc_ids else self.id
+
+    def add_detail(self, msg: str):
+        self.details.append(msg)
```
gruebel, line 119: you can remove the method, because the attribute `details` is not defined as private and therefore can and should be accessed directly, like `self.details.append(msg)` inside the actual check.

gruebel (commented):
nice work, most of my suggestions are Python specific, which makes the code more pythonic 🙂

gruebel:
if you have issues with making `mypy` pass, then just reach out here and I will fix the issues, because it can be quite hard, if you didn't add type hints before.

checkov/common/output/github_actions_record.py
eternalblue: makes sense, guess the method signature is evaluated once and cached?

docs/3.Custom Policies/Python Custom Policies.md
```diff
@@ -89,7 +89,28 @@ def scan_resource_conf(self, conf):
...
+        if key:
+            # The following line sets the evaluated keys
+            self.evaluated_keys = ['storage_encrypted/[0]']
+            return CheckResult.PASSED
+        
+    self.details.append("'storage_encrypted' was not found on the resource configuration")
+    
+    return CheckResult.FAILED
+```
+
```
gruebel, line 112: I saw that you also had an example CLI output for your details feature. Could you add it here to showcase, how it will look like 🚀

gruebel (approved):
great work 🥇 have a final request, which is not code related 😄

lirshindalman (approved):

docs/3.Custom Policies/Python Custom Policies.md
eternalblue: done

PR #3274, opened 2022-07-28
Use enforcement rule settings from the platform

mikeurbanski1:

## Description

Adds a flag (off / false by default) to use enforcement rules from the BC / PC platform in the local run.

The platform enforcement rules provide two meaningful settings for Checkov:
- Hard fail threshold: equivalent to `--hard-fail-on <SEVERITY>`
- Soft fail threshold: equivalent to `--check <SEVERITY>` (in the platform, this is the value that determines the level at which checks run or get skipped, so it is not related to the `--soft-fail-on` option in checkov)

If you use the `--use-platform-enforcement-rules` flag and none of the `--check`, `--skip-check`, `--soft-fail`, etc flags, then it is functionally equivalent to using the two options in the bullets above, EXCEPT that the respective thresholds are applied on a per-runner basis instead of globally. This is because it is possible to define different levels for different categories in the enforcement rules settings. Each runner is mapped to one of these categories.

If you combine the enforcement rules with the `--check` etc flags, then the behavior is as follows:
- If you use `--soft-fail`, it overrides the enforcement rule hard fail threshold for all runners.
- If you use `--check` and / or `--skip-check` with only check IDs (not severities), then it combines those lists with the soft fail threshold from the enforcement rule. 
- If you use `--check` and / or `--skip-check` with a severity, then those severities override the enforcement rule soft fail threshold for all runners.
- If you use `--soft-fail-on` and / or `--hard-fail-on` with only check IDs (not severities), then it combines those lists with the hard fail threshold from the enforcement rule.
- If you use `--soft-fail-on` and / or `--hard-fail-on` with a severity, then those severities override the enforcement rule hard fail threshold for all runners.

The logic to actually fetch the enforcement rules and select the correct one for this repo is pretty straightforward and only makes up a small part of this code.

In order to pre-calculate the hard/soft fail logic at a more convenient place, that logic was moved out of `get_exit_code` into a method that converts enforcement rules and local arguments to standard thresholds, which get passed to `get_exit_code` such that the logic for that method did not actually change (minus the parts that were moved).

The `should_run_check` method still handles the logic for calculating the check/skip check results, but gets enforcement rule types for each runner passed to it earlier on. Each check registry that calls `should_run_check` has had a `report_type` field added to it that it can pass to the runner filter in order to determine which enforcement rules to apply. Because the graph check registries are not hardcoded for each framework, this value gets passed dynamically during graph check evaluation from the runner.

Pretty much all of the changes to runner and registry classes simply handle the logic from the last paragraph. There are some differences in the constructs of the runners that make the behavior for this slightly different per runner, but the intent is the same across all.

The testing for this logic is pretty modular, and is roughly broken down as follows:
- Tests for the logic to convert enforcement rules and hard/soft fail arguments into the values evaluated by `get_exit_code`
- Tests for `should_run_check` that combine enforcement rules and check/skip check arguments
- Tests for each registry to check that it knows its check type
- Complete tests for each runner to show that it correctly passes the check type to the runner filter and applies enforcement rules (including for both graph and python checks, where applicable)

## Checklist:


checkov/main.py
```diff
@@ -467,6 +471,17 @@ def add_parser_args(parser: ArgumentParser) -> None:
                     'custom policies and suppressions if using an API token. Note: it will prevent BC platform IDs from '
                     'being available in Checkov.',
                action='store_true')
+    parser.add('--use-platform-enforcement-rules', action='store_true',
```
nimrodkor, line 474: @mikeurbanski1 I don't like the word `platform` here. Perhaps `use-enforcement-rules` and then in the description explain it is configured in either Bridgecrew or Prisma Cloud? IDK

nimrodkor (commented):
Very nice! Hopefully, only small stuff breaks downstream 😆

checkov/main.py
mikeurbanski1: Yeah, I was 50/50 on this. I will remove platform.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -545,6 +545,12 @@ def get_customer_run_config(self) -> None:
                 logging.error(error_message)
                 raise BridgecrewAuthError(error_message)
             self.customer_run_config_response = json.loads(request.data.decode("utf8"))
+
+            # TODO remove
+            # url = f"{self.api_url}/api/v1/enforcement-rules"
+            # request = self.http.request("GET", url, headers=headers)
+            # self.customer_run_config_response['enforcementRules'] = json.loads(request.data.decode("utf8"))
+
```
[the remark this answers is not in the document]
mikeurbanski1: Totally.

checkov/common/bridgecrew/code_categories.py
```diff
@@ -0,0 +1,51 @@
...
+from checkov.common.bridgecrew.check_type import CheckType
+
+
+@dataclass
+class CodeCategoryType:
+    IAC = "IAC"
+    OPEN_SOURCE = "OPEN_SOURCE"
+    SECRETS = "SECRETS"
+    IMAGES = "IMAGES"
+    SUPPLY_CHAIN = "SUPPLY_CHAIN"
```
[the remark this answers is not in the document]
mikeurbanski1: I am not sure what you mean. These are the 5 internal values from enforcement rules APIs.

```
{
            "id": "3bcbaed0-b77e-4e9c-8c89-fa454f4648ae",
            "creationDate": "2022-04-24T12:14:53.913Z",
            "name": "Security default findings",
            "createdBy": "mike+github@bridgecrew.io",
            "mainRule": true,
            "editable": true,
            "codeCategories": {
                "SUPPLY_CHAIN": {
                    "softFailThreshold": "LOW",
                    "hardFailThreshold": "LOW",
                    "commentsBotThreshold": "LOW"
                },
                "OPEN_SOURCE": {
                    "softFailThreshold": "HIGH",
                    "hardFailThreshold": "CRITICAL",
                    "commentsBotThreshold": "LOW"
                },
                "IAC": {
                    "softFailThreshold": "HIGH",
                    "hardFailThreshold": "CRITICAL",
                    "commentsBotThreshold": "LOW"
                },
                "IMAGES": {
                    "softFailThreshold": "CRITICAL",
                    "hardFailThreshold": "CRITICAL",
                    "commentsBotThreshold": "LOW"
                },
                "SECRETS": {
                    "softFailThreshold": "CRITICAL",
                    "hardFailThreshold": "CRITICAL",
                    "commentsBotThreshold": "LOW"
                }
            },
            "repositories": []
        }
```

mikeurbanski1:
> Very nice! Hopefully, only small stuff breaks downstream 😆

It shouldn't, because it requires the flag, and won't affect anything related to the actual upload and scan in the platform (that can still just use the current logic for applying enforcement rules)

gruebel (approved):
looks good 🍻

docs/2.Basics/Hard and soft fail.md
```diff
@@ -62,3 +62,17 @@ The table below shows how different values of `--soft-fail`, `--soft-fail-on`, a
 |False|LOW,CKV_789|CKV_123|1|The explicit match of the hard fail criteria results in a hard fail|
 |False|CKV_789|HIGH|1|CKV_789 explicitly matches a soft fail criteria, which overrides the hard fail. But CKV_123 is not in either list, so defaults to the value of `--soft-fail`, which is false|
 |True|CKV_789|HIGH|0|CKV_789 explicitly matches a soft fail criteria, which overrides the hard fail. But CKV_123 is not in either list, so defaults to `--soft-fail`, which is true|
+
+# Platform enforcement rules
+
+Checkov can download [enforcement rules](https://docs.bridgecrew.io/changelog/enforcement-rules-settings) that you configure in the Bridgecrew or Prisma Cloud platform. This allows you to centralize the failure and check threshold configurations, instead of defining them in each pipeline.
```
tsmithv11, line 68: Probably better to point to the docs than the changelog: https://docs.bridgecrew.io/docs/enforcement

docs/2.Basics/Suppressing and Skipping Policies.md
```diff
@@ -245,3 +245,16 @@ kube-system namespace:
 ```sh
 checkov -d . --skip-check kube-system
 ```
+
+# Platform enforcement rules
+
+Checkov can download [enforcement rules](https://docs.bridgecrew.io/changelog/enforcement-rules-settings) that you configure in the Bridgecrew or Prisma Cloud platform. This allows you to centralize the failure and check threshold configurations, instead of defining them in each pipeline.
```
tsmithv11, line 251: Same here: https://docs.bridgecrew.io/docs/enforcement

checkov/main.py
```diff
@@ -467,6 +471,17 @@ def add_parser_args(parser: ArgumentParser) -> None:
                     'custom policies and suppressions if using an API token. Note: it will prevent BC platform IDs from '
                     'being available in Checkov.',
                action='store_true')
+    parser.add('--use-enforcement-rules', action='store_true',
+               help='Use the Enforcement Rules configured in the platform for hard / soft fail logic. With this option, '
```
tsmithv11, line 475: Small suggestion, but can you use "Enforcement rules"?

tsmithv11 (commented):
Code looks good. Nothing to add there. Added suggestions for docs.

kartikp10 (approved):
LGTM🏅

PR #3282, opened 2022-07-31
Error handling on types failures

PelegLi:

## Description

several minor fixes for type mismatches referencing not initialized variables

## Checklist:

### DIFF
--- a/checkov/helm/runner.py
+++ b/checkov/helm/runner.py
@@ -155,10 +155,16 @@ class Runner(BaseRunner):
         target_dir.replace("//", "/")
         chart_name = chart_meta.get('name', chart_meta.get('Name'))
         chart_version = chart_meta.get('version', chart_meta.get('Version'))
-        if target_dir.endswith('/'):
-            target_dir = target_dir[:-1]
-        if target_dir.endswith(chart_name):
-            target_dir = target_dir[:-len(chart_name)]
+        try:
+            if target_dir.endswith('/'):
+                target_dir = target_dir[:-1]
+            if target_dir.endswith(chart_name):
+                target_dir = target_dir[:-len(chart_name)]
+        except Exception:
+            logging.info(
+                f"Error processing helm chart {chart_name} at dir: {chart_dir}. Working dir: {target_dir}.",
+                exc_info=True,
+            )
         logging.info(
             f"Processing chart found at: {chart_dir}, name: {chart_name}, version: {chart_version}")
         # dependency list is nicer to parse than dependency update.
--- a/checkov/kustomize/runner.py
+++ b/checkov/kustomize/runner.py
@@ -229,8 +229,8 @@ class Runner(BaseRunner):
                         self.templateRendererCommand = self.kubectl_command
                         return None
             
-            except Exception as e:
-                logging.debug(f"An error occured testing the {self.kubectl_command} command: {e}")
+            except Exception:
+                logging.debug(f"An error occured testing the {self.kubectl_command} command", exc_info=True)
                 pass
 
         elif shutil.which(self.kustomize_command) is not None:
@@ -248,8 +248,8 @@ class Runner(BaseRunner):
                 else:
                     return self.check_type
 
-            except Exception as e:
-                logging.debug(f"An error occured testing the {self.kustomize_command} command: {e}")
+            except Exception:
+                logging.debug(f"An error occured testing the {self.kustomize_command} command", exc_info=True)
                 pass
         
         else:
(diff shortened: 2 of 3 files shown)


ChanochShayner (approved):

checkov/helm/runner.py
```diff
@@ -155,10 +155,16 @@ def _convert_chart_to_k8s(self, chart_item):
...
         chart_version = chart_meta.get('version', chart_meta.get('Version'))
-        if target_dir.endswith('/'):
-            target_dir = target_dir[:-1]
-        if target_dir.endswith(chart_name):
-            target_dir = target_dir[:-len(chart_name)]
+        try:
+            if target_dir.endswith('/'):
+                target_dir = target_dir[:-1]
+            if target_dir.endswith(chart_name):
+                target_dir = target_dir[:-len(chart_name)]
```
YaaraVerner, line 162: what was the error here? the length of the string must be enough (the `endswith` function should verify it)

checkov/terraform/checks/resource/gcp/CloudArmorWAFACLCVE202144228.py
```diff
@@ -24,14 +24,17 @@ def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
...
+                    expr = match[0].get("expr")
+                    if expr and expr[0].get("expression") == ["evaluatePreconfiguredExpr('cve-canary')"]:
+                        if rule.get("action") == ["allow"]:
+                            return CheckResult.FAILED
+                        if rule.get("preview") == [True]:
+                            return CheckResult.FAILED
+
+                        return CheckResult.PASSED
+                except Exception:
+                    return CheckResult.UNKNOWN
```
YaaraVerner, line 37: can you check the specific edge case? and add test for it :)

YaaraVerner (commented):
had some comments :)

achiar99 (approved):

checkov/helm/runner.py
PelegLi: TypeError: endswith first arg must be str or a tuple of str, not NoneType

gruebel: if `target_dir` is  `None`, then a few lines higher `target_dir.replace("//", "/")` should also fail.
probably just do a `None` check after line 154.
```python
target_dir = chart_dir.replace(self.root_folder, f'{self.target_folder_path}/')
if target_dir is None:
    raise ...
```

checkov/kustomize/runner.py
```diff
@@ -229,7 +229,7 @@ def check_system_deps(self):
                         self.templateRendererCommand = self.kubectl_command
                         return None
             
-            except Exception:
+            except Exception as e:
                 logging.debug(f"An error occured testing the {self.kubectl_command} command: {e}")
```
gruebel, line 233:
```suggestion
            except Exception:
                logging.debug(f"An error occured testing the {self.kubectl_command} command", exc_info=True)
```
then we have the complete stack trace

checkov/kustomize/runner.py
```diff
@@ -248,7 +248,7 @@ def check_system_deps(self):
                 else:
                     return self.check_type
 
-            except Exception:
+            except Exception as e:
```
gruebel, line 251: same here

checkov/terraform/checks/resource/gcp/CloudArmorWAFACLCVE202144228.py
gruebel: that would be great, I'm also curious what exactly is failing here and why.

checkov/helm/runner.py
PelegLi: @gruebel but it didn't fail on a few lines higher, it fail on `if target_dir.endswith(chart_name):`

checkov/terraform/checks/resource/gcp/CloudArmorWAFACLCVE202144228.py
PelegLi: the case was that `rule` is a list, so the `try-except` catches this case of unexpected type and also unexpected `expr` type, in case there will be

checkov/helm/runner.py
gruebel: ah got it, the problem is the `chart_name`, it could be `None`. This should work on line 156
```python
chart_name = chart_meta.get('name', chart_meta.get('Name'))
if chart_name is None:
    raise ...
```

PR #3299, opened 2022-08-03
Fix subdirectory module clone using PAT

kartikp10:

## Description

Fixes an issue where module sources that have a subdirectory in them are not cloned using the PAT-based loaders. 

Previous behavior:
For a module sourced from a subdirectory (as shown below), the git_loader strips the username and token values required for HTTP basic auth. 

```hcl
module "a" {
   source = "git::https://example.com/repo.git//modules/test"
}
```
Therefore, the above example would result in git clone being attempted with "`https://example.com/repo.git` as opposed to `https://username:token@example.com/repo.git`

## Checklist:


mikeurbanski1:
It seems fine, but it's not quite clear what the buggy vs fixed behavior is. What is an example module source that did not work before but works now? Is it a reference to a top-level module that also has sub-modules? Or is it a reference to a submodule in a repo (i.e. where a `//` is used)? Or both?

mikeurbanski1 (approved):

PR #3329, opened 2022-08-09
change github org schema and add UT

maxamel:
This PR fixes an issue with checkov github VCS flow.
Github graphql API can return null for some objects (unless they are explicitly defined as non null).
samlIdentityProvider can be null, but since our schema does not take it into account, it fails json validation and ignores the organization. This silent ignore causes this issue to remain hidden.

## Checklist:

### DIFF
--- a/tests/github/test_dal.py
+++ b/tests/github/test_dal.py
@@ -1,0 +1,28 @@
+import os
+from unittest import mock
+
+from pytest_mock import MockerFixture
+
+from checkov.github.dal import Github
+
+
+@mock.patch.dict(os.environ, {"GITHUB_ORG": "simpleOrg"}, clear=True)
+def test_org_security(mocker: MockerFixture):
+    dal = Github()
+    mock_data = {
+        "data": {
+            "organization": {
+                "name": "Bridgecrew",
+                "login": "Bridgecrew-dev",
+                "description": "",
+                "ipAllowListEnabledSetting": "DISABLED",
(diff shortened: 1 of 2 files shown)


tests/github/test_dal.py
```diff
@@ -0,0 +1,28 @@
...
+                "description": "",
+                "ipAllowListEnabledSetting": "DISABLED",
+                "ipAllowListForInstalledAppsEnabledSetting": "DISABLED",
+                "requiresTwoFactorAuthentication": False,
+                "samlIdentityProvider": None
+            }
+        }
+    }
+    graphql_mock = mocker.MagicMock(return_value=mock_data)
+    mocker.patch("checkov.common.vcs.base_vcs_dal.BaseVCSDAL._request_graphql", graphql_mock)
```
gruebel, line 26:
```suggestion
    mocker.patch("checkov.common.vcs.base_vcs_dal.BaseVCSDAL._request_graphql", return_value=mock_data)
```
if I'm not wrong, this should be enough 🙂  it makes sense to `MagicMock` when you actually want to validate that it was accessed or have more complex interactions with it.

gruebel (approved):
nice find 🚀 are there more areas, where we need to expect a `null` value?

Eliran-Turgeman (approved):

maxamel:
Thanks @gruebel, for this particular flow we shouldn't expect nulls any more.
We'll need to check regarding other flows (like gitlab, etc.).

PR #3335, opened 2022-08-10
chore: Removing dict[BlockTypeAlias, BicepElements] type error

luca-digrazia:
**"filename"**: "checkov/bicep/graph_builder/graph_to_tf_definitions.py"
**"warning_type"**: "Incompatible variable type [9]"
  "warning_message": " BLOCK_TYPE_TO_BICEP_ELEMENTS_MAP is declared to have type `typing.Dict[typing.Union[typing_extensions.Literal['module'], typing_extensions.Literal['output'], typing_extensions.Literal['param'], typing_extensions.Literal['resource'], typing_extensions.Literal['targetScope'], typing_extensions.Literal['var']], BicepElements]` but is used as type `typing.Dict[str, BicepElements]`."
**"warning_line"**: 15
**"fix"**: Removing dict[BlockTypeAlias, BicepElements]


gruebel:
hi @luca-digrazia thanks for opening the PR, but where do you get the error from? If it is some kind of type checker, then we use internally `mypy`, which can correctly handle this.

luca-digrazia:
> 

Hi, thanks for the reply. Yes, it is a Pyre warning and a fix for it.

gruebel (approved):
basically this is a `pyre` limitation and the type hinting was correct. Nevertheless `mypy` implies the correct type for the map and keeps working correctly, so I'm fine with the change.

PR #3342, opened 2022-08-10
chore: add flake8-type-checking plugin

gruebel:

## Description

- updated flake8, because `dlint` removed the upper bound in their new version 🎉 
- fixed couple of flake8 issues
- added a flake8 plugin to help with identifying an import issue, when it is under the `TYPE_CHECKING` block, but actually used in the code

Only drawback of the `flake8-type-checking` plugin it only works with Python 3.8, so either use Python 3.8 or use `pre-commit` to invoke it. If it finds an issue with an import it will look like this

```shell
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

checkov/bicep/runner.py:35:1: TC004 Move import 'NetworkxConnector' out of type-checking block. Import is used for more than type hinting.

```

## Checklist:


ayajbara (approved):

checkov/bicep/graph_builder/graph_to_tf_definitions.py
```diff
@@ -4,13 +4,12 @@
...
 from typing import Any, TYPE_CHECKING, cast
 
-from pycep.typing import BicepJson
-
 from checkov.bicep.graph_builder.graph_components.block_types import BlockType, BlockTypeAlias
 from checkov.bicep.graph_builder.local_graph import BicepElements, BicepElementsAlias
 
 if TYPE_CHECKING:
     from checkov.bicep.graph_builder.graph_components.blocks import BicepBlock
+    from pycep.typing import BicepJson
```
itai1357, line 12: why do we need it inside the if statement?
I thought that the "if TYPE_CHECKING" is for avoiding circular dependences

itai1357 (approved):

checkov/bicep/graph_builder/graph_to_tf_definitions.py
gruebel: basically during runtime the `if` block is ignored, so none of the imports is actually imported, which of course benefits performance and in some cases helps to prevent circular dependencies. Sadly the plugin can't find circular dependencies, it just helps to optimize the imports, which are need to type hints.

itai1357: thanks!

PR #3367, opened 2022-08-15
Fix string container use case in GHA

Eliran-Turgeman:

## Description

We assume that the keyword “container” in GHA workflow must be specified as follows:

```
container:
  image: <image_name>
```

But there is an additional option:
```
container: <image_name>
```

so “container” is parsed as a string and not as a dict, which will fail sca_image runner on such workflow file

Added two more resources and uts for GHA runner

## Checklist:


gruebel (approved):
good find 💪 can you rename the test file `image_referencer_tests.py` to `test_image_referencer.py` that's the preferred naming convention for pytest tests 🙂

lirshindalman (approved):
:rocket:

PR #3380, opened 2022-08-18
Fix AllowUnsecureCommandsOnJob check GHA

Eliran-Turgeman:

## Description

`env` keyword can also be specified as such

```
env: ${{ matrix.config.env }}
```

which breaks the check since it's parsed as a string

## Checklist:

### DIFF
--- a/checkov/github_actions/checks/job/AllowUnsecureCommandsOnJob.py
+++ b/checkov/github_actions/checks/job/AllowUnsecureCommandsOnJob.py
@@ -23,7 +23,7 @@ class AllowUnsecureCommandsOnJob(BaseGithubActionsCheck):
         if "env" not in conf or not conf["env"]:
             return CheckResult.PASSED, conf
         env_variables = conf.get("env", {})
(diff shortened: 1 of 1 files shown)


nimrodkor (approved):

checkov/github_actions/checks/job/AllowUnsecureCommandsOnJob.py
```diff
@@ -23,7 +23,7 @@ def scan_entity_conf(self, conf: Dict[str, Any]) -> Tuple[CheckResult, Dict[str,
         if "env" not in conf or not conf["env"]:
             return CheckResult.PASSED, conf
         env_variables = conf.get("env", {})
-        if env_variables.get("ACTIONS_ALLOW_UNSECURE_COMMANDS", False):
+        if isinstance(env_variables, dict) and env_variables.get("ACTIONS_ALLOW_UNSECURE_COMMANDS", False):
```
arielkru, line 26: So if this is false, I think we should return `CheckResult.UNKOWN`, since we don't parse the variable

arielkru (approved):
🚀

PR #3406, opened 2022-08-23
fix: Parse only openapi files

PelegLi:

## Description

OpenAPI runner loads and parses many YML files unrelated to openapi (most YMLs aren't). Parsing many irrelevant files is time consuming, so in order to decrease the number of irrelevant files parsed, a soft pre-validation on the raw text file is added before parsing, excluding YMLs clearly unrelated to openapi.


checkov/openapi/runner.py
```diff
@@ -46,7 +46,14 @@ def parse_format(
         func: _ParseFormatJsonCallable | _ParseFormatYamlCallable,
     ) -> tuple[dict[str, Any] | list[dict[str, Any]], list[tuple[int, str]]] | None:
         try:
-            parsed_file = func(self, f)
+            if f.endswith(".json"):
+                parsed_file = func(self, f)
```
gruebel, line 50: why not also do it for JSON files?

gruebel (commented):
nice, always good to have a bit of pre evaluation.

rotemavni (commented):
@PelegLi Can you create a test for the runner to scan a dir with multiple files, some for OpenAPI and some not, and validate the check results are as expected (as they were before the change)?

checkov/openapi/runner.py
PelegLi: 1. The main pain for openapi resides in the YAMLs, I want to check first if it improves performance and doesn't raise any other issues.
2. YAML runner allows passing the file content to the parser, so we don't need to open and load the file twice. JSON runner currently does not allow passing the content, and modifying it requires a deeper and riskier change.

gruebel: sounds good 🙂

PelegLi:
> @PelegLi Can you create a test for the runner to scan a dir with multiple files, some for OpenAPI and some not, and validate the check results are as expected (as they were before the change)?

Done

gruebel (approved):
☕

ayajbara (approved):

PR #3411, opened 2022-08-24
fix(terraform): do not evaluate OCI policy statements

behnazh-w:

[//]: # "
    # PR Title
    Additionally we encourage adding a scope to the prefix, which indicates the targeted framework, otherwise 'general' will be assumed.
    ex.
"

## Description

Checkov currently evaluates the `statements` attributes of a Terraform resource, which contain IAM policies. However, the syntax of the policies differ for different providers, which results in unexpected evaluations. Oracle Cloud Infrastructure uses a [special syntax](https://docs.oracle.com/en-us/iaas/Content/Identity/Concepts/policysyntax.htm) for the policy statements, but Checkov seems to try to evaluate it as a JSON object. 

Here is a solution which prevents evaluating the `statements` attributes when the provider is set to `OCI` through an environment variable.

Closes https://github.com/bridgecrewio/checkov/issues/3412

### DIFF
--- a/checkov/terraform/graph_builder/variable_rendering/renderer.py
+++ b/checkov/terraform/graph_builder/variable_rendering/renderer.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 import re
@@ -26,7 +28,7 @@ if TYPE_CHECKING:
     from checkov.terraform.graph_builder.local_graph import TerraformLocalGraph
 
 ATTRIBUTES_NO_EVAL = ["template_body", "template"]
-VAR_TYPE_DEFAULT_VALUES = {
+VAR_TYPE_DEFAULT_VALUES: dict[str, list[Any] | dict[str, Any]] = {
     'list': [],
     'map': {}
 }
@@ -137,7 +139,7 @@ class TerraformVariableRenderer(VariableRenderer):
             if value is not None:
                 return value
 
-        if attributes.get(CustomAttributes.BLOCK_TYPE) in [BlockType.VARIABLE, BlockType.TF_VARIABLE]:
+        if attributes.get(CustomAttributes.BLOCK_TYPE) in (BlockType.VARIABLE, BlockType.TF_VARIABLE):
             var_type = attributes.get('type')
             default_val = attributes.get("default")
             if default_val is None:
@@ -152,8 +154,8 @@ class TerraformVariableRenderer(VariableRenderer):
         return None
 
     @staticmethod
-    def get_default_placeholder_value(var_type):
-        if not var_type or type(var_type) != str:
+    def get_default_placeholder_value(var_type: Any) -> list[Any] | dict[str, Any] | None:
+        if not var_type or not isinstance(var_type, str):
             return None
         match = TYPE_REGEX.match(var_type)
         return VAR_TYPE_DEFAULT_VALUES.get(match.group(2)) if match else None
@@ -267,7 +269,7 @@ class TerraformVariableRenderer(VariableRenderer):
                 for inner_val in lst_curr_val:
                     if (
                         isinstance(inner_val, str)
-                        and not any(c in inner_val for c in ["{", "}", "[", "]", "="])
+                        and not any(c in inner_val for c in ("{", "}", "[", "]", "="))
                         or attribute in ATTRIBUTES_NO_EVAL
                     ):
                         evaluated_lst.append(inner_val)
--- a/checkov/terraform/graph_builder/variable_rendering/renderer.py
+++ b/checkov/terraform/graph_builder/variable_rendering/renderer.py
@@ -27,7 +27,6 @@ from checkov.terraform.graph_builder.variable_rendering.evaluate_terraform impor
 if TYPE_CHECKING:
     from checkov.terraform.graph_builder.local_graph import TerraformLocalGraph
 
-ATTRIBUTES_NO_EVAL = ["template_body", "template"]
 VAR_TYPE_DEFAULT_VALUES: dict[str, list[Any] | dict[str, Any]] = {
     'list': [],
     'map': {}
@@ -46,7 +45,6 @@ class TerraformVariableRenderer(VariableRenderer):
         if os.environ.get("CKV_TERRAFORM_PROVIDER") == "OCI":
             # OCI policy statements have a special syntax and should not be evaluated.
             self.attributes_no_eval.add("statements")
-
 
     def evaluate_vertex_attribute_from_edge(self, edge_list: List[Edge]) -> None:
         multiple_edges = len(edge_list) > 1
--- a/checkov/terraform/graph_builder/variable_rendering/renderer.py
+++ b/checkov/terraform/graph_builder/variable_rendering/renderer.py
@@ -42,6 +42,12 @@ class TerraformVariableRenderer(VariableRenderer):
     def __init__(self, local_graph: "TerraformLocalGraph") -> None:
         super().__init__(local_graph)
 
+        self.attributes_no_eval = {"template_body", "template"}
+        if os.environ.get("CKV_TERRAFORM_PROVIDER") == "OCI":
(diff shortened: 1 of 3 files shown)


behnazh-w:
@gruebel Is there anything I can do on my end to help merging this PR?

gruebel:
hey @behnazh-w I adjusted your code a bit. Please have a look and you can remove your change in the GitHub workflow file 🙂

checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -25,8 +27,7 @@
 if TYPE_CHECKING:
     from checkov.terraform.graph_builder.local_graph import TerraformLocalGraph
 
-ATTRIBUTES_NO_EVAL = ["template_body", "template"]
-VAR_TYPE_DEFAULT_VALUES = {
+VAR_TYPE_DEFAULT_VALUES: dict[str, list[Any] | dict[str, Any]] = {
```
behnazh-w, line 30: @gruebel Because these changes seem to be type and linting fixes that are not related to my changes, do you mind if I add them in a separate commit to have a clean history? That would mean that the two commits shouldn't be squashed though in this PR.

checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
behnazh-w, line 1: Just out of curiosity, what's the purpose of importing `annotations` here? Is it for performance reasons as described in [PEP 563](https://peps.python.org/pep-0563/)?

tests/terraform/graph/graph_builder/test_oci_policy.py
```diff
@@ -1,43 +1,38 @@
...
+from pathlib import Path
+from unittest import mock
 
-from checkov.common.graph.db_connectors.networkx.networkx_db_connector import (
-    NetworkxConnector,
-)
+from checkov.common.graph.db_connectors.networkx.networkx_db_connector import NetworkxConnector
 from checkov.terraform.graph_manager import TerraformGraphManager
 
+@mock.patch.dict(os.environ, {"CKV_TERRAFORM_PROVIDER": "OCI"})
```
behnazh-w, line 10: I also considered mocking the `env` variable but because `attributes_no_eval` is extended at module load time, the test wouldn't have exercised my change exactly. So I decided to set the `env` variable before running `pytest` in the GHA workflow to make sure it doesn't have any side effects. If you think mocking is enough for testing, I'm fine with it.

tests/terraform/graph/graph_builder/test_oci_policy.py
```diff
@@ -1,43 +1,38 @@
...
+def test_oci_policy_statements_with_provider_env_var():
```
behnazh-w, line 11: I followed the other tests and created a class to be consistent. But if a method is enough, I'm happy with that.

tests/terraform/graph/graph_builder/test_oci_policy.py
```diff
@@ -1,43 +1,38 @@
...
-                str(
-                    local_graph.vertices[0].config["oci_identity_policy"]["example"][
-                        "statements"
-                    ]
-                ),
-                "[[\"allow group group-admin-001 to use groups in tenancy where target.group.name != 'Administrators'\"]]",
-            )
+    # then
+    statements = local_graph.vertices[0].config["oci_identity_policy"]["example"]["statements"][0]
+    assert statements == ["True"]
```
behnazh-w, line 38: I can add a comment to clarify that `"True"` is a wrong value due to special syntax in OCI policy.

behnazh-w:
@gruebel Thanks a lot for making the effort. I have left some comments and questions.

behnazh-w:
@gruebel thanks again for reviewing this PR. I separated the changes to two commits: first one is for fixing the linting issues and the second one fixes the policy evaluation. 

It's ready to be merged. Just a reminder not to squash the two commits to avoid mixing the linting fixes with the policy fix.

checkov/terraform/graph_builder/variable_rendering/renderer.py
gruebel: mostly for better and easier type hinting 🙂 you can use built-in types instead of the typing variants, like `list` instead of `List`.

tests/terraform/graph/graph_builder/test_oci_policy.py
gruebel: yeah it is kind of mixed. we started with `untitest` style and still do them, but as a test runner we use `pytest` and then you actually can just use functions, which I prefer 🙂

tests/terraform/graph/graph_builder/test_oci_policy.py
gruebel: yeah, it only works because I changed the global variable to be an instance variable, otherwise this approach wouldn't work, because they are only loaded once.

checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -40,6 +41,11 @@ class TerraformVariableRenderer(VariableRenderer):
     def __init__(self, local_graph: "TerraformLocalGraph") -> None:
         super().__init__(local_graph)
 
+        self.attributes_no_eval = {"template_body", "template"}
+        if os.environ.get("CKV_TERRAFORM_PROVIDER") == "OCI":
```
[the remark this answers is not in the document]
nimrodkor: Suggestion - have attributes_no_eval be a method, that gets the vertex index, and checks if the vertex at this index is an OCI terraform resource - starts with `oci_`. If so - does not render statements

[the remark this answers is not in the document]
behnazh-w: > Suggestion - have attributes_no_eval be a method, that gets the vertex index, and checks if the vertex at this index is an OCI terraform resource - starts with `oci_`. If so - does not render statements

@nimrodkor Thanks for the suggestion.  Please see commit 1016cdacf0f545210b24c727c1986b609df8ab01.

nimrodkor (approved):

checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -40,6 +41,26 @@ class TerraformVariableRenderer(VariableRenderer):
...
+        :param attribute: the attribute to check
+        :param vertex_index: the index of the current vertex
+        :return bool: True if the attribute should not be evaluated and False otherwise
+        """
+        if attribute in {"template_body", "template"}:
+            return True
+
+        # OCI policy statements have a special syntax and should not be evaluated.
+        # Check if the vertex at this index is an OCI terraform resource.
+        vertex_attributes = self.local_graph.get_vertex_attributes_by_index(vertex_index)
```
rotemavni, line 57: @behnazh-w How do you feel about fetching the attributes only if `attribute == "statements"`? I would be more efficient (this way we don't access the graph for attributes that should not be skipped)

rotemavni (commented):
Thank you for you contribution! Had just one comment.

checkov/terraform/graph_builder/variable_rendering/renderer.py
behnazh-w: > @behnazh-w How do you feel about fetching the attributes only if `attribute == "statements"`? I would be more efficient (this way we don't access the graph for attributes that should not be skipped)

Thanks @rotemavni for the feedback. Please see commit eaeceb10c88a27b04050676d5f2144bacfcfcf5c.

rotemavni (approved):

checkov/terraform/graph_builder/variable_rendering/renderer.py
rotemavni: Thank you @behnazh-w, I approved the PR

PR #3427, opened 2022-08-27
chore: moving tests that depend sca_package-report to use exact the same input report

itai1357:

## Description
moving tests that depend sca_package-report to use exact the same input report
the target is making sure we are testing the dependencies of sca_package report in case we have some change in our report.

## References:
https://github.com/pytest-dev/pytest-mock/pull/182/files (PR for enabling "package_mocker" which I used in the conftest.py)

## Checklist:


functional_tests/sca_package/test_sca_package.py
```diff
@@ -0,0 +1,102 @@
...
+    ##############################################################################################################
+    # phase 1: creating report (without asserting it, as it already happens in the UTs for sca_package's runner) #
+    ##############################################################################################################
+    report = Runner().run(root_folder=EXAMPLES_DIR)
+
+    ####################################################
+    # phase 2: creating cli output based on the report #
+    ####################################################
+
+    cli_output = report.print_console(False, False, None, None, False)
```
gruebel, line 30: I actually understood the intention differently, to test a full run, like a user would invoke it.
The current approach invokes the phases (from phase 2) independently with the internal functions to create the output, so why create one long test and not for each phase one test, which is basically the current practice in https://github.com/bridgecrewio/checkov/blob/master/tests/sca_package/test_output.py and https://github.com/bridgecrewio/checkov/blob/master/tests/sca_package/test_output_reports.py
So, what is the benefit of adding a long test compared to multiple small ones, which test each output on its own? If the initial report creation is time consuming, then you can create a fixture, where you create the report once and use it to test each of the output possibilities.

itai1357: let assume you do some change in the sca_package runner.
it may affect also the cli-output and the box-reports

in the current approach, you have to adjust at least 4 tests, as each has its own inputs, and of course you may miss any
in the suggested approach, you have just to adjust 1 tests!

this is my idea. regarding what you said about the "fixture", we indeed can use the same input for all the tests that depends on sca_package-runner report.
but we will have to move the bom-report tests (those that relevant to sca_package) from test/common to test/sca_package for being under the same conftest.py
what do you say?

gruebel: if the output tests are mainly target for SCA, then no problem to move them there. leveraging a fixture to be passed to single tests also makes local testing faster, because you only need to run the test for which output you currently adjust and then run all the tests in the file/folder to check for regressions.

itai1357: @gruebel fixed

tests/sca_package/test_output_reports.py
```diff
@@ -7,11 +7,111 @@
...
+from checkov.common.output.csv import CSVSBOM, FILE_NAME_OSS_PACKAGES
 
 EXAMPLES_DIR = Path(__file__).parent / "examples"
 
 
+def test_console_output(sca_package_report):
+    console_output = sca_package_report.print_console(False, False, None, None, False)
+    print()
+    print(console_output)
+    print()
```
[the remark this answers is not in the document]
itai1357: sure!!

Saarett (approved):

PR #3443, opened 2022-08-30
feat: support injection of custom plugins

maxamel:
This PR adds support for injecting custom plugins from the outside. This is part of secret detection capability.
Checkov reads the detector from file and loads it together with its own internal detectors.

## Checklist:

### DIFF
--- a/checkov/secrets/runner.py
+++ b/checkov/secrets/runner.py
@@ -181,7 +181,7 @@ class Runner(BaseRunner[None]):
             secrets_duplication: dict[str, bool] = {}
             for _, secret in secrets:
                 if hasattr(secret, 'check_id'):
-                    check_id = secret.check_id
+                    check_id = secret.check_id      # type: ignore
                 else:
                     check_id = SECRET_TYPE_TO_ID.get(secret.type)
                 if not check_id:
--- a/checkov/secrets/runner.py
+++ b/checkov/secrets/runner.py
@@ -127,10 +127,10 @@ class Runner(BaseRunner[None]):
                     'limit': ENTROPY_KEYWORD_LIMIT
                 }
             ]
-        custom_plugins = os.getenv("CHECKOV_CUSTOM_DETECTOR_PLUGINS", None)
+        custom_plugins = os.getenv("CHECKOV_CUSTOM_DETECTOR_PLUGINS_PATH", None)
         logging.info(f"Custom detector flag set to {custom_plugins}")
         if custom_plugins:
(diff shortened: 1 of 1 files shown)


lirshindalman (approved):

checkov/secrets/runner.py
```diff
@@ -128,6 +128,22 @@ def run(
...
             ]
+        custom_plugins = os.getenv("CHECKOV_CUSTOM_DETECTOR_PLUGINS_PATH", None)
+        logging.info(f"Custom detector flag set to {custom_plugins}")
+        if custom_plugins:
+            detector_path = f"{custom_plugins}/platform_regex_detector.py"
+            file_exists = exists(detector_path)
+            if file_exists:
+                logging.info(f"Custom detector found at {detector_path}. Loading...")
+                plugins_used.append({
+                    'name': 'PlatformRegexDetector',
```
Eliran-Turgeman, line 139: Does that mean that all the plugins we add in this way would be named `PlatformRegexDetector`?

maxamel: No, that's just the name of the common infra plugin I created. Each detector plugin will have his own name and will be loaded by platform_regex_detector itself.

Eliran-Turgeman (approved):

checkov/secrets/runner.py
Eliran-Turgeman: 🔥

checkov/secrets/runner.py
```diff
@@ -128,6 +128,22 @@ def run(
                     'limit': ENTROPY_KEYWORD_LIMIT
                 }
             ]
+        custom_plugins = os.getenv("CHECKOV_CUSTOM_DETECTOR_PLUGINS_PATH", None)
```
gruebel, line 131:
```suggestion
        custom_plugins = os.getenv("CHECKOV_CUSTOM_DETECTOR_PLUGINS_PATH")
```
`None` is always the default, so no need to extra add it here 😄

checkov/secrets/runner.py
```diff
@@ -128,6 +128,22 @@ def run(
...
+        logging.info(f"Custom detector flag set to {custom_plugins}")
+        if custom_plugins:
+            detector_path = f"{custom_plugins}/platform_regex_detector.py"
```
gruebel, line 134:
```suggestion
            detector_path = f"{custom_plugins}/custom_regex_detector.py"
```
since we talk about custom detectors here, lets also name it like that

gruebel (commented):
nice, looks good, added some minor suggestions

PR #3457, opened 2022-09-02
fix(terraform): gcp postgresql default values

### LINKED ISSUE #3323: GoogleCloudPostgreSql checks falsely pass when not declared at all
**Describe the issue**
Checks regarding GoogleCloudPostgreSql (database_flags) seems to return failure only when explicitly set while not taking into account default behaviour from cloud provider. Instead, I would suggest we invert the logic of the tests and also add missing flags as failure. I've tested this with `log_lock_waits` and `log_connections` and by looking at code it seems to expand to more. I would like to volunteer to fine comb all those checks according to official documentation and maybe even add some more checks. 

**Examples**
It should be `if (flag['name'] == 'log_lock_waits') and (flag['value'] == 'on'):` and also fix `pass2` in example as `fail2` and add it in test.

 - Checkov Version 2.1.90

losisin:

## Description

Fix checks when using default (undeclared) settings from GCP.

Fixes #3323 

## Edited policies

`CKV_GCP_51`, `CKV_GCP_52`, `CKV_GCP_53` and `CKV_GCP_54`.

### Description

When scanning terraform plan with checkov and running infrastructure with something like steampipe, result should match. Here is a list of all available PostgreSQL database flags with their respective default values: https://cloud.google.com/sql/docs/postgres/flags#list-flags-postgres

## Checklist:


gruebel (commented):
in general looks good, just need to fix the tests and then it is ready to be merged 🚀

gruebel (approved):
nice, great job 🥇

Saarett (approved):

PR #3460, opened 2022-09-02
platform(general): trim API urls

mikeurbanski1:

[//]: # "
    # PR Title
    Additionally we encourage adding a scope to the prefix, which indicates the targeted framework, otherwise 'general' will be assumed.
    ex.
"

## Description

Trims the BC and Prisma API URL inputs.
## Checklist:

### DIFF
--- a/checkov/common/util/http_utils.py
+++ b/checkov/common/util/http_utils.py
@@ -9,7 +9,7 @@ import time
 import os
 import aiohttp
 import asyncio
-from typing import Any, TYPE_CHECKING, cast, Optional
+from typing import Any, TYPE_CHECKING, cast, Optional, overload
 
 from urllib3.response import HTTPResponse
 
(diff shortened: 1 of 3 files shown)


gruebel (approved):
I fixed the `mypy` issue for you 🙂

checkov/common/util/http_utils.py
```diff
@@ -25,11 +25,27 @@
...
 
+@overload
+def normalize_bc_url(url: None) -> None:
+    ...
+
+
+@overload
+def normalize_bc_url(url: str) -> str:
+    ...
+
```
itai1357, line 36: HI! I 
never saw such a using of overload.
can you please share why those 2 function are necessary?
it seem that the third one is enough

gruebel: the `@overload` functionality in Python is only for typing purposes. It gives a possibility to differentiate between what kind of types go in and what to what should come out. In this example it is clear, if `None` goes in, `None` should come out, same for `str`.
If you don't add the `@overload` annotation, then the type checker will assume that either `str` or `None` could return, which can be ok, but then you have handle it downstream and always check, if it `is not None`

maxamel (approved):

itai1357 (approved):

checkov/common/util/http_utils.py
itai1357: cool!

PR #3485, opened 2022-09-07
chore: Unpin dpath

jayvdb:
**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

## Description

This allows using the latest `dpath`, and also the current minimum version of `dpath`, as these have always existed in `dpath.util`; c.f. https://github.com/dpath-maintainers/dpath-python/blob/build%2C1.5%2C0/dpath/util.py

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my feature, policy, or fix is effective and works
- [x] New and existing tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #3490, opened 2022-09-08
Revert "chore: Unpin dpath"

nimrodkor:
Reverts bridgecrewio/checkov#3485

Sorry @jayvdb , but this is breaking downstream dependencies for us. Reverting for now, until we find a better solution.
CC: @gruebel


jayvdb:
@nimrodkor hi, could you let me know what "downstream" is ?

Issue #3222, opened 2022-07-14
Move CLI output summary to the bottom of the output


tsmithv11:
**Describe the issue**
If you have more than a few failed checks, it can be difficult to scroll to the top to see the summary or in a terminal, it may even be cut off. Can we move the summary to the bottom after the policy violations?

nimrodkor:
Sensational! I love this. 
The relevant part of our code that needs to be refactored:
https://github.com/bridgecrewio/checkov/blob/5fdaf23e404b247f686ce6e07ec8f98032380a2d/checkov/common/output/report.py#L245
Would appreciate a contribution here!

PR #3497, opened 2022-09-12
feat(general): add flag for summary position

marynaKK:

[//]: # "
    # PR Title
    ex.
"

## Description

Fixes https://github.com/bridgecrewio/checkov/issues/3222

### Description
Added a flag to enable changing the summary position within the report. Default behavior didn't change - summary at the top. adding the flag --summary-position-bottom will move it to the bottom.

## Checklist:

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -575,7 +575,7 @@ def add_parser_args(parser: ArgumentParser) -> None:
                action='append',
                help='black file list to filter out from the secret scanner')
     parser.add('--summary-position-bottom', action='store_true',
-               help='change the summary to be appended in the bottom after the policy violations')
+               help='Change the summary to be appended in the bottom after the policy violations')
 
 
 def get_external_checks_dir(config: Any) -> Any:
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -574,6 +574,8 @@ def add_parser_args(parser: ArgumentParser) -> None:
                env_var='CKV_SECRETS_SCAN_BLACK_LIST',
                action='append',
                help='black file list to filter out from the secret scanner')
+    parser.add('--summary-position-bottom', action='store_true',
+               help='change the summary to be appended in the bottom after the policy violations')
 
 
 def get_external_checks_dir(config: Any) -> Any:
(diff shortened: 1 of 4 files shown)


arielkru:
Very nice @marynaKK !
I see James suggested the default position to be at the bottom and it makes sense to me, shouldn't we implement it that way?
https://github.com/bridgecrewio/checkov/issues/3222#issuecomment-1184213825

marynaKK:
@arielkru You make sense :) I thought to preserve the current behavior, the flag default is top, and if it is used the summary moves to the bottom.

nimrodkor (commented):
Is there a test for this @marynaKK ? If not - can we write one?

gruebel:
if we change the default behaviour, then we should mark this PR as a breaking change -> `break(general): ...`

checkov/main.py
```diff
@@ -574,6 +574,8 @@ def add_parser_args(parser: ArgumentParser) -> None:
                env_var='CKV_SECRETS_SCAN_BLACK_LIST',
                action='append',
                help='black file list to filter out from the secret scanner')
+    parser.add('--summary-position-bottom', action='store_true',
```
gruebel, line 577:
```suggestion
    parser.add('--summary-position', choices=["top", "bottom"],
```
we could change it to allow two different choices top or bottom and then set the default to top or bottom. of course define them as constants and then they can be evaluated before printing to console.

nimrodkor (approved):
No tests in that area then... Only update the docs:
https://github.com/bridgecrewio/checkov/blob/96962a3c3b2301b4c1567ded3ffa7765767f26a7/docs/2.Basics/CLI%20Command%20Reference.md#L8
@gruebel This is **not** a breaking change, only if you set the arg do you get a different behavior

gruebel (commented):
nice, great job on the first PR 💪

gruebel:
@nimrodkor that's why I wrote, if we change the default behaviour 🙂 if we don't, we don't need 😄

marynaKK:
@gruebel Your suggestion is more clear - I'll keep it in mind to keep things simpler :)
changed it.

gruebel (approved):
great work ☕ 🍰

PR #3518, opened 2022-09-16
feat(imageReferencer): Split `PRESENT_CACHED_RESULTS` env var to 2 feature flag like vars

arielkru:

[//]: # "
    # PR Title
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/sca_image/runner.py
+++ b/checkov/sca_image/runner.py
@@ -174,11 +174,12 @@ class Runner(PackageRunner):
             if image_referencer.is_workflow_file(abs_fname):
                 images = image_referencer.get_images(file_path=abs_fname)
                 for image in images:
-                    if not strtobool(os.getenv('CHECKOV_PRESENT_CACHED_RESULTS', "False")):
+                    if strtobool(os.getenv('CKV_CREATE_SCA_IMAGE_REPORTS_FOR_IR', "True")):
                         image_report = self.get_image_report(dockerfile_path=abs_fname, image=image,
                                                              runner_filter=runner_filter)
                         merge_reports(report, image_report)
-                    else:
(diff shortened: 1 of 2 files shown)


checkov/sca_image/runner.py
```diff
@@ -174,11 +174,12 @@ def iterate_image_files(self, abs_fname: str, report: Report, runner_filter: Run
             if image_referencer.is_workflow_file(abs_fname):
                 images = image_referencer.get_images(file_path=abs_fname)
                 for image in images:
-                    if not strtobool(os.getenv('CHECKOV_PRESENT_CACHED_RESULTS', "False")):
+                    if strtobool(os.getenv('CKV_CREATE_SCA_IMAGE_REPORTS_FOR_IR', "True")):
```
gruebel, line 177: can you change the prefix for both env vars to `CHECKOV_...` we try to use the full name for internal, not really meant to be changed by a user and `CKV_...` is used for user related flags and behaviour changes.

gruebel (approved):
noice 🍻

checkov/sca_image/runner.py
Eliran-Turgeman, line 177: Probably a nit, but is `CKV_CREATE_SCA_IMAGE_REPORTS_FOR_IR` supposed to be True by default?

If we want to allow this functionality by default then I think the env var name should express the negation of the functionality, something like - `CHECKOV_NO_SCA_IMAGE_REPORTS_FOR_IR`, if that makes sense

gruebel: I know what you mean, but I don't think there is any issue with having an env var which is by default enabled, we also do it in other areas. It can also become confusing that to similar env vars needs to be configured differently. And lastly they are not really meant to be used by normal users 🙂

arielkru: My thought was like Anton's, that since these two env vars come as a logical pair, it'd be clearer to have them in the same direction (both FF that enable a feature). And yes, the first by default is on and the second is off.

Eliran-Turgeman (approved):

PR #3535, opened 2022-09-19
fix(general): handle fixes for cloned OOTB policies

mikeurbanski1:

[//]: # "
    # PR Title
    ex.
"

## Description

Handles a case where fixes get returned with a custom policy ID, when an OOTB fixable policy is cloned. Checkov would fail to map them to the correct record, and logged an error message. This also resulted in the IDE extensions thinking the scan failed.

This fixes the core issue and also logs the actual error, if it occurs, at a lower level, because it does not break the overall scan.

## Checklist:

### DIFF
--- a/tests/common/integration_features/test_fixes_integration.py
+++ b/tests/common/integration_features/test_fixes_integration.py
@@ -105,11 +105,10 @@ class TestFixesIntegration(unittest.TestCase):
         self.assertTrue(all(r.fixed_definition is not None for r in report.failed_checks))
 
     def setUp(self) -> None:
-        global _old_check_metadata
-        _old_check_metadata = metadata_integration.check_metadata
+        self._old_check_metadata = metadata_integration.check_metadata
 
     def tearDown(self) -> None:
-        metadata_integration.check_metadata = _old_check_metadata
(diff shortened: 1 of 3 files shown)


tests/common/integration_features/test_fixes_integration.py
```diff
@@ -26,6 +38,102 @@ def test_integration_valid(self):
...
+        fixes_integration.post_runner(report)
+
+        self.assertTrue(all(r.fixed_definition is not None for r in report.failed_checks))
+
+    def setUp(self) -> None:
+        global _old_check_metadata
+        _old_check_metadata = metadata_integration.check_metadata
+
+    def tearDown(self) -> None:
+        metadata_integration.check_metadata = _old_check_metadata
```
gruebel, line 112:
```suggestion
    def setUp(self) -> None:
        self._old_check_metadata = metadata_integration.check_metadata

    def tearDown(self) -> None:
        metadata_integration.check_metadata = self._old_check_metadata
```
this should do the trick without needing to use global 🙂

gruebel (approved):
🍻

rotemavni (approved):

PR #3585, opened 2022-09-29
fix(github): fix variables initialization

marynaKK:

## Description

The initialization of the variables in the init of GitHub class is done before the type is set to string (by assigning the "" string). This caused to no checks at all from the GitHub framework (as no repo/account/rtc. are set in the first place)

## Checklist:

### DIFF
--- a/checkov/github/dal.py
+++ b/checkov/github/dal.py
@@ -9,13 +9,15 @@ from checkov.github.schemas.org_security import schema as org_security_schema
 
 
 class Github(BaseVCSDAL):
+    github_conf_dir_path: str
+    github_org_security_file_path: str
+    github_branch_protection_rules_file_path: str
+    github_org_webhooks_file_path: str
+    github_repository_webhooks_file_path: str
+    github_repository_collaborators_file_path: str
+    org: str
+
     def __init__(self) -> None:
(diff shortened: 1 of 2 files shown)


checkov/github/dal.py
```diff
@@ -10,14 +10,13 @@
...
     def __init__(self) -> None:
-        super().__init__()
-
         self.github_conf_dir_path = ""
         self.github_org_security_file_path = ""
         self.github_branch_protection_rules_file_path = ""
         self.github_org_webhooks_file_path = ""
         self.github_repository_webhooks_file_path = ""
         self.github_repository_collaborators_file_path = ""
+        super().__init__()
```
gruebel, line 19: can you add a comment the init has to be at the end and also add or extend an existing UT to check some/all attributes are filled with something?

maxamel (approved):

checkov/github/dal.py
marynaKK: Hi! thanks for the review :) I've added a test for this (all paths should be initialized even if not used)
Also, I moved the type declaration (by Nimrod's suggestion) to be in the class instead of the init method.
💪 🐶

PR #3596, opened 2022-09-30
fix(dockerfile): Check EXPOSE 22/tcp

### LINKED ISSUE #3298: CKV_DOCKER_1 rule not failed when using EXPOSE 22/TCP
**Issue**
If it is related to an existing check, CKV_DOCKER_1.
Dockerfile EXPOSE can accept port using tcp proto by defaut if not specified or you can also specify the protocol with /tcp or /udp.

When using "EXPOSE 22/tcp" the rule does not detect the SSH port. something like '22/tcp'.split('/')[0] will return port without protocol or we can test both '22' or '22/tcp' in the rule.

**Examples**
Issue can be replicated with a Dockerfile which uses "EXPOSE 22/tcp".

**Version**
 - Checkov Version 2.1.87

**Specification**
EXPOSE specification https://docs.docker.com/engine/reference/builder/#expose

turingbeing:

[//]: # "
    # PR Title
    ex.
"

## Description

- Updated EXPOSE check to look for `22` or `22/tcp`
- Updated failing dockerfile to test.
- Didn't include a check for `22/udp` as that doesn't seem valid in the context of this port.

Fixes #3298 

## Checklist:


checkov/dockerfile/checks/ExposePort22.py
```diff
@@ -13,7 +13,7 @@ def __init__(self):
     def scan_entity_conf(self, conf):
         i = 0
         for expose_term in conf:
-            if "22" in expose_term['value'].split(' '):
+            if "22" | "22/tcp" in expose_term['value'].split(' '):
```
gruebel, line 16: this syntax doesn't do what you think it should do 🙂 
```suggestion
            if any(port in expose_term['value'].split(' ') for port in ("22", "22/tcp")):
```
you need to do something like that

tests/dockerfile/resources/expose_port/fail/Dockerfile
```diff
@@ -4,5 +4,6 @@ COPY package*.json ./
 RUN npm install
 COPY . .
 EXPOSE 3000 22
+EXPOSE 22/tcp
```
gruebel, line 7: this is actually not the right place, this check doesn't have a proper test file with examples yet. They need to be added here https://github.com/bridgecrewio/checkov/tree/master/tests/dockerfile/checks

checkov/dockerfile/checks/ExposePort22.py
turingbeing: Thanks for the review I will rectify tomorrow I can now why that wouldn't work

tests/dockerfile/resources/expose_port/fail/Dockerfile
turingbeing: @gruebel I have made the code changes, and created the tests. But I actually ran into a bit of an issue, which I ran into last time, but thought it might have just been me not fully understanding the setup.

When I try and setup the dev requirements using `pipenv install --dev` I just get a wall of Red Text.

I believe it's because the `Pipfile.lock` doesn't contain appropriate hashed for my platform `darwin` `arm64`

Should I create a separate issue for this?

turingbeing: `pyston` was the main trouble maker, but according to this release note, I'd need to update the `Pipenv` so we can download the `arm64` package

[v2.3.3](https://github.com/pyston/pyston/releases/tag/pyston_2.3.3)

turingbeing: @gruebel I could still do with some help on the above, I have the code ready but reluctant to push without testing

gruebel: hey @turingbeing what exactly is your issue at the moment? you can't run the tests or you don't know where to add a test for your change?

turingbeing: I've added the tests, but can't run them or install some of the dev-requirements because of architecture.

To make it easier I might just dig out my Linux Laptop and execute them there.

I'll push up my changes meantime but won't request a re-review until I've confirmed the tests pass

gruebel: gotcha, yeah just push your changes and I will check it anyway. it is not such a big change in total 😄

gruebel (approved):
🙂

YaaraVerner (approved):

PR #3611, opened 2022-10-03
chore(Dockerfile): simplify dependency to speed up the build

PeterDaveHello:
## Description

There's no need to have `wget` and `curl` in the build process, especially they're both going to be removed at the end of the build, simplify the dependency in Dockerfile can speed up the Docker image build process, save some time and network bandwidth.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,22 +8,21 @@ RUN set -eux; \
     apt-get install -y --no-install-recommends \
             ca-certificates \
             git \
-            wget \
             curl \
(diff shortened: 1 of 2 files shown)


gruebel (approved):
thanks for the contribution 🚀 it is true that wget is not needed here, but the speed improvement is minimal, probably under 1 sec, because the wget package has a size of ~300kb.

tronxd (approved):

PR #3621, opened 2022-10-06
platform(general): Get resources from platform and filter taggable resources for policies

mikeurbanski1:

[//]: # "
    # PR Title
    ex.
"

## Description

- Gets the list of all resource types and their attributes from the platform
- Also gets a list of attributes that we want to filter by resource type (this is just tags / labels for the various providers right now)
- Gets the provider from the updated platform policy metadata response
- Checkov will apply the logic to replace any policies using these attributes with a resource type of "all" with the resource types for the given attribute and provider (if available).

So, for example, if you have a platform policy with provider `aws` and a condition that checks `tags` with resource types `all`, then the check will only apply to taggable AWS Terraform resources. If you use the attribute `Tags.*.Key`, then it will only apply to CloudFormation.

If you have a Checkov policy that uses `all` with the attribute `tags`, then it will apply it to AWS, Azure, and Ali (because checkov policies do not have a provider in the spec).

This also removes the local list of resource types from checkov, which was pretty much totally dead code in reality, because platform custom policies didn't even have the provider in the API response.

Requires some platform changes (some done, some in progress) before this can be 100% safely merged (but it _should_ still work)

## Checklist:


mikeurbanski1:
I am in typing hell

checkov/common/bridgecrew/integration_features/features/attribute_resource_types_integration.py
```diff
@@ -0,0 +1,108 @@
...
+if TYPE_CHECKING:
+    from checkov.common.bridgecrew.platform_integration import BcPlatformIntegration
+
+
+ALL_TYPES = '__all__'
+
+
+class AttributeResourceTypesIntegration(BaseIntegrationFeature):
+    def __init__(self, bc_integration: BcPlatformIntegration) -> None:
+        super().__init__(bc_integration=bc_integration, order=1)  # must be after policy metadata
```
tronxd, line 20: `CustomPoliciesIntegration` order's is already 1. Can we assign a unique order?

checkov/common/bridgecrew/integration_features/features/attribute_resource_types_integration.py
```diff
@@ -0,0 +1,108 @@
...
+        if not attr:
+            return None
+        if '.' in attr:
+            attr = attr[0:attr.index('.')]
+
+        resource_types = self.attribute_resources.get(attr, None)
+        if not resource_types:
+            return None
+
+        return resource_types.get(provider or '__all__')
```
tronxd, line 69:
```suggestion
        return resource_types.get(provider or ALL_TYPES)
```

checkov/common/bridgecrew/integration_features/features/attribute_resource_types_integration.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _build_attribute_resource_map(self, resource_definitions: _ResourceDefinitions) -> None:
```
tronxd, line 71: Can we add a pydoc to the function?

checkov/common/checks_infra/checks_parser.py
```diff
@@ -133,18 +134,33 @@
...
-        check.id = raw_check.get("metadata", {}).get("id", "")
-        check.name = raw_check.get("metadata", {}).get("name", "")
-        check.category = raw_check.get("metadata", {}).get("category", "")
-        check.frameworks = raw_check.get("metadata", {}).get("frameworks", [])
-        check.guideline = raw_check.get("metadata", {}).get("guideline")
+
+        metadata = raw_check.get("metadata", {})
+
+        # the first approach comes from the custom policy integration
+        provider = metadata.get("scope", {}).get("provider")\
```
tronxd, line 141:
```suggestion
        provider = metadata.get("scope", {}).get("provider")
```

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -85,7 +85,11 @@ def get_attribute_matches(self, vertex: Dict[str, Any]) -> List[str]:
         if self.is_jsonpath_check:
             parsed_attr = self.parsed_attributes.get(self.attribute)
             if parsed_attr is None:
-                parsed_attr = parse(self.attribute)
+                try:
+                    parsed_attr = parse(self.attribute)
+                except Exception as e:
+                    print(e)
```
tronxd, line 91:
```suggestion
                    logging.error(e)
```

tronxd (approved):
LGTM, had some comments

mikeurbanski1:
@gruebel @nimrodkor @rotemavni added you to sanity check whether this can be safely merged. I ran some tests against prod (where the platform changes do not exist) and it seems to work fine in checkov. And I don't think anything here will break a platform scan in the checkov wrapper.

gruebel (approved):
nice, great work. The only question I had, you already answered in the description 🥇

PR #3645, opened 2022-10-12
fix(kubernetes): minor fix for getting entity type from template

PelegLi:

## Description

avoid exception when extracting entity type from a k8s template

## Checklist:

### DIFF
--- a/checkov/kubernetes/checks/resource/base_registry.py
+++ b/checkov/kubernetes/checks/resource/base_registry.py
@@ -6,7 +6,7 @@ class Registry(BaseCheckRegistry):
         super().__init__(report_type)
 
     def extract_entity_details(self, entity):
-        kind = entity["kind"]
+        kind = entity.get("kind")
         conf = entity
         return kind, conf
 
--- a/checkov/kubernetes/checks/resource/base_registry.py
+++ b/checkov/kubernetes/checks/resource/base_registry.py
@@ -6,7 +6,7 @@ class Registry(BaseCheckRegistry):
         super().__init__(report_type)
 
     def extract_entity_details(self, entity):
-        kind = entity.get("kind")
+        kind = entity.get("kind") or ""
         conf = entity
         return kind, conf
 


checkov/kubernetes/checks/resource/base_registry.py
```diff
@@ -6,7 +6,7 @@ def __init__(self, report_type: str):
         super().__init__(report_type)
 
     def extract_entity_details(self, entity):
-        kind = entity["kind"]
+        kind = entity.get("kind")
```
gruebel, line 9: is it ok, if it is `None`? not that it results in errors in other areas downstream 😄

tronxd (commented):
If `kind` is None then we would not be able to run the K8 check properly since we rely on a non-null value for it. In what use case does it happen?

checkov/kubernetes/checks/resource/base_registry.py
PelegLi: The only place this method is called from is [here](https://github.com/bridgecrewio/checkov/blob/master/checkov/kubernetes/checks/resource/base_registry.py#L14), and it means that it would not be able to get checks by entity type and that the results for this unknown entity would be empty

PelegLi:
> If `kind` is None then we would not be able to run the K8 check properly since we rely on a non-null value for it. In what use case does it happen?

Maybe a malformed template. If we cannot infer the entity type then we can't get the relevant checks

tronxd:
@PelegLi Please add a UT that replicates the issue and ensure the runner is skipping it without crashing

tronxd (approved):

checkov/kubernetes/checks/resource/base_registry.py
gruebel: ok, gotcha. This is more of a type hint concern, but leave it as it is, I will see how it will be, when I add type hints to `kubernetes`, because `get_checks()` actually expects a `str`, which it isn't, but it also doesn't fail, which makes sense from the code.
```suggestion
        kind = entity.get("kind") or ""
```
maybe this would be better for downstream handling

gruebel (approved):
☕

PR #3648, opened 2022-10-12
fix(terraform): local_authentication_disabled - cosmodb check to look at SQL Api only CKV_AZURE_140

JamesWoolfenden:
The current check is invalid, only the SQL api of cosmosdb can have this property set, if you try the results on any other type:

`Error: creating Database Account: (Name "pike-cosmos-db" / Resource Group "pike"): creating/updating CosmosDB Account "pike-cosmos-db" (Resource Group "pike"): documentdb.DatabaseAccountsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="DisableLocalAuth is only allowed to be configured for SQL API account.\r\nActivityId: 18c78879-d6e5-465f-90b0-756d721115fa, Microsoft.Azure.Documents.Common/2.14.0"`

Now we check the kind of the db and ignore if not **GlobalDocumentDB**

### DIFF
--- a/checkov/terraform/checks/resource/azure/CosmosDBLocalAuthDisabled.py
+++ b/checkov/terraform/checks/resource/azure/CosmosDBLocalAuthDisabled.py
@@ -1,10 +1,8 @@
-from typing import Any
-
-from checkov.common.models.enums import CheckCategories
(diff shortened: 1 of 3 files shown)


nimrodkor (approved):

gruebel (approved):
🙂

PR #3706, opened 2022-10-20
fix(gitlab): Modify gitlab ci resource id

Eliran-Turgeman:

[//]: # "
    # PR Title
    ex.
"

## Description

* Fixed SuspectCurlInScript check not having lines and code block via definitions_raw object
* Changed resource id to better reflect the resource, for example:
`*.rules.*.rules.CKV_GITLABCI_2[7:9]` -> `planOnlySubset`
For the following source code:
```
planOnlySubset:
  script: echo "This job creates double pipelines!"
  rules:
    - changes:
        - $DOCKERFILES_DIR/*
    - if: $CI_PIPELINE_SOURCE == "push"
    - if: $CI_PIPELINE_SOURCE == "merge_request_event"

...
```
* Added matching resource ids for images in gitlab ci

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/common/checks/base_check_registry.py
+++ b/checkov/common/checks/base_check_registry.py
@@ -197,6 +197,3 @@ class BaseCheckRegistry:
                                 self.logger.error(f"Cannot load external check '{check_name}' from {check_full_path}", exc_info=True)
                             finally:
                                 BaseCheckRegistry.__loading_external_checks = False
-
-    def set_definitions_raw(self, definitions_raw: list[tuple[int, str]]) -> None:
-        self.definitions_raw = definitions_raw
--- a/checkov/common/checks/base_check_registry.py
+++ b/checkov/common/checks/base_check_registry.py
@@ -34,6 +34,7 @@ class BaseCheckRegistry:
         self.wildcard_checks: Dict[str, List[BaseCheck]] = defaultdict(list)
         self.check_id_allowlist: Optional[List[str]] = None
         self.report_type = report_type
+        self.definitions_raw: list[tuple[int, str]] | None = None
 
     def register(self, check: BaseCheck) -> None:
         # IMPLEMENTATION NOTE: Checks are registered when the script is loaded
@@ -196,3 +197,6 @@ class BaseCheckRegistry:
                                 self.logger.error(f"Cannot load external check '{check_name}' from {check_full_path}", exc_info=True)
                             finally:
                                 BaseCheckRegistry.__loading_external_checks = False
+
+    def set_definitions_raw(self, definitions_raw: list[tuple[int, str]]) -> None:
+        self.definitions_raw = definitions_raw
--- a/checkov/common/runners/object_runner.py
+++ b/checkov/common/runners/object_runner.py
@@ -100,6 +100,9 @@ class Runner(BaseRunner[None]):  # if a graph is added, Any needs to replaced
         for file_path in self.definitions.keys():
             self.pbar.set_additional_data({'Current File Scanned': os.path.relpath(file_path, root_folder)})
             skipped_checks = collect_suppressions_for_context(self.definitions_raw[file_path])
+
+            if registry.report_type == CheckType.GITLAB_CI:
+                registry.set_definitions_raw(self.definitions_raw[file_path])
             results = registry.scan(file_path, self.definitions[file_path], skipped_checks, runner_filter)  # type:ignore[arg-type] # this is overridden in the subclass
             for key, result in results.items():
                 result_config = result["results_configuration"]
--- a/checkov/common/runners/object_runner.py
+++ b/checkov/common/runners/object_runner.py
@@ -102,7 +102,7 @@ class Runner(BaseRunner[None]):  # if a graph is added, Any needs to replaced
             skipped_checks = collect_suppressions_for_context(self.definitions_raw[file_path])
 
             if registry.report_type == CheckType.GITLAB_CI:
-                registry.set_definitions_raw(self.definitions_raw[file_path])
+                registry.definitions_raw = self.definitions_raw[file_path]
             results = registry.scan(file_path, self.definitions[file_path], skipped_checks, runner_filter)  # type:ignore[arg-type] # this is overridden in the subclass
             for key, result in results.items():
                 result_config = result["results_configuration"]
--- a/checkov/gitlab_ci/common/resource_id_utils.py
+++ b/checkov/gitlab_ci/common/resource_id_utils.py
@@ -1,0 +1,20 @@
+from __future__ import annotations
+from typing import Any
+
+from checkov.common.util.consts import START_LINE, END_LINE
+
+
+def generate_resource_key_recursive(conf: dict[str, Any] | list[str] | str, key: str, start_line: int,
+                                    end_line: int) -> str:
+    if not isinstance(conf, dict):
+        return key
+
+    for k, value in conf.items():
+        if k == START_LINE or k == END_LINE or isinstance(value, str):
+            continue
+        elif isinstance(value, list):
+            return f'{key}.{k}' if key else k
+        elif value[START_LINE] <= start_line <= end_line <= value[END_LINE]:
+            return generate_resource_key_recursive(value, f'{key}.{k}' if key else k, start_line, end_line)
+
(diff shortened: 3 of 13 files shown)


checkov/common/checks/base_check_registry.py
```diff
@@ -196,3 +197,6 @@ def load_external_checks(self, directory: str) -> None:
                                 self.logger.error(f"Cannot load external check '{check_name}' from {check_full_path}", exc_info=True)
                             finally:
                                 BaseCheckRegistry.__loading_external_checks = False
+
+    def set_definitions_raw(self, definitions_raw: list[tuple[int, str]]) -> None:
```
gruebel, line 201: no need to create a method for this, just directly set the attribute, when needed

checkov/common/runners/object_runner.py
```diff
@@ -100,6 +100,9 @@ def run(
         for file_path in self.definitions.keys():
             self.pbar.set_additional_data({'Current File Scanned': os.path.relpath(file_path, root_folder)})
             skipped_checks = collect_suppressions_for_context(self.definitions_raw[file_path])
+
+            if registry.report_type == CheckType.GITLAB_CI:
+                registry.set_definitions_raw(self.definitions_raw[file_path])
```
gruebel, line 105: any reason, why limit it to just GitLab CI?

checkov/gitlab_ci/common/resource_id_utils.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+from typing import Any
+
+from checkov.common.util.consts import START_LINE, END_LINE
```
gruebel, line 4: very nice 😄

checkov/gitlab_ci/common/resource_id_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def generate_resource_key_recursive(conf: dict[str, Any] | list[str] | str, key: str, start_line: int,
+                                    end_line: int) -> str:
+    if not isinstance(conf, dict):
+        return key
+
+    for k, value in conf.items():
+        if isinstance(value, dict) and value[START_LINE] <= start_line <= end_line <= value[END_LINE]:
+            return generate_resource_key_recursive(value, f'{key}.{k}' if key else k, start_line, end_line)
```
gruebel, line 14: can you move the ternary operator `f'{key}.{k}' if key else k` to a separate variable before invoking it, feels a bit to tight here 😄

checkov/gitlab_ci/image_referencer/provider.py
```diff
@@ -43,6 +44,8 @@ def extract_images_from_workflow(self) -> list[Image]:
                                         name=image_name,
                                         start_line=start_line,
                                         end_line=end_line,
+                                        related_resource_id=generate_resource_key_recursive(self.workflow_config, '',
```
gruebel, line 47: can use keyword arguments here, usually with 3+ arguments it is hard to understand which argument is which 😵

checkov/gitlab_ci/image_referencer/provider.py
```diff
@@ -52,6 +55,8 @@ def extract_images_from_workflow(self) -> list[Image]:
                                 name=image_name,
                                 start_line=start_line,
                                 end_line=end_line,
+                                related_resource_id=generate_resource_key_recursive(self.workflow_config, '',
```
gruebel, line 58: same here

checkov/gitlab_ci/runner.py
```diff
@@ -45,6 +47,14 @@ def is_workflow_file(self, file_path: str) -> bool:
     def included_paths(self) -> Iterable[str]:
         return (".gitlab-ci.yml", ".gitlab-ci.yaml")
 
+    def get_resource(self, file_path: str, key: str, supported_entities: Iterable[str], definitions: dict[str, Any] | None = None) -> str:
+        start_line, end_line = Runner.get_start_and_end_lines(key)
+        file_config: dict[str, Any] | None = force_dict(self.definitions[file_path])
```
gruebel, line 52: is the type hint really needed here? can you try it out without it.

checkov/gitlab_ci/runner.py
```diff
@@ -45,6 +47,14 @@ def is_workflow_file(self, file_path: str) -> bool:
...
+        if not file_config:
+            return key
+        resource_id: str = generate_resource_key_recursive(file_config, '', start_line, end_line)
```
gruebel, line 55: also here keyword arguments, please 🥺

checkov/gitlab_ci/runner.py
```diff
@@ -91,3 +101,15 @@ def extract_images(
...
 
         return images
+
+    @staticmethod
+    def get_start_and_end_lines(key: str) -> list[int]:
+        check_name = key.split('.')[-1]
+        try:
+            start_end_line_bracket_index = check_name.index('[')
+            closing_bracket_index = check_name.index(']')
+            if closing_bracket_index - start_end_line_bracket_index == 1:
```
gruebel, line 111:
```suggestion
            if "[" not in check_name or "[]" in check_name:
```
how about this? then you also don't need the `try/except` block

checkov/yaml_doc/base_registry.py
```diff
@@ -44,10 +44,13 @@ def _scan_yaml_array(
                         )
             if isinstance(analyzed_entities, list):
                 for item in analyzed_entities:
+                    item_dict = item
+                    if isinstance(item, str):
+                        item_dict = self.set_lines_for_item(item)
```
gruebel, line 49: did you do this, because of typing issues? instead of just
```suggestion
                    if isinstance(item, str):
                        item = self.set_lines_for_item(item)
```

checkov/yaml_doc/base_registry.py
```diff
@@ -253,3 +256,29 @@ def get_result_key(self, check: BaseCheck,
...
         pass
+
+    def set_lines_for_item(self, item: str) -> dict[int | str, str | int] | str:
+        if not self.definitions_raw:
+            return item
+
+        item_lines = item.rstrip().split("\n")
+        first_line, last_line = item_lines[0], item_lines[-1]
+
+        is_single_line = True if len(item_lines) == 1 else False
```
gruebel, line 267: why not a special handling for single lines, instead of trying to integrate in the code for multi lines? something like
```python
if len(item_lines) == 1:
    return {...}
```

gruebel (commented):
nice work, left a couple of comments

checkov/common/runners/object_runner.py
Eliran-Turgeman: I wasn't sure if that's needed for all frameworks, I added it because some gitlab ci check names were missing the resource lines and code blocks and I needed the definitions to resolve them.
I think we should gradually apply it to more frameworks as we find issues with their resource ids, WDYT?

Eliran-Turgeman:
@gruebel Addressed all your comments, thanks! 😄

checkov/yaml_doc/base_registry.py
```diff
@@ -253,3 +255,31 @@ def get_result_key(self, check: BaseCheck,
...
+        item_dict: dict[int | str, str | int] = {
+            idx: line for idx, line in enumerate(item_lines)
+        }
+
+        if len(item_lines) == 1:
+            item_line = item_lines[0]
+            for idx, line in self.definitions_raw:
+                if item_line in line:
+                    item_dict[STARTLINE_MARK] = idx
+                    item_dict[ENDLINE_MARK] = idx
```
gruebel, line 273: shouldn't here be a `return item_dict` on the same indent as the `for` loop?

gruebel (approved):
last one 😄

checkov/common/runners/object_runner.py
gruebel: gotcha, gotcha. let's leave it for now as it is. if we need it for something else, then lets add it for all checks

checkov/yaml_doc/base_registry.py
Eliran-Turgeman: Yeah, good catch 😄

lirshindalman (approved):

PR #3707, opened 2022-10-20
feat(circleci): CircleCI Image Reference using Mixin class

### LINKED ISSUE #3687: Adjust CircleCI Pipelines runner to use the new Image Referencer implementation
**Describe the issue**
Currently the CircleCI Pipelines runner uses the old Image Referencer implementation and it should be migrated to use the `Mixin` flavour.

marynaKK:

[//]: # "
    # PR Title
    ex.
"

## Description

Changes the runner of CircleCI to inherit ImageReferencerMixin.
Also added some missing scanning to the provider to be more comprehensive. 

Fixes  #3687

Please notice I didn't change the testing- regular checks were not touched.
also, resource_id for CircleCI can be improved, but I didn't include it in this PR.

## Checklist:


checkov/circleci_pipelines/image_referencer/provider.py
```diff
@@ -0,0 +1,38 @@
...
+
+class CircleCIProvider:
+    __slots__ = ("workflow_config", "file_path")
+
+    def __init__(self, workflow_config: dict[str, Any], file_path: str):
+        self.file_path = file_path
+        self.workflow_config = workflow_config
+
+    def extract_images_from_workflow(self) -> list[Image]:
+        images: list[Image] = []
```
marynaKK, line 16: maybe should be a set? Is a List in other implementations

gruebel: code-wise it should be possible. at the moment there was no need to do it, because getting a duplicated `Image` instance is rather rare and shouldn't happen. But what we don't do and this would be an improvement to only send unique image names to our platform. Are you interested to do this change in a separate PR?

checkov/circleci_pipelines/runner.py
```diff
@@ -42,92 +46,47 @@ def is_workflow_file(self, file_path: str) -> bool:
...
+        runner_filter = runner_filter or RunnerFilter()
+        report = super().run(root_folder=root_folder, external_checks_dir=external_checks_dir,
+                             files=files, runner_filter=runner_filter, collect_skip_comments=collect_skip_comments)
+        if runner_filter.run_image_referencer:
+            if files:
+                # 'root_folder' shouldn't be empty to remove the whole path later and only leave the shortened form
+                root_folder = os.path.split(os.path.commonprefix(files))[0]
+
+            image_report = self.check_container_image_references(
+                graph_connector=None,
```
gruebel, line 66:
```suggestion
```
no need to set it, because it is `None` by default

tests/circleci_pipelines/resources/.circleci/config.yml
```diff
@@ -84,6 +106,37 @@ jobs:
...
+    executor:
+      name: win/server-2019
+    steps:
+      - some-step
+
+  test-executor-machine-server-2019-cuda:
+    executor:
+      name: win/windows-server-2019-cuda
+    steps:
+      - some-step
```
gruebel, line 138: same here, all are CircleCI internal machine names, nothing we can actually scan

gruebel (commented):
good first try 💪 but I think you went the wrong way with the images, which sadly is not what we actually wanted to extract 😢

checkov/circleci_pipelines/runner.py
```diff
@@ -1,21 +1,24 @@
...
 
 if TYPE_CHECKING:
     from checkov.common.checks.base_check_registry import BaseCheckRegistry
+    from networkx import DiGraph
 
 WORKFLOW_DIRECTORY = "circleci"
 
 
-class Runner(YamlRunner, ImageReferencer):
+class Runner(ImageReferencerMixin["dict[str, dict[str, Any] | list[dict[str, Any]]]"], YamlRunner):
```
Eliran-Turgeman, line 21: What is the significance of the type expression in `ImageReferencerMixin`?

Eliran-Turgeman (commented):
Great job! 🚀 

Added a single comment, and also can you add more tests? (for example, testing the extraction of images from the config.yml)

checkov/circleci_pipelines/runner.py
gruebel: the extra type expressions allows the `definitions` type hint to be adjustable, I needed to do it for the `dockerfile` runner, because the definitions type is different and I didn't want to add an other case to the already long type hint.

gruebel (approved):
nice work 🐶

Eliran-Turgeman (approved):
🔥

PR #3716, opened 2022-10-23
fix(gitlab): vcs gitlab groups retrieval

maxamel:
## Description

This PR fixes an issue in vcs gitlab groups retrieval. One part is in schema validation, and the second in the logic of failing checks.

## Checklist:

### DIFF
--- a/checkov/common/vcs/vcs_schema.py
+++ b/checkov/common/vcs/vcs_schema.py
@@ -15,6 +15,6 @@ class VCSSchema():
         try:
             validate(instance=data, schema=self.schema)
         except jsonschema.exceptions.ValidationError as e:
-            logging.debug(f"validation error {e}", exc_info=True)
+            logging.debug("validation error", exc_info=True)
             return False
         return True
--- a/checkov/common/vcs/vcs_schema.py
+++ b/checkov/common/vcs/vcs_schema.py
@@ -14,7 +14,7 @@ class VCSSchema():
     def validate(self, data: dict[str, Any] | list[dict[str, Any]]) -> bool:
         try:
             validate(instance=data, schema=self.schema)
-        except jsonschema.exceptions.ValidationError:
-            logging.debug("validation error", exc_info=True)
+        except jsonschema.exceptions.ValidationError as e:
+            logging.debug(f"validation error {e}", exc_info=True)
             return False
         return True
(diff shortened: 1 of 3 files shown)


checkov/common/vcs/vcs_schema.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, schema: dict[str, Any]) -> None:
     def validate(self, data: dict[str, Any] | list[dict[str, Any]]) -> bool:
         try:
             validate(instance=data, schema=self.schema)
-        except jsonschema.exceptions.ValidationError:
-            logging.debug("validation error", exc_info=True)
+        except jsonschema.exceptions.ValidationError as e:
+            logging.debug(f"validation error {e}", exc_info=True)
```
gruebel, line 18: the `exc_info=True` argument gives the whole stack trace, is something missing or why do also add the error message again to the log message?

nimrodkor (approved):

checkov/gitlab/checks/two_factor_authentication.py
```diff
@@ -24,9 +24,9 @@ def __init__(self) -> None:
     def scan_entity_conf(self, conf: list[dict[str, Any]], entity_type: str) -> tuple[CheckResult, list[dict[str, Any]]] | None:  # type:ignore[override]
         if schema.validate(conf):
             for group in conf:
-                if group.get("require_two_factor_authentication", False) is True:
-                    return CheckResult.PASSED, conf
-            return CheckResult.FAILED, conf
+                if group.get("require_two_factor_authentication") is False:
```
gruebel, line 27: is `require_two_factor_authentication` always set and can't be `None`? or is the default `True`?

marynaKK, line 27: I think that maybe `require_two_factor_authentication` is not a required field (although the schema down here states it is required - from the docs: https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/groups.md) 
Hence in case there is no `require_two_factor_authentication` the schema will be filtered and will not get here

maxamel: Looks like the docs state that `require_two_factor_authentication` is not mandatory on PUT/POST operations, since it defaults to False. But on retrieval it is always returned.

gruebel (approved):
☕

checkov/gitlab/checks/two_factor_authentication.py
marynaKK: You are correct! missed the difference :)

maxamel: Yes it is set. I think it defaults to False, since creating a group without it yields `require_two_factor_authentication=False`

PR #3794, opened 2022-11-04
feat(terraform): Add resources based checks to AWS resource exposure policy CKV_AWS_109

jjchavanne:

[//]: # "
    # PR Title
    ex.
"

## Description

- fixed path reference in the YAML contribution docs page 
- added graph resource based checks for CKV_AWS_109

Fixes # (issue)

## New/Edited policies (Delete if not relevant)

- CKV_AWS_109

### Description
- https://cloudsplaining.readthedocs.io/en/latest/glossary/resource-exposure/

### Fix
- remove the permissions or wildcard permissions if they are not necessary

## Checklist:


gruebel:
🤔 the check itself is a bit different compared to his sibling `CKV_AWS_109` because the original one also considers `constraints`. Any thoughts on this?

jjchavanne:
You are right, the existing data check only fails if resource is also wildcard/"*", meaning no constraints, however there is still risk even if scoped down to a single resource.  The first example given here is a good ref: https://cloudsplaining.readthedocs.io/en/latest/glossary/resource-exposure/

_- s3:PutBucketPolicy, s3:PutBucketAcl, and s3:PutObjectAcl grant permissions to modify the properties of S3 buckets or objects for new or existing objects in an S3 bucket, which could expose objects to rogue actors or to the internet._

Summary: There is risk even with these permissions on a single resource and therefore suggest to either remove the resource constraint on the data check or have 2 policies with different severities, one with higher severity where there are no resource constraints.   I think I'd prefer only one policy though.

gruebel:
the original check and yours are targeting 2 different things so I'm fine with 2 separate policies.

I somehow feel that the chosen name and the flagged permissions are a bit misleading
```
Ensure AWS IAM policy does not allow actions that permit modification of resource-based policies that can otherwise expose AWS resources to the public.
```
ex.
- `codestar:DeleteProject` this doesn't expose anything to the public, just deletes a project, which then not accessible anymore
- `iam:DeleteAccessKey` just removes the access for the user, no exposure
- `iam:DeleteUser` user gone, no possibility of resource exposure
- `ram:DeleteResourceShare` nothing bad, just whoever it was shared with has no access anymore

So I'm actually not sure, what we try to protect the user from? your example makes total sense, but as my example hopefully show it is not very clear why these permissions are in the list.

jjchavanne:
Hey guys, I'm going to close this PR, as I think this deserves a bigger discussion around way forward with cloudsplaining policies as well as some of the points that @gruebel made above. 
cc: @tsmithv11 @mikeurbanski1

PR #3795, opened 2022-11-04
platform(general): rationalize policy metadata error handling behavior

mikeurbanski1:

[//]: # "
    # PR Title
    ex.
"

## Description

This change causes checkov to fail if the user uses a platform API key, but the connection itself fails (e.g. SSL or auth error), and they don't use `--include-all-checkov-checks`.

Today, the behavior is that the policy metadata download will fail with an error, but then the run will proceed. However, Checkov thinks all the policies are local policies, and does not run any of them (by design when using an API key). The result is blank output (minus the connection error message in the logs). This causes issues primarily in the IDEs, where there is no error message popup from the failed run, so it just looks like nothing happened.

This change will cause the run to actually fail if no policies are going to get evaluated.

It also handles some general conflicting logic with the use of `--skip-download`, which causes similar behavior (but without an error).

## Checklist:

- [ ] I have added tests that prove my feature, policy, or fix is effective and works  *(I tested all the cases locally)*

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -136,6 +136,25 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
     if config.output is None:
         config.output = ['cli']
 
+    if config.bc_api_key and not config.include_all_checkov_policies:
+        if config.skip_download and not config.external_checks_dir:
+            print('You are using an API key along with --skip-download but not --include-all-checkov-policies or --external-checks-dir. '
+                  'With these arguments, Checkov cannot fetch metadata to determine what is a local Checkov-only '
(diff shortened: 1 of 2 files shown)


checkov/main.py
```diff
@@ -256,7 +275,19 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
     if config.skip_download or BC_SKIP_MAPPING.upper() == "TRUE":
         bc_integration.skip_download = True
 
-    bc_integration.get_platform_run_config()
+    try:
+        bc_integration.get_platform_run_config()
+    except Exception as e:
```
gruebel, line 280:
```suggestion
    except Exception:
```

gruebel (approved):
looks good 🥇

marynaKK (approved):

PR #3802, opened 2022-11-06
platform(sca): add new sca package scan

ayajbara:
new SCA package scan will contain a new flow of sca package scanning that will work when environment variable RUN_NEW_SCA_PACKAGE_SCAN is true


checkov/main.py
```diff
@@ -75,6 +76,8 @@
 logger = logging.getLogger(__name__)
 checkov_runners = [value for attr, value in CheckType.__dict__.items() if not attr.startswith("__")]
 
+RUN_NEW_SCA_PACKAGE_SCAN = os.getenv('RUN_NEW_SCA_PACKAGE_SCAN', '').lower() == 'true'
```
AdamDev, line 79: the name "New" is not good, today it's new, tomorrow there is a newer.
maybe "RUN_SCA_PACKAGE_SCAN_V2"

checkov/sca_package_2/scanner.py
```diff
@@ -0,0 +1,33 @@
...
+
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Any
+
+from checkov.common.bridgecrew.platform_integration import bc_integration
+
+from checkov.common.util.tqdm_utils import ProgressBar
+
+SLEEP_DURATION = 2
```
AdamDev, line 11: Are those in use?

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -343,6 +347,8 @@ def persist_repository(
                 for file_path in f_names:
                     _, file_extension = os.path.splitext(file_path)
                     if file_extension in SUPPORTED_FILE_EXTENSIONS or file_path in SUPPORTED_FILES:
+                        if RUN_NEW_SCA_PACKAGE_SCAN and file_extension in SUPPORTED_PACKAGE_FILES:
```
AdamDev, line 350: this function is to persist the files.
If we don't want to persist the files in case the new sca package scan.
I think we should no call this function at all in that case.

checkov/main.py
AdamDev, line 79: the name "New" is not good, today it's new, tomorrow there is a newer.
maybe "RUN_SCA_PACKAGE_SCAN_V2"

checkov/sca_package_2/scanner.py
AdamDev, line 11: Are those in use?

checkov/sca_package_2/scanner.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, pbar: ProgressBar | None = None, root_folder: str | Path | None = None) -> None:
+        self._base_url = bc_integration.api_url
+        if pbar:
+            self.pbar = pbar
+        else:
+            self.pbar = ProgressBar('')
+            self.pbar.turn_off_progress_bar()
+        self.root_folder = root_folder
+
+    def scan(self) -> Sequence[dict[str, Any]]:
```
AdamDev, line 25: Can you tell why the following 3 methods are empty?

checkov/sca_package_2/runner.py
```diff
@@ -0,0 +1,151 @@
...
+        for file in files or []:
+            file_path = Path(file)
+            if not file_path.exists():
+                logging.warning(f"File {file_path} doesn't exist")
+                continue
+            package_files_to_persist.append(FileToPersist(file, os.path.relpath(file, root_folder)))
+
+        logging.info(f"{len(package_files_to_persist)} sca package files found.")
+        bc_integration.persist_files(package_files_to_persist)
+        return bool(package_files_to_persist)
```
AdamDev, line 151: I think we should try catch this code:
`bc_integration.persist_files(package_files_to_persist)`

and return the results of it.

at the current implementation, it checks if there are files to persist and not if it succeed.

checkov/main.py
ayajbara: it will be removed after the implementation is finished, but good idea :)

checkov/sca_package_2/scanner.py
ayajbara: will use them in the next PRs

checkov/common/bridgecrew/platform_integration.py
ayajbara: we do not persisting the package files here because they already uploaded in the scanner

checkov/sca_package_2/scanner.py
ayajbara: will be implemented next PRs.

checkov/main.py
```diff
@@ -75,6 +76,8 @@
 logger = logging.getLogger(__name__)
 checkov_runners = [value for attr, value in CheckType.__dict__.items() if not attr.startswith("__")]
 
+RUN_SCA_PACKAGE_SCAN_V2 = os.getenv('RUN_SCA_PACKAGE_SCAN_V2', '').lower() == 'true'
```
gruebel, line 79:
```suggestion
RUN_SCA_PACKAGE_SCAN_V2 = os.getenv('CHECKOV_RUN_SCA_PACKAGE_SCAN_V2', '').lower() == 'true'
```
we usually prefix internal env vars with `CHECKOV` which the user should only change, if he knows what he is doing 😄

gruebel: and move the const to here `checkov/common/util/consts.py` then import it from there, so you don't need to retrieve it in different areas

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -333,6 +335,8 @@ def persist_repository(
                 f_name = os.path.basename(f)
                 _, file_extension = os.path.splitext(f)
                 if file_extension in SUPPORTED_FILE_EXTENSIONS or f_name in SUPPORTED_FILES:
+                    if RUN_SCA_PACKAGE_SCAN_V2 and file_extension in SUPPORTED_PACKAGE_FILES:
+                        continue
```
gruebel, line 339: how about moving it directly after `_, file_extension = os.path.splitext(f)`

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -343,6 +347,8 @@ def persist_repository(
                 for file_path in f_names:
                     _, file_extension = os.path.splitext(file_path)
                     if file_extension in SUPPORTED_FILE_EXTENSIONS or file_path in SUPPORTED_FILES:
+                        if RUN_SCA_PACKAGE_SCAN_V2 and file_extension in SUPPORTED_PACKAGE_FILES:
+                            continue
```
gruebel, line 351: same here

checkov/sca_package_2/runner.py
```diff
@@ -0,0 +1,151 @@
...
+                    file_path_str = str(file_path)
+                    package_files_to_persist.append(
+                        FileToPersist(file_path_str, os.path.relpath(file_path_str, root_folder)))
+
+        for file in files or []:
+            file_path = Path(file)
+            if not file_path.exists():
+                logging.warning(f"File {file_path} doesn't exist")
+                continue
+            package_files_to_persist.append(FileToPersist(file, os.path.relpath(file, root_folder)))
```
gruebel, line 147: shouldn't we also check here `if file_path.name in SUPPORTED_PACKAGE_FILES` otherwise we upload non package files.

checkov/sca_package_2/runner.py
```diff
@@ -0,0 +1,151 @@
...
+
+        logging.info(f"{len(package_files_to_persist)} sca package files found.")
+        bc_integration.persist_files(package_files_to_persist)
+        return bool(package_files_to_persist)
```
gruebel, line 151:
```suggestion
        return package_files_to_persist
```
how about this and then change the type hint to
```python
->  list[FileToPersist]:
```
your if condition where you call the function will still work, because it will check if it empty or not.

gruebel (commented):
great work 🍰 , added a few comments

AdamDev (approved):

PR #3812, opened 2022-11-07
platform(sca): Run package scan using API

ayajbara:

[//]: # "
    # PR Title
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/sca_package_2/scanner.py
```diff
@@ -24,10 +28,35 @@ def __init__(self, pbar: ProgressBar | None = None, root_folder: str | Path | No
...
+        return []
 
     def run_scan(self) -> dict[str, Any]:
-        pass
+        logging.info("Start to scan package files.")
+
+        request_body = {
+            "branch": "",
+            "commit": "",
+            "path": os.path.join(bc_integration.repo_path, '') if bc_integration.repo_path else "",
```
gruebel, line 40: why `os.path.join(bc_integration.repo_path, '')` and not just `bc_integration.repo_path`?

checkov/sca_package_2/scanner.py
```diff
@@ -24,10 +28,35 @@ def __init__(self, pbar: ProgressBar | None = None, root_folder: str | Path | No
...
+            "repoId": bc_integration.repo_id,
+            "id": bc_integration.timestamp,
+            "repositoryId": ""
+        }
+
+        response = request_wrapper(
+            "POST", f"{self._base_url}/api/v1/vulnerabilities/cli/scan",
```
gruebel, line 47: can you move `f"{self._base_url}/api/v1/vulnerabilities/cli/scan"` to the `__init__` then it can be used for the polling to0.

checkov/sca_package_2/scanner.py
```diff
@@ -24,10 +28,35 @@ def __init__(self, pbar: ProgressBar | None = None, root_folder: str | Path | No
...
+            should_call_raise_for_status=True
+        )
+
+        response_json = response.json()
+
+        if not response_json["startedSuccessfully"]:
+            logging.info("Failed to run package scanning.")
+            return dict()
+
+        return self.poll_scan_result()
```
gruebel, line 59: I think the polling should be under `def scan()`, just return the id here so, it can be used by `poll_scan_result()`

gruebel (commented):
nice, looking forward when it is completely implemented 🚀

checkov/sca_package_2/scanner.py
ayajbara: because `bc_integration.repo_path` is without a slash in the end and if the slash not exists `os.path.join(bc_integration.repo_path, '')` will add it

checkov/sca_package_2/scanner.py
ayajbara: sure

AdamDev (approved):

checkov/sca_package_2/scanner.py
gruebel: ok, got it

PR #3823, opened 2022-11-08
feat(general): add reporting contributor metrics

maxamel:
This PR adds capability to report contributor metrics in case we are in prisma context. They are reported by running git log in case of CI integration and sending them to external API.

## Checklist:


checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
     return None
 
 
+def report_contributor_metrics(repository: str) -> None:  # ignore: type
+
+    def _parse_gitlog() -> dict[str, Any] | None:
+        process = subprocess.Popen(['git', 'shortlog', '-ne', '--all', '--since', '"90 days ago"', '--pretty="commit-%ct"', '--reverse'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)  # nosec
+        out, err = process.communicate()
+        if err:
+            logger.info('Failed to collect contributor metrics')
```
gruebel, line 434: can you also log the `err` itself here

checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
...
+            return None
+        # split per contributor
+        list_of_contributors = out.decode('utf-8').split('\n\n')
```
gruebel, line 437: it would nice to move this into a normal function, then it easy to mock it and then write some UTs with expected output and what should come out at the end.

checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
...
+                "contributors": list(map(lambda contributor: _process_contributor(contributor),
+                                         list(filter(lambda x: x, list_of_contributors))
+                                         ))
+                }
+
+    def _process_contributor(contributor: str) -> str:
+        splittedList = contributor.split('\n')
+        user = splittedList[0]
+        commit = splittedList[1]
+        return user[0:user.find('(')] + commit[commit.find('-') + 1:len(commit)].strip('\"')
```
Eliran-Turgeman, line 448: The `0` and `len(commit)` can be omitted

```suggestion
        return user[:user.find('(')] + commit[commit.find('-') + 1:].strip('\"')
```

checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
...
+
+    request_body = _parse_gitlog()
+    if request_body:
+        response = request_wrapper(
+            "POST", f"{bc_integration.api_url}/api/v1/contributors/report",
+            headers=bc_integration.get_default_headers("POST"), data=json.dumps(request_body)
+        )
+        if response.status_code < 300:
```
Eliran-Turgeman, line 456: Just a suggestion, you can use `response.ok` instead (it checks if the status code is below 400)

checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
     return None
 
 
+def report_contributor_metrics(repository: str) -> None:  # ignore: type
+
+    def _parse_gitlog() -> dict[str, Any] | None:
+        process = subprocess.Popen(['git', 'shortlog', '-ne', '--all', '--since', '"90 days ago"', '--pretty="commit-%ct"', '--reverse'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)  # nosec
```
nimrodkor, line 431: @maxamel Can we use gitpython instead of subprocess?
https://pypi.org/project/GitPython/

checkov/main.py
```diff
@@ -249,6 +251,14 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
                                                         source_version=source_version,
                                                         repo_branch=config.branch,
                                                         prisma_api_url=config.prisma_api_url)
+
+            if source.name in [BCSourceType.GITHUB_ACTIONS, BCSourceType.JENKINS, BCSourceType.CIRCLECI, BCSourceType.CODEBUILD] \
```
nimrodkor, line 255: @maxamel Maybe we should add to the `BCSource` enum another field, `is_ci`, and have it by default `False` but for these 4 have it as `True`?

nimrodkor (commented):
Very nice @maxamel

checkov/main.py
```diff
@@ -415,6 +425,40 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
     return None
 
 
+def report_contributor_metrics(repository: str) -> None:  # ignore: type
```
Eliran-Turgeman, line 428: I think it might be worth extracting it to a different module, IMO it's not too related to the main script

checkov/main.py
gruebel: I also wanted to suggest it, but it looks like that `shortlog` doesn't work very well with `gitpython` https://github.com/gitpython-developers/GitPython/discussions/1434

checkov/main.py
maxamel: I actually used it in the beginning but then I noticed it marks anything below 400 as ok.
I would expect only 200/202/204 to be an OK response.

checkov/main.py
maxamel: Done

checkov/main.py
maxamel: Done

maxamel:
Extracted reporting logic to separate file, added UTs, extended BcSource to include collaborator reporting flag, and added env variable protection on this flow.

nimrodkor (approved):
I think this is more concise, but it is a style issue. Approved anyway @maxamel

checkov/common/bridgecrew/bc_source.py
```diff
@@ -24,16 +25,16 @@ class BCSourceType:
...
+    BCSourceType.VSCODE: SourceType(BCSourceType.VSCODE, False, False),
+    BCSourceType.JETBRAINS: SourceType(BCSourceType.JETBRAINS, False, False),
+    BCSourceType.CLI: SourceType(BCSourceType.CLI, True, False),
+    BCSourceType.KUBERNETES_WORKLOADS: SourceType(BCSourceType.KUBERNETES_WORKLOADS, True, False),
+    BCSourceType.GITHUB_ACTIONS: SourceType(BCSourceType.GITHUB_ACTIONS, True, True),
+    BCSourceType.DISABLED: SourceType(BCSourceType.VSCODE, False, False),
+    BCSourceType.CODEBUILD: SourceType(BCSourceType.CODEBUILD, True, True),
+    BCSourceType.JENKINS: SourceType(BCSourceType.JENKINS, True, True),
+    BCSourceType.CIRCLECI: SourceType(BCSourceType.CIRCLECI, True, True),
+    BCSourceType.ADMISSION_CONTROLLER: SourceType(BCSourceType.ADMISSION_CONTROLLER, False, False)
```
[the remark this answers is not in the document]
maxamel: Cool fixed

Eliran-Turgeman (approved):

PR #3830, opened 2022-11-08
feat(general): Get taggable resources from platform

mikeurbanski1:

[//]: # "
    # PR Title
    ex.
"

## Description

- Gets the list of all resource types and their attributes from the platform
- Also gets a list of attributes that we want to filter by resource type (this is just tags / labels for the various providers right now)
- Gets the provider from the updated platform policy metadata response
- Checkov will apply the logic to replace any policies using these attributes with a resource type of "all" with the resource types for the given attribute and provider (if available).

So, for example, if you have a platform policy with provider `aws` and a condition that checks `tags` with resource types `all`, then the check will only apply to taggable AWS Terraform resources. If you use the attribute `Tags.*.Key`, then it will only apply to CloudFormation.

If you have a Checkov policy that uses `all` with the attribute `tags`, then it will apply it to AWS, Azure, and Ali (because checkov policies do not have a provider in the spec).

This also removes the local list of resource types from checkov, which was pretty much totally dead code in reality, because platform custom policies didn't even have the provider in the API response.

Requires some platform changes that must be merged together.

## Checklist:


gruebel (approved):
🍻

tronxd (approved):

PR #3853, opened 2022-11-11
fix(terraform): Handled nested unrendered vars

mikeurbanski1:

[//]: # "
    # PR Title
    ex.
"

## Description

Adds logic to handle unrendered variables for custom policies that use nested attributes, like checking for a specific tag.

Previously, `tags.component equals ...` for `tags = local.unrendered_var` would fail because the value of that attribute is `None`, and this doesn't get picked up as an unrendered var. The fix is to check each attribute part (first `tags`, then `tags.component`) to see if it maps to an unrendered var.

Additionally, , a policy like `tags.component exists` would fail for `tags = local.unrendered_var`, because of the `is_value_attribute_check` condition. This PR modifies that behavior to only care about `is_value_attribute_check` when we are evaluating the literal attribute in the policy (`tags.component`, not just `tags`).

This PR also adds some explicit tests for simple and nested unrendered variables for conditions with and without `value`, since it seemed like those were missing. (They are just reflected in changed results of other tests for other things).

## Checklist:


rotemavni (approved):
Nice!

gruebel (approved):
noice 🍺

PR #3865, opened 2022-11-13
fix(kubernetes): handle empty spec object in k8s templates

PelegLi:

## Description

handle failing checks due to unexpected empty spec object.
no UT for this specific case wasn't added since a YAML with null value is not parsed in checkov (considered as a parsing error)

## Checklist:

### DIFF
--- a/checkov/kubernetes/checks/resource/base_spec_check.py
+++ b/checkov/kubernetes/checks/resource/base_spec_check.py
@@ -50,6 +50,6 @@ class BaseK8Check(BaseCheck):
     @staticmethod
     def get_inner_entry(conf: Dict[str, Any], entry_name: str) -> Dict[str, Any]:
         spec = {}
-        if conf.get("spec") and conf.get("spec").get("template"):
+        if conf.get("spec") and isinstance(conf["spec"], dict) and conf.get("spec").get("template"):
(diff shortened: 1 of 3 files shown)


achiar99 (approved):

checkov/kubernetes/checks/resource/base_spec_check.py
```diff
@@ -50,6 +50,6 @@ def wrapper(self, conf, entity_type=None):
     @staticmethod
     def get_inner_entry(conf: Dict[str, Any], entry_name: str) -> Dict[str, Any]:
         spec = {}
-        if conf.get("spec") and conf.get("spec").get("template"):
+        if conf.get("spec") and isinstance(conf["spec"], dict) and conf.get("spec").get("template"):
```
rotemavni, line 53:
```suggestion
        if conf.get("spec") and conf.get("spec", {}).get("template"):
```

rotemavni (commented):
@PelegLi Nice, can you please add a test?

checkov/kubernetes/checks/resource/base_spec_check.py
PelegLi: this doesn't solve the issue - it returns a None object and does not use the default empty dict

rotemavni (approved):

PR #3867, opened 2022-11-13
platform(sca): Fix bom report

AdamDev:

[//]: # "
    # PR Title
    ex.
"

## Description

On creating the BOM report, the delimiter was a comma character, it cause an issue because there are some licenses that contain comma characters.

Fixes # (issue)

Instead of encode and decode the licenses to string, just pass them as a list.

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -22,10 +22,11 @@ def is_raw_formatted(licenses: str) -> bool:
 
 
 def format_licenses_to_string(licenses_lst: list[str]) -> str:
-    if licenses_lst and len(licenses_lst) > 0:
+    if licenses_lst and len(licenses_lst) > 1:
         joined_str = '","'.join(licenses_lst)
-        print('format_licenses_to_string res', f'"{joined_str}"')
         return f'"{joined_str}"'
+    elif licenses_lst and len(licenses_lst) == 1:
+        return licenses_lst[0]
     else:
         return 'Unknown'
 
@@ -38,7 +39,6 @@ def format_string_to_licenses(licenses_str: str) -> list[str]:
         licenses_str = licenses_str[1:-1] if licenses_str.startswith('"') and licenses_str.endswith('"') else licenses_str
         license_lst = licenses_str.split('","')
 
-        print('format_string_to_licenses:', license_lst)
         return license_lst
     else:
         return []
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -18,11 +18,10 @@ class ImageDetails(SCADetails):
 
 
 def is_raw_formatted(licenses: str) -> bool:
-    return licenses.__contains__('","')
+    return '","' in licenses
 
 
 def format_licenses_to_string(licenses_lst: list[str]) -> str:
-    print('format_licenses_to_string - start', licenses_lst)
     if isinstance(licenses_lst, list):
         if len(licenses_lst) > 1:
             joined_str = '","'.join(licenses_lst)
@@ -33,7 +32,6 @@ def format_licenses_to_string(licenses_lst: list[str]) -> str:
 
 
 def format_string_to_licenses(licenses_str: str) -> list[str]:
-    print('format_string_to_licenses - start', licenses_str)
     if licenses_str == 'Unknown':
         return [licenses_str]
     elif licenses_str:
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field
 
 UNKNOWN_LICENSE = 'Unknown'
 
+
 @dataclass
 class SCADetails:
     package_types: dict[str, str] = field(default_factory=dict)
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -16,13 +16,8 @@ class ImageDetails(SCADetails):
     name: str | None = ''
     related_resource_id: str | None = ''
 
-def fotmat_licenses_to_string(licenses_lst:list[str]) -> str:
+def format_licenses_to_string(licenses_lst:list[str]) -> str:
     if licenses_lst and len(licenses_lst) > 0:
-        # in case we have a quote inside the license, then we need to escape it,
-        # in CSV the escape is 2 double quotes
-        # for license in licenses_lst:
-        #   license = license.replace(r'"', r'"""')
-
         joined_str = '","'.join(licenses_lst)
         return f'"{joined_str}"'
     else:
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -36,7 +36,7 @@ def format_string_to_licenses(licenses_str: str) -> list[str]:
     print('format_string_to_licenses - start', licenses_str)
     if licenses_str == 'Unknown':
         return [licenses_str]
-    elif licenses_str and len(licenses_str) > 0:
+    elif licenses_str:
         # remove first and last quotes
         licenses_str = licenses_str[1:-1] if licenses_str.startswith('"') and licenses_str.endswith('"') else licenses_str
         license_lst = licenses_str.split('","')
--- a/checkov/common/output/common.py
+++ b/checkov/common/output/common.py
@@ -15,3 +15,24 @@ class ImageDetails(SCADetails):
     image_id: str = ''
     name: str | None = ''
     related_resource_id: str | None = ''
+
+def fotmat_licenses_to_string(licenses_lst:list[str]) -> str:
+    if licenses_lst and len(licenses_lst) > 0:
+        # in case we have a quote inside the license, then we need to escape it,
+        # in CSV the escape is 2 double quotes
+        # for license in licenses_lst:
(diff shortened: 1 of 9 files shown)


tests/sca_package/test_runner.py
```diff
@@ -67,7 +67,7 @@ def test_run(sca_package_report):
     cve_record_with_2_license = next((c for c in report.failed_checks if c.resource == "path/to/requirements.txt.flask" and c.check_name == "SCA package scan"), None)
     assert cve_record_with_2_license is not None
     assert "licenses" in cve_record_with_2_license.vulnerability_details
-    assert cve_record_with_2_license.vulnerability_details["licenses"] == "OSI_APACHE, DUMMY_OTHER_LICENSE"
+    assert cve_record_with_2_license.vulnerability_details["licenses"] == '"OSI_APACHE","DUMMY_OTHER_LICENSE"'
```
Saarett, line 70: Can we add another test and not only replace the first one?

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
     name: str | None = ''
     related_resource_id: str | None = ''
+
+
+def is_raw_formatted(licenses: str) -> bool:
+    return licenses.__contains__('","')
+
+
+def format_licenses_to_string(licenses_lst: list[str]) -> str:
+    print('format_licenses_to_string - start', licenses_lst)
```
Saarett, line 25: Is the log still needed?

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
+        joined_str = '","'.join(licenses_lst)
+        return f'"{joined_str}"'
+    elif licenses_lst and len(licenses_lst) == 1:
+        return licenses_lst[0]
+    else:
+        return 'Unknown'
+
+
+def format_string_to_licenses(licenses_str: str) -> list[str]:
+    print('format_string_to_licenses - start', licenses_str)
```
Saarett, line 36: Is the log still needed?

checkov/common/output/csv.py
```diff
@@ -157,8 +159,23 @@ def persist_report_oss_packages(self, file_name: str, is_api_key: bool, output_p
             is_api_key=is_api_key,
         )
 
+    @staticmethod
+    def arrange_rows(rows: list[dict[str, Any]]) -> None:
```
Saarett, line 163: Have you tested the performance of this method on a real case, a big CSV file? I'm worried it would be pretty slow

Saarett (commented):
Great job! Commented 😄

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
     image_id: str = ''
     name: str | None = ''
     related_resource_id: str | None = ''
+
+
+def is_raw_formatted(licenses: str) -> bool:
+    return licenses.__contains__('","')
```
gruebel, line 21:
```suggestion
    return '","' in licenses
```
you typically don't use the internal methods directly.

checkov/common/output/common.py
gruebel, line 25:
```suggestion
    logging.debug('format licenses to string', licenses_lst)
```
if you really want to log it, then please use `logging`

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
+
+def format_licenses_to_string(licenses_lst: list[str]) -> str:
+    print('format_licenses_to_string - start', licenses_lst)
+    if licenses_lst and len(licenses_lst) > 1:
+        joined_str = '","'.join(licenses_lst)
+        return f'"{joined_str}"'
+    elif licenses_lst and len(licenses_lst) == 1:
+        return licenses_lst[0]
+    else:
+        return 'Unknown'
```
gruebel, line 32:
```suggestion
    if isinstance(licenses_lst, list):
        if len(licenses_lst) > 1:
            joined_str = '","'.join(licenses_lst)
            return f'"{joined_str}"'
        elif licenses_lst:
            return licenses_lst[0]
    return 'Unknown'
```
depends on what you want to do and how sure you are that a list is passed to the function.
```suggestion
    if len(licenses_lst) > 1:
        joined_str = '","'.join(licenses_lst)
        return f'"{joined_str}"'
    elif licenses_lst:
        return licenses_lst[0]
    else:
        return 'Unknown'
```

checkov/common/output/common.py
gruebel, line 36: same here

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
+
+
+def format_string_to_licenses(licenses_str: str) -> list[str]:
+    print('format_string_to_licenses - start', licenses_str)
+    if licenses_str == 'Unknown':
```
gruebel, line 37: can you create a constant for 'Unknown' and then use it here and in the above function.

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
+        return [licenses_str]
+    elif licenses_str and len(licenses_str) > 0:
```
gruebel, line 39:
```suggestion
    elif licenses_str:
```
no need for the right side, because the left side checks if it is not empty

checkov/common/output/csv.py
```diff
@@ -157,8 +159,23 @@ def persist_report_oss_packages(self, file_name: str, is_api_key: bool, output_p
...
+        # we search for formatted rows and covert them back into csv formatted file.
```
gruebel, line 164:
```suggestion
        # we search for formatted rows and convert them back into csv formatted file.
```

checkov/common/output/csv.py
```diff
@@ -157,8 +159,23 @@ def persist_report_oss_packages(self, file_name: str, is_api_key: bool, output_p
...
+        for row in rows:
+            for key in row.keys():
+                val = str(row[key])
+
+                if is_raw_formatted(val):
+                    row[key] = ', '.join(format_string_to_licenses(val))
+                    val = str(row[key])
+                row[key] = val[1:-1] if val.startswith('"') and val.endswith('"') else row[key]
+                row[key] = '' if val == 'None' else row[key]
```
gruebel, line 173:
```suggestion
            for key, value in row.items():
                val = str(value)

                if is_raw_formatted(val):
                    val = ', '.join(format_string_to_licenses(val))
                val = val[1:-1] if val.startswith('"') and val.endswith('"') else val
                row[key] = '' if val == 'None' else val
```
no need to always write to `row[key]` access it

checkov/common/output/csv.py
```diff
@@ -157,8 +159,23 @@ def persist_report_oss_packages(self, file_name: str, is_api_key: bool, output_p
...
+
     @staticmethod
     def write_section(file: str, header: list[str], rows: list[dict[str, Any]], is_api_key: bool) -> None:
+        CSVSBOM.arrange_rows(rows)
```
gruebel, line 177: did you test it against a bigger example to see how the performance implications are to process all entries in all rows?

checkov/common/output/common.py
```diff
@@ -15,3 +15,32 @@ class ImageDetails(SCADetails):
...
+        # remove first and last quotes
+        licenses_str = licenses_str[1:-1] if licenses_str.startswith('"') and licenses_str.endswith('"') else licenses_str
+        license_lst = licenses_str.split('","')
```
ayajbara, line 42: what if the `licenses_str` contains ","?

tests/sca_package/test_runner.py
gruebel, line 70: especially a test with a comma inside a license would be great, can be a fake license.

gruebel (commented):
nice work, added some comments

checkov/common/output/common.py
AdamDev: not really

checkov/common/output/common.py
AdamDev: there are 2 methods: 
```
format_licenses_to_string() - encode the licenses array to string, with "," delimiter.
format_string_to_licenses() - decode back the licenses string to list.
```

so licenses_str is the result of format_licenses_to_string()

tests/sca_package/test_runner.py
AdamDev: yes

checkov/common/output/common.py
AdamDev: I removed it

checkov/common/output/csv.py
AdamDev: I tested it with CSV with 7000 rows, it finished in milliseconds.

checkov/common/output/csv.py
AdamDev: I tested it with CSV with 7000 rows, it finished in milliseconds.

Saarett (approved):
Great job! 🎖️

PR #3868, opened 2022-11-13
feat(terraform): Mask secret values in Terraform plan file reports by resource

arielkru:

[//]: # "
    # PR Title
    ex.
"

## Description

- This PR enables masking secret values that are not classified as secrets.
- The solution is a general one, implemented only in tf plan runner for now, for 3 types of resource & attributes that their values should be hidden (`azurerm_key_vault_secret`: `value`, `aws_secretsmanager_secret_version`: `secret_string`, `google_kms_secret_ciphertext`: `plaintext`).
- Can be extended to future runners by adding the `resource_attributes_to_omit` map when calling `omit_secret_value_from_checks()`

## Checklist:

### DIFF
--- a/checkov/common/typing.py
+++ b/checkov/common/typing.py
@@ -13,7 +13,9 @@ _ScannerCallableAlias: TypeAlias = Callable[
     [str, "BaseCheck", "_SkippedCheck", "dict[str, Any]", str, str, "dict[str, Any]"], None
 ]
 
-ResourceAttributesToOmit: TypeAlias = Dict[str, str]
+_Resource: TypeAlias = str
+_Attribute: TypeAlias = str
+ResourceAttributesToOmit: TypeAlias = Dict[_Resource, _Attribute]
 
 
 class _CheckResult(TypedDict, total=False):
--- a/checkov/common/typing.py
+++ b/checkov/common/typing.py
@@ -13,7 +13,7 @@ _ScannerCallableAlias: TypeAlias = Callable[
     [str, "BaseCheck", "_SkippedCheck", "dict[str, Any]", str, str, "dict[str, Any]"], None
 ]
 
-ResourceAttributesToOmit: TypeAlias = dict[str, str]
+ResourceAttributesToOmit: TypeAlias = Dict[str, str]
 
 
 class _CheckResult(TypedDict, total=False):
--- a/checkov/common/typing.py
+++ b/checkov/common/typing.py
@@ -13,9 +13,7 @@ _ScannerCallableAlias: TypeAlias = Callable[
     [str, "BaseCheck", "_SkippedCheck", "dict[str, Any]", str, str, "dict[str, Any]"], None
 ]
 
-_Resource = str
-_Attribute = str
-ResourceAttributesToOmit = Dict[_Resource, _Attribute]
+ResourceAttributesToOmit: TypeAlias = dict[str, str]
 
 
 class _CheckResult(TypedDict, total=False):
--- a/checkov/common/typing.py
+++ b/checkov/common/typing.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, Dict
 from typing_extensions import TypeAlias, TypedDict
 
 if TYPE_CHECKING:
@@ -13,6 +13,9 @@ _ScannerCallableAlias: TypeAlias = Callable[
     [str, "BaseCheck", "_SkippedCheck", "dict[str, Any]", str, str, "dict[str, Any]"], None
 ]
 
+_Resource = str
+_Attribute = str
+ResourceAttributesToOmit = Dict[_Resource, _Attribute]
 
 class _CheckResult(TypedDict, total=False):
     result: "CheckResult" | tuple["CheckResult", dict[str, Any]]
--- a/checkov/common/typing.py
+++ b/checkov/common/typing.py
@@ -17,6 +17,7 @@ _Resource = str
 _Attribute = str
 ResourceAttributesToOmit = Dict[_Resource, _Attribute]
 
+
 class _CheckResult(TypedDict, total=False):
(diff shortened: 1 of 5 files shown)


checkov/common/typing.py
```diff
@@ -13,6 +13,10 @@
     [str, "BaseCheck", "_SkippedCheck", "dict[str, Any]", str, str, "dict[str, Any]"], None
 ]
 
+_Resource = str
+_Attribute = str
+ResourceAttributesToOmit = Dict[_Resource, _Attribute]
```
gruebel, line 18:
```suggestion
ResourceAttributesToOmit: TypeAlias = dict[str, str]
```
something like this is called a `TypeAlias` and should be marked as such 🙂 
I don't think you need `_Resource` and `_Attribute`, at least I didn't see you using it somewhere.

checkov/common/util/secrets.py
```diff
@@ -133,20 +133,28 @@ def omit_secret_value_from_line(secret: str, line_text: str) -> str:
...
-
-        for idx, line in entity_code_lines:
-            censored_line = omit_multiple_secret_values_from_line(secrets, line)
-            censored_code_lines.append((idx, censored_line))
-    else:
-        censored_code_lines = entity_code_lines
+        secrets.update([str(secret) for key, secret in entity_config.items() if key.startswith(f'{check.id}_secret')])
+
+    if resource_attributes_to_omit and check.entity_type in resource_attributes_to_omit and \
+            resource_attributes_to_omit.get(check.entity_type, None) in entity_config:
```
gruebel, line 146:
```suggestion
            resource_attributes_to_omit.get(check.entity_type) in entity_config:
```
`None` is default anyway

checkov/common/util/secrets.py
```diff
@@ -133,20 +133,28 @@ def omit_secret_value_from_line(secret: str, line_text: str) -> str:
...
+        secret = entity_config.get(resource_attributes_to_omit.get(check.entity_type, {}), [])  # type:ignore[arg-type]
```
gruebel, line 147:
```suggestion
        secret = entity_config.get(resource_attributes_to_omit.get(check.entity_type), [])  # type:ignore[arg-type]
```
`{}` as a default doesn't make sense here, either nothing or an empty string `resource_attributes_to_omit.get(check.entity_type, "")`

checkov/common/util/secrets.py
```diff
@@ -133,20 +133,28 @@ def omit_secret_value_from_line(secret: str, line_text: str) -> str:
...
+        if isinstance(secret, list) and len(secret) != 0:
```
gruebel, line 148:
```suggestion
        if isinstance(secret, list) and len(secret) != 0:
```
```suggestion
        if isinstance(secret, list) and secret:
```
not empty check, without using `len` is faster

checkov/common/util/secrets.py
```diff
@@ -133,20 +133,28 @@ def omit_secret_value_from_line(secret: str, line_text: str) -> str:
...
+            secrets.add(secret[0])
+
+    if len(secrets) == 0:
```
gruebel, line 151:
```suggestion
    if not secrets:
```
same here

checkov/terraform/plan_runner.py
```diff
@@ -31,6 +32,12 @@
     "CKV_GCP_82",
 }
 
+RESOURCE_ATTRIBUTES_TO_OMIT = {
+    'azurerm_key_vault_secret': 'value',
+    'aws_secretsmanager_secret_version': 'secret_string',
+    'google_kms_secret_ciphertext': 'plaintext'
```
gruebel, line 38: should we also add SSM parameter to the list or in a follow up PR? https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssm_parameter#encrypted-string-using-default-ssm-kms-key as I remember they are saved as plain text in the plan file, but I could be wrong.

checkov/terraform/plan_runner.py
```diff
@@ -115,16 +122,20 @@ def run_block(self, entities,
...
                 results = registry.scan(scanned_file, entity, [], runner_filter, report_type=CheckType.TERRAFORM_PLAN)
                 for check, check_result in results.items():
                     if check.id in TF_LIFECYCLE_CHECK_IDS:
                         # can't be evaluated in TF plan
                         continue
-
+                    censored_code_lines = omit_secret_value_from_checks(check, check_result,
+                                                                        entity_code_lines,
+                                                                        entity_config,
+                                                                        resource_attributes_to_omit=RESOURCE_ATTRIBUTES_TO_OMIT)
```
gruebel, line 135: can you also use named parameters for the other values

gruebel (commented):
nice work, added few minor suggestions

Eliran-Turgeman (approved):

checkov/common/typing.py
arielkru: I dont use them, my intention was to indicate the expected structure of `ResourceAttributesToOmit` if anyone wants to use it in the future, so its clear that the key should be a resource name and the value its matching attribute name.
Think its better not to mention that? @gruebel

checkov/terraform/plan_runner.py
arielkru: Good catch, so is that true also for `aws_db_instance`: `password` in plan files? @gruebel

checkov/common/typing.py
gruebel: gotcha, sounds good, then
```python
_Resource: TypeAlias = str
_Attribute: TypeAlias = str
ResourceAttributesToOmit: TypeAlias = dict[_Resource, _Attribute]
```

checkov/terraform/plan_runner.py
gruebel: good question, yeah it is `All arguments including the username and password will be stored in the raw state as plain-text.`

gruebel: lets create a new GitHub issue with example resources, which can omit secrets in the TF plan file

PR #3874, opened 2022-11-14
chore: Stop bundling optional dependency pyston and move to extras

### LINKED ISSUE #3649: Checkov installation on Ubuntu
**Describe the issue**
If it is related to an existing check, please note the relevant check ID.
Also, explain the logic for this addition / change.

**Examples**

 - Checkov Version [e.g. 22]

**Additional context**

Hey, 
Tried to install checkov on Ubuntu 18.04 and got this error
Could not find a version that satisfies the requirement pyston-autoload==2.3.5; python_version < "3.11" and (sys_platform == "linux" or sys_platform == "darwin") and platform_machine == "x86_64" (from checkov) (from versions: )
No matching distribution found for pyston-autoload==2.3.5; python_version < "3.11" and (sys_platform == "linux" or sys_platform == "darwin") and platform_machine == "x86_64" (from checkov)

nikolaik:
## Description

This fixes checkov install on alpine and probably others where the version markers are unable to omit [musl](https://www.musl-libc.org/) and similarly incompatible install targets.

checkov can still bundle pyston by installing/requiring `checkov[pyston]`

Fixes #3649 

This can be tested by cloning this pr on a failing machine and doing `pip install .` and `pip install .[pyston]`.

Some reasons for why this change is a good one:
- Allows more users (like alpine users) to install the latest checkov using pip, without patches.
- Pyston is working to upstream their change into cpython, so bundling pyston will provide less value over time, once these patches hit cpython. See https://blog.pyston.org/2022/09/29/announcing-3-7-3-10-support-and-a-new-direction/ .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,10 @@ setup(
             "GitPython==3.1.7",
             "bandit",
             "jsonschema",
+        ],
+        "pyston": [
+            "pyston-autoload==2.3.5; python_version < '3.11' and (sys_platform == 'linux' or sys_platform == 'darwin') and platform_machine == 'x86_64'",
+            "pyston==2.3.5; python_version < '3.11' and (sys_platform == 'linux' or sys_platform == 'darwin') and platform_machine == 'x86_64'",
         ]
     },
     install_requires=[
@@ -70,8 +74,6 @@ setup(
         "prettytable>=3.0.0",
         "pycep-parser==0.3.9",
         "charset-normalizer",
-        "pyston-autoload==2.3.5; python_version < '3.11' and (sys_platform == 'linux' or sys_platform == 'darwin') and platform_machine == 'x86_64'",
-        "pyston==2.3.5; python_version < '3.11' and (sys_platform == 'linux' or sys_platform == 'darwin') and platform_machine == 'x86_64'",
         "schema",
         "requests>=2.26.0",
     ],


nikolaik:
@nimrodkor Should I move pyston to dev dependencies in `Pipfile`?

gruebel:
hey @nikolaik thanks for the contribution, but this basically removes `pyston` completely. Most users will not try to run `pip install checkov[pyston]` and there is no way I know of to do the opposite of adding some parameter to not install a package. We will also leverage more and more Python packages with C extensions or even add Cython in the mix, which will result in `checkov` not being installable on Alpine at all.

Python on Alpine is usually not recommended due to the usage of musl libc compared to glibc, which results in weird errors and degraded performance. Here you can find some info around that https://github.com/docker-library/python/issues/509

Long story short, we don't target Alpine as the OS to run `checkov` on, if it works great, if not then use our Docker image or move to a newer Debian or Fedora based OS.

nikolaik:
Thanks for reviewing!

> Python on Alpine is usually not recommended due to the usage of musl libc compared to glibc, which results in weird errors and degraded performance. Here you can find some info around that [docker-library/python#509](https://github.com/docker-library/python/issues/509)

I see the concern with slower performance under alpine, especially with https://superuser.com/questions/1219609/why-is-the-alpine-docker-image-over-50-slower-than-the-ubuntu-image Found another discussion on the topic here: https://github.com/edgedb/edgedb/pull/4273 In my experience I have not had any issues with running python things under alpine.
 
> Long story short, we don't target Alpine as the OS to run `checkov` on, if it works great, if not then use our Docker image or move to a newer Debian or Fedora based OS.

With python 3.11 pyston deps are no longer matched using the markers, so bumping to python 3.11 when using alpine is a valid workaround for now.

As long as checkov deps provide musllinux wheels then installation will just work, even though there is not official support.

PR #3907, opened 2022-11-21
feat(secrets): Add Terraform multiline secrets handling

### LINKED ISSUE #1660: False positive for `CKV_SECRET_6` for `google_secret_manager_secret_version` resources
**Describe the bug**
Checkov 2.0.446 is detecting a false positive regarding the secret manager usage.

Basically it is showing multiple failed check for resources declared in the following format:

```hcl
data "google_secret_manager_secret_version" "mysecret" {
  secret = "somesecretid"
}
```

With the error:

```
Check: CKV_SECRET_6: "Base64 High Entropy String"
        FAILED for resource: 8b50635275125444f17b28140b0eb5f529861e4b
        File: /secrets.tf:6-7
        Guide: https://docs.bridgecrew.io/docs/git_secrets_6
```

**Expected behavior**
Checkov should not detect this is a failed check, because this is the secret-id, and not the secret value.

 - OS: ubuntu
 - Checkov Version  2.0.446

gruebel:

[//]: # "
    # PR Title
    ex.
"

## Description

- added a new multiline parser for Terraform hcl files
- since Terraform can leverage heredoc and can have embedded JSON or YAML in it, I iterate over the possible multiline parser
- added a method to ignore secrets, which are related to a `data` block

Fixes #1660

## Checklist:


checkov/secrets/parsers/terraform/multiline_parser.py
```diff
@@ -0,0 +1,40 @@
...
+class TerraformMultiLineParser(BaseMultiLineParser):
+    def consecutive_lines_in_same_object(
+        self,
+        raw_context: CodeSnippet | None,
+        other_line_idx: int,
+    ) -> bool:
+        if not raw_context or not 0 <= other_line_idx < len(raw_context.lines):
+            return False
+
+        return True
```
Eliran-Turgeman, line 25:
```suggestion
        return raw_context and 0 <= other_line_idx < len(raw_context.lines)
```

Eliran-Turgeman (approved):
🔥

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -151,37 +195,54 @@ def analyze_line(
...
+                        context=context,
+                        raw_context=raw_context,
+                    ):
+                        # to reduce false-positives we check the context of the finding
+                        return set()
+
+                    return potential_secrets
 
                 # not so classic key-value pair, from multiline, that is only in an array format.
                 # The scan is one-way backwards, so no duplicates expected.
```
marynaKK, line 223: my mistake- but can we change the comment here to something like this:
The scan is both forward and backword, searching for a pair for a potential secret.

checkov/secrets/plugins/ignore_secret.py
```diff
@@ -0,0 +1,43 @@
...
+    for line_index in range(secret_line_index - 1, -1, -1):
+        if raw_context.lines[line_index].lstrip().startswith('data "'):
+            # a data block is typically used to get remote information,
+            # therefore can retrieve a secret, but has not a hardcoded secret
+            return True
+
+    return False
+
+
+def ignore_secret(
```
marynaKK, line 27: If we adding a mechanism for ignoring some data- maybe it can be added to the multiline parser as a function for all classes?
that's the same as skipping comments, but in this case, that's skipping specific blocks with data as a prefix.

checkov/secrets/plugins/ignore_secret.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import TYPE_CHECKING
+
+from detect_secrets.util.filetype import FileType
+
+if TYPE_CHECKING:
+    from detect_secrets.util.code_snippet import CodeSnippet
+
+
+def ignore_terraform_data_block(
+    secret_line_index: int,
```
marynaKK, line 12: no need for `secret_line_index` - there is `raw_context.target_index`
I wanted to use as much from code snippet as possible - if we will change how it works in the future :)

marynaKK (approved):
Looks great! :) 🚀 🍒 
I am a bit concerned about parsing lines that are not so beautiful (edge cases that are close to the formatted file). but I guess that's not a big deal :)

checkov/secrets/parsers/terraform/multiline_parser.py
gruebel: this happens, when you write your code several times and forget to simplify it, after removing a lot of code 😄

checkov/secrets/parsers/terraform/multiline_parser.py
```diff
@@ -0,0 +1,40 @@
...
+
+import re
+from typing import TYPE_CHECKING
+
+from checkov.common.parsers.multiline_parser import BaseMultiLineParser
+
+if TYPE_CHECKING:
+    from detect_secrets.util.code_snippet import CodeSnippet
+
+START_OBJ = re.compile(r"\s*\w*\s*=?\s*{")
```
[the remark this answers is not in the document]
gruebel: since I use `re.match()` with this pattern, it is not needed to `^` because it always checks against the beginning of the string

checkov/secrets/parsers/terraform/multiline_parser.py
```diff
@@ -0,0 +1,40 @@
...
+END_OBJ = re.compile(r"\s*}")
```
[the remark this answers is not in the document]
gruebel: because Terraform blocks are a bit different, then json syntax. no need for `,` handling for example

checkov/secrets/plugins/entropy_keyword_combinator.py
gruebel: sure 😄

checkov/secrets/plugins/ignore_secret.py
gruebel: yeah, that was my problem, it is related to a specific file type, but I was not sure, if it is a good idea to add it to the multi parser, but I will revisit it 🤔

checkov/secrets/plugins/ignore_secret.py
gruebel: nice catch 🚀

PR #3929, opened 2022-11-23
fix(general): update checks_metadata structure

arielkru:

[//]: # "
    # PR Title
    ex.
"

## Description
check_meatada.json should be a dict of resourceIds as keys, with each of the resourceIds containing a dict with check_ids as keys and the metadata as values.

## Checklist:

### DIFF
--- a/checkov/common/bridgecrew/wrapper.py
+++ b/checkov/common/bridgecrew/wrapper.py
@@ -42,7 +42,7 @@ def _extract_checks_metadata(report: Report, full_repo_object_key: str) -> dict[
     for check in list(itertools.chain(report.passed_checks, report.failed_checks, report.skipped_checks)):
         metadata_key = f'{check.file_path}:{check.resource}'
         check_meta = dict({k: getattr(check, k, "") for k in check_metadata_keys},
-             **{'file_object_path': full_repo_object_key + check.file_path})
+                          **{'file_object_path': full_repo_object_key + check.file_path})
         metadata[metadata_key][check.check_id] = check_meta
 
     return metadata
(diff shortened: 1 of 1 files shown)


checkov/common/bridgecrew/wrapper.py
```diff
@@ -37,9 +38,14 @@ def _put_json_object(s3_client: BaseClient, json_obj: Any, bucket: str, object_p
 
 
 def _extract_checks_metadata(report: Report, full_repo_object_key: str) -> dict[str, dict[str, Any]]:
-    return {check.check_id: dict({k: getattr(check, k, "") for k in check_metadata_keys},
-                                 **{'file_object_path': full_repo_object_key + check.file_path}) for check in
-            list(itertools.chain(report.passed_checks, report.failed_checks, report.skipped_checks))}
+    metadata = defaultdict(dict)
+    for check in list(itertools.chain(report.passed_checks, report.failed_checks, report.skipped_checks)):
```
marynaKK, line 42: no need for list - the itertoolss.chain is enough :)

checkov/common/bridgecrew/wrapper.py
```diff
@@ -37,9 +38,14 @@ def _put_json_object(s3_client: BaseClient, json_obj: Any, bucket: str, object_p
...
+        metadata_key = f'{check.file_path}:{check.resource}'
+        check_meta = dict({k: getattr(check, k, "") for k in check_metadata_keys},
```
marynaKK, line 44: dict() can be replaced with {} : https://katrin-affolter.ch/Programming/Python_dictionaries
much more faster and better :)

marynaKK (commented):
had 2 comments :) 🍒 
Also, a Q as I do not fully understand - are the resource_id the same as the one we are changing in places like gha, azure pipelines, IR? if they are - they are different from the id that is used here. Is this an issue?

arielkru:
> had 2 comments :) 🍒 Also, a Q as I do not fully understand - are the resource_id the same as the one we are changing in places like gha, azure pipelines, IR? if they are - they are different from the id that is used here. Is this an issue?

@marynaKK good q, it shouldn't be an issue because the resourceId here doesnt change the resourceIds in the reports, it just fits the way we expect resource keys to appear in the checks_metadata.json file.
So for example an azure_pipelines resource is saved under the key: `/azure-pipelines.yaml:jobs[0](FailTag)` and that's ok

nimrodkor (approved):

PR #3938, opened 2022-11-24
fix(terraform): CKV_AZURE_144 passes on defaults

### LINKED ISSUE #3936: CKV_AZURE_144 passes on defaults
**Describe the issue**
If it is related to an existing check, please note the relevant check ID.
Also, explain the logic for this addition / change.

The check CKV_AZURE_144 passes if the property "public_network_access_enabled" is not explicitly set since it assumes that it defaults to false. This seems to not be the case at least for AzureRM < 3.0.0. Right now we have publicly accessible Workspaces for which the check passes since the property is not set.

**Examples**

The Module Code:

image

The actual Workspace:
image

Since the defaults seem to be subject to change the check should probably fail if the property is not set.

Opulski:

[//]: # "
    # PR Title
    ex.
"

## Description

Fixes #3936

### DIFF
--- a/checkov/terraform/checks/resource/azure/MLPublicAccess.py
+++ b/checkov/terraform/checks/resource/azure/MLPublicAccess.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import Any
 
-from checkov.common.models.enums import CheckCategories
+from checkov.common.models.enums import CheckCategories, CheckResult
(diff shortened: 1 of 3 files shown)


gruebel (approved):
nice, thanks for the contribution 🥇

tronxd (approved):

PR #3969, opened 2022-12-02
fix(terraform): Add some more supported keys to bigquery public acl check ignore list to avoid false positive

dharada1:

## Description

CKV_GCP_15 checks if access block with only the role key found in the statefile. This happens when manually adding "allUsers" to the dataset. ( It is said in existing code comment ) 
In the code, there is valid key list that determines the access block is in such a situation. This list should be maintained each time when google adds new supported keys.

BigQuery dataset access block supports more keys than it's implemented on the checkov rule currently. 
If you have these keys on your dataset resource, checkov returns false-positve result.

- https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets?hl=ja
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset#nested_access

Actually I encountered this problem on bridgecrew's runtime scan first time, and I noticed checkov has same problem. Now I'm checking if it reproduces in terraform & checkov.

## Checklist:


gruebel:
thanks, looks good so far. Would be great if you can also add a test for it 🙂

dharada1:
@gruebel Thanks ! I'll work on it next week 🙏

dharada1:
@gruebel 

I added tests for new key types and run all tests by `pipenv run python -m coverage run -m pytest tests/` .

tests/terraform/checks/resource/gcp/test_GoogleBigQueryDatasetPublicACL.py
```diff
@@ -51,6 +51,83 @@ def test_success(self):
...
+    def test_success_dataset(self):
+        resource_conf = {"dataset_id": ["example_dataset"],
+                         "friendly_name": ["test"],
+                         "description": ["This is a test description"],
+                         "location": ["EU"],
+                         "default_table_expiration_ms": [3600000],
+                         "access": [{"dataset": {"dataset": {"datasetId": "foo","projectId": "bar"},"targetTypes": ["VIEWS"]}}]
+                         }
+        scan_result = check.scan_resource_conf(conf=resource_conf)
+        self.assertEqual(CheckResult.PASSED, scan_result)
```
dharada1, line 107: These ones don't require "role" . Instead, they requires each object schema.
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets?hl=ja

tests/terraform/checks/resource/gcp/test_GoogleBigQueryDatasetPublicACL.py
```diff
@@ -51,6 +51,83 @@ def test_success(self):
...
+    def test_success_iam_member(self):
+        resource_conf = {"dataset_id": ["example_dataset"],
+                         "friendly_name": ["test"],
+                         "description": ["This is a test description"],
+                         "location": ["EU"],
+                         "default_table_expiration_ms": [3600000],
+                         "access": [{"role": ["READER"], "iam_member": ["deleted:serviceAccount:foo@bar.iam.gserviceaccount.com?uid=1234567890"]}]
+                         }
+        scan_result = check.scan_resource_conf(conf=resource_conf)
+        self.assertEqual(CheckResult.PASSED, scan_result)
```
dharada1, line 118: This happens when you delete service account.
Existing service account is appear with user_by_email.

gruebel:
> > My code follows the style guidelines of this project
> 
> I'm not sure if there is style guidelines in this project 🤔

it mainly means, linting and typing job passes and it fits the general approach of the other code, ex. how we write other checks 🙂

gruebel (approved):
nice job 🎄 I removed the `iam_member` case, because it is not supported by the provider. I also changed the syntax of the test file, because it still used the old way.

marynaKK (approved):
🎄 🍒

PR #3482, opened 2022-09-07
feat(general): `intersects/not_intersects` operators (solvers)

kadoshms:
**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

## Description

Introducing 2 new operators - `intersects` and `not_intersects` operators, that given 2 values, check if those values intersect.

*Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.*

Fixes # (issue)

### Description
*Include a description of what makes it a violation and any relevant external links.*

### Fix
*How does someone fix the issue in code and/or in runtime?*

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my feature, policy, or fix is effective and works
- [ ] New and existing tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4004, opened 2022-12-07
feat(kubernetes): Added CKV2_K8S_1 to ensure RoleBinding do not allow privilege escalation to a ServiceAccount/Node

bo156:

[//]: # "
    # PR Title
    ex.
"

## Description

Used graph connection in k8s to ensure no ServiceAccount/Node is allowed to perform an action on a RoleBinding, thus allowing him to set its own permissions.

## New/Edited policies

- New - CKV2_K8S_1

### Description
Used graph connection in k8s to ensure no ServiceAccount/Node is allowed to perform an action on a RoleBinding, thus allowing him to set its own permissions.

## Checklist:


YaaraVerner (approved):
Nice job!

checkov/common/checks_infra/solvers/attribute_solvers/not_within_attribute_solver.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional, Any, Dict
+
+from checkov.common.checks_infra.solvers.attribute_solvers.within_attribute_solver import WithinAttributeSolver
+from checkov.common.graph.checks_infra.enums import Operators
+
+
+class NotWithinAttributeSolver(WithinAttributeSolver):
```
gruebel, line 7: it actually would have been better to add the new operator in a separate PR, because you also need to add tests for it, like here #3482 doing so would also add it to the changelog and now it will be a hidden feature addition.

bo156: Completely agree so I removed this solver, instead I used a `not` operator on the `within` one.
I would have just dropped the commit but I made a mess and pushed something else with it, so I just added another commit removing it :/

PelegLi (approved):
Nice!

PR #4017, opened 2022-12-08
feat(github):  more CIS checks- part2

marynaKK:

## Description

Added 3 checks:
Please refer to the docs documented for each check to verify the logic I chose makes sense. 
As checkov runs on files for checks, I decided to add a default file for the checks that pass for GitHub om default. 

## Checklist:


marynaKK (commented):
.

marynaKK (commented):
fix naming

checkov/github/checks/issue_deletion_is_limited.py
```diff
@@ -0,0 +1,32 @@
...
+        categories = (CheckCategories.SUPPLY_CHAIN,)
+        super().__init__(
+            id=id,
+            name=name,
+            categories=categories,
+            supported_entities=["*"],
+            block_type=BlockType.DOCUMENT
+        )
+
+    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> CheckResult | None:  # type:ignore[override]
```
gruebel, line 26: does this check always pass? if yes, then we don't need it. I would rather add a blank file with comment about not needing to implement it.

checkov/github/checks/repository_creation_is_limited.py
```diff
@@ -0,0 +1,51 @@
...
+            if len(found_parameters) != 3 or members_can_create_repositories:
+                return CheckResult.FAILED
+            return CheckResult.PASSED
+        return None
+
+    def get_evaluated_keys(self) -> list[str]:
+        return [
+            "members_can_create_public_repositories",
+            "members_can_create_private_repositories",
+            "members_can_create_internal_repositories"
```
gruebel, line 47: it feels like this check should be split into 3, because allowing to create a public is the worst from a security perspective, then internal and lastly private ones.

checkov/github/checks/repository_deletion_is_limited.py
```diff
@@ -0,0 +1,32 @@
...
+        categories = (CheckCategories.SUPPLY_CHAIN,)
+        super().__init__(
+            id=id,
+            name=name,
+            categories=categories,
+            supported_entities=["*"],
+            block_type=BlockType.DOCUMENT
+        )
+
+    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> CheckResult | None:  # type:ignore[override]
```
gruebel, line 26: also here, it doesn't look like we need this check

checkov/github/dal.py
```diff
@@ -158,6 +169,14 @@ def persist_branch_metadata(self) -> None:
         if branch_metadata:
             BaseVCSDAL.persist(path=self.github_conf_file_paths["branch_metadata"][0], conf=branch_metadata)
 
+    def persist_organization_metadata(self) -> None:
+        org_metadata = self.get_organization_metadata()
+        if org_metadata:
+            BaseVCSDAL.persist(path=self.github_conf_file_paths["org_metadata"][0], conf=org_metadata)
+
+    def persist_github_default_empty_file(self) -> None:
+        BaseVCSDAL.persist(path=self.github_conf_file_paths["default_github"][0], conf={CKV_GITHUB_DEFAULT: None})
```
gruebel, line 178: is there any problem just to set the `conf` as an empty `dict`?

checkov/github/checks/repository_creation_is_limited.py
marynaKK: I agree. Should have mentioned I was not sure about this. splitting to 3 checks :)

checkov/github/dal.py
marynaKK: yes- in base_registry, if there is no entity (in this case all the file content is `{}`) , in line 111-112, so no result is returned for the check. 

that's why I chose to put something default for all checks that might need this. 
if we don't run them in the first place as they always pass- no need for this file :)

checkov/github/checks/issue_deletion_is_limited.py
marynaKK: Yes, this check always passes for github. I think so - that's why I added documentation. would be happy for a second opinion here.

checkov/github/repository_creation_is_limited.py
```diff
@@ -0,0 +1,40 @@
...
+            matches = jsonpath_expression.find(conf)
+            if matches and not matches[0].value:
+                return CheckResult.PASSED
+            # default value for all 3 public, internal and private is true.
+            return CheckResult.FAILED
+        return None
+
+    @abstractmethod
+    def get_evaluated_keys(self) -> list[str]:
+        pass
```
Eliran-Turgeman, line 40: can be removed if not implemented here, and also I think you missed the creation of the check instance

Eliran-Turgeman (approved):

checkov/github/checks/issue_deletion_is_limited.py
marynaKK: removed!

gruebel (approved):
🎄

PR #4033, opened 2022-12-10
feat(terraform_plan): support jsonpath queries in AWS IAM policy strings for Terraform plan

gruebel:
**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

[//]: # "
    # PR Title
    Be aware that we use the title to create changelog automatically and therefore only allow specific prefixes
    - break:    to indicate a breaking change, this supersedes any of the types
    - feat:     to indicate new features or checks
    - fix:      to indicate a bugfix or handling of edge cases of existing checks
    - docs:     to indicate an update to our documentation
    - chore:    to indicate adjustments to workflow files or dependency updates
    - platform: to indicate a change needed for the platform
    Additionally a scope is needs to be added to the prefix, which indicates the targeted framework, in doubt choose 'general'.
    ex.
    feat(terraform): add CKV_AWS_123 to ensure that VPC Endpoint Service is configured for Manual Acceptance
"

## Description

- added an extra behaviour for specific fields in AWS IAM resources to transform the strings to `dict`
- due to performance implications it is only done for the resource types and the specific field
- had to fork `jsonpath-ng` and fix an issue, which didn't allow filtering on a single `dict`

Fixes #4032

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my feature, policy, or fix is effective and works
- [x] New and existing tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #4043, opened 2022-12-11
chore: replace deprecated LegacyVersion with own one

gruebel:

[//]: # "
    # PR Title
    ex.
"

## Description

- the deprecated LegacyVersion was finally removed, therefore I copied the original implementation and created a drop in replacement
- after merging PR #4033 I will update the dependencies

## Checklist:


Eliran-Turgeman (approved):
☃️

checkov/common/packaging/version.py
```diff
@@ -0,0 +1,141 @@
...
+    @property
+    def is_postrelease(self) -> bool:
+        return False
+
+    @property
+    def is_devrelease(self) -> bool:
+        return False
+
+
+_legacy_version_component_re = re.compile(r"(\d+ | [a-z]+ | \.| -)", re.VERBOSE)
```
matansha, line 89: are the spaces part of the regular expression? 
also, i suggest to add flag for case insensitive

checkov/common/packaging/version.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @property
+    def public(self) -> str:
+        return self._version
+
+    @property
+    def base_version(self) -> str:
+        return self._version
+
+    @property
```
matansha, line 52: Why do you need all this meaningless properties?

gruebel: so it is equivalent with the `Version` class inside the `packaging`  library

checkov/common/packaging/version.py
gruebel: the spaces don't seem to do anything for this use case, just kept them as they were. but will add ignorecase to it

matansha: Are you sure? the space are meaningful in regex. did you check this? if yes, i will approve it

gruebel: yeap, I tested it. It didn't change the output with or without the spaces.

gruebel: in general you are right that spaces are meaningful, but not for what it is used here

matansha (approved):

PR #4057, opened 2022-12-12
feat(github): more CIS checks- part3 

marynaKK:

## Description
This pr has many changes, hence not all policies of 1.3.* are included here.
main changes:

1. Added checks CKV_GITHUB_26, CKV_GITHUB_27, CKV_GITHUB_28
To support these checks, changes and additions were made:
- multiple API replies could be parsed by the same scheme, causing files to be parsed in a check even if the check should not parse them
example: collaborators vs. admin collaborators - the count changes depending on the check, and so does the check result.
- no code was showing in the code snippet for some checks - fixed
-  added repo\org additional collected info from the DAL layer to each check definition- allows combining parameters from multiple API calls in a single check.
- added env variable `GITHUB_MAX_ADMINS_ORG` - LMK if the naming should be changed, or if this is not the correct way to let the user control it.

2. Added base check of OrganizationCheck - hence removed the RepositoryCreationIsLimited base check, and changed 3 checks that were derived from it.

## Checklist:


checkov/github/base_github_org_check.py
```diff
@@ -0,0 +1,45 @@
...
+            if org_schema.validate(conf):
+                evaluated_key = self.get_evaluated_keys()[0].replace("/", ".")
+                jsonpath_expression = parse(f"$..{evaluated_key}")
+                matches = jsonpath_expression.find(conf)
+                if matches:
+                    if matches[0].value in self.get_allowed_values():
+                        return CheckResult.PASSED
+                    return CheckResult.FAILED
+                return self.missing_attribute_result
+        return None
```
nimrodkor, line 37: @marynaKK Should retrun `CheckResult.UNKNOWN`, we don't return `None` from checks

checkov/github/checks/minimum_admins_in_org.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> CheckResult | None:  # type:ignore[override]
+        ckv_metadata, conf = self.resolve_ckv_metadata_conf(conf=conf)
+        if 'org_admins' in ckv_metadata.get('file_name', ''):
+            if org_members.validate(conf):
+                if len(conf) <= ckv_metadata['org_complementary_metadata']['max_admins_count']:
+                    return CheckResult.PASSED
+                else:
+                    return CheckResult.FAILED
+        return None
```
nimrodkor, line 33: @marynaKK Here as well

checkov/github/checks/repository_collaborators.py
```diff
@@ -21,21 +21,22 @@ def __init__(self) -> None:
...
+                    if isinstance(item, dict):
+                        permissions = item.get("permissions", {})
+                        admin = permissions.get('admin', False)
+                        if admin:
+                            admin_collaborators += 1
+                if admin_collaborators >= 2:
+                    return CheckResult.PASSED
+                else:
+                    return CheckResult.FAILED
         return None
```
nimrodkor, line 39: @marynaKK Here as well, I see we missed this

checkov/github/runner.py
```diff
@@ -53,3 +54,22 @@ def require_external_checks(self) -> bool:
...
+            file_name = Path(file_path).stem
+            ckv_metadata = {
+                'file_name': file_name,
+                'org_complementary_metadata': self.github.org_complementary_metadata,
+                'repo_complementary_metadata': self.github.repo_complementary_metadata,
+            }
+            if isinstance(definition, dict):
+                definition[CKV_METADATA] = ckv_metadata
+            elif isinstance(definition, list):
+                definition.append(ckv_metadata)
```
nimrodkor, line 75: @marynaKK Why not use `force_list` and have 1 way to handle this, at all times?

checkov/json_doc/runner.py
```diff
@@ -45,7 +45,7 @@ def _parse_file(
         return parse(filename=f, file_content=file_content)
 
     def get_start_end_lines(self, end: int, result_config: dict[str, Any], start: int) -> tuple[int, int]:
-        if not isinstance(result_config, DictNode):
+        if not (isinstance(result_config, DictNode) or isinstance(result_config, ListNode)):
```
nimrodkor, line 48:
```suggestion
        if not isinstance(result_config, (ListNode, DictNode)):
```

nimrodkor (commented):
Very nice @marynaKK !

checkov/github/runner.py
marynaKK: I didn't want to impact existing checks more than I already did in this pr.
Is a good idea actually  :)

marynaKK: I tried to- this is mixing between data in form of a list and an addition to an existing list. 
I still will have if-else clauses, but that would be on modified data -> possible additional errors.

nimrodkor: So make it a dict always. Does that solve it?

marynaKK: no- it will change the schema of the data, then I can no longer use the validate schema, or I would have to change back the data to the original schema.
why is this wrong to support both cases?

nimrodkor: @marynaKK Because it means multiple places across the code need to always take into account the structure of the data may be different. Data should be structured consistently. Please find a solution to this

checkov/github/checks/webhooks_https_orgs.py
```diff
@@ -23,18 +23,20 @@ def __init__(self) -> None:
...
+            if org_webhooks_schema.validate(conf):
+                for item in conf:
+                    if isinstance(item, dict):
+                        item_config = item.get("config", {})
+                        if not item_config:
+                            continue
+                        url = item_config.get('url', '')
+                        insecure_ssl = item_config.get('insecure_ssl', '0')
+                        secret = item_config.get('secret', '')
+                        if re.match("^http://", url) or insecure_ssl != '0' and secret != '********':  # nosec
```
gruebel, line 37: can you move `"^http://"` to `re.compile()` at the top, so it is scannable by CodeQL

checkov/github/checks/webhooks_https_repos.py
```diff
@@ -23,17 +23,19 @@ def __init__(self) -> None:
...
+        if 'repository_webhooks' in ckv_metadata.get('file_name', ''):
+            if repository_webhooks_schema.validate(conf):
+                for item in conf:
+                    if isinstance(item, dict):
+                        item_config = item.get("config", {})
+                        if not item_config:
+                            continue
+                        url = item_config.get('url', '')
+                        insecure_ssl = item_config.get('insecure_ssl', '0')
+                        if re.match("^http://", url) or insecure_ssl != '0':
```
gruebel, line 36: same here, or put it in some utils/consts file

checkov/github/dal.py
```diff
@@ -54,6 +58,11 @@ def discover(self) -> None:
         self.default_branch_cache = {}
         self.org = os.getenv('GITHUB_ORG', '')
 
+        self.discover_optional_variables()
+
+    def discover_optional_variables(self) -> None:
+        self.org_complementary_metadata['max_admins_count'] = os.getenv('GITHUB_MAX_ADMINS_ORG', 2)
```
gruebel, line 64: maybe I missed something, but why do we set this via env var?

marynaKK: thought it is a fluid count that can change on how big is the repository. and I am not sure what number to put here.

nimrodkor (approved):

gruebel (approved):
🎄

PR #4076, opened 2022-12-14
feat(terraform): Omit values from graph checks

arielkru:

[//]: # "
    # PR Title
    ex.
"

## Description

This change omits attributes (when defined) from graph checks as well.
Checkov output example after this change for a graph check with sensitive attributes:
image

## Checklist:


checkov/common/util/secrets.py
```diff
@@ -169,31 +168,34 @@ def omit_secret_value_from_checks(check: BaseCheck, check_result: dict[str, Chec
...
-    return censored_definitions
+def omit_secret_value_from_graph_checks(check: BaseGraphCheck, check_result: dict[str, CheckResult] | _CheckResult,
+                                        entity_code_lines: list[tuple[int, str]],
+                                        entity_config: dict[str, Any] | ParameterAttributes | ResourceAttributes,
+                                        resource_attributes_to_omit: ResourceAttributesToOmit | None = None) -> \
+        list[tuple[int, str]]:
+    secrets = set()  # a set, to efficiently avoid duplicates in case the same secret is found in the following conditions
+    censored_code_lines = []
+
+    if check.category == CheckCategories.SECRETS and check_result.get('result') == CheckResult.FAILED:  # type:ignore[comparison-overlap]
```
nimrodkor, line 179: @arielkru CheckCategories.SECRETS is an int, but self.category is defined as `self.category = ""`. Can we straighten this out and avoid the `# type:ignore[comparison-overlap]`?

checkov/terraform/runner.py
```diff
@@ -241,12 +243,18 @@ def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter)
                         referrer_id = self._find_id_for_referrer(tf_path)
                         if referrer_id:
                             resource = f'{referrer_id}.{resource_id}'
+                    entity_config = self.get_graph_resource_entity_config(entity, entity_context)
+                    censored_code_lines = entity_context.get('code_lines') if not resource_attributes_to_omit else \
+                        omit_secret_value_from_graph_checks(check=check, check_result=check_result,
```
nimrodkor, line 248: @arielkru This means we will need to implement it for every framework (i.e. CFN, K8s etc)?
Is there a common solution?

nimrodkor (commented):
Nice @arielkru !

checkov/common/util/secrets.py
```diff
@@ -169,31 +168,34 @@ def omit_secret_value_from_checks(check: BaseCheck, check_result: dict[str, Chec
...
+        secrets.update([str(secret) for key, secret in entity_config.items() if
+                        key.startswith(f'{check.id}_secret')])
```
gruebel, line 181:
```suggestion
        secrets = {
            str(secret) for key, secret in entity_config.items()
            if key.startswith(f'{check.id}_secret')
        }
```
this should do the job without needing the temporal list

checkov/common/util/secrets.py
```diff
@@ -169,31 +168,34 @@ def omit_secret_value_from_checks(check: BaseCheck, check_result: dict[str, Chec
...
+
+    if resource_attributes_to_omit:
+        for resource_type_to_omit in [resource for resource in check.resource_types if resource in resource_attributes_to_omit]:
+            for attribute_to_omit in [attr for attr in resource_attributes_to_omit.get(resource_type_to_omit) if attr in entity_config]:  # type:ignore[union-attr]
```
gruebel, line 185: please use just normal loops here, to create a list comprehension and then iterating over it directly without reusing it somewhere is beneficial performance-wise

checkov/terraform/runner.py
arielkru: @nimrodkor simplified this to `censored_code_lines = omit_secret_value_from_graph_checks(..)` that can be implemented in any framework if needed. Currently we don't have neither the `resource_attributes_to_omit` set for any other framework or secret category graph based checks, so I think it's not relevant in the rest (right?). This is implemented in terraform runner but is actually just for terraform plan graph checks.

nimrodkor (approved):

lirshindalman (approved):

PR #4040, opened 2022-12-11
feat(kubernetes): added deployment name to pod resource id

PelegLi:
**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

## Description

concatenate the Deployment resource's name to the nested Pod's resource id and adding it to the Pod's config in `_parent_resource`

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my feature, policy, or fix is effective and works
- [x] New and existing tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4100, opened 2022-12-18
feat(kubernetes): Support graph edges for nested (related) Pod resources.

metahertz:

## Description
Within a Deployment, pods are defined in the abstract. As they will be dynamically created at runtime based on the direction of the Deployment manifest (and it's ReplicationController). 

The pod spec, nested within a k8s Deployment object, has critical security information which we'd like to write rules against and relate to other objects (in the case of the usecase which created this PR, ensuring Pod's have an associated SecurityPolicy binding, even if they are defined within a Deployment).

This change is the UX-facing second part of #4040 which enabled us to find "parent"/"related" objects within a graph node when created from a Deployment.

Here we edit the Kubernetes runner to ensure we refer to the Real, on-disk Deployment object yaml file when referring to the pod spec, including correct line numbers for the Checkov report output.

Fixes # Internal BCD-292

## New/Edited policies 
- CKV2_K8S_6

### Description
A Pod must have an associated NetworkPolicy. Without a networkpolicy, a Pod may have complete ingress and egress network access within the cluster. Allowing an attacker easy lateral movement or exposing unnecessary open services within the cluster networking fabric.

### Fix
Match a NetworkPolicy object to the Deployment/Pod via labels and selectors:

```
kind: NetworkPolicy
metadata:
  name: test-network-policy
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: internal-proxy
```

## Checklist:

- [NA] I have made corresponding changes to the documentation
- [NA] Any dependent changes have been merged and published in downstream modules


tests/kubernetes/graph/checks/test_checks/EnsureK8SWorkerHandlesRelatedPodsFromDeploymentGraphs.yaml
```diff
@@ -0,0 +1,18 @@
+metadata:
+  id: "CKV2_K8S_6"
+  name: "Minimize the admission of pods which lack an associated NetworkPolicy"
+  # This test is to check functionality of the k8s runner looking up "virtual/related pods" from K8s Deployments for the graph edges. Rather than the policy itself
```
bo156, line 4: If that's a different check from the main one you added, I would give it a different id and name indicating it's a test check, right now it just seems like a copy from the regular one and is not very clear what's the difference (I had to stare at it for a minute).

I have to say though that you could probably use the original one without adding this, so why write them both?

checkov/kubernetes/runner.py
```diff
@@ -251,7 +252,19 @@ def mutate_kubernetes_graph_results(
                 entity_file_path = entity[CustomAttributes.FILE_PATH]
                 entity_file_abs_path = _get_entity_abs_path(root_folder, entity_file_path)
                 entity_id = entity[CustomAttributes.ID]
-                entity_context = self.context[entity_file_path][entity_id]
+                # Deal with nested pods within a deployment.
```
bo156, line 255: I would separate this block to a function with a relevant name, which will simplify the readability of the code.

bo156 (changes requested):

bo156 (approved):

PR #4104, opened 2022-12-19
feat(sca): protecting checkov with try/catch wrapping

itai1357:

## Description
the main goal is catching errors in sensitive places, and handling them so the customer will not get them

in some cases, we can't do anything, so we safely terminate the run with status code 2
in others (mostly in cases we are using the cache), we still can give the customer its accurate data, so we catch the error and continue the scan  

## Checklist:


checkov/common/bridgecrew/vulnerability_scanning/integrations/docker_image_scanning.py
```diff
@@ -30,8 +30,7 @@ async def report_results_async(
             docker_image_name=kwargs["docker_image_name"],
         )
 
-    # can be removed, if image scanning is also using report_results_asyncio
-    # make sure to remove it only after validating it isn't used from the platform
+    # can be removed after it is fully moved to the platform
```
nimrodkor, line 33: @itai1357 can we move it completely?

itai1357: @nimrodkor will be in other PR, and after implementing it in platform

checkov/sca_package/scanner.py
```diff
@@ -86,32 +90,47 @@ async def run_scan_multi(
                     res['repository'] = str(input_paths_as_list[idx])
                     scan_results[idx] = res
 
-        return scan_results
+        # checking whether there are None results that indicates for error. id there is any, we return None
+        scan_results_without_nones: list[dict[str, Any]] = []
```
AdamDev, line 94: Product question, in case we have an error in a single scan result, don't we still want to show all other results?

itai1357: good question. but as I understood, 
The logic here - if the results are partial, we have to terminate the run and exit code should be 2

@nimrodkor @gruebel am I right? or maybe we still want to report the result for the successful files?

nimrodkor: @itai1357 is correct

gruebel (approved):
nice job 🎄

ayajbara (approved):

PR #4119, opened 2022-12-20
feat(sca): changed setup bridgecrew credentials 

ayajbara:

[//]: # "
    # PR Title
    ex.
"

## Description

the new sca package scanning is not working as expected in IDEs, since we blocked uploading the package files that we need to scan and in the new package scanning we need these files so we can scan them in Platform.

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/common/bridgecrew/platform_integration.py
```diff
@@ -257,8 +257,10 @@ def setup_bridgecrew_credentials(
                     region_name=region,
                     config=config,
                 )
-                self.platform_integration_configured = True
-                self.use_s3_integration = True
+                if not is_package_scan:
+                    self.skip_fixes = True  # no need to run fixes on CI integration
```
nimrodkor, line 261: Do we need to run fixes on package scanning in CI integration? I don't think so...

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -257,8 +257,10 @@ def setup_bridgecrew_credentials(
...
+                    self.platform_integration_configured = True
+                    self.use_s3_integration = True
```
nimrodkor, line 263: @ayajbara Why would those 2 be false if this is a package scan?

checkov/sca_package_2/runner.py
```diff
@@ -149,5 +149,14 @@ def upload_package_files(
                     package_files_to_persist.append(FileToPersist(file, os.path.relpath(file, root_folder)))
 
         logging.info(f"{len(package_files_to_persist)} sca package files found.")
+        if package_files_to_persist and bc_integration.bc_source and not bc_integration.bc_source.upload_results:
```
nimrodkor, line 152: What happens is `package_files_to_persist` is not empty, `bc_integration.bc_source` exists and `bc_integration.bc_source.upload_results` is True? We don't upload the package files? Something seems off in this predicate

nimrodkor (changes requested):
@ayajbara The changes to the if statements are not clear, so even if they work - they will be hard to maintain. Let's try to refactor it.

checkov/sca_package_2/runner.py
ayajbara: if upload_results is True, we dont to do anything since we call setup_bridgecrew_credentials function before we run any runner and it will initialize the s3 path and s3 client. and only if upload_results if False and there are package files to scan we will call setup_bridgecrew_credentials function to initialize the s3 path and s3 client for the first time.

checkov/common/bridgecrew/platform_integration.py
ayajbara: the name of the var maybe is not a best choice. 
but it used when we run package scanning and we didnt setup the bridgecrew credentials yet, since we need this setup for the new package scanning, I setup the credentials in the runner and using this var. 
and in this case these variables will be False since we need to block "upload results"

PR #4125, opened 2022-12-21
feat(gha): support on directive in workflow files

maxamel:
This PR adds support for "on" block in GHA graphs, including adding vertices, building definitions from files, and building definitions context.

## Checklist:


checkov/github_actions/utils.py
```diff
@@ -75,18 +75,22 @@ def is_schema_valid(config: dict[str, Any] | list[dict[str, Any]]) -> bool:
...
 
     for file_path in file_paths:
-        result = parse_file(f=file_path)
-        # result should be tuple of dict representing the file payload structure and list of lines of the payload
-        if result is not None:
-            definitions[str(file_path)] = result[0]
-            definitions_raw[str(file_path)] = result[1]
+        str_file_path = str(file_path)
+        process = str_file_path in files_set if files_set else True
+        if process:
```
nimrodkor, line 88: @maxamel Please use a prefix that hints this is a boolean. Also, `process` is about as generic as it should be. Perhaps `should_parse` is a better name?

tests/github_actions/gha/.github/workflows/failed.yaml
```diff
@@ -1,14 +1,15 @@
 name: read-only
 
 on:
-  workflow_dispatch:
+  pull_request:
+    types: [ opened, synchronize, labeled, unlabeled ]
```
nimrodkor, line 5: @maxamel Can we add a test case for this on block?
```
on:
  pull_request:
    types: [ opened, synchronize, labeled, unlabeled ]
    workflow_dispatch:
```

I wonder how it will treat the `workflow_dispatch:` line and want to make sure multiple attributes don't collide

nimrodkor (commented):
Very nice @maxamel !

tests/github_actions/gha/.github/workflows/failed.yaml
maxamel: Added a test. Results look fine IMO

tests/github_actions/gha/.github/workflows/multiple_on_descendants.yaml
```diff
@@ -0,0 +1,27 @@
+name: Handle branches
+on:
+  pull_request:
+    types: [ opened, synchronize, labeled, unlabeled ]
+    workflow_dispatch:
```
nimrodkor, line 5:
```suggestion
  pull_request:
    types: [ opened, synchronize, labeled, unlabeled ]
  workflow_dispatch:
```

gruebel (approved):
nice work 🎄 please also add the new resource to the docs 🙂 https://github.com/bridgecrewio/checkov/blob/main/docs/3.Custom%20Policies/YAML%20Custom%20Policies.md#github-actions

tests/github_actions/gha/.github/workflows/multiple_on_descendants.yaml
maxamel: @nimrodkor this is a different structure than the one previously suggested. With this structure, the definition context will not contain the workflow_dispatch. It is not related to the fact we have multiple blocks under "on", since removing the "pull_request" also does not keep the workflow_dispatch. It's probably related to the fact the workflow_dispatch is empty. Let me know if the expectation is to see the workflow_dispatch in the final definition_context (even in case it's empty).

nimrodkor: This is the correct structure, so highly recommend changing the test file. 
Regarding the test result - I don't mind whether we actually have a "workflow_dispatch" node here, but I do want consistent behavior :). If it is consistent, let's leave it as is

maxamel: cool, changed the file and adjusted the test.

PR #4130, opened 2022-12-21
feat(general): Make code blocks for json check results focused on the relevant part

arielkru:

[//]: # "
    # PR Title
    ex.
"

## Description

github checks were including all the vcs config file as the code block for every check result.
This change makes every github check base class to check if there is a relevant result configuration that is smaller than the whole json file, and make it the result code block and lines range.
example for checkov results after the change:
image
 
## Checklist:


lirshindalman (approved):

checkov/common/runners/object_runner.py
```diff
@@ -203,7 +203,7 @@ def add_python_check_results(
                         bc_check_id=check.bc_id,
                         check_name=check.name,
                         check_result=result,
-                        code_block=self.definitions_raw[file_path][start - 1:end + 1],
+                        code_block=self.definitions_raw[file_path][start - 1:end + 2],
```
marynaKK, line 206: why `end+2`?
do we know for sure that always `len(self.definitions_raw[file_path]) >=  end+2` ?

marynaKK, line 206: why `end+2`?
won't this affect other code blocks too?

checkov/github/base_github_configuration_check.py
```diff
@@ -40,3 +41,14 @@ def resolve_ckv_metadata_conf(conf: dict[str, Any]) -> tuple[dict[str, Any], dic
                 del new_conf[CKV_METADATA]
                 return ckv_metadata, new_conf
         return {}, conf
+
+    def get_result_configuration(
+            self, evaluated_key: str, conf: dict[str, Any]) -> dict[str, Any] | str | list[str | dict[str, Any]]:
+        # if the evaluated key points to a key within an object, the result config should be its parent, for context
+        # example 1 - key: `object.nested_object.key` -> json_path should be `$.object.nested_object`
```
gruebel, line 48: is there a case, where even the `nested_object` doesn't exist?

arielkru: If it doesn't exist then 'evaluated_conf' (line 52) would be empty and the function would just return the entire conf.
But maybe it's better in that case to return the `object` is that your suggestion? @gruebel

checkov/common/runners/object_runner.py
arielkru: It would affect all the yaml and json code blocks (that inherit this object runner).
I did this cause I noticed that when we do +1 with the way the lines are parsed we don't include in the results the closing bracket (which is usually a line below), so this would include always the next line.
But maybe in some yaml configs the next line is not related so I agree it might not be wanted, i'll revert it.
We could change the json parsing but im not sure its worth it, the current parsing is enough just looks a bit weird
great comment

marynaKK (approved):

nimrodkor (changes requested):
This should leverage the `file_line_range` of the record, not build a new flow @arielkru

checkov/github/base_github_configuration_check.py
gruebel: yeah correct, that was my thought 🙂 but probably the whole `conf` is also fine

arielkru:
@nimrodkor But it does
the record's `file_line_range` is set in `checkov/common/runners/object_runner.py`, based on `start` and `end`:
https://github.com/bridgecrewio/checkov/blob/ea90cfc866f9038bce27f3b4b63275a0a793f617/checkov/common/runners/object_runner.py#L208
and `start` and `end` get the correct focused values only if there's a `result_config`:
https://github.com/bridgecrewio/checkov/blob/ea90cfc866f9038bce27f3b4b63275a0a793f617/checkov/common/runners/object_runner.py#L176-L177
and `result_config` is set for the right lines when the check returns a tuple of `CheckResult, result_config`

I could add a test to check that, but Im not sure its really needed cause it just gets into an existing flow.
Attaching a screenshot of a report example for a focused resource after this change:
image
And this is its printed result:
image

  
Is anything missing?

checkov/github/base_github_negative_branch_security.py
```diff
@@ -27,9 +27,12 @@ def __init__(self, id: str, name: str, missing_attribute_result: CheckResult = C
         )
         self.missing_attribute_result = missing_attribute_result
 
-    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> CheckResult:  # type:ignore[override]
+    def scan_entity_conf(  # type:ignore[override]
+            self, conf: dict[str, Any], entity_type: str) -> \
+            CheckResult | tuple[CheckResult, dict[str, Any] | str | list[str | dict[str, Any]]]:
```
nimrodkor, line 32:
```suggestion
            tuple[CheckResult, dict[str, Any]]
```

checkov/common/checks/base_check.py
```diff
@@ -84,7 +84,8 @@ def run(
 
     @multi_signature()
     @abstractmethod
-    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> CheckResult | tuple[CheckResult, dict[str, Any]]:
+    def scan_entity_conf(self, conf: dict[str, Any], entity_type: str) -> \
+            CheckResult | tuple[CheckResult, dict[str, Any] | str | list[str | dict[str, Any]]]:
```
nimrodkor, line 88: @arielkru This return type is just impossible. No idea what to expect here TBH. This is a bad sign

checkov/github/base_github_negative_branch_security.py
```diff
@@ -27,9 +27,12 @@ def __init__(self, id: str, name: str, missing_attribute_result: CheckResult = C
...
         if branch_security_schema.validate(conf):
             evaluated_key = self.get_evaluated_keys()[0].replace("/", ".")
+            evaluated_conf = self.get_result_configuration(evaluated_key, conf)
```
nimrodkor, line 35: @arielkru The fact you set the same change over and over is another bad sign

nimrodkor (commented):
I understand this is the current implementation. I'm challenging you to think of something which is more centralized, less hacky, and more consistent from schema, behavior and implementation perspective

checkov/github/base_github_configuration_check.py
arielkru: @gruebel implemented your suggestion 👍

nimrodkor (approved):

PR #4133, opened 2022-12-22
feat(sca): run old package scanning for IDE scan 

ayajbara:


Saarett:
@ayajbara Great job! Do we have some tests to make sure it works as expected?

checkov/main.py
```diff
@@ -201,6 +201,24 @@ def run(banner: str = checkov_banner, argv: List[str] = sys.argv[1:]) -> Optiona
...
+
+    if config.list:
+        # This speeds up execution by not setting up upload credentials (since we won't upload anything anyways)
+        logger.debug('Using --list; setting source to DISABLED')
+        source = SourceTypes[BCSourceType.DISABLED]
+
+    if CHECKOV_RUN_SCA_PACKAGE_SCAN_V2 and source.upload_results:
+        del DEFAULT_RUNNERS[-1]
+    else:
+        del DEFAULT_RUNNERS[-2]
```
Saarett, line 220: Rather than deleting the value from the list, perhaps add it here, e.g.
DEFAULT_RUNNERS.append(sca_package_runner())
reason is, in the future we might add more runners, and then your approach won't work because you'll delete the new runner

checkov/main.py
```diff
@@ -79,8 +79,7 @@
 logger = logging.getLogger(__name__)
 checkov_runners = [value for attr, value in CheckType.__dict__.items() if not attr.startswith("__")]
 
-
-DEFAULT_RUNNERS = (
+DEFAULT_RUNNERS = [
```
gruebel, line 82: can you add a comment here, that the `sca_package` runner is added during the run method

gruebel (approved):

checkov/main.py
ayajbara: yeah sure

Saarett (approved):

PR #4138, opened 2022-12-22
fix(terraform): handle entity_configs that miss paths from definition_paths

arielkru:

[//]: # "
    # PR Title
    # Allowed prefixs: argo|arm|azure|bicep|bitbucket|circleci|cloudformation|dockerfile|github|gha|gitlab|helm|kubernetes|kustomize|openapi|sca|secrets|serverless|terraform|general|graph|terraform_plan    
"

## Description

This change is supposed to avoid errors in case of broken entity configs / definition paths (based on a real case)

## Checklist:


checkov/terraform/runner.py
```diff
@@ -631,5 +631,5 @@ def get_graph_resource_entity_config(self, entity, entity_context):
         definition_path = entity_context.get('definition_path', [])
         entity_config = entity['config_']
         for path in definition_path:
-            entity_config = entity_config[path]
+            entity_config = entity_config.get(path, entity_config)
```
achiar99, line 634: @arielkru i think we need to understand why this key doesnt exist. if its not related to secrets ill take this

checkov/terraform/runner.py
```diff
@@ -629,7 +629,10 @@ def extract_images(
 
     def get_graph_resource_entity_config(self, entity, entity_context):
         definition_path = entity_context.get('definition_path', [])
-        entity_config = entity['config_']
+        entity_config = entity.get('config_', {})
```
gruebel, line 632:
```suggestion
        entity_config = entity.get(CustomAttributes.CONFIG, {})
```

checkov/terraform/runner.py
gruebel: also curious about it, seems more like a bug

arielkru:
Fixed by @YaaraVerner 's https://github.com/bridgecrewio/checkov/pull/4146

PR #4166, opened 2022-12-28
chore: add missing type hints under cloudformation

### LINKED ISSUE #3661: Add type hints to cloudformation
**Describe the issue**

- enable mypy for `checkov/cloudformation`
- add missing type hints

gruebel:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

- finished up type hinting `cloudformation`
- `cloudformation/checks` is skipped intentionally, because I still have to find a solution for our decorator usage

Fixes #3661 

## Checklist:

### DIFF
--- a/checkov/cloudformation/graph_builder/graph_components/blocks.py
+++ b/checkov/cloudformation/graph_builder/graph_components/blocks.py
@@ -36,9 +36,9 @@ class CloudformationBlock(Block):
     def update_attribute(
         self, attribute_key: str,
         attribute_value: Any,
-        change_origin_id: int,
+        change_origin_id: int | None,
         previous_breadcrumbs: List[BreadcrumbMetadata],
-        attribute_at_dest: str,
+        attribute_at_dest: str | None,
         transform_step: bool = False,
     ) -> None:
         super().update_attribute(
@@ -82,7 +82,7 @@ class CloudformationBlock(Block):
     @staticmethod
     def _should_add_previous_breadcrumbs(change_origin_id: Optional[int],
                                          previous_breadcrumbs: List[BreadcrumbMetadata],
-                                         attribute_at_dest: Optional[str]):
+                                         attribute_at_dest: Optional[str]) -> bool:
         return (
             change_origin_id is not None
             and attribute_at_dest is not None
@@ -90,5 +90,5 @@ class CloudformationBlock(Block):
         )
 
     @staticmethod
-    def _should_set_changed_attributes(change_origin_id: Optional[int], attribute_at_dest: Optional[str]):
+    def _should_set_changed_attributes(change_origin_id: Optional[int], attribute_at_dest: Optional[str]) -> bool:
         return change_origin_id is not None and attribute_at_dest is not None
--- a/checkov/cloudformation/graph_builder/graph_components/generic_resource_encryption.py
+++ b/checkov/cloudformation/graph_builder/graph_components/generic_resource_encryption.py
@@ -1,13 +1,15 @@
-from typing import Dict, List, Union, Any
+from __future__ import annotations
+
 from checkov.common.graph.graph_builder import EncryptionTypes
 from checkov.common.graph.graph_builder.graph_components.generic_resource_encryption_base import GenericResourceEncryptionBase
+from checkov.common.util.data_structures_utils import get_empty_list_str
 
 
 class GenericResourceEncryption(GenericResourceEncryptionBase):
     def __init__(
         self,
         resource_type: str,
-        attribute_values_map: Dict[str, Union[List[bool], List[str]]],
+        attribute_values_map: dict[str, list[bool] | list[str]],
         enabled_by_default: bool = False,
     ) -> None:
         super().__init__(resource_type,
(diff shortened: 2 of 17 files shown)


checkov/cloudformation/graph_builder/graph_components/generic_resource_encryption.py
```diff
@@ -1,13 +1,15 @@
...
 from checkov.common.graph.graph_builder.graph_components.generic_resource_encryption_base import GenericResourceEncryptionBase
+from checkov.common.util.data_structures_utils import get_empty_list_str
 
 
 class GenericResourceEncryption(GenericResourceEncryptionBase):
     def __init__(
         self,
         resource_type: str,
-        attribute_values_map: Dict[str, Union[List[bool], List[str]]],
+        attribute_values_map: dict[str, list[bool] | list[str]],
```
bo156, line 12: Can you teach me what's the difference here?
I know they both work, I'm just not sure why it should be important to change it to the second option

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
```diff
@@ -26,13 +37,13 @@ class CloudformationVariableRenderer(VariableRenderer):
 
     def __init__(self, local_graph: "CloudformationLocalGraph") -> None:
         super().__init__(local_graph)
-        self.edge_evaluation_methods = {
+        self.edge_evaluation_methods: "dict[str, Callable[[Any, dict[str, Any]], tuple[str | None, str | None]]]" = {
```
bo156, line 40: perhaps create a `TypeVar` here and move to the top of the module?

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
```diff
@@ -264,14 +281,15 @@ def _evaluate_findinmap_connection(value: List[str], dest_vertex_attributes: Dic
...
+    def _evaluate_getatt_connection(
+        value: Any, dest_vertex_attributes: Dict[str, Any]
+    ) -> tuple[str | None, str | None]:
         # value = [ "logicalNameOfResource", "attributeName" ]
         try:
             if isinstance(value, list) and len(value) == 2:
                 resource_name = value[0]
                 attribute_at_dest = value[1]
-                dest_name = dest_vertex_attributes.get(CustomAttributes.BLOCK_NAME).split('.')[-1]
+                dest_name = dest_vertex_attributes[CustomAttributes.BLOCK_NAME].split('.')[-1]
```
bo156, line 292: Don't we want to check what happens if we there is no such value in the dict?

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
```diff
@@ -448,13 +469,13 @@ def _handle_edge_list_evaluation_functions(self, edge_list: List[Edge], origin_v
 
             if len(evaluated_edges) == len(edge_list):
                 # succeeded in evaluation all of the edges
-                for evaluated_value in evaluated_edges:
+                for evaluated_edge in evaluated_edges:
```
bo156, line 472: nice :)

bo156 (approved):

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
gruebel: it was not checked before and would have resulted in a `AttributeError` of `None` having no attribute `split`
if this key is not there, then we have a general problem, which is better reflected through directly accessing it.

checkov/cloudformation/graph_builder/graph_components/blocks.py
```diff
@@ -36,9 +36,9 @@ def __init__(
     def update_attribute(
         self, attribute_key: str,
         attribute_value: Any,
-        change_origin_id: int,
+        change_origin_id: int | None,
```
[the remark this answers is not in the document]
gruebel: using `|` was introduced in Python 3.10, I think. And is the preferred way to use instead of `Optional`, but at the end they both do the same. You just don't need to import it 🙂

checkov/cloudformation/graph_builder/graph_components/generic_resource_encryption.py
gruebel: as mentioned above it is the preferred way by `typeshed` https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#conventions

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
gruebel: I actually used a `TypeAlias`, that's what you mean, right?, initially and changed that line so often 😄 sure , can bring it back

checkov/cloudformation/graph_builder/variable_rendering/renderer.py
gruebel: I know great change 😄

rotemavni (approved):

PR #4175, opened 2023-01-01
feat(gha): add support for gha existing graph

maxamel:
This PR adds support in checkov for gha existing graph.
The graph needs to be provided from the outside and it will be used to run checks instead of building locally.

## Checklist:


checkov/bicep/runner.py
```diff
@@ -68,7 +68,7 @@ def __init__(
 
         self.context: dict[str, dict[str, Any]] = {}
         self.definitions: dict[Path, BicepJson] = {}  # type:ignore[assignment]  # need to check, how to support subclass differences
-        self.definitions_raw: dict[Path, list[tuple[int, str]]] = {}
+        self.definitions_raw: dict[Path, list[tuple[int, str]]] = {}    # type:ignore[assignment]
```
nimrodkor, line 71: @maxamel Why is this ignore needed?

checkov/common/runners/object_runner.py
```diff
@@ -118,33 +118,43 @@ def run(
...
+        else:
+            logging.info("Going to use existing graph")
+            if self.check_type == CheckType.GITHUB_ACTIONS and isinstance(self.definitions, dict):
+                # populate gha metadata dict
+                for key, definition in self.definitions.items():
+                    if isinstance(definition, dict):
+                        workflow_name = definition.get('name', '')
+                        triggers = self._get_triggers(definition)
+                        jobs = self._get_jobs(definition)
+                        self.map_file_path_to_gha_metadata_dict[key] = {"triggers": triggers, "workflow_name": workflow_name, "jobs": jobs}
```
nimrodkor, line 152: @maxamel Please extract to the child github actions runner, and have an abstract method / method that does nothing in the parent class

checkov/bicep/runner.py
maxamel: Because the parent class now has this field as well, and it's defined differently in each child class. 
I think it's the same case as the above line with `self.definitions`, and it's ignored there for the same reason.

marynaKK (approved):

checkov/github_actions/runner.py
```diff
@@ -144,3 +144,14 @@ def extract_images(
             images.extend(manager.extract_images_from_workflow())
 
         return images
+
+    def populate_metadata_dict(self) -> None:
+        if self.check_type == CheckType.GITHUB_ACTIONS and isinstance(self.definitions, dict):
```
nimrodkor, line 149: This is the GHA runner...
```suggestion
        if isinstance(self.definitions, dict):
```

nimrodkor (approved):

PR #4193, opened 2023-01-04
feat(general): Add resource attributes to omit arg

pazbechor:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Adding the option to supply Checkov a config file (JSON format) that has resource to attributes. 
By using this config, one can mask desired values from is scanning (for example: Secret values in the config) 
*

## Checklist:


checkov/common/util/secrets.py
```diff
@@ -139,21 +140,29 @@ def omit_secret_value_from_line(secret: str | None, line_text: str) -> str:
...
         secrets.update([str(secret) for key, secret in entity_config.items() if
                         key.startswith(f'{check.id}_secret')])
 
-    if resource_attributes_to_omit and check.entity_type in resource_attributes_to_omit:
-        for attribute_to_omit in [attr for attr in resource_attributes_to_omit.get(check.entity_type) if attr in entity_config]:  # type:ignore[union-attr]
-            secret = entity_config.get(attribute_to_omit)
+    if resource_attributes_to_omit:
+        universal_mask = resource_attributes_to_omit.get(RESOURCE_ATTRIBUTES_TO_OMIT_UNIVERSAL_MASK, set())
+        resource_masks = resource_attributes_to_omit.get(check.entity_type, set())
+        resource_masks.update(universal_mask)
```
arielkru, line 161:
```suggestion
        resource_masks = resource_attributes_to_omit.get(check.entity_type, set())
        if RESOURCE_ATTRIBUTES_TO_OMIT_UNIVERSAL_MASK in resource_attributes_to_omit:
            resource_masks.update(resource_attributes_to_omit[RESOURCE_ATTRIBUTES_TO_OMIT_UNIVERSAL_MASK])
```

As most of the runs won't have the RESOURCE_ATTRIBUTES_TO_OMIT_UNIVERSAL_MASK (it's a specific opt in case by users), I think its better to avoid the set update + variable saving when its not relevant

checkov/common/util/secrets.py
```diff
@@ -183,13 +195,21 @@ def omit_secret_value_from_graph_checks(check: BaseGraphCheck, check_result: dic
         }
 
     if resource_attributes_to_omit:
+        # Universal mask ('*') might exist in resource_attributes_to_omit. If it does exist, we need to mask the all
```
arielkru, line 198:
```suggestion
        # Universal mask ('*') might exist in resource_attributes_to_omit. If it does exist, we need to mask all the
```

checkov/common/util/secrets.py
```diff
@@ -183,13 +195,21 @@ def omit_secret_value_from_graph_checks(check: BaseGraphCheck, check_result: dic
...
         for resource in check.resource_types:
-            if resource in resource_attributes_to_omit:
-                for attribute in resource_attributes_to_omit.get(resource):  # type:ignore[union-attr]
-                    if attribute in entity_config:
-                        secret = entity_config.get(attribute)
-                        if isinstance(secret, list) and secret:
-                            secrets.add(secret[0])
+            resource_masks = set(resource_attributes_to_omit.get(resource, set()))
+            # resource_masks should contain all mask rules that should apply on this resource
+            resource_masks.update(universal_mask)
```
arielkru, line 204: Variable `universal_mask` saving and the update function can be skipped here as well for most scans if we first check if univeral mask exists like in the previous suggestion

checkov/common/util/secrets.py
```diff
@@ -183,13 +195,21 @@ def omit_secret_value_from_graph_checks(check: BaseGraphCheck, check_result: dic
...
+            if not resource_masks:
+                continue
+            # If entity is one that should be masked, we add it the value to secrets
+            for entity in entity_config.keys():
```
arielkru, line 208: we use `entity` to describe resources (the wrapping hierarchy), so I'd use here a different name like `attribute` which is more accurate to describe the child hierarchy of the entity

README.md
```diff
@@ -294,8 +294,26 @@ Run secrets scanning on all files in MyDirectory. Skip CKV_SECRET_6 check on jso
...
+ { "azurerm_key_vault_secret": ["value"]}
+```
+ - config file that contains (detailed below) will cause the {RESOURCE} that has value attribute to be masked.
+ ```sh
+ { "*": ["value"]}
+```
+Usage example:
+```sh
+checkov -d --framework terraform_plan --resource-attr-to-omit PATH_TO_FILE.json
+```
```
arielkru, line 311: I think the intention was to add the config params to the existing checkov config file, so we don't require users to create new files just for this.
So the existing config file could now be extended to look like this:
```
framework:
- all
secrets-scan-file-type: []
summary-position: top
mask:
- password_variable                # this is a univeral mask
- resource_type/attribute        # this is a specific resource attribute mask
```
Unless these specs were changed, the config structure should be changed so it doesnt require a new file

arielkru (commented):
Very nice!
I'm not sure about the config file implementation, worth getting @tsmithv11 review on this.
And I added some small comments

checkov/common/util/secrets.py
pazbechor: I don't totally under stand :| 
I get both universal mask & resource mask - resource mask with universal mask (set update)
in case universal mask is not in the resource_attributes_to_omit it will update set with the empty set (o(1))
@arielkru

gruebel (commented):
I'm actually not sure, if should allow passing '*' for resource types, also the these are not all runners and some runners have both Python checks and graph cehcks.

checkov/terraform/plan_runner.py
```diff
@@ -146,22 +147,33 @@ def run(
...
         return report
 
-    def _get_graph_report(self, root_folder: str, runner_filter: RunnerFilter,
-                          tf_local_graph: Optional[TerraformLocalGraph],
-                          resource_attributes_to_omit: ResourceAttributesToOmit) -> Report:
+    @staticmethod
+    def _extend_resource_attributes_to_omit(runner_filter: RunnerFilter):
+        for k, v in RESOURCE_ATTRIBUTES_TO_OMIT.items():
+            # It's ok as runner_filter is ALWAYS default dict with set() as value
+            runner_filter.resource_attr_to_omit[k].update(v)
```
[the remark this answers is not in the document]
pazbechor: what do @tsmithv11  think ?

checkov/common/util/secrets.py
arielkru: @pazbechor I agree it's o(1) so no big problem here, just thought of a tiny performance improvement so it works either way.
Let's say 90% of the runs won't have a global attribute, so as the code is now, these 3 instructions will always happen:
1. get `'*'` from`resource_attributes_to_omit`
2. save an empty set to `universal_mask` variable
3. update `resource_masks` with an empty set

So for 90% of the cases (just a rough assumption) the last 2 instructions are redundant and we can skip them by just checking first #1 and if its None, don't do the other 2.
But again, just a minor improvement

checkov/common/util/secrets.py
arielkru: same

checkov/common/util/secrets.py
pazbechor: we talked about it :)

checkov/terraform/plan_runner.py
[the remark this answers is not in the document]
pazbechor: Talked with @tsmithv11 - he decided to extend it

README.md
```diff
@@ -294,7 +294,17 @@ Run secrets scanning on all files in MyDirectory. Skip CKV_SECRET_6 check on jso
...
-
+One can mask values from scanning results by supplying a configuration file (using --config-file flag) with mask entry.
+The masking can apply on resource & value (or multiple values, seperated with a comma). 
+Examples:
+```sh
+mask:
+- aws_instance:user_data
+- azurerm_key_vault_secret:admin_password,user_passwords
+```
+In the example above, the following resource types will be masked:
```
arielkru, line 305:
```suggestion
In the example above, the following values will be masked:
```

arielkru (approved):
🚀

gruebel (approved):
🍻

PR #4211, opened 2023-01-08
fix(terraform_plan): fix module resources ids

YaaraVerner:

## Description
use the address as the resource id if the `CHECKOV_ENABLE_NESTED_MODULES` env variable is enabled.
add the last referred module (if it exists) to the resource id if the `CHECKOV_ENABLE_NESTED_MODULES` env variable is off.
use the old plan resource id for plan enrichment from terraform dir (without deep analysis)

## Checklist:


checkov/common/output/report.py
```diff
@@ -593,6 +595,13 @@ def handle_skipped_checks(
                 report.failed_checks.remove(record)
         return report
 
+    @staticmethod
+    def get_plan_resource_raw_id(resource_id: str) -> str:
+        resource_raw_id = ".".join(resource_id.split(".")[-2:])
```
bo156, line 600: didn't you did something just like that somewhere else?

checkov/terraform/plan_utils.py
```diff
@@ -102,3 +102,7 @@ def get_entity_context(definitions, definitions_raw, definition_path, full_file_
                 entity_context['address'] = resource_defintion[TF_PLAN_RESOURCE_ADDRESS]
                 return entity_context
     return entity_context
+
+
+def get_resource_id_without_nested_modules(address: str) -> str:
+    return ".".join(address.split(".")[-4:])
```
bo156, line 108: the random numbers are a bit wierd for anyone not familiar with the build, can you use a const?

checkov/common/output/report.py
```diff
@@ -593,6 +595,13 @@ def handle_skipped_checks(
...
+        if '[' in resource_raw_id:
```
achiar99, line 601: i think it will be better to wrap with the env var

bo156 (approved):

achiar99 (approved):

checkov/common/output/report.py
gruebel: a doc string with an example would be great 🙂

YaaraVerner: no

checkov/terraform/plan_utils.py
gruebel: also here a doc string with an example, then it more clear, what actually is happening

checkov/terraform/runner.py
```diff
@@ -241,6 +242,11 @@ def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter,
                             if referrer_id:
                                 resource = f'{referrer_id}.{resource_id}'
                         definition_context_file_path = get_tf_definition_key_from_module_dependency(full_file_path, module_dependency, module_dependency_num)
+                    elif entity.get(CustomAttributes.TF_RESOURCE_ADDRESS) and entity.get(CustomAttributes.TF_RESOURCE_ADDRESS) != resource_id:
+                        # for plan resources
+                        resource = entity.get(CustomAttributes.TF_RESOURCE_ADDRESS)
```
gruebel, line 247:
```suggestion
                        resource = entity[CustomAttributes.TF_RESOURCE_ADDRESS]
```

gruebel (commented):
nice work 💪

gruebel (approved):

PR #3962, opened 2022-12-01
feat(terraform): Adding yaml based build time policies for corresponding PC run time policies

ssiddardha:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license

This PR has 6 Yaml - based checks with detailed description below

Checkov Title: Ensure AWS Config recorder is enabled

PC Policy ID - ca5c571e-6930-44af-a47b-ebde3ac20ca5

PC Policy Title - AWS Config recording is disabled

[...]

PR #4228, opened 2023-01-10
fix(terraform): Improve CKV2_AWS_46 (CloudFront with S3 and OAI)

ankon:

## Description

The check CKV2_AWS_46 seems to produce false positives, and was hard to understand.

See also https://github.com/bridgecrewio/checkov/pull/3962

## Checklist:


checkov/terraform/checks/graph_checks/aws/CloudFrontS3OriginConfigWithOAI.yaml
```diff
@@ -19,12 +19,11 @@ definition:
     - cond_type: attribute
       resource_types:
         - aws_cloudfront_distribution
-      attribute: "origin.*.s3_origin_config"
+      attribute: "origin.*.s3_origin_config.origin_access_identity"
       operator: "exists"
```
ankon, line 23: The more I think about it: this probably should check "not empty", not just "exists". 

Same below. 

WDYT?

gruebel: you can change it to a 'not empty', but a not empty `origin_access_identity` is a broken configuration, but correct me, if I'm wrong.

gruebel (commented):
thanks, changes look good, just have a general question 🙂

Issue #4230, opened 2023-01-11
Checkov Taking 10x longer to check code since 2.2.252


darren-johnson:
**Describe the issue**
We have been using Checkov for over a year now with very few issues. However yesterday when the docker container version changed from 2.2.252 the checks went from 3 mins in our pipeline to over 30 mins! I am able to repeat this locally too.

**Examples**
Please share an example code sample (in the IaC of your choice) + the expected outcomes.

**Version (please complete the following information):**
 - Checkov Version [e.g. 22]

**Additional context**

gruebel:
hey @darren-johnson thanks for reaching out. We enabled a new feature yesterday and colleague of my is working on a performance improvement. I will share the version number with the fix, when it is released.

gruebel:
@darren-johnson the fix was merged and is included in version ` >= 2.2.256`. Would be great, if you can try it out again 🙂 
if it didn't get any better, then it would be great, if you could share more information about the flags you use and what kind of IaC you are checking against.

YaaraVerner:
hey @darren-johnson, the fix is in version `2.2.256`, please let us know if it resolved the issue.

darren-johnson:
Just tested and it seems to be back to 3 minutes again.  Thanks for your help!

PR #4231, opened 2023-01-11
fix(terraform): improve cross-variable edges performance

YaaraVerner:

## Description
when adding a new edge, check if a new cross_variable_edge is exist in the origin out_edges instead of all edges

Relates #4230 

## Checklist:

### DIFF
--- a/checkov/terraform/graph_builder/local_graph.py
+++ b/checkov/terraform/graph_builder/local_graph.py
@@ -291,11 +291,13 @@ class TerraformLocalGraph(LocalGraph[TerraformBlock]):
             return False
         edge = Edge(origin_vertex_index, dest_vertex_index, label)
         if cross_variable_edges:
-            if edge in self.edges or self.vertices[edge.dest].block_type != BlockType.RESOURCE or \
-                    self.vertices[edge.origin].block_type != BlockType.RESOURCE:
+            if self.vertices[dest_vertex_index].block_type != BlockType.RESOURCE or \
+                    self.vertices[origin_vertex_index].block_type != BlockType.RESOURCE:
+                return False
+            if edge in self.out_edges[origin_vertex_index]:
                 return False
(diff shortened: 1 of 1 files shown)


checkov/terraform/graph_builder/local_graph.py
```diff
@@ -291,11 +291,13 @@ def _create_edge(self, origin_vertex_index: int, dest_vertex_index: int, label:
...
-            if edge in self.edges or self.vertices[edge.dest].block_type != BlockType.RESOURCE or \
-                    self.vertices[edge.origin].block_type != BlockType.RESOURCE:
+            if self.vertices[dest_vertex_index].block_type != BlockType.RESOURCE or \
+                    self.vertices[origin_vertex_index].block_type != BlockType.RESOURCE:
+                return False
+            if edge in self.out_edges[origin_vertex_index]:
                 return False
             edge.label = CROSS_VARIABLE_EDGE_PREFIX + edge.label
-            if edge in self.edges:
+            if edge in self.out_edges[origin_vertex_index]:
```
gruebel, line 300: isn't this redundant, because you check it on line 297 already?

YaaraVerner: no, because it's with a different label (with the cross variable)

gruebel: ah got it, yeah we compare the origin ID + dest ID + label for equality

gruebel (approved):
🥇

checkov/terraform/graph_builder/local_graph.py
bo156: It is a bit misleading though, maybe create s a new variable and check it?
For example `edge_with_cross_variable_label_prefix`

bo156 (approved):

PR #4234, opened 2023-01-11
feat(azure): Add image referencer in azure pipelines

lirshindalman:

add Azure Pipelines Image Referencer 

## Checklist:


checkov/common/images/image_referencer_manager.py
```diff
@@ -0,0 +1,20 @@
...
+
+from checkov.common.images.image_referencer_provider import Provider
+
+from typing import Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from checkov.common.images.image_referencer import Image
+
+
+class ImageReferencerManager:
```
gruebel, line 11:
```suggestion
class WorkflowImageReferencerManager:
```
this IR manager is specific to workflow files, so let's also give him an appropriate name 🙂

checkov/common/images/image_referencer_provider.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import Any
+
+from checkov.common.images.image_referencer import Image
+
+
+class Provider:
```
gruebel, line 9:
```suggestion
class WorkflowImageReferencerProvider:
```
same here

checkov/azure_pipelines/image_referencer/provider.py
```diff
@@ -0,0 +1,56 @@
...
+        for key, sub_job in job.items():
+            if key == self.supported_keys:
+                images.append(self.crete_image(sub_job, start_line, end_line))
+            elif isinstance(sub_job, dict):
+                images.extend(self.extract_images_from_dict(sub_job))
+            elif isinstance(sub_job, list):
+                images.extend(self.extract_images_from_list(sub_job))
+        return images
+
+    def crete_image(self, container: dict[str, Any] | str, start_line: int, end_line: int) -> Image:
```
gruebel, line 41:
```suggestion
    def create_image(self, container: dict[str, Any] | str, start_line: int, end_line: int) -> Image:
```

checkov/azure_pipelines/common/resource_id_utils.py
```diff
@@ -0,0 +1,40 @@
...
+            block_name = block_to_inspect.get('displayName',
+                                              block_to_inspect.get('name',
+                                                                   block_to_inspect.get('job',
+                                                                                        block_to_inspect.get('stage',
+                                                                                                             False))))
+            inspected_key = f'{inspected_key}({block_name})' if block_name else inspected_key
+            if block_to_inspect[START_LINE] == start_line:
+                return inspected_key
+            return generate_resource_key_recursive(start_line, end_line, block_to_inspect, resource_key=inspected_key)
+        return None
```
gruebel, line 25: it would be better to move the function to the root file level instead of using an inline function. you will need to pass `start_line` and `end_line` as normal as normal parameter. Currently the function is recreated each time you call `generate_resource_key_recursive()`

checkov/common/images/image_referencer_manager.py
```diff
@@ -0,0 +1,20 @@
...
+
```
gruebel, line 12: you can define here `__slots__` so you don't need to do it in each of the subclasses.

checkov/azure_pipelines/image_referencer/provider.py
```diff
@@ -0,0 +1,56 @@
...
+        images = []
+        start_line, end_line = AzurePipelinesProvider._get_start_end_lines(job)
+        for key, sub_job in job.items():
+            if key == self.supported_keys:
+                images.append(self.crete_image(sub_job, start_line, end_line))
+            elif isinstance(sub_job, dict):
+                images.extend(self.extract_images_from_dict(sub_job))
+            elif isinstance(sub_job, list):
+                images.extend(self.extract_images_from_list(sub_job))
+        return images
```
gruebel, line 39: the biggest drawback of this approach is that we basically do a deep search for the key `container`. Depending on the size of the file, this can be rather time consuming. Wouldn't it be much faster just to directly search under the keys, where we expect a `container` reference?
ex
```python
for job in self.workflow_config.get("jobs"):
    container = job.get("container")
    if container:
        images.append(self.create_image(container, start_line, end_line))
```
and for `stages` something similar.

gruebel (commented):
nice work 🍻 added a couple of comments

checkov/azure_pipelines/image_referencer/provider.py
```diff
@@ -0,0 +1,56 @@
...
+            image_name = container
+        elif isinstance(container, dict):
+            if 'image' in container:
+                image_name = container['image']
+        return Image(
+            file_path=self.file_path,
+            name=image_name,
+            start_line=start_line,
+            end_line=end_line,
+            related_resource_id=generate_resource_key_recursive(file_conf=self.workflow_config,
```
Eliran-Turgeman, line 53: Let's check if `image_name` is not empty, and the generated resource key is not `None`

tests/azure_pipelines/image_referencer/test_runner.py
```diff
@@ -0,0 +1,54 @@
...
+from checkov.common.bridgecrew.bc_source import get_source_type
+from checkov.azure_pipelines.runner import Runner
+
+from checkov.runner_filter import RunnerFilter
+from pytest_mock import MockerFixture
+
+RESOURCES_PATH = Path(__file__).parent / "resources/single_image"
+
+
+def test_gitlab_workflow(mocker: MockerFixture, image_cached_result, license_statuses_result):
```
Eliran-Turgeman, line 14:
```suggestion
def test_azure_pipelines_workflow(mocker: MockerFixture, image_cached_result, license_statuses_result):
```

checkov/azure_pipelines/image_referencer/provider.py
Eliran-Turgeman: I think all the relevant paths to an image are listed here https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops

You can also use jmespath to traverse the json with keywords - example in `checkov/circleci_pipelines/image_referencer/provider.py`

lirshindalman: the problem is that i can have 
     'stage'-> 'job'->'steps'->'task'->'container' 
or  'job' -> 'container' 
or 'job' -> 'container' ->'image'
and not directly in 'job' or in 'stage'.
@gruebel

lirshindalman: the problem is that i can have
'stage'-> 'job'->'steps'->'task'->'container'
or 'job' -> 'container'
or 'job' -> 'container' ->'image'
and not directly in 'job' or in 'stage'.

Eliran-Turgeman (approved):

checkov/common/images/image_referencer_provider.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class WorkflowImageReferencerProvider:
+    def __init__(self, workflow_config: dict[str, Any], file_path: str):
+        self.workflow_config = workflow_config
+        self.file_path = file_path
+
+    @staticmethod
+    def _get_start_end_lines(entity: dict[str, Any]) -> tuple[int, int]:
+        return entity.get('__startline__', 0), entity.get('__endline__', 0)
```
gruebel, line 16: can you use the consts `START_LINE` and `END_LINE`. check that you reference the right ones 😄

gruebel (approved):

PR #4237, opened 2023-01-11
fix(terraform): default value for CKV_AZURE_5

### LINKED ISSUE #4227: False positive for check CKV_AZURE_5: "Ensure RBAC is enabled on AKS clusters"
**Describe the issue**
The check CKV_AZURE_5 for terraform resource `azurerm_kubernetes_cluster` ensures that RBAC is enabled in the kubernetes cluster.
Depending on how the `role_based_access_control_enabled` property is set, the check result is exact or not :
- `role_based_access_control_enabled = true`: the check passes. It's ok.
- `role_based_access_control_enabled = false`: the check fails. It's ok.
- `role_based_access_control_enabled` not defined : check fails. It's NOT ok as default value of this property is `true` (see https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#role_based_access_control_enabled)

**Examples**
This example will fails but it shouldn't:
```
resource "azurerm_resource_group" "foo" {
  name     = "foo"
  location = "West Europe"
}

resource "azurerm_kubernetes_cluster" "foo" {
  name                = "foo"
  resource_group_name = azurerm_resource_group.foo.name
  location            = azurerm_resource_group.foo.location
  dns_prefix          = "foo"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }

  identity {
    type = "SystemAssigned"
  }

  # role_based_access_control_enabled = true
}
```

 - Checkov Version : `2.2.252` (latest docker image)

**Additional context**
The problem is in this source file : https://github.com/bridgecrewio/checkov/blob/48abe40926c97bd2e6f8c80491369be462ce3edd/checkov/terraform/checks/resource/azure/AKSRbacEnabled.py#L19-L29

It returns `false` if the property is not found in the resource. It shouldn't be the case as the default value of the property is `true`

fredgate:

## Description

The check CKV_AZURE_5 should succeed if the `role_based_access_control_enabled` property is not set in resource `azurerm_kubernetes_cluster` as its default value is `true`.

Fixes #4227 

## Checklist:

### DIFF
--- a/checkov/terraform/checks/resource/azure/AKSRbacEnabled.py
+++ b/checkov/terraform/checks/resource/azure/AKSRbacEnabled.py
@@ -23,10 +23,10 @@ class AKSRbacEnabled(BaseResourceCheck):
         ]
 
         for key in self.evaluated_keys:
-            if dpath.search(conf, key) and dpath.get(conf, key)[0]:
-                return CheckResult.PASSED
+            if dpath.search(conf, key):
+                return CheckResult.PASSED if dpath.get(conf, key)[0] else CheckResult.FAILED
 
-        return CheckResult.FAILED
+        return CheckResult.PASSED
 
 
 check = AKSRbacEnabled()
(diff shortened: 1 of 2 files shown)


gruebel (approved):
thanks 🍻

ChanochShayner (approved):

PR #4238, opened 2023-01-12
fix(secrets): Change secret validation status message

Eliran-Turgeman:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description
* Add 'Unavailable' status and use it as default
* Change record message for secret validation status

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/common/output/record.py
```diff
@@ -208,6 +210,24 @@ def to_string(self, compact: bool = False, use_bc_ids: bool = False) -> str:
...
 
+    def _get_secret_validation_status_message(self) -> str:
+        warning_sign_unicode = '\u26a0'
+        text_by_secret_validation_status_status = {
+            ValidationStatus.VALID.value: colored(f'\t{warning_sign_unicode} This secret has been validated'
+                                                  f' and should be prioritized\n', "red"),
+            ValidationStatus.INVALID.value: '\tThis is not a valid secret and can be de-prioritized\n',
+            ValidationStatus.UNKNOWN.value: '\tWe were not able to validate this secret\n',
+            ValidationStatus.UNAVAILABLE.value: ''
+        }
```
marynaKK, line 221: maybe this can be declared once as a const? :)
also maybe better to move it to the secrets record file?

Eliran-Turgeman: Moved it to secrets_record :)
I don't think it should be a const as there's no other function that needs it

gruebel: since you invoke the function each time a secret verifier status is needed, the dict will be recreated. therefore moving it to a const, like @marynaKK suggested, makes it more performant. Also is there any reason why `warning_sign_unicode` is a variable and not just inlined?

Eliran-Turgeman: Good point, will move to const.

It's a variable just for readability's sake

checkov/common/output/record.py
```diff
@@ -197,11 +197,12 @@ def to_string(self, compact: bool = False, use_bc_ids: bool = False) -> str:
...
                 hasattr(self, 'validation_status') and \
-                os.getenv("CKV_VALIDATE_SECRETS"):
-            secret_validation_status_string = f'\tStatus: {self.validation_status}\n'
+                os.getenv("CKV_VALIDATE_SECRETS") and \
+                hasattr(self, '_get_secret_validation_status_message'):  # for typing purposes, can't check with isinstance cause of circular dependency
+            secret_validation_status_string = self._get_secret_validation_status_message()
 
         if self.check_result["result"] == CheckResult.FAILED and code_lines and not compact:
-            return f"{check_message}{status_message}{severity_message}{detail}{file_details}{secret_validation_status_string}{caller_file_details}{guideline_message}{code_lines}{evaluation_message}"
+            return f"{check_message}{status_message}{secret_validation_status_string}{severity_message}{detail}{file_details}{caller_file_details}{guideline_message}{code_lines}{evaluation_message}"
```
marynaKK, line 205: A small general question here :)
the parts of the validation status of the secret are not related to all other records. I am not sure, but don't we prefer to change the ending of this method, so the main record class will not even know there is a `_get_secret_validation_status_message` method? 
and the `secrets_record` as the subclass will add the logic it needs?

I couldn't find a reason why this could be a bad idea so thought of asking. WDYT? :)

gruebel: definitely would be more clean, but needs some adjustments, like adding an extra method, which is called to add this validation string and in the base calls it just returns an empty string.

Eliran-Turgeman: Definitely a good idea.

I could just override `to_string` by copy-pasting it from Record, but that will result in a lot of code duplication.

Definitely, there's room for better abstractions here, but I think it should be handled separately since it won't be a small change.
I think we can start by copy-pasting `to_string` to `secrets_record` and I'll work on better abstracting it.

WDYT? 
@marynaKK @gruebel

gruebel: let's not copy it, instead create a a new issue and add a comment to your code part to improve this part by leveraging the inheritance.

gruebel (approved):
🐰

Eliran-Turgeman:
Waiting for dependent changes to be merged first

PR #4250, opened 2023-01-15
fix(secrets): Change secret validation status message

Eliran-Turgeman:
Reverts bridgecrewio/checkov#4248

### DIFF
--- a/checkov/common/output/secrets_record.py
+++ b/checkov/common/output/secrets_record.py
@@ -1,10 +1,25 @@
 from __future__ import annotations
+
+import logging
 from typing import Optional, List, Tuple, Dict, Any
+
+from termcolor import colored
+
+from checkov.secrets.consts import ValidationStatus
 
 from checkov.common.bridgecrew.severities import Severity
 
 from checkov.common.output.record import Record
 from checkov.common.typing import _CheckResult
+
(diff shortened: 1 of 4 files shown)


checkov/secrets/consts.py
```diff
@@ -6,6 +6,7 @@ class ValidationStatus(Enum):
     VALID = 'Valid'
     INVALID = 'Invalid'
     UNKNOWN = 'Unknown'
+    UNAVAILABLE = 'UNAVAILABLE'
```
nimrodkor, line 9: @Eliran-Turgeman should be `Unavailable`?

Eliran-Turgeman: Yes!

checkov/common/output/secrets_record.py
```diff
@@ -63,3 +78,13 @@ def __init__(self,
                          definition_context_file_path=definition_context_file_path
                          )
         self.validation_status = validation_status
+
+    def _get_secret_validation_status_message(self) -> str:
+        message = None
+        if hasattr(self, 'validation_status') and self.validation_status:
```
marynaKK, line 84: no need to check for `hasattr` if its in the secret_record.py :)

marynaKK (approved):

gruebel (approved):
🏖️

PR #4256, opened 2023-01-16
fix(terraform): fix empty list in CKV_AZURE_183 check

micheelengronne:

## Description

fix dns_servers configuration when declared as an empty list.

## Checklist:

### DIFF
--- a/checkov/terraform/checks/resource/azure/VnetLocalDNS.py
+++ b/checkov/terraform/checks/resource/azure/VnetLocalDNS.py
@@ -23,7 +23,7 @@ class VnetLocalDNS(BaseResourceCheck):
         if "dns_servers" in conf and conf["dns_servers"] and isinstance(conf["dns_servers"], list):
             dns_servers = conf["dns_servers"][0]
             for ip in dns_servers:
-                if "address_space" in conf and conf["address_space"] and isinstance(conf["address_space"], list):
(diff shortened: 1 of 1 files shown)


checkov/terraform/checks/resource/azure/VnetLocalDNS.py
```diff
@@ -23,7 +23,7 @@ def scan_resource_conf(self, conf: dict[str, list[Any]]) -> CheckResult:
         if "dns_servers" in conf and conf["dns_servers"] and isinstance(conf["dns_servers"], list):
             dns_servers = conf["dns_servers"][0]
             for ip in dns_servers:
-                if "address_space" in conf and conf["address_space"] and isinstance(conf["address_space"], list):
+                if "address_space" in conf and conf["address_space"] and isinstance(conf["address_space"], list) and len(conf["address_space"]) > 0:
```
gruebel, line 26: hey @micheelengronne this is already done here
```
if conf["address_space"]:
```
checks for a non empty list. Actually it checks for truthiness and this can only be true, if it is not not empty, `None`, etc.

micheelengronne: Actually I was possibly drunk when I wrote this patch because it is not even the correct check. It is conf["dns_servers"] emptiness we should check because when it is empty it still triggers this check.

micheelengronne: And you are right, empty list is Falsy so I should go back to my usecase and figure out what happens.

PR #4260, opened 2023-01-16
feat(secrets): fix to keyword combinator to reduce FPs

maxamel:
This PR fixes keyword combinator to filter out potential secrets which are detected on the keyword in case that keyword is a hint. We re-run the secret detection on the value only. In addition increased the entropy level threshold.
All in all this should reduce FPs whilst keeping TP on the same level. 

## Checklist:


checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -220,18 +220,39 @@ def analyze_line(
...
+                            # Re-run secret detection on cut line
+                            value_secrets = self.detect_secret(
+                                scanners=self.high_entropy_scanners_iac,
+                                filename=filename,
+                                line=line,
+                                line_number=line_number,
+                                kwargs=kwargs
+                            )
+                            detected_secrets.remove(detected_secret)
+                            detected_secrets = detected_secrets.union(value_secrets)
```
gruebel, line 253:
```suggestion
                            detected_secrets.update(value_secrets)
```
should be the same without creating a new set 🙂

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -220,18 +220,39 @@ def analyze_line(
...
+                            break
+                    return detected_secrets
```
gruebel, line 255: can you move this part to a small method 🙂

gruebel (commented):
nice, looks good 🏖️

checkov/secrets/plugins/entropy_keyword_combinator.py
maxamel: post processing is now in a new method

gruebel (approved):

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -269,6 +275,14 @@ def analyze_line(
 
         return set()
 
+    def postprocess_secrets(self, detected_secrets: set[PotentialSecret], line: str) -> None:
+        for detected_secret in detected_secrets:
+            if detected_secret.secret_value and line.replace('"', '').replace("'", '').startswith(
```
nimrodkor, line 280: @maxamel Do we need to clean up more characters? whitespace?

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -220,18 +220,24 @@ def analyze_line(
...
+                    detected_secrets = self.detect_secret(
                         scanners=self.high_entropy_scanners_iac,
                         filename=filename,
                         line=line,
                         line_number=line_number,
                         kwargs=kwargs
                     )
+                    # postprocess detected secrets - filter out potential secrets on keyword and re-run secret detection
+                    # on their value only
+                    self.postprocess_secrets(detected_secrets, line)
```
nimrodkor, line 239: Let's try a more explicit name? For example:
```suggestion
                    self.remove_fp_secrets_in_keys(detected_secrets, line)
```

checkov/secrets/plugins/entropy_keyword_combinator.py
```diff
@@ -269,6 +275,14 @@ def analyze_line(
 
         return set()
 
+    def postprocess_secrets(self, detected_secrets: set[PotentialSecret], line: str) -> None:
+        for detected_secret in detected_secrets:
+            if detected_secret.secret_value and line.replace('"', '').replace("'", '').startswith(
+                    detected_secret.secret_value):
+                # Found keyword prefix as potential secret
+                detected_secrets.remove(detected_secret)
+                break
```
nimrodkor, line 284: @maxamel why not go over the entire secret? Why just on the first one?

checkov/secrets/plugins/entropy_keyword_combinator.py
maxamel: No, since if you remove whitespace from line you'd have to do it for secret value as well.

checkov/secrets/plugins/entropy_keyword_combinator.py
maxamel: Because after you cut the first part, if you continue iterating you'll just remove all the potential secrets and end up with nothing. We're interested only in the prefix since that indicates a keyword has been flagged.

nimrodkor (approved):

PR #4296, opened 2023-01-22
fix(terraform): fix in variable rendering

YaaraVerner:

## Description
create a function for conditional expression division instead of regex to catch more edge cases (such as dictionary as value).
fix in evaluate_conditional_expression - do not choose the first value if we didn't succeed to evaluate expression.
fix in evaluate_map - do not replace '=' if it is '==' or '!='

## Checklist:


ChanochShayner (approved):
💯

tronxd (approved):

PR #4323, opened 2023-01-24
feat(sca): Add a --support flag

reemodo:
## Description

For better supporting customer and in a second step OSS users we would like to add a new support flag, which automatically sends the logs to our backend.

add a support flag --support
a copy of the logs should be uploaded next to the run metadata info, only when the flag is used
extend run metadata with following info
OS system info, ex. Darwin Kernel Version 21.6.0
CPU architecture info, ex. x86_64
Python implementation, ex CPython
System environment variables, just the names, not their values

## Checklist:


checkov/common/bridgecrew/wrapper.py
```diff
@@ -96,11 +96,16 @@ def persist_checks_results(
...
 
 def persist_run_metadata(
-        run_metadata: dict[str, str | list[str]], s3_client: BaseClient, bucket: str, full_repo_object_key: str
+        run_metadata: dict[str, str | list[str]] | str, s3_client: BaseClient, bucket: str, full_repo_object_key: str
 ) -> None:
-    object_path = f'{full_repo_object_key}/{checkov_results_prefix}/run_metadata.json'
     try:
-        s3_client.put_object(Bucket=bucket, Key=object_path, Body=json.dumps(run_metadata, indent=2))
+        if isinstance(run_metadata, str):
+            object_path = f'{full_repo_object_key}/{checkov_results_prefix}/log_test.txt'
```
mikeurbanski1, line 103: `log_test`?

checkov/common/util/ext_argument_parser.py
```diff
@@ -115,6 +115,12 @@ def add_parser_args(self) -> None:
             action="version",
             version=version,
         )
+        self.add(
+            "--support",
+            action="store_true",
+            help="store your logs to support ASAP",
```
mikeurbanski1, line 121:
```suggestion
            help="Enable debug logs and upload the logs to the server. Requires a Bridgecrew or Prisma Cloud API key.",
```

docs/2.Basics/CLI Command Reference.md
```diff
@@ -60,4 +60,5 @@ nav_order: 2
...
...                                                                                                                                                                                                                                                                | Enable saving you command logs due to support you ASAP                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
mikeurbanski1, line 64:
```suggestion
| `--support`                                                                                                                                                                                                                                                                                                                                                                | Enable debug logs and upload the logs to the server. Requires a Bridgecrew or Prisma Cloud API key.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```

checkov/main.py
```diff
@@ -144,7 +146,11 @@ def run(banner: str = checkov_banner, argv: list[str] = sys.argv[1:]) -> int | N
...
         "checkov_executable": sys.argv[0],
-        "args": parser.format_values(sanitize=True).split('\n')
+        "args": parser.format_values(sanitize=True).split('\n'),
+        "OS_system_info": platform.platform(),
+        "CPU_architecture": platform.processor(),
+        "Python_implementation": platform.python_implementation()
+
     }
 
     logger.debug(f'Run metadata: {json.dumps(run_metadata, indent=2)}')
```
mikeurbanski1, line 156: Can we add some logic somewhere around here to ensure the following:

- If --support is set, then check for a non-empty API key and log a warning if it is not set: "The --support flag requires a platform API key in order to upload the logs."
- If --support is set, set the log level to DEBUG (the logger is already initialized, so you'll have to update it)

mikeurbanski1 (changes requested):
Nice job! I made a few suggestions to clean it up a bit.

checkov/main.py
mikeurbanski1: We need to set debug logs at least before we print the run metadata, because that is the first piece of critical information

checkov/main.py
```diff
@@ -572,349 +580,358 @@ def run(self, banner: str = checkov_banner) -> int | None:
...
+            elif not self.config.quiet:
+                print(f"{banner}")
 
-            bc_integration.onboarding()
-        return None
+                bc_integration.onboarding()
+            return None
+        finally:
+            if self.config.support:
+                bc_integration.persist_run_metadata(log_stream.getvalue())
```
mikeurbanski1, line 934: Are there going to be memory issues / buffer size issues with this approach? For a large repo, the log size can be very large.

checkov/main.py
Saarett: @mikeurbanski1 Just a quick thought regarding the 2nd one, can it be something else and not only DEBUG?

checkov/main.py
Saarett: @mikeurbanski1 Good point, do you have a repo to share so we can test it out?

checkov/main.py
reemodo: Do you mean if the user runs `checkov --support` a log warning should appear in his console?

checkov/main.py
reemodo: I have tested a large repo. The log file size of it was 32MB. It ran without issues and took 30 seconds to send the request and receive the response. And because I am using the stream, the log file will contain only the current command logs. So I don't think the size of logs for one command can cause any buffers issues. And if you have any other big repos to test, it will be great.

gruebel: since this is pure text, it should be possible to lower the size by creating a compressed tar file and upload it.

checkov/main.py
gruebel: as far as I can tell you already implemented the first part, by logging an error, if someone uses `--support` flag without API key, but please move it to the beginning of the actual run method not in the `finally` block.
for the second one, you need to adjust the log level of the `stream_handler` to `DEBUG` this should be possible by just adding
```python
stream_handler.setLevel(logging.DEBUG)
```

checkov/main.py
```diff
@@ -572,349 +580,361 @@ def run(self, banner: str = checkov_banner) -> int | None:
...
-            bc_integration.onboarding()
-        return None
+                bc_integration.onboarding()
+            return None
+        finally:
+            if self.config.support:
+                if not self.config.bc_api_key:
+                    self.parser.error("--bc-api-key argument is required when using --support")
+                else:
+                    bc_integration.persist_run_metadata(log_stream.getvalue())
```
Saarett, line 937: @reemodo So we invoke a function called "persist_**run_metdata**" in order to persist **logs**?
Perhaps we should create another function, or change the name of the current one?

checkov/main.py
reemodo: The second point, I have already added it. 
The first one, if I put it at the beginning of the run method. It will send the stream string to our file in s3 and it will not contains all the logs. And if I put it at the end of the run method without (try and finally) and the user has an error or expiation the run will stop when rising the bug to the console and the lines of handling the support flag will not be run or save. So, I used `try and finally` to run the support flag with all logs even if there are exceptions they will be saved.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -452,6 +454,15 @@ def persist_run_metadata(self, run_metadata: dict[str, str | list[str]]) -> None
             return
         persist_run_metadata(run_metadata, self.s3_client, self.bucket, self.repo_path)
 
+    def persist_logs_stream(self, logs_stream: StringIO) -> None:
```
reemodo, line 457: @Saarett I add new functions to handle the support flag.

checkov/common/bridgecrew/wrapper.py
```diff
@@ -101,6 +104,25 @@ def persist_run_metadata(
     object_path = f'{full_repo_object_key}/{checkov_results_prefix}/run_metadata.json'
     try:
         s3_client.put_object(Bucket=bucket, Key=object_path, Body=json.dumps(run_metadata, indent=2))
+
+    except Exception:
+        logging.error(f"failed to persist run metadata into S3 bucket {bucket}", exc_info=True)
+        raise
+
+
+def persist_logs_stream(logs_stream: StringIO, s3_client: BaseClient, bucket: str, full_repo_object_key: str) -> None:
```
reemodo, line 113: @gruebel The tar compression works successfully.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -452,6 +454,15 @@ def persist_run_metadata(self, run_metadata: dict[str, str | list[str]]) -> None
...
+        if not self.use_s3_integration:
+            return
+        if not self.bucket or not self.repo_path:
+            logging.error(f"Something went wrong: bucket {self.bucket}, repo path {self.repo_path}")
+            return
+        print(logs_stream.getvalue())
```
Saarett, line 463: This can be removed

Saarett (approved):
Great job!

checkov/common/bridgecrew/platform_integration.py
itai1357: I'm a little afraid of implementing here function that named "persist_logs_stream" while you also import it from other file.
1. did you checked that it behaves as expected?
2. unless you use the same names on purpose (maybe there is some pattern that I'm not familiar with), I suggest to use different names

checkov/common/bridgecrew/wrapper.py
```diff
@@ -101,6 +104,25 @@ def persist_run_metadata(
...
+        info = tar.tarinfo(name='logs_file.txt')
+        bio.seek(0)
+        info.size = logs_stream.tell()
+        tar.addfile(info, bio)
+    file_io.seek(0)
+    object_path = f'{full_repo_object_key}/{checkov_results_prefix}/logs_file.tar.gz'
+    try:
+        s3_client.put_object(Bucket=bucket, Key=object_path, Body=file_io)
     except Exception:
         logging.error(f"failed to persist run metadata into S3 bucket {bucket}", exc_info=True)
```
itai1357, line 127:
```suggestion
        logging.error(f"failed to persist logs stream into S3 bucket {bucket}", exc_info=True)
```

checkov/common/bridgecrew/platform_integration.py
itai1357: nevermind. my bad. Now I noticed that the current method is class-one. it is ok

checkov/common/bridgecrew/wrapper.py
itai1357: @reemodo I'm highly recommend:
1. moving the compression logic to here: `checkov/common/util/file_utils.py`
A suggestion
2. creating UTs for the compression part + making sure you can decompress it to the original data:

checkov/main.py
```diff
@@ -572,349 +580,361 @@ def run(self, banner: str = checkov_banner) -> int | None:
...
-            print(f"{banner}")
+                return exit_code
+            elif not self.config.quiet:
+                print(f"{banner}")
 
-            bc_integration.onboarding()
-        return None
+                bc_integration.onboarding()
+            return None
+        finally:
```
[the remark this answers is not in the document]
itai1357: I think that should be helpful:
`logging.error(<some error massage>, exc_info=True)`

if it doesn't, look here:
https://www.geeksforgeeks.org/how-to-print-exception-stack-trace-in-python/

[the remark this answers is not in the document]
gruebel: yeah, an `except Exception` block would be great to catch some strange error and then log it, like suggested.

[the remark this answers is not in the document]
reemodo: Sure, but I will use `BaseException` to catch the system-exiting exceptions SystemExit, KeyboardInterrupt and GeneratorExit.

mikeurbanski1 (approved):

PR #4327, opened 2023-01-25
feat(general): igraph library support

tronxd:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

This PR enable the support of the (now experimental ) `igraph` library, which is proven to be more CPU and memory efficient from the `networkx` library.
The graph library to use is controlled by the `CHECKOV_GRAPH_FRAMEWORK` environment variable, which accepts two values - `NETWORKX` and `IGRAPH` (defaults to `NETWORKX`), and uses the appropriate library by its value. 

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/common/checks_infra/solvers/connections_solvers/connection_exists_solver.py
```diff
@@ -25,45 +27,82 @@ def __init__(
...
-                destination_attributes['connected_node'] = origin_attributes
-                continue
+                if origin_attributes and destination_attributes_list:
+                    for destination_attributes in destination_attributes_list:
+                        if origin_attributes in self.excluded_vertices or destination_attributes in self.excluded_vertices:
+                            failed.extend([origin_attributes, destination_attributes])
+                        elif origin_attributes in self.unknown_vertices or destination_attributes in self.unknown_vertices:
+                            unknown.extend([origin_attributes, destination_attributes])
+                        else:
+                            passed.extend([origin_attributes, destination_attributes])
```
YaaraVerner, line 69: it's looks like the same logic as in lines 83-89, can you extract it to a function?

checkov/common/checks_infra/solvers/connections_solvers/connection_exists_solver.py
```diff
@@ -25,45 +27,82 @@ def __init__(
...
+                        output_edges = graph_connector.edges(v, data=True)
+                        _, output_destination, _ = next(iter(output_edges))
+                        output_destination = graph_connector.nodes(data=True)[output_destination]
+                        output_destination_type = output_destination.get(CustomAttributes.RESOURCE_TYPE)
+                        if self.is_associated_edge(
+                            origin_attributes.get(CustomAttributes.RESOURCE_TYPE), output_destination_type
+                        ):
+                            passed.extend([origin_attributes, output_destination])
+                    except StopIteration:
+                        continue
```
YaaraVerner, line 104: shouldn't there be similar logic to handle output blocks to the igraph implementation as well?

checkov/common/graph/db_connectors/igraph/igraph_db_connector.py
```diff
@@ -0,0 +1,73 @@
...
+    def networkit_from_local_graph(self, local_graph: LocalGraph[_Block]) -> Graph:
+        # colors = {
+        #     "Pod": "blue",
+        #     "Deployment": "magenta",
+        #     "Service": "olive",
+        # }
+        #
+        # random_colors = ("red", "blue", "olive", "gold", "magenta", "purple", "black")
+        # random_colors_len = len(random_colors)
+        # randrange(len(random_colors))
```
YaaraVerner, line 39: is this relevant?

YaaraVerner (commented):
Nice work! had a few comments 😊

checkov/common/graph/db_connectors/igraph/igraph_db_connector.py
tronxd: Can be used for visualization, so I'll leave it here for now

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -45,6 +46,24 @@ def run(self, graph_connector: DiGraph) -> Tuple[List[Dict[str, Any]], List[Dict
...
+
+            for data in graph_connector.vs.select(**select_kwargs)["attr"]:
+                result = self.get_operation(vertex=data)
+                # A None indicate for UNKNOWN result - the vertex shouldn't be added to the passed or the failed vertices
+                if result is None:
+                    unknown_vertices.append(data)
+                elif result:
+                    passed_vertices.append(data)
+                else:
+                    failed_vertices.append(data)
```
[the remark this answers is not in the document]
gruebel: multi threading wouldn't help, because it is CPU bound, only multiprocessing could give some boost. If we need it, then we should do it in a separate PR and it could become tricky, because we also leverage multiprocessing on the runner level.

checkov/common/checks_infra/solvers/connections_solvers/connection_exists_solver.py
gruebel: as i remember it was not needed, but if we have a test case and it doesn't work with the current implementation, then we should fix it. Basically `igraph` returns back the whole connection path and not one at a time like `networkx` so our implementation is different between them.

checkov/common/checks_infra/solvers/connections_solvers/connection_exists_solver.py
gruebel: sounds good, maybe it also makes sense to create separate functions just for the DFS part for both of them, so it is too overloaded here

checkov/common/graph/db_connectors/igraph/igraph_db_connector.py
gruebel: that was my left over, which I wanted to use for adding kind of debugging capabilities

setup.py
```diff
@@ -30,6 +30,7 @@
             "GitPython==3.1.7",
             "bandit",
             "jsonschema",
+            "parameterized"
```
gruebel, line 33: no need for that, I think we actually can remove the whole dev block, we don't use it anywhere and there are so many missing libs here 🙈

gruebel (approved):
nice, looks good 🍻

YaaraVerner (approved):

PR #4353, opened 2023-01-29
feat(general): Create 3d combinations post runner

arielkru:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

3d policies are a new kind of policy that checks for combinations between IaC, CVE and (TBD) secrets scan results.
* 3d policies are fetched from bc_integration and then run agaisnt all the scan reports
* the 3d policy runner inherits a new post runner class
* Implemented a new format for 3d policies output
* extracted some common functions to reduce duplications
* changed sca_image and sca_package outputs to have uppercase severities

## Checklist:


checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def run(  # type:ignore[override]
+            self,
+            checks: list[Base3dPolicyCheck] | None = None,
+            scan_reports: list[Report] | None = None,
+            runner_filter: RunnerFilter | None = None
+    ) -> Report | list[Report]:
```
tronxd, line 32: It looks like that only a scan report is returned, in which case it's a list of reports?

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+        reports_by_fw = {report.check_type: report for report in scan_reports}
+        for check in checks:
+            records = self.run_check(check, reports_by_fw)
+            for record in records:
+                report.add_record(record=record)
+
+        self.pbar.close()
+        return report
+
+    def run_check(self, check: Base3dPolicyCheck, reports_by_fw: dict[str, Report]) -> list[Record]:
```
tronxd, line 54: Can we rename to `collect_check` or something like it? We don't actually run the check since we already have a report

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+        records = []
+        iac_results_map = self.solve_check_iac(check, reports_by_fw)
+        cve_results_map = self.solve_check_cve(check, reports_by_fw)
+
+        for iac_resource, iac_records in iac_results_map.items():
+            for cve_resource in cve_results_map:
+                if iac_resource == cve_resource:
+                    # This means we found the combination on the same resource -> create a violation for that resource
+                    check_result = CheckResult.FAILED
+                    record = self.get_record(check, iac_records[0], check_result)
```
tronxd, line 64:
```suggestion
                    record = self.get_record(check, iac_resource, check_result)
```

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+                    record.set_guideline(check.guideline)
+                    records.append(record)
+
+        self.pbar.update()
+        return records
+
+    def solve_check_iac(self, check: Base3dPolicyCheck, reports_by_fw: dict[str, Report]) -> dict[str, list[Record]]:
+        iac_results_map: dict[str, list[Record]] = {}
+        if check.iac:
+            for fw, bc_check_ids in check.iac.items():
```
tronxd, line 74:
```suggestion
            for framework, bc_check_ids in check.iac.items():
```
Let's not use shorthands like `fw`

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+                fw_report = reports_by_fw.get(fw)
+                if fw_report:
+                    fw_records = fw_report.failed_checks
+                    for record in fw_records:
+                        if record.bc_check_id in bc_check_ids:
+                            resource_id = f'{record.file_abs_path}:{record.resource}'
```
tronxd, line 80: If this comes from a report, shouldn't the `resource_id` be already computed?

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,130 @@
...
+        if check.cve:
+            cve_report = reports_by_fw.get(CheckType.SCA_IMAGE)
+            if cve_report:
+                image_results = cve_report.image_cached_results
+                risk_factor = None
+                # currently checks only for a single risk factor condition on the cves, to be extended
+                for attribute, value in check.cve.items():
+                    if attribute == CVEAttribute.RISK_FACTORS:
+                        risk_factor = value[0]
+
```
tronxd, line 98: risk factor may not be the only CVE property that we might want to check - can we have a map between the combinations attribute (`CVEAttribute`) to the sca_image report and use it?

checkov/common/bridgecrew/integration_features/features/policies_3d_integration.py
```diff
@@ -0,0 +1,92 @@
...
+            for policy in policies:
+                try:
+                    logging.debug(f"Loading 3d policy id: {policy.get('id')}")
+                    converted_check = self._convert_raw_check(policy)
+                    check = self.platform_policy_parser.parse_raw_check(converted_check)
+                    check.severity = Severities[policy['severity']]
+                    check.bc_id = check.id
+                    checks.append(check)
+                except Exception:
+                    logging.debug(f"Failed to load 3d policy id: {policy.get('id')}", exc_info=True)
```
lirshindalman, line 81: maybe add the except we get in the log

lirshindalman (approved):
nice!! :rocket:

checkov/policies3d/runner.py
arielkru: Good idea

checkov/policies3d/runner.py
arielkru: No because the resource id in the iac reports and in the image cache reports are different, so I construct it here in the same format so the comparison works

checkov/policies3d/runner.py
arielkru: I extended to list in case its relevant in the future but you're right that it's not needed. In second thought it can be removed now and if it'd be relevant in the future it'd be added, so removing

checkov/policies3d/runner.py
arielkru: It is still a check (just a post runner type of check) and running it is just running its logic, so I dont think run check is wrong here, but I get the confusion, changed it to collect

checkov/policies3d/runner.py
arielkru: Notice your change swaps a record type for a string type, and I need the record here..

checkov/common/bridgecrew/integration_features/features/policies_3d_integration.py
arielkru: That's what `exc_info=True` does :)

tronxd (approved):
👑

checkov/policies3d/runner.py
tronxd: Is it guaranteed that `iac_records[0]` is the matching record for the resource? Shouldn't we match it?

arielkru: We need the iac record just for some attributes we insert to our own 3d policy record, and these attributes should be the same in all of the records for the same resource, so I just take the first one

checkov/common/bridgecrew/code_categories.py
```diff
@@ -40,6 +40,7 @@ class CodeCategoryType:
     CheckType.SERVERLESS: CodeCategoryType.IAC,
     CheckType.TERRAFORM: CodeCategoryType.IAC,
     CheckType.TERRAFORM_PLAN: CodeCategoryType.IAC,
+    CheckType.POLICY_3D: CodeCategoryType.IMAGES  # actually a combination of all code categories but currently image scannning is common to all of the 3d policies
```
marynaKK, line 43: why not create another code category?

checkov/common/bridgecrew/integration_features/features/policies_3d_integration.py
```diff
@@ -0,0 +1,92 @@
...
+
+if TYPE_CHECKING:
+    from checkov.common.bridgecrew.platform_integration import BcPlatformIntegration
+    from checkov.common.output.report import Report
+    from checkov.common.typing import _BaseRunner
+
+
+class Policies3DIntegration(BaseIntegrationFeature):
+    def __init__(self, bc_integration: BcPlatformIntegration) -> None:
+        super().__init__(bc_integration=bc_integration, order=11)
```
marynaKK, line 21: maybe should add enums for the order? why is it 11?

marynaKK (commented):
SO NICE! 🍒

checkov/common/bridgecrew/code_categories.py
gruebel: @tsmithv11 thoughts about the code category?

checkov/policies3d/checks_infra/base_parser.py
```diff
@@ -0,0 +1,8 @@
+from typing import Dict, Any
+
+from checkov.policies3d.checks_infra.base_check import Base3dPolicyCheck
+
+
+class Base3dPolicyCheckParser:
+    def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -> Base3dPolicyCheck:
+        raise NotImplementedError
```
gruebel, line 8:
```suggestion
    @abstractmethod
    def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -> Base3dPolicyCheck:
        pass
```

checkov/policies3d/output.py
```diff
@@ -0,0 +1,131 @@
...
+                    "severity": severity_str
+                }
+            )
+
+            if package_name in package_cves_details_map:
+                package_cves_details_map[package_name]["cves"].sort(key=compare_cve_severity, reverse=True)
+                package_cves_details_map[package_name]["current_version"] = package_version
+                package_cves_details_map[package_name]["image_name"] = image_name
+
+        if len(package_cves_details_map.keys()):
```
gruebel, line 51:
```suggestion
        if package_cves_details_map:
```
no need to actually calculate the length

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,140 @@
...
+CVE_CHECK_TO_REPORT_ATTRIBUTE = {
+    CVECheckAttribute.RISK_FACTORS: CVEReportAttribute.RISK_FACTORS
+}
+
+class Policy3dRunner(BasePostRunner):
+    check_type = CheckType.POLICY_3D  # noqa: CCE003  # a static attribute
+
+    def __init__(self) -> None:
+        super().__init__()
```
gruebel, line 33:
```suggestion
```
if you don't change anything, this can be removed

checkov/policies3d/runner.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def run(  # type:ignore[override]
```
gruebel, line 35: I know why you added the `type:ignore`, I will adjust it.

gruebel (approved):
nice work 🍻 looking forward to see more adoption of it. Do you know, if we have to do some extra work for other output types?

checkov/policies3d/runner.py
gruebel: fixed it and other typing/linting issues

checkov/policies3d/runner.py
gruebel: fixed 🙂

checkov/common/bridgecrew/code_categories.py
tsmithv11: We spoke about it. I guess we can't lead with "3" so it makes sense to go with POLICY_3D

tsmithv11: I think we should have a new code category. Some of the policies to come won't have images.

arielkru: added

checkov/common/bridgecrew/integration_features/features/policies_3d_integration.py
nimrodkor: Sadly, it doesn't log the actual message... We tried it @arielkru

tsmithv11 (approved):

PR #4370, opened 2023-01-30
feat(sca): ignore package.json file when yarn.lock exists

ayajbara:
## Description

currently if we have yarn.lock file and package.json, we build a DT for yarn.lock and scan package.json.

Fixes # (issue)

### Description

### Fix

## Checklist:


itai1357 (approved):
this bug causes me to think, maybe we need to get the scannable files from the DT-setup and use API call for getting the scanned file, for avoiding sure errors in the future

AdamDev (approved):

PR #4385, opened 2023-02-01
feat(secrets): extract new detector_utils file from entropy keyword combinator

maxamel:
This PR extracts utils from entropy keyword combinator so we can use them for other probability-based detectors and plugins. We are deleting a few methods from entropy keyword combinator and adding them to detector_utils.
## Checklist:

### DIFF
--- a/checkov/secrets/plugins/detector_utils.py
+++ b/checkov/secrets/plugins/detector_utils.py
@@ -5,7 +5,6 @@ import re
 from re import Pattern
 from typing import Any, TYPE_CHECKING
 
-from detect_secrets.plugins.high_entropy_strings import HighEntropyStringsPlugin
 from detect_secrets.util.filetype import FileType
 from detect_secrets.plugins.keyword import DENYLIST
 from detect_secrets.plugins.keyword import AFFIX_REGEX
@@ -23,6 +22,7 @@ if TYPE_CHECKING:
     from checkov.secrets.parsers.multiline_parser import BaseMultiLineParser
     from detect_secrets.core.potential_secret import PotentialSecret
     from detect_secrets.util.code_snippet import CodeSnippet
+    from detect_secrets.plugins.high_entropy_strings import HighEntropyStringsPlugin
 
 MAX_KEYWORD_LIMIT = 500
 
--- a/checkov/secrets/plugins/detector_utils.py
+++ b/checkov/secrets/plugins/detector_utils.py
@@ -196,14 +196,14 @@ def detect_secret(
         line_number: int = 0,
         **kwargs: Any,
 ) -> set[PotentialSecret]:
-    for entropy_scanner in scanners:
-        matches = entropy_scanner.analyze_line(filename, line, line_number, **kwargs)
+    for scanner in scanners:
+        matches = scanner.analyze_line(filename, line, line_number, **kwargs)
         if matches:
             return matches
     return set()
 
 
-def analyze_multiline(
+def analyze_multiline_keyword_combinator(
         filename: str,
         scanners: tuple[HighEntropyStringsPlugin, ...],
         multiline_parser: BaseMultiLineParser,
--- a/checkov/secrets/plugins/detector_utils.py
+++ b/checkov/secrets/plugins/detector_utils.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
     from checkov.secrets.parsers.multiline_parser import BaseMultiLineParser
     from detect_secrets.core.potential_secret import PotentialSecret
     from detect_secrets.util.code_snippet import CodeSnippet
-    from detect_secrets.plugins.high_entropy_strings import HighEntropyStringsPlugin
+    from detect_secrets.plugins.base import BasePlugin
 
 MAX_KEYWORD_LIMIT = 500
 
@@ -190,7 +190,7 @@ def extract_from_string(pattern: dict[Pattern[str], int] | None, string: str) ->
 
 
 def detect_secret(
-        scanners: tuple[HighEntropyStringsPlugin, ...],
+        scanners: tuple[BasePlugin, ...],
         filename: str,
         line: str,
         line_number: int = 0,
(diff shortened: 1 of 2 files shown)


checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+def detect_secret(
+        scanners: tuple[HighEntropyStringsPlugin, ...],
+        filename: str,
+        line: str,
+        line_number: int = 0,
+        **kwargs: Any,
+) -> set[PotentialSecret]:
+    for entropy_scanner in scanners:
+        matches = entropy_scanner.analyze_line(filename, line, line_number, **kwargs)
```
marynaKK, line 200: if this method is going to be static, it might use other scanners then entropy-based, so it should not use the naming it had previously when it was part of the `EntropyKeywordCombinator` class.

checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        if matches:
+            return matches
+    return set()
+
+
+def analyze_multiline(
+        filename: str,
+        scanners: tuple[HighEntropyStringsPlugin, ...],
+        multiline_parser: BaseMultiLineParser,
+        line_number: int = 0,
```
marynaKK, line 210: not sure setting a valid line number is a good idea... maybe None is a better option in case of no line?

checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    for entropy_scanner in scanners:
+        matches = entropy_scanner.analyze_line(filename, line, line_number, **kwargs)
+        if matches:
+            return matches
+    return set()
+
+
+def analyze_multiline(
+        filename: str,
+        scanners: tuple[HighEntropyStringsPlugin, ...],
```
marynaKK, line 208: maybe change the typing to `BasePlugin` ?

checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+        entropy_on_value = detect_secret(
+            scanners=scanners,
+            filename=filename,
+            line=secret_adjust,
+            line_number=line_number,
+            kwargs=kwargs
+        )
+
+        if entropy_on_value:
```
marynaKK, line 232: it's no more just entropy, but a `potential_secret` which may be detected by scanner other than entropy scanner

checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        **kwargs: Any,
+) -> set[PotentialSecret]:
+    for entropy_scanner in scanners:
+        matches = entropy_scanner.analyze_line(filename, line, line_number, **kwargs)
+        if matches:
+            return matches
+    return set()
+
+
+def analyze_multiline(
```
marynaKK, line 206: maybe this method needs to be called something else? like `analyze_multiline_keyword_combinator`
cause now it will be in the `detector_utils` and it will not be coupled with the entropy_combinator class, so it's out of context now if the name stays the same.

checkov/secrets/plugins/detector_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+if TYPE_CHECKING:
+    from checkov.secrets.parsers.multiline_parser import BaseMultiLineParser
+    from detect_secrets.core.potential_secret import PotentialSecret
+    from detect_secrets.util.code_snippet import CodeSnippet
+    from detect_secrets.plugins.high_entropy_strings import HighEntropyStringsPlugin
+
+MAX_KEYWORD_LIMIT = 500
+
+DENY_LIST_REGEX = r'|'.join(DENYLIST)
```
marynaKK, line 29: maybe all the constants here should be in a separate class for keyword_combinator? (maybe also the `analyze_multiline` method?) as they are not general detector utils for secrets as general?

marynaKK (commented):
Had some comments :) Not sure everything crucial :)  🍒

checkov/secrets/plugins/detector_utils.py
maxamel: Can you clarify this? The method isn't static, and I want to leave naming as-is to prevent breaking the consistency of using the same method name for methods with similar purpose (we use it in other parts of the code as well, not just entropy)

checkov/secrets/plugins/detector_utils.py
maxamel: It could lead to runtime errors since analyze_multiline calls analyze_line which calls analyze_string which isn't implemented in BasePlugin. By enforcing HighEntropyStringsPlugin we avoid this. In general, using higher objects in the hierarchy is usually safer.

checkov/secrets/plugins/detector_utils.py
maxamel: I think analyze_multiline reflects it's general nature :)
The new policy which isn't keyword combinator is going to use it, and any future plugins we may introduce should also be able to use it, since it doesn't contain any keyword combinator specific logic.

checkov/secrets/plugins/detector_utils.py
maxamel: The issue is that the extraction is happening to open these directives for usage in other places.
In general, any non-regex detector should be able to use them as they are generic and aren't targeted at keyword combinator.

checkov/secrets/plugins/detector_utils.py
maxamel: Thanks, I removed it as it should be a mandatory field.

checkov/secrets/plugins/detector_utils.py
gruebel: if it is just a typing issue, then we can create a Protocol, which "enforces" the plugin to specific methods.

maxamel: Thanks @gruebel , I see Protocol is available from 3.8 and we should be able to support 3.7+, no?

gruebel: you can import it via `from typing_extensions import Protocol`

gruebel: ok, I checked the `detect-secrets` code. we can use `BasePlugin` as type hint here, because it is an abstract class, so a subclass needs to implement `analyze_string()` anyway 😄 I add the commit

gruebel (approved):
🍻

marynaKK (approved):

PR #4398, opened 2023-02-02
feat(terraform): [Foreach/Count Handling] Render dynamic foreach/count statement

ChanochShayner:

## Description
Render dynamic foreach/count statements for resources

## Checklist:


checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -23,12 +25,19 @@ def handle_foreach_rendering(self, foreach_blocks: dict[str, list[int]]) -> None
         self._handle_foreach_rendering_for_resource(foreach_blocks.get(BlockType.RESOURCE))
 
     def _handle_foreach_rendering_for_resource(self, resources_blocks: list[int]) -> None:
+        block_index_to_statement = self._get_statements(resources_blocks)
+        # new_resources = self._create_new_foreach_resources(block_index_to_statement)
```
bo156, line 29: You have commented code here

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -23,12 +25,19 @@ def handle_foreach_rendering(self, foreach_blocks: dict[str, list[int]]) -> None
...
+    def _get_statements(self, resources_blocks: list[int]) -> dict[int, Optional[list[str] | dict[str, Any] | int]]:
+        block_index_to_statement: dict[int, Optional[list[str] | dict[str, Any] | int]] = {}
         for block_index in resources_blocks:
             foreach_statement = self._get_foreach_statement(block_index)
+            block_index_to_statement[block_index] = foreach_statement
             # empty foreach_statement -> leave the main resource
-            if foreach_statement is None:
-                continue
-            # new_resources = self._create_new_foreach_resources(i, foreach_statement)
+        blocks_to_render = [block_idx for block_idx, statement in block_index_to_statement.items() if statement is None]
```
bo156, line 37: just for understanding - why do we want to render only if `statement is None`? 
Intuitively I would assume it should be the opposite

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
             return self._handle_static_count_statement(count_statement)
         return
 
-    def _create_new_foreach_resources(self, block_index: int, foreach_statement: list[str] | dict[str, Any]) -> None:
+    def _handle_dynamic_statement(self, blocks_to_render: list[int]) -> dict[int, Optional[list[str] | dict[str, Any] | int]]:
+        rendered_statements_by_idx: dict[int, Optional[list[str] | dict[str, Any] | int]] = {}
```
bo156, line 119: perhaps create an alias to this type and call it something like `FOR_EACH_BLOCK_TYPE`?

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
...
+        sub_graph = self._build_sub_graph(blocks_to_render)
+        self._render_sub_graph(sub_graph, blocks_to_render)
+        for block_idx in blocks_to_render:
+            if not self._is_static_statement(block_idx, sub_graph):
+                rendered_statements_by_idx[block_idx] = None
+            else:
+                rendered_statements_by_idx[block_idx] = self._handle_static_statement(block_idx, sub_graph)
```
bo156, line 126: kind of wierd that a function called `_handle_dynamic_statement` is calling `_handle_static_statement`.
This implies this function is more like a general `_handle_statement`, and that handling a dynamic statement is basically the function - `lambda x: retrun None`

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
...
+        return rendered_statements_by_idx
+
+    @staticmethod
+    def _render_sub_graph(sub_graph: TerraformLocalGraph, blocks_to_render: list[int]) -> None:
+        renderer = TerraformVariableRenderer(sub_graph)
+        renderer.vertices_index_to_render = blocks_to_render
+        renderer.render_variables_from_local_graph()
+
+    def _build_sub_graph(self, blocks_to_render: list[int]) -> TerraformLocalGraph:
+        from checkov.terraform.graph_builder.local_graph import TerraformLocalGraph
```
tronxd, line 136: Move import to top of the file

bo156, line 136: why do you use a local import here?

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
...
+        sub_graph = TerraformLocalGraph(self.local_graph.module)
+        sub_graph.vertices = [{}] * len(self.local_graph.vertices)
+        for i, block in enumerate(self.local_graph.vertices):
+            if block.block_type == BlockType.RESOURCE and i not in blocks_to_render:
+                sub_graph.vertices[i] = {}
+            else:
+                sub_graph.vertices[i] = block  # type:ignore
```
tronxd, line 143: You can use a `defaultdict` instead of the if/else init

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
...
+                rendered_statements_by_idx[block_idx] = self._handle_static_statement(block_idx, sub_graph)
+        return rendered_statements_by_idx
+
+    @staticmethod
+    def _render_sub_graph(sub_graph: TerraformLocalGraph, blocks_to_render: list[int]) -> None:
+        renderer = TerraformVariableRenderer(sub_graph)
+        renderer.vertices_index_to_render = blocks_to_render
+        renderer.render_variables_from_local_graph()
+
+    def _build_sub_graph(self, blocks_to_render: list[int]) -> TerraformLocalGraph:
```
tronxd, line 135: Can we add a test to this method?

checkov/terraform/graph_builder/foreach_handler.py
bo156, line 143: add a type to your initialization instead of ignoring

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -107,5 +115,38 @@ def _handle_static_statement(self, block_index: int) -> Optional[list[str] | dic
...
+        from checkov.terraform.graph_builder.local_graph import TerraformLocalGraph
+        sub_graph = TerraformLocalGraph(self.local_graph.module)
+        sub_graph.vertices = [{}] * len(self.local_graph.vertices)
```
bo156, line 138: instead of this initialization you can just create an empty list and append to it inside the for loop

checkov/terraform/graph_builder/variable_rendering/renderer.py
```diff
@@ -80,7 +80,10 @@ def attributes_no_eval(self, attribute: str, vertex_index: int) -> bool:
     def evaluate_vertex_attribute_from_edge(self, edge_list: List[Edge]) -> None:
         multiple_edges = len(edge_list) > 1
         edge = edge_list[0]
-        origin_vertex_attributes = self.local_graph.vertices[edge.origin].attributes
+        try:
+            origin_vertex_attributes = self.local_graph.vertices[edge.origin].attributes
+        except AttributeError:
```
bo156, line 85: why would it raise an attribute error here? (asking more to understand than adding a review comment)

bo156 (changes requested):

checkov/terraform/graph_builder/foreach_handler.py
ChanochShayner: It can be at the top due to circular imports

ChanochShayner: It can be at the top due to circular imports

tronxd: So try solving the circular import.. that's not a solution

checkov/terraform/graph_builder/foreach_handler.py
ChanochShayner: First, we are running over all the resources and getting the statics -> all the None is probably a dynamic -> building a sub-graph for all the dynamic and rendering them -> need to validate if now they are static - this is why we are checking static in the dynamic flow

checkov/terraform/graph_builder/foreach_handler.py
ChanochShayner: changed.

checkov/terraform/graph_builder/foreach_handler.py
ChanochShayner: we need the length to equal to the local graph + every vertice should be in the same index like the local graph

checkov/terraform/graph_builder/variable_rendering/renderer.py
ChanochShayner: In the subgraph, we are adding a non/{} vertices.

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -137,16 +137,14 @@ def _build_sub_graph(self, blocks_to_render: list[int]) -> TerraformLocalGraph:
         sub_graph = TerraformLocalGraph(self.local_graph.module)
         sub_graph.vertices = [{}] * len(self.local_graph.vertices)
         for i, block in enumerate(self.local_graph.vertices):
-            if block.block_type == BlockType.RESOURCE and i not in blocks_to_render:
-                sub_graph.vertices[i] = {}
-            else:
-                sub_graph.vertices[i] = block  # type:ignore
+            if not (block.block_type == BlockType.RESOURCE and i not in blocks_to_render):
+                sub_graph.vertices[i] = block  # type: ignore
```
tronxd, line 141: Remove the `# type: ignore`

tronxd (approved):

checkov/terraform/graph_builder/foreach_handler.py
bo156: ok :)

checkov/terraform/graph_builder/foreach_handler.py
bo156: perhaps this method could just move to `local_graph.py`? it's not really related to for_each and might also be uses as a util elsewhere.

checkov/terraform/graph_builder/foreach_handler.py
bo156: why is that a problem if we are going through the vertices in order?
from my perspective it seems like an extra line of code for no reason, but obviously you and @tronxd disagree so I would love to hear why :)

bo156 (approved):

checkov/terraform/graph_builder/foreach_handler.py
ChanochShayner: It doesn't help because it will still be circular -
local_graph should use my funcs.

PR #4430, opened 2023-02-06
platform(secrets): Add secrets custom regex on file

pazbechor:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Adding the option to apply regex policies on file (till now one could only apply regex policies for single line)

## Checklist:


checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -26,9 +26,16 @@ class CustomRegexDetector(RegexBasedDetector):
     def __init__(self) -> None:
         self.regex_to_metadata: dict[str, dict[str, Any]] = dict()
         self.denylist = set()
+        self.multiline_denylist = set()
+        self.multiline_regex_to_metadata: dict[str, dict[str, Any]] = dict()
+        self._did_analyzed_file: dict[str, bool] = dict()
```
nimrodkor, line 31:
```suggestion
        self._analyzed_files: set[str] = set()
```

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,84 @@ def analyze_line(
...
+            string_to_analyze=line,
+            output=output,
+            line_number=line_number,
+            context=raw_context,
+            is_multiline=False,
+            **kwargs
+        )
+
+        if not self._did_analyzed_file.get(filename):
+            self._did_analyzed_file[filename] = True
```
nimrodkor, line 65:
```suggestion
        if filename not in self._analyzed_files:
            self._analyzed_files.add(filename)
```

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,84 @@ def analyze_line(
...
+            file_content = None
+            try:
+                with open(filename, 'r') as f:
+                    file_content = f.read()
+            except Exception:
+                logging.error(
```
nimrodkor, line 71: Errors sometimes cause people to think checkov stopped working. Warning, however, tells there was a specific issue here that doesn't affect the entire run. Let's use `warning`
```suggestion
                logging.warning(
```

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,84 @@ def analyze_line(
...
                 logging.info(
                     f'Finding for check {ps.check_id} are not 5-100 characters in length, was ignored')  # type: ignore
 
-        return output
-
-    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type:ignore[override]
-        for regex in self.denylist:
+    def analyze_string(self, string: str, regex_denylist: Optional[Set] = None, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type: ignore # type:ignore[override]
+        if regex_denylist is None:
+            regex_denylist = self.denylist
```
nimrodkor, line 129: @pazbechor Please fetch the `regex_denylist` from the kwargs, and then create a local list unifying the self.denylist with the regex_denylist. I don't want to break signatures

nimrodkor (commented):
@pazbechor Nice! I like the approach. Can you add tests?

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -26,9 +26,16 @@ class CustomRegexDetector(RegexBasedDetector):
     def __init__(self) -> None:
         self.regex_to_metadata: dict[str, dict[str, Any]] = dict()
         self.denylist = set()
+        self.multiline_denylist = set()
```
lirshindalman, line 29:
```suggestion
        self.multiline_deny_list = set()
```

checkov/secrets/plugins/custom_regex_detector.py
pazbechor: pls check if what I did is fine :)

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,85 @@ def analyze_line(
...
                 logging.info(
                     f'Finding for check {ps.check_id} are not 5-100 characters in length, was ignored')  # type: ignore
 
-        return output
-
-    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type:ignore[override]
-        for regex in self.denylist:
+    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type: ignore # type:ignore[override]
+        regex_denylist: Optional[Set[Pattern[str]]] = set()
+        regex_denylist = kwargs.get("regex_denylist", self.denylist)  # type: ignore
```
nimrodkor, line 130: Let's try to reduce mypy ignores please 🙏 . Also, please don't do `# type: ignore` as it is a general, dangerous suppression
```suggestion
    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type:ignore[override]
        regex_denylist: Set[Pattern[str]] = kwargs.get("regex_denylist") or self.denylist  # type:ignore[assignment]
        for regex in regex_denylist:
```

nimrodkor (approved):

checkov/secrets/plugins/custom_regex_detector.py
pazbechor: talked about it   :)

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,85 @@ def analyze_line(
...
-            ps.check_id = self.regex_to_metadata[regex.pattern]["Check_ID"]  # type:ignore[attr-defined]
-            if len(cast(str, ps.secret_value)) in range(MIN_CHARACTERS, MAX_CHARACTERS) or not regex_data['isCustom']:
+            regex_data = current_regex_to_metadata[regex.pattern]
+            # Wrap multiline match with fstring + ''
+            match = f"'{match}'" if is_multiline else match
+            ps = PotentialSecret(
+                type=regex_data["Name"],
+                filename=filename,
+                secret=match,
+                line_number=line_number,
```
maxamel, line 116: If we're scanning the entire file, what line_number will be shown in case of a match?

pazbechor: yup, I don't think it's relevant as it will indicate the error any way. Do you think it should be changed to 0 or something like that ?

maxamel: Yes default of zero for multiline match sounds reasonable

lirshindalman (approved):
nice!!

checkov/secrets/plugins/custom_regex_detector.py
nimrodkor: Still need to take the mypy reductions @pazbechor

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,85 @@ def analyze_line(
...
                 output.add(ps)
             else:
                 logging.info(
                     f'Finding for check {ps.check_id} are not 5-100 characters in length, was ignored')  # type: ignore
 
-        return output
-
-    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type:ignore[override]
-        for regex in self.denylist:
+    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type: ignore # type:ignore[override]
```
nimrodkor, line 128:
```suggestion
    def analyze_string(self, string: str, **kwargs: Optional[Dict[str, Any]]) -> Generator[Tuple[str, Pattern[str]], None, None]:  # type:ignore[override]
```

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -43,26 +50,85 @@ def analyze_line(
...
+        regex_denylist: Set[Pattern[str]] = set()
```
nimrodkor, line 129: This line is garbage @pazbechor

PR #4449, opened 2023-02-08
fix(terraform): fix CKV2_AWS_46 (false positive on connection with s3 via logging_config)

hfastek:

## Description

False positive on:
`aws_cloudfront_distribution` with origin **other** than S3, but _connected_ via `logging_config` with `aws_s3_bucket`.

## Edited policies:
```
  id: "CKV2_AWS_46"
  name: "Ensure AWS Cloudfront Distribution with S3 have Origin Access set to enabled"
```

### Fix
Changed too wide _connection_ condition to specific attribute:

```
attribute: "origin.*.domain_name"
```

## Checklist:

### DIFF
--- a/checkov/terraform/checks/graph_checks/aws/CLoudFrontS3OriginConfigWithOAI.yaml
+++ b/checkov/terraform/checks/graph_checks/aws/CLoudFrontS3OriginConfigWithOAI.yaml
@@ -4,18 +4,12 @@ metadata:
   category: "IAM"
 definition:
   or:
-    - and:
-        - cond_type: filter
-          attribute: resource_type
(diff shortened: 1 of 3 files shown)


gruebel:
thanks for the contribution, the proposed change is sadly a bit too restrictive, because the bucket could also be passed via a `local` or `var` reference.

hfastek:
True, and it is decision between false positive (`logging_config`) and false negative (local, var).
I guess false positive (as is now) with skip would be better option.
Or maybe you have other idea how to fix above false positive?

gruebel:
sadly not yet, need to take a deeper look into it

PR #4456, opened 2023-02-08
platform(general): Use new enforcement categories

mikeurbanski1:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Updates checkov to use the new enforcement rules from the platform. This is meant to be a short term solution to unblock the release. As you will see, it suffers from some pretty major flaws related to the fact that we do not have a 1-1 mapping from runners to code categories (the SCA runners have two categories). This means that when handling `should_run_check` and `get_exit_code` operations, we can no longer just consider the runner, but also have to consider the types of checks involved as well.

It is ugly, moreso because of the type hints, but you will see by comparing the regular (non-SCA) branches and SCA branches that all we are really doing is duplicating the logic while splitting it for licenses and vulns (repeating for each map entry), so the actual added complexity is not significant. It's just verbose.

## Checklist:


gruebel (approved):
it ain't pretty, but we should handle the SCA special case in a follow up PR

ayajbara (approved):

PR #4472, opened 2023-02-09
Updated RDSDeletionProtection.py for aws_db_instance

Jhanvi12:
I have updated rule CKV_AWS_139 (File -> RDSDeletionProtection.py) to check deletion protection for aws_db_instance also.


JamesWoolfenden (changes requested):
Needs associate tests and new testcase/data for each

checkov/terraform/checks/resource/aws/IAMStarResourcePolicyDocument.py
```diff
@@ -0,0 +1,38 @@
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.common.util.type_forcers import force_list, extract_policy_dict
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from typing import List
+
+
+class IAMStarResourcePolicyDocument(BaseResourceCheck):
```
gruebel, line 7: this looks like a wrong addition 😄

checkov/terraform/checks/resource/aws/RDSDeletionProtection.py
```diff
@@ -7,7 +7,7 @@ class RDSDeletionProtection(BaseResourceValueCheck):
     def __init__(self):
         name = "Ensure that RDS clusters have deletion protection enabled"
         id = "CKV_AWS_139"
-        supported_resources = ['aws_rds_cluster']
+        supported_resources = ['aws_rds_cluster' ,'aws_db_instance']
```
gruebel, line 10: @JamesWoolfenden what do you think? looking at the name of the check we would either need to change/extend the name or create a separate check for it.

JamesWoolfenden: id go with a separate check one for db instance, we went that way for encryption and CMK checks and it lets have the control on suppression.

JamesWoolfenden: a new check for 'aws_db_instance' and new test cases and test for that new check

Jhanvi12:
The rule is AWS DB Instance deletion protection .In need to merge from my local branch master to main bridgecrew master branch

Jhanvi12:
> Needs associate tests and new testcase/data for each

Hello created test case for that , the link is https://github.com/Jhanvi12/checkov/blob/master/tests/terraform/checks/resource/aws/test_AWSDatabaseInstanceDeletionProtection.py

JamesWoolfenden:
you're trying to merge into the wrong branch, thats why theres so many commit, main is your target branch.

PR #4476, opened 2023-02-10
fix(cloudformation): Updated AWS_CKV_7 to not require rotation on asymmetric keys

### LINKED ISSUE #4475: AWS_CKV_7 False Positive on assymetric key check in Cloudformation
**Describe the issue**
In terraform, the check avoids false positives with an extra check against symmetric keys before checking whether rotation is enabled.  This same check hasn't been configured for cloudformation:

```
        # Only symmetric keys support auto rotation. The attribute is optional and defaults to symmetric.
        spec = conf.get('customer_master_key_spec')
        if not spec or 'SYMMETRIC_DEFAULT' in spec:
        else:
            return CheckResult.PASSED
```

**Examples**

```
RSASigningKey:
  Type: 'AWS::KMS::Key'
  Properties:
    Description: RSA-3072 asymmetric KMS key for signing and verification
    KeySpec: RSA_3072
    KeyUsage: SIGN_VERIFY
    KeyPolicy:
      Version: 2012-10-17
      Id: key-default-1
      Statement:
        - Sid: Enable IAM User Permissions
          Effect: Allow
          Principal:
            AWS: 'arn:aws:iam::111122223333:root'
          Action: 'kms:*'
          Resource: '*'
        - Sid: Allow administration of the key
          Effect: Allow
          Principal:
            AWS: 'arn:aws:iam::111122223333:role/Admin'
          Action:
            - 'kms:Create*'
            - 'kms:Describe*'
            - 'kms:Enable*'
            - 'kms:List*'
            - 'kms:Put*'
            - 'kms:Update*'
            - 'kms:Revoke*'
            - 'kms:Disable*'
            - 'kms:Get*'
            - 'kms:Delete*'
            - 'kms:ScheduleKeyDeletion'
            - 'kms:CancelKeyDeletion'
          Resource: '*'
        - Sid: Allow use of the key
          Effect: Allow
          Principal:
            AWS: 'arn:aws:iam::111122223333:role/Developer'
          Action:
            - 'kms:Sign'
            - 'kms:Verify'
            - 'kms:DescribeKey'
          Resource: '*'
```

 - Checkov Version [e.g. 2.3.0]

**Additional context**

This blocks checkov working for assymetric keys in CFN.

andyloughran:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Updated cloudformation and terraform checks to not expect key rotation for asymmetric keys as listed here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html#cfn-kms-key-keyspec

Fixes #4475 

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/KMSRotation.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSRotation.py
@@ -19,10 +19,8 @@ class KMSRotation(BaseResourceValueCheck):
             if 'KeySpec' in conf['Properties'].keys():
                 spec = conf['Properties']['KeySpec']
                 if not spec or 'SYMMETRIC_DEFAULT' in spec or 'HMAC' in spec:
-                    print("Symmetric Key")
                     return super().scan_resource_conf(conf)
                 else:
-                    print("Found Assymetric Key")
                     return CheckResult.PASSED
         return super().scan_resource_conf(conf)
 
--- a/checkov/cloudformation/checks/resource/aws/KMSRotation.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSRotation.py
@@ -1,4 +1,4 @@
-from checkov.common.models.enums import CheckCategories
+from checkov.common.models.enums import CheckCategories, CheckResult
 from checkov.cloudformation.checks.resource.base_resource_value_check import BaseResourceValueCheck
 
 
@@ -12,6 +12,19 @@ class KMSRotation(BaseResourceValueCheck):
 
     def get_inspected_key(self) -> str:
         return "Properties/EnableKeyRotation"
+
+    def scan_resource_conf(self, conf):
+        # Only symmetric keys support auto rotation. The attribute is optional and defaults to symmetric.
+        if 'Properties' in conf.keys():
+            if 'KeySpec' in conf['Properties'].keys():
+                spec = conf['Properties']['KeySpec']
+                if not spec or 'SYMMETRIC_DEFAULT' in spec or 'HMAC' in spec:
+                    print("Symmetric Key")
+                    return super().scan_resource_conf(conf)
+                else:
+                    print("Found Assymetric Key")
+                    return CheckResult.PASSED
+        return super().scan_resource_conf(conf)
 
 
 check = KMSRotation()
--- a/checkov/cloudformation/checks/resource/aws/KMSRotation.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSRotation.py
@@ -15,13 +15,12 @@ class KMSRotation(BaseResourceValueCheck):
 
     def scan_resource_conf(self, conf):
         # Only symmetric keys support auto rotation. The attribute is optional and defaults to symmetric.
-        if 'Properties' in conf.keys():
-            if 'KeySpec' in conf['Properties'].keys():
-                spec = conf['Properties']['KeySpec']
-                if not spec or 'SYMMETRIC_DEFAULT' in spec or 'HMAC' in spec:
-                    return super().scan_resource_conf(conf)
-                else:
-                    return CheckResult.PASSED
+        properties = conf.get("Properties")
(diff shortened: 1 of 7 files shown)


checkov/cloudformation/checks/resource/aws/KMSRotation.py
```diff
@@ -13,5 +13,16 @@ def __init__(self) -> None:
...
+    def scan_resource_conf(self, conf):
+        # Only symmetric keys support auto rotation. The attribute is optional and defaults to symmetric.
+        if 'Properties' in conf.keys():
+            if 'KeySpec' in conf['Properties'].keys():
+                spec = conf['Properties']['KeySpec']
+                if not spec or 'SYMMETRIC_DEFAULT' in spec or 'HMAC' in spec:
+                    return super().scan_resource_conf(conf)
+                else:
+                    return CheckResult.PASSED
+        return super().scan_resource_conf(conf)
```
gruebel, line 25:
```suggestion
        properties = conf.get("Properties")
        if properties and isinstance(properties, dict):
            spec = properties.get("KeySpec")
            if spec and isinstance(spec, str):
                if 'SYMMETRIC_DEFAULT' not in spec and 'HMAC' not in spec:
                    return CheckResult.UNKNOWN
        return super().scan_resource_conf(conf)
```
the result should be `UNKNOWN`, because as you said, you can't enable key rotation for asymmetric keys, so the check is actually not valid for this setup. And I also suggest a bit different logic to be more on the safer side 🙂

checkov/terraform/checks/resource/aws/KMSRotation.py
```diff
@@ -16,7 +16,7 @@ def get_inspected_key(self):
     def scan_resource_conf(self, conf):
         # Only symmetric keys support auto rotation. The attribute is optional and defaults to symmetric.
         spec = conf.get('customer_master_key_spec')
-        if not spec or 'SYMMETRIC_DEFAULT' in spec:
+        if not spec or 'SYMMETRIC_DEFAULT' in spec or 'HMAC' in spec:
             return super().scan_resource_conf(conf)
         else:
             return CheckResult.PASSED
```
gruebel, line 22: also would be great, if you can change the the result to `UNKNOWN`

gruebel (commented):
nice, great work 💪  Added 2 comments

checkov/terraform/checks/resource/aws/KMSRotation.py
andyloughran: Heya @gruebel,
Re-read the document and agree.  This check isn't relevant for assymetric keys, so once they're identified as not symmetric, the check should not apply.  Is `UNKNOWN` the right result in that instance?

andyloughran: Done the switch

andyloughran:
@gruebel after reading the enum documentation, I'm considering `SKIPPED` ahead of `UNKNOWN`, as arguably the check should be re-scoped to only check Symmetric keys ?

gruebel:
> @gruebel after reading the enum documentation, I'm considering `SKIPPED` ahead of `UNKNOWN`, as arguably the check should be re-scoped to only check Symmetric keys ?

`SKIPPED` is actually misleading here, because we use it to indicate a check was intentionally skipped by the user not by us. For this case we use `UNKNOWN` 🙂

andyloughran:
> > @gruebel after reading the enum documentation, I'm considering `SKIPPED` ahead of `UNKNOWN`, as arguably the check should be re-scoped to only check Symmetric keys ?
> 
> `SKIPPED` is actually misleading here, because we use it to indicate a check was intentionally skipped by the user not by us. For this case we use `UNKNOWN` 🙂

Pukka, I will defer to your wisdom and experience on this - thanks for the review! :)

bo156 (approved):

ayajbara (approved):

PR #4487, opened 2023-02-13
feat(terraform): support more json encoded objects as part of terraform resource and fix evaluation of true/false in json

bo156:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Changed `base_attribute_solver` to always try to load a json object. Up until now we only did it for `policy` attribute, but as many places could have this behavior I changed it to be the default.
I also added a new test that checks it using a custom policy for `aws_emr_cluster`

## Checklist:

### DIFF
--- a/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
@@ -84,7 +84,7 @@ class BaseAttributeSolver(BaseSolver):
         for attr in attr_parts:
             attr_to_check = f'{attr_to_check}.{attr}' if attr_to_check else attr
             value_to_check = vertex.get(attr_to_check)
-            value_to_check = self._render_json_str(value_to_check, attr, vertex)
+            value_to_check = self._render_json_str(value_to_check)
 
             # we can only check is_attribute_value_check when evaluating the full attribute
             # for example, if we have a policy that says "tags.component exists", and tags = local.tags, then
@@ -199,11 +199,27 @@ class BaseAttributeSolver(BaseSolver):
         return False
 
     @staticmethod
-    def _render_json_str(value_to_check: Any, attr: str, vertex: Dict[str, Any]) -> Any:
-        if attr == 'policy' and vertex.get('resource_type', '').endswith('policy'):
-            try:
-                value_to_check = json.loads(value_to_check)
-                return value_to_check
-            except Exception as e:
-                logging.info(f'cant parse policy str to object, {str(e)}')
+    def _render_json_str(value_to_check: Any) -> Any:
+        """
+        Tries to render objects containing json dict, for example:
+        - resource "aws_iam_policy" "example1" {
+            policy == <<POLICY
+                {
+                ... json-data
+                }
+            POLICY
+        }
+        - resource "aws_emr_security_configuration" "example2" {
+            configuration = <<EOF
+                {
+                ... json-data
+                }
+            EOF
+        }
+        """
+        try:
+            value_to_check = json.loads(value_to_check)
+            return value_to_check
+        except Exception as e:
+            logging.info(f'cant parse policy str to object, {str(e)}')
         return value_to_check
--- a/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
@@ -218,8 +218,10 @@ class BaseAttributeSolver(BaseSolver):
         }
         """
         try:
-            value_to_check = json.loads(value_to_check)
-            return value_to_check
+            # Check if the value looks like a json object, and if it is try to parse it
+            if isinstance(value_to_check, str) and value_to_check.strip().startswith('{'):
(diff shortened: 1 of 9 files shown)


checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -199,11 +199,27 @@ def _is_variable_dependant(value: Any, source: str) -> bool:
...
+        - resource "aws_emr_security_configuration" "example2" {
+            configuration = <<EOF
+                {
+                ... json-data
+                }
+            EOF
+        }
+        """
+        try:
+            value_to_check = json.loads(value_to_check)
```
gruebel, line 221: I don't think this is a good idea, because it will also change none JSON strings, like `"true" -> True` or `"12345" -> 12345`. We should at least do a string lookup, before trying to cast to dict/list.
Also not sure, if someone does it, but if someone actually wants the value to be a string, then we will now break it.
ex. `name = "[\"some name\"]"` so this wouldn't be a feature, it would be a breaking change.

gruebel (changes requested):

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
bo156: got you, my intention was we don't want to add support for each attribute terraform allow you to add as an inlne configuration (to reduce future maintenance cost).
Do you have any idea how to do it better?
If not, I will just regress it to the way it was before and add support for `configuration` as well

tests/terraform/graph/checks/resources/CustomAwsEMRSecurityConfiguration/main.tf
```diff
@@ -0,0 +1,91 @@
+provider "aws" {
```
tronxd, line 1: Add a test of edge cases where the attribute is `"true"` (every boolean attribute), and "`5`" (you can use data retention attributes which accepts a number)

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
tronxd, line 221: I'm worrying about performance degradation here, since we will try parsing every attribute into a JSON.
Implement a mechanism which lookups a JSON object starting delimiters (`{` for object , `[` for an array of objects), and parse into a JSON only when there's a match

bo156: Do you think it will be better to do it like that if you also relate to what @gruebel said about this part?
I mainly try to understand if it is even ok to try to manipulate each attribute like that, even if it does look like a json object.

gruebel: if we target a more general solution then this would be the minimum to do to make sure, we don't try to parse everything as json.
the other approach would be to maintain a map of resource + argument, which should be parsed as json.

tests/terraform/graph/checks/resources/CustomAwsEMRSecurityConfiguration/main.tf
bo156: done

tronxd (approved):

checkov/terraform/checks/resource/aws/EMRClusterIsEncryptedKMS.py
```diff
@@ -16,7 +16,7 @@ def scan_resource_conf(self, conf):
         if 'configuration' not in conf:
             return CheckResult.SKIPPED
         configuration = conf['configuration'][0]
-        if "SSE-KMS" in configuration:
+        if "SSE-KMS" in str(configuration):
```
gruebel, line 19: ok, this is needed, because it is now a `dict`? could change it to a `BaseValueResourceCheck`? then the inspected key should be something like `configuration. EncryptionConfiguration.AtRestEncryptionConfiguration. S3EncryptionConfiguration. EncryptionMode` and expected value `SSE-KMS`

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -199,11 +199,29 @@ def _is_variable_dependant(value: Any, source: str) -> bool:
...
+            configuration = <<EOF
+                {
+                ... json-data
+                }
+            EOF
+        }
+        """
+        try:
+            # Check if the value looks like a json object, and if it is try to parse it
+            if isinstance(value_to_check, str) and value_to_check.startswith('{'):
```
gruebel, line 222: let's go with that 👍

checkov/terraform/checks/resource/aws/EMRClusterIsEncryptedKMS.py
bo156: Yeah it's beacuse it'ss a dict :)
I actually checked it out but I thought there might be more places to place `SSE-KMS` in (maybe not this field alone I mean). If there are no more places I will change it

gruebel (approved):
🍪

PR #4493, opened 2023-02-14
feat(terraform): Assign/override main vertices by the first new vertice.

ChanochShayner:

## Description
Assign/override main vertices by the first new vertice to avoid mismatch in the vertices indexes.

## Checklist:

### DIFF
--- a/checkov/terraform/graph_builder/foreach_handler.py
+++ b/checkov/terraform/graph_builder/foreach_handler.py
@@ -158,7 +158,7 @@ class ForeachHandler(object):
     def _create_new_resources_count(self, statement: int, block_idx: int) -> None:
         main_resource = self.local_graph.vertices[block_idx]
         for i in range(statement):
-            resource_idx = block_idx if i == 0 else None
+            resource_idx = block_idx if i == 0 else -1
             self._create_new_resource(main_resource, i, resource_idx=resource_idx)
 
     @staticmethod
@@ -207,7 +207,7 @@ class ForeachHandler(object):
             main_resource: TerraformBlock,
             new_value: int | str,
             new_key: Optional[str] = None,
-            resource_idx: Optional[int] = None
+            resource_idx: int = -1
     ):
         new_resource = deepcopy(main_resource)
         block_type, block_name = new_resource.name.split('.')
@@ -224,7 +224,7 @@ class ForeachHandler(object):
 
         idx_to_change = new_key or new_value
         self._add_index_to_block_properties(new_resource, idx_to_change)
-        if resource_idx is not None:
+        if resource_idx != -1:
             self.local_graph.vertices[resource_idx] = new_resource
         else:
             self.local_graph.vertices.append(new_resource)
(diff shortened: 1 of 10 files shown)


checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -216,21 +224,21 @@ def _create_new_resource(self, main_resource: TerraformBlock, new_value: int | s
...
+            self.local_graph.vertices.append(new_resource)
 
-    def _create_new_resources_foreach(self, statement: list[str] | dict[str, Any], main_resource: TerraformBlock) -> None:
+    def _create_new_resources_foreach(self, statement: list[str] | dict[str, Any], block_idx: int) -> None:
+        main_resource = self.local_graph.vertices[block_idx]
         if isinstance(statement, list):
-            for new_value in statement:
-                self._create_new_resource(main_resource, new_value, new_key=new_value)
+            for i, new_value in enumerate(statement):
+                resource_idx = block_idx if i == 0 else None
```
tronxd, line 236:
```suggestion
                resource_idx = block_idx if i == 0 else -1
```
Let's keep indices with their original type

tronxd (approved):

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -216,21 +224,21 @@ def _create_new_resource(self, main_resource: TerraformBlock, new_value: int | s
 
         idx_to_change = new_key or new_value
         self._add_index_to_block_properties(new_resource, idx_to_change)
-        self.local_graph.vertices.append(new_resource)
+        if resource_idx is not None:
+            self.local_graph.vertices[resource_idx] = new_resource
```
bo156, line 228: This logic is a bit scary in my opinion, as it basically relies on the fact that you need to pass the correct input to this function in order for it to work, meaning sending `resoucre_idx = None` or not pass it at all.
Instead you can move line 161 to be just here, so the logic of assigning `None` to `resource_idx` will be part of the function itself and not separated from it.

checkov/terraform/graph_builder/foreach_handler.py
bo156, line 236: seems like you are also doing the same here, so moving this logic like I said earlier will actually save you this line as well

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -216,21 +224,21 @@ def _create_new_resource(self, main_resource: TerraformBlock, new_value: int | s
...
-            for new_key, new_value in statement.items():
-                self._create_new_resource(main_resource, new_value, new_key=new_key)
-
-    def _delete_main_resource(self, block_index_to_statement: FOR_EACH_BLOCK_TYPE) -> None:
-        new_vertices = [
-            block for i, block in enumerate(self.local_graph.vertices) if i not in block_index_to_statement or not block_index_to_statement[i]
-        ]
-        self.local_graph.vertices = new_vertices
+            for i, (new_key, new_value) in enumerate(statement.items()):
+                resource_idx = block_idx if i == 0 else None
```
bo156, line 240: again

bo156 (changes requested):

checkov/terraform/graph_builder/foreach_handler.py
```diff
@@ -200,7 +201,14 @@ def _build_key_to_val_changes(new_val: str, new_key: str):
             EACH_KEY: new_key
         }
 
-    def _create_new_resource(self, main_resource: TerraformBlock, new_value: int | str, new_key: Optional[str] = None):
+    def _create_new_resource(
+            self,
+            main_resource: TerraformBlock,
```
YaaraVerner, line 206: you don't need to get the main_resource as a parameter, you are getting the `resource_idx` so you can get it directly from the graph

bo156 (approved):

PR #4514, opened 2023-02-16
fix(secrets):  import git

lirshindalman:

Failed to initialize: Bad git executable

## Checklist:

### DIFF
--- a/checkov/secrets/scan_git_history.py
+++ b/checkov/secrets/scan_git_history.py
@@ -1,12 +1,17 @@
 from __future__ import annotations
 
-import git
 import logging
 from typing import TYPE_CHECKING, Dict
 from detect_secrets.core import scan
 
 if TYPE_CHECKING:
     from detect_secrets import SecretsCollection
+
+try:
+    import git
+    git_import_error = None
+except ImportError as e:
+    git_import_error = e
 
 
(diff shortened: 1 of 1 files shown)


Eliran-Turgeman:
I think we should set env var: GIT_PYTHON_REFRESH=quiet
https://github.com/gitpython-developers/GitPython/issues/816

checkov/secrets/scan_git_history.py
```diff
@@ -1,16 +1,26 @@
...
 
 if TYPE_CHECKING:
     from detect_secrets import SecretsCollection
 
+os.environ["GIT_PYTHON_REFRESH"] = "quiet"
+try:
+    import git
+    git_import_error = None
+except ImportError as e:
+    git_import_error = e
```
marynaKK, line 16: can we add here a log for the import error also, not only in the function? :)

marynaKK (approved):
🍒

checkov/secrets/scan_git_history.py
lirshindalman: I don't want to add the same log twice. if we don't use the function, the warning is irrelevant.

Eliran-Turgeman (approved):

PR #4590, opened 2023-03-01
fix(cloudformation): KMSKeyWildCardPrincipal modification - Check for wildcards inside of lists

MattJaccino:
My company just found a case of a dev using a wildcard inside of an AWS principal list to evade this rule.  Since the `principal_value` in this code could be either a string or a list, checking for membership as well as equality would prevent the circumnavigation of the rule.

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"
# cloudformation: KMSKeyWildCardPrincipal modification - Check for wildcards inside of lists

## Description

My company just found a case of a dev using a wildcard inside of an AWS principal list to evade this rule.  Since the `principal_value` in this code could be either a string or a list, checking for membership as well as equality would prevent the circumnavigation of the rule.

Fixes # (issue)

Ability to skate by the check by placing the wildcard within a list

### Description
It undermines the rule

### Fix
Check for membership as well

## Checklist:

### DIFF
--- a/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
@@ -49,7 +49,7 @@ class KMSKeyWildCardPrincipal(BaseResourceValueCheck):
                 for principal in principals_list:
                     if isinstance(principal, dict):
                         for principal_value in principal.values():
-                            if principal_value == '*':
+                            if principal_value == '*' or '*' in principal_value:
                                 return CheckResult.FAILED
                     else:
                         if principal == '*':
--- a/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
+++ b/checkov/cloudformation/checks/resource/aws/KMSKeyWildCardPrincipal.py
@@ -49,7 +49,7 @@ class KMSKeyWildCardPrincipal(BaseResourceValueCheck):
                 for principal in principals_list:
                     if isinstance(principal, dict):
                         for principal_value in principal.values():
-                            if principal_value == '*' or '*' in principal_value:
+                            if principal_value == '*' or (isinstance(principal_value, list) and '*' in principal_value):
                                 return CheckResult.FAILED
                     else:
                         if principal == '*':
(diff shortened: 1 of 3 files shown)


gruebel:
thanks for the PR, could you also add a resource example here?

MattJaccino:
Hi Anton,

Here's a snippet of a YAML CFN template that would pass the check as it is currently in Checkov even though it contains a wildcarded principal:

```yaml
  SomeKmsKey:
    Type: 'AWS::KMS::Key'
    Properties:
      KeyPolicy:
        Version: 2012-10-17
        Id: key-default-1
        Statement:
          - Sid: Enable IAM User Permissions
            Effect: Allow
            Principal:
              AWS: 
                - '*' # <-- wildcarded principal without condition
                - 'arn:aws:iam::123456789012:root'
            Action: 'kms:*'
            Resource: '*'
```

MattJaccino:
added a type check to resolve the errors in the tests, but also made me think that a wildcard can still be added if you take advantage of intrinsics in cfn

`!Join [ '', ['', '*' ] ]` would resolve to `*` after processing, but Checkov would just see it as an object.

This could be handled if all intrinsics in templates get "evaluated" prior to scanning, the way you already with with Refs with parameters with default values.

gruebel:
😄 sneaky dev
change looks good, can you also add it as a test?

gruebel:
> added a type check to resolve the errors in the tests, but also made me think that a wildcard can still be added if you take advantage of intrinsics in cfn
> 
> `!Join [ '', ['', '*' ] ]` would resolve to `*` after processing, but Checkov would just see it as an object.
> 
> This could be handled if all intrinsics in templates get "evaluated" prior to scanning, the way you already with with Refs with parameters with default values.

yeah support of evaluating this and also other intrinsic functions would be great, but currently not on our short term roadmap.

MattJaccino:
> 😄 sneaky dev change looks good, can you also add it as a test?

think I got the right file, yeah? added it to `tests/cloudformation/checks/resource/aws/example_KMSKeyWildCardPrincipal/KMSKeyWildCardPrincipal-FAILED-AWS-Wildcard.yaml`

gruebel:
yeah, folder looks good, but you also need to adjust the actual test here `tests/cloudformation/checks/resource/aws/test_KMSKeyWildCardPrincipal.py` just increment the failed value by 1 🙂

MattJaccino:
done

gruebel (approved):
thanks, great jon 🏆

gruebel:
Only thing, which you forgot is to add it to the test file 😄 
https://github.com/bridgecrewio/checkov/blob/main/tests/ansible/checks/graph_checks/test_yaml_policies.py

Eliran-Turgeman (approved):

PR #4610, opened 2023-03-05
feat(gitlab_ci): fix wrong resource in gitlab-ci

maxamel:
This PR fixes wrong resource given to gitlab_ci findings.
We now skip string/list findings in resource generation of gitlab_ci and require only jobs(i.e. dicts) to be processed.
Note - multiple images per resource are not handled in this PR and will need to be handled in a future task. 

## Checklist:

### DIFF
--- a/tests/gitlab_ci/test_runner.py
+++ b/tests/gitlab_ci/test_runner.py
@@ -21,7 +21,7 @@ class TestRunnerValid(unittest.TestCase):
         )
         self.assertEqual(len(report.failed_checks), 5)
         self.assertEqual(report.parsing_errors, [])
-        self.assertEqual(len(report.passed_checks), 7)
+        self.assertEqual(len(report.passed_checks), 9)
         self.assertEqual(report.skipped_checks, [])
         report.print_console()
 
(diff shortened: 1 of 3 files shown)


marynaKK (approved):

tests/gitlab_ci/test_runner.py
```diff
@@ -55,8 +55,25 @@ def test_runner_image_check(self):
...
+            runner_filter=RunnerFilter(framework=['gitlab_ci'], checks=['CKV_GITLABCI_3'])
+        )
+        self.assertEqual(len(report.failed_checks), 0)
+        self.assertEqual(report.parsing_errors, [])
         self.assertEqual(len(report.passed_checks), 4)
         self.assertEqual(report.skipped_checks, [])
+        self.assertEqual(report.passed_checks[0].resource, 'prebuild')
+        self.assertEqual(report.passed_checks[1].resource, 'build.image')
+        self.assertEqual(report.passed_checks[2].resource, 'deploy')
+        self.assertEqual(report.passed_checks[3].resource, 'deploy')
```
Eliran-Turgeman, line 76: shouldn't both resource ids be `deploy.services`?

maxamel: Depends on the specific case, but we're providing the most focused resource we can reliably compute. The reason we don't see `deploy.services` is because services in our example is a list, which doesn't have start/end lines, so cannot be found within the encapsulating dict. The image, on the other hand, is a dict which has start/end lines so that's why we output `build.image`. But as I wrote in the description, we'll need further work around this to handle multiple images per resource, which will force us to compute more focused resources, in which case for lists we'll probably compute something like `deploy.services[1]`. 😄

nimrodkor (approved):

gruebel (approved):

PR #4646, opened 2023-03-12
fix(general): check suppression id instead of policy id

PelegLi:

## Description

Fixing a case in which a suppressed custom policy fails

## Checklist:


rotemavni (commented):
If I understand correctly, we are changing the behavior of `get_policy_level_suppressions` to be a list of ids instead of `policyId`.

My concern here is when looking at where we use `get_policy_level_suppressions` - in `main.py`, we use it to perform `runner_filter.set_suppressed_policies`.
Those are assigned to `suppressed_policies` and it might affect the important `should_run_check` method.
How will the condition `(bc_check_id in self.suppressed_policies and bc_check_id not in self.bc_cloned_checks)` (line 234 in `runner_filter`) work given that change?

tests/common/integration_features/test_suppressions_integration.py
```diff
@@ -1052,7 +1052,7 @@ def test_get_policy_level_suppressions(self):
                                                       'resourceId': '/src/BC_AWS_LOGGING_7.tf:aws_cloudtrail.cloudtrail8'}],
                  'checkovPolicyId': 'acme_AWS_1668010000289'}]}
 
-        expected_suppressions = ['BC_AWS_GENERAL_175', 'BC_AWS_LOGGING_2', 'BC_K8S_26']
+        expected_suppressions = ['404088ed-4251-41ac-8dc1-45264af0c461', 'b68013bc-2908-4c9a-969d-f1640d4aca11', '271c1a79-2333-4a12-bf7d-55ec78468b94']
```
bo156, line 1055: TBH the previous format looks more readable.
Can you share why is this needed?

rotemavni (approved):

tests/common/integration_features/test_suppressions_integration.py
PelegLi: the previous is wrong, so readability is not a factor here

tronxd (approved):

PR #4647, opened 2023-03-12
fix(gitlab): Modify GitLab CI resource ids

Eliran-Turgeman:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/gitlab_ci/common/resource_id_utils.py
+++ b/checkov/gitlab_ci/common/resource_id_utils.py
@@ -3,16 +3,52 @@ from typing import Any
 
 from checkov.common.util.consts import START_LINE, END_LINE
 
+IMAGE_BLOCK_NAMES = ('image', 'services')
+
 
 def generate_resource_key_recursive(conf: dict[str, Any] | list[str] | str, key: str, start_line: int,
                                     end_line: int) -> str:
+    return _generate_resource_key_recursive(conf, key, start_line, end_line, set(), 0)
+
+
+def _generate_resource_key_recursive(conf: dict[str, Any] | list[str] | str, key: str, start_line: int,
+                                    end_line: int, scanned_image_blocks: set[str], depth: int) -> str:
     if not isinstance(conf, dict):
         return key
 
     for k, value in conf.items():
+        if k in IMAGE_BLOCK_NAMES:
+            scanned_image_blocks.add(k)
+
         if isinstance(value, dict) and value[START_LINE] <= start_line <= end_line <= value[END_LINE]:
             next_key = f'{key}.{k}' if key else k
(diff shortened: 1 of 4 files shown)


lirshindalman (approved):

checkov/gitlab_ci/common/resource_id_utils.py
```diff
@@ -3,16 +3,54 @@
...
+
+            for idx, value_dict in enumerate(value):
+                if value_dict[START_LINE] <= start_line <= end_line <= value_dict[END_LINE]:
+                    next_key += f'.{idx + 1}'
+                    break  # There can be only one match in terms of line range
+
+            return _generate_resource_key_recursive(value, next_key, start_line, end_line, scanned_image_blocks,
+                                                    depth + 1)
+
+        if any(block_name in conf.keys()
```
maxamel, line 39: Can this be extracted out of the encapsulating for loop?
Seems we're doing the same for every conf iteration

checkov/gitlab_ci/common/resource_id_utils.py
```diff
@@ -3,16 +3,54 @@
 
 from checkov.common.util.consts import START_LINE, END_LINE
 
+IMAGE_BLOCK_NAMES = ('image', 'services')
+
 
 def generate_resource_key_recursive(conf: dict[str, Any] | list[str] | str, key: str, start_line: int,
                                     end_line: int) -> str:
+    return _generate_resource_key_recursive(conf, key, start_line, end_line, set(), 0)
```
maxamel, line 11: Instead of creating another method you can just set the last two arguments as optional params in generate_resource_key_recursive and set their  default value to set() and zero.

gruebel: sounds good, but `scanned_image_blocks` should be `None` as default, otherwise you have a mutable object as default, which will result in some surprises, not good ones 😄

maxamel:
Seems the test results of  **test_runner_image_resources** should also be changed somehow, since it currently also produces same resource for several images 😄

checkov/gitlab_ci/common/resource_id_utils.py
Eliran-Turgeman: we need to do it at every iteration because we don't know in what depth these block names will appear

Eliran-Turgeman:
> Seems the test results of **test_runner_image_resources** should also be changed somehow, since it currently also produces same resource for several images 😄

That's a limitation we have since strings do not have the start line and end line fields.
I think we should merge this one as it will solve some of the edge cases.

maxamel (approved):

PR #4659, opened 2023-03-13
feat(secrets): allow scanning just partial history of commits

omryMen:

## Description

Currently doesn't have tests - added a task for that.
this was tested locally with platform 

allow scanning just partial history of commits - from commit - last_commit_scanned until current head

## Checklist:


checkov/secrets/scan_git_history.py
```diff
@@ -84,7 +88,11 @@ def _get_commits_diff(self) -> Dict[str, Dict[str, str | Dict[str, str]]]:
         except Exception as e:
             logging.error(f"Folder {self.root_folder} is not a GIT project {e}")
             return commits_diff
-        commits = list(repo.iter_commits(repo.active_branch))
+        if last_commit_sha:
+            start = repo.head.commit.hexsha
+            commits = list(repo.iter_commits(start + '..' + last_commit_sha))
```
maxamel, line 93: Shouldn't we take the commit right after last_commit_sha ?
I thought last_commit_sha is already scanned so this will create a duplication, no ?

maxamel (approved):
Nice! Added one small comment.

lirshindalman (approved):

checkov/secrets/scan_git_history.py
lirshindalman: no, we need the last commit we already scan to compare the new commit to

omryMen: it takes until that one, not included
it's like taking from a list [:last_idx] so last will not be inside

PR #4660, opened 2023-03-13
feat(terraform): s3 bucket has event notifications

JamesWoolfenden:
To capture:

*[S3.11] S3 buckets should have event notifications enabled
Related requirements: NIST.800-53.r5 CA-7, NIST.800-53.r5 SI-3(8), NIST.800-53.r5 SI-4, NIST.800-53.r5 SI-4(4)


gruebel (approved):
🍻

bo156 (approved):

PR #4813, opened 2023-03-31
fix(terraform): fix false positive CKV2_GCP_20 (fails for any non-MySQL instance)

nicolst:

## Description

CKV2_GCP_20 fails for any non-MySQL instance because of how the conditions are set up. This PR fixes this issue and brings it more in line with conditions used in similar checks.

## New/Edited policies
```
id: "CKV2_GCP_20"
name: "Ensure MySQL DB instance has point-in-time recovery backup configured"
category: "BACKUP_AND_RECOVERY"
```

### Description
Name suggests this policy should only apply to MySQL instances, but because of how the conditions are set up, it also fails for any PostgreSQL instance (even if it has point-in-time recovery enabled).

### Fix
Changed the condition from failing if the instance is not of type MYSQL, to succeeding if it is not. This seems to be more consistent with how it is implemented in other checks, e.g. [CKV2_GCP_13](https://github.com/bridgecrewio/checkov/blob/main/checkov/terraform/checks/graph_checks/gcp/GCPPostgreSQLDatabaseFlaglog_durationIsSetToON.yaml), [CKV2_GCP_14](https://github.com/bridgecrewio/checkov/blob/main/checkov/terraform/checks/graph_checks/gcp/GCPPostgreSQLDatabaseFlaglog_executor_statsIsSetToOFF.yaml).

The original check also checked whether the instance type started with "MYSQL_" or "mysql_", but instance types should always be in full caps, and the aforementioned checks do not check for both of these cases.

## Checklist:

### DIFF
--- a/checkov/terraform/checks/graph_checks/gcp/GCPMySQLdbInstancePoint_In_TimeRecoveryBackupIsEnabled.yaml
+++ b/checkov/terraform/checks/graph_checks/gcp/GCPMySQLdbInstancePoint_In_TimeRecoveryBackupIsEnabled.yaml
@@ -4,23 +4,16 @@ metadata:
   category: "BACKUP_AND_RECOVERY"
 
 definition:
-  and:
+  or:
     - cond_type: "attribute"
-      resource_types: "google_sql_database_instance"
+      resource_types: 
+        - "google_sql_database_instance"
+      attribute: "database_version"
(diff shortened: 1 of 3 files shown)


gruebel (approved):
nice, good job 🏆

bo156 (approved):

PR #4838, opened 2023-04-04
feat(secrets): Add fields to record of secrets in git history

marynaKK:
added 3 fields: `added_date`, `removed_date` and `added_by`
retrieved from git for each secret found.

## Checklist:


checkov/common/output/secrets_record.py
```diff
@@ -57,7 +57,10 @@ def __init__(self,
                  definition_context_file_path: Optional[str] = None,
                  validation_status: Optional[str] = None,
                  added_commit_hash: Optional[str] = None,
-                 removed_commit_hash: Optional[str] = None
+                 removed_commit_hash: Optional[str] = None,
+                 created_by: Optional[str] = None,
+                 removed_date: Optional[str] = None,
+                 create_date: Optional[str] = None
```
gruebel, line 63: this more about consistency in wording, but `added_by` and `added_date` would fit better to  `added_commit_hash`
This would mean to change in all the places you touched.

checkov/secrets/scan_git_history.py
```diff
@@ -153,13 +157,15 @@ def _run_scan_one_commit(commit: Dict[str, str | Dict[str, str]]) -> Tuple[List[
                     f"Found {len(file_results)} secrets in file path {file_name} in commit {commit_hash}")
                 results.append(RawStore(file_results=file_results, file_name=file_name, commit=commit,
                                         commit_hash=commit_hash, type=FILE_RESULTS_STR,
-                                        rename_from='', rename_to=''))
+                                        rename_from='', rename_to='', committer=commit.get(COMMIT_COMMITTER),
+                                        committed_datetime=commit.get(COMMIT_COMMITTED_DATETIME)))
```
omryMen, line 161: maybe we should put all metadata under one key ? 
meta: date create, date delete, commiter, hash ?

checkov/secrets/git_history_store.py
```diff
@@ -71,24 +81,30 @@ def _add_new_secret(self, secret_key: str,
...
-             'code_line': code_line
-             })
+        enriched_potential_secret: EnrichedPotentialSecret = {
+            'added_commit_hash': commit_hash,
+            'removed_commit_hash': '',
+            'potential_secret': secret,
+            'code_line': code_line,
+            'created_by': commit.get(COMMIT_COMMITTER, ''),
+            'removed_date': '',
+            'create_date': commit.get(COMMIT_COMMITTED_DATETIME, '')
```
gruebel, line 91: would be weird to not have the committer and datetime here, could create debug log for now, if we don't have that info

checkov/common/output/secrets_record.py
marynaKK: this would mean to change it also in platform + in the db migration
I can fix it but this will be the overhead 😅

checkov/secrets/scan_git_history.py
marynaKK: I can add another `TypedDict`
I thought `EnrichedPotentialSecretMetadata` can be enough
change it here too?

omryMen: I mean: currently the commit wiill look like 
{ hash: 'fsafsags', date: '2010..',  added_by: 'maryna' ... 'file.py': 'some diffs', 'file2.py': 'some diffs' }
what I mean is to do:
{ commit_meta: { hash: 'fsafsags', date: '2010..',  added_by: 'maryna' ...} 
files: 'file.py': 'some diffs', 'file2.py': 'some diffs' }

don't have to add the 'files' key, but the others can be united under one key and then it's more consolidated

omryMen (approved):

checkov/secrets/git_types.py
```diff
@@ -0,0 +1,76 @@
...
+    def remove_file(self, filename: str) -> None:
+        if self.files.get(filename):
+            del self.files[filename]
+
+
+class CommitMetadata:
+    __slots__ = ("commit_hash_key", "committer", "committed_datetime")
+
+    def __init__(self, commit_hash_key: str, committer: str, committed_datetime: str):
+        self.commit_hash_key: str = commit_hash_key
```
omryMen, line 61: I think it should be just commit_hash (without key)

lirshindalman (approved):

PR #4865, opened 2023-04-10
feat(general): 3D policies syntax refactor

Eliran-Turgeman:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Refactoring 3D policies syntax

New syntax example:

```
{
  "version": "v1",
  "definition": [
    {
      "cves": {
        "or": [
          {
            "and": [
              {
                "risk_factors": "DoS"
              },
              {
                "risk_factors": "Remote Execution"
              }
            ]
          }
        ]
      }
    },
    {
      "iac": {
        "or": [
          {
            "violation_id": "BC_K8S_29"
          },
          {
            "violation_id": "BC_K8S_23"
          }
        ]
      }
    }
  ]
}
```

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/policies_3d/checks_parser.py
+++ b/checkov/policies_3d/checks_parser.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
+import copy
 from typing import Dict, Any
 
 from checkov.common.output.record import Record
 from checkov.policies_3d.checks_infra.base_parser import Base3dPolicyCheckParser
 from checkov.policies_3d.checks_infra.base_check import Base3dPolicyCheck
-from checkov.policies_3d.syntax.syntax import Predicament, RiskFactorCVEContains
+from checkov.policies_3d.syntax.syntax import Predicament, RiskFactorCVEContains, ViolationIdEquals
 from checkov.sca_image.models import ReportCVE
+import itertools
 
 SUPPORTED_LOGICAL_OPERATORS = ['and', 'or']
 CVE_ATTRIBUTE_TO_PREDICATE_CLASS = {'risk_factors': RiskFactorCVEContains}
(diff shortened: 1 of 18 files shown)


checkov/policies_3d/checks_parser.py
```diff
@@ -18,3 +37,121 @@ def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -
...
+                value = [value] if isinstance(value, str) else value
+                top_level_predicament.predicates.append(RiskFactorCVEContains(value, cve_report))
+
+        nested_definition = cve_definition[top_level_logical_op]
+        nested_logical_op = None
+        for key, value in nested_definition[0].items():
+            if key in SUPPORTED_LOGICAL_OPERATORS:
+                nested_logical_op = key
+            elif key == 'risk_factors':
+                value = [value] if isinstance(value, str) else value
```
arielkru, line 88: you can also use `value = force_list(value)` here, simpler

checkov/policies_3d/checks_parser.py
```diff
@@ -18,3 +37,121 @@ def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -
...
+            nested_definition = nested_definition[0][nested_logical_op]
+            if not isinstance(nested_definition, list):
+                return None
+
+            nested_predicament = Predicament(logical_op=nested_logical_op)
+            for definition in nested_definition:
+                for key, value in definition.items():
+                    if key == 'risk_factors':
+                        value = [value] if isinstance(value, str) else value
+                        nested_predicament.predicates.append(RiskFactorCVEContains(value, cve_report))
```
arielkru, line 102: Maybe extract this to a separated function like `add_predicate(predicate_key, predicate_value, predicament)` or something similar. 2 benefits:
1. No duplicate code (this is very similar to lines 87-89)
2. In that way it can be a generic parsing so it can get any cve predicate (other than risk factor as well), and choose the right predicate by a map of key string to CVE predicate. That way adding new predicates would only require to add a new CVEPredicate inheritance, and this code would support it out of the box.

checkov/policies_3d/checks_parser.py
```diff
@@ -18,3 +37,121 @@ def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -
...
+
+        top_level_predicament = Predicament(logical_op='')
+        top_level_logical_op = ''
+        for key, value in iac_definition.items():
+            if key in SUPPORTED_LOGICAL_OPERATORS:
+                top_level_logical_op = key
+                top_level_predicament.logical_op = key
+            elif key == 'violation_id':
+                value = [value] if isinstance(value, str) else value
+                top_level_predicament.predicates.append(ViolationIdEquals(iac_record, value))
```
arielkru, line 127: The extracted function would also help here. So it could also get a value of "type" and then it would get the mapping by itself of cve / secret / iac depending on the type. The rest of the logic is identical

checkov/policies_3d/checks_parser.py
```diff
@@ -18,3 +37,121 @@ def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -
...
+                Predicament(
+                    logical_op='and',
+                    predicaments=[predicament for predicament in combination]
+                )
+            )
+
+        return check
+
+
+    def _create_cve_predicament(self, cve_report: ReportCVE) -> Predicament | None:
```
arielkru, line 64: It looks like the 3 functions of `_create_x_predicament` can be consolidated into one, as most of the logic is similar. So it can just be a function called `_create_predicament` and then handle the differences by a type argument or some mappings.
Another option is to create the abstract function for most of the similarity, and let each of these specific functions call the abstract one with specific values. makes sense?

tests/policies_3d/conftest.py
```diff
@@ -136,6 +136,44 @@ def policy_3d_3() -> dict[str, Any]:
...
+          'code': """{
+            "version": "v1",
+            "definition": [
+              {
+                "cves": {
+                  "or": [
+                    {
+                      "and": [
+                        {
+                          "risk_factors": "DoS"
```
arielkru, line 152: should be `risk_factor` as it can get only a single value

arielkru (commented):
Very nice!

omryMen:
isn't this version is v2 already :)

checkov/policies_3d/checks_parser.py
Eliran-Turgeman: extracted into `_create_predicate`

checkov/policies_3d/checks_parser.py
Eliran-Turgeman: extracted to `_create_module_predicament`

checkov/policies_3d/runner.py
```diff
@@ -28,13 +34,161 @@ class CVEReportAttribute(str, Enum):
...
+        true_iac_records = []
+        true_secrets_records = []
+
+        records_ids = set()
+
+        any_record_data_source = None
+
+        all_predicates = []
+        for predicament in check.predicaments:
+            all_predicates.extend(predicament.get_all_children_predicates())
```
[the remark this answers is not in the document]
Eliran-Turgeman: Good call, changed to a set

arielkru (approved):
🚀

PR #4889, opened 2023-04-13
fix(secrets): fix slicing in githistory 

marynaKK:
## Checklist:

### DIFF
--- a/checkov/common/util/decorators.py
+++ b/checkov/common/util/decorators.py
@@ -22,7 +22,10 @@ def time_it(func: Callable[P, T]) -> Callable[P, T]:
         end = default_timer()
 
         func_path = f"{func.__code__.co_filename.replace('.py', '')}.{func.__name__}"
-        logging.info(f"'{func_path}' took: {timedelta(seconds=end - start)}")
+        info = f"'{func_path}' took: {timedelta(seconds=end - start)}\n"
+        with open('time_it.txt', 'a') as f:
+            f.writelines(info)
+        logging.info(info)
 
         return output
     return wrapper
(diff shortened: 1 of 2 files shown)


checkov/common/util/decorators.py
```diff
@@ -22,7 +22,10 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
         end = default_timer()
 
         func_path = f"{func.__code__.co_filename.replace('.py', '')}.{func.__name__}"
-        logging.info(f"'{func_path}' took: {timedelta(seconds=end - start)}")
+        info = f"'{func_path}' took: {timedelta(seconds=end - start)}\n"
+        with open('time_it.txt', 'a') as f:
```
omryMen, line 26: how will we see it in the cloud ?

checkov/secrets/git_history_store.py
```diff
@@ -124,7 +124,8 @@ def get_added_and_removed_commit_hash(self, key: str, secret: PotentialSecret) -
             enriched_secrets = self.secrets_by_file_value_type[secret_key]
             chosen_secret = enriched_secrets[0]
             if len(enriched_secrets) > 1:
-                added, removed, _file = key.split("_")
+                res = key.split("_")
```
omryMen, line 127: did you change the key ?

checkov/common/util/decorators.py
marynaKK: actually, this is for local runs :) if we want it in s3 and saved for later - it's not from here.
I can remove it, but personally, it helped me track runs and not lose the latest run accidently.

checkov/secrets/git_history_store.py
marynaKK: No. in case of files like: `my_name_have_underscore.py`
The previous code would fail, as there are 6 fields to extract and not 3.

checkov/common/util/decorators.py
omryMen: so maybe somehow we can do it file for local and console out for cloud ?

marynaKK: I am doing both :) I didnt removed the previous code

omryMen: oh cool, missed that

omryMen (approved):

PR #4914, opened 2023-04-17
fix(secrets): Save fetched policy destination from current work dir to temp

pazbechor:
### Description
PR regarding changing the dest directory for fetched policies in checkov

### Fix
Fetching the files to temporary directory instead of working directory (in case WORKDIR has not been supplied)

## Checklist:


checkov/secrets/runner.py
```diff
@@ -124,7 +125,7 @@ def run(
         # load runnable plugins
         customer_run_config = bc_integration.customer_run_config_response
         plugins_index = 0
-        work_path = str(os.getenv('WORKDIR', current_dir))
+        work_path = str(os.getenv('WORKDIR', tempfile.gettempdir()))
```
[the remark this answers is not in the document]
pazbechor: The tempfile.gettempdir() does not provide empty directory,

nimrodkor (approved):

checkov/secrets/runner.py
```diff
@@ -124,7 +125,12 @@ def run(
         # load runnable plugins
         customer_run_config = bc_integration.customer_run_config_response
         plugins_index = 0
-        work_path = str(os.getenv('WORKDIR', current_dir))
+        work_dir_obj = None
+        work_path = str(os.getenv('WORKDIR')) if os.getenv('WORKDIR') else None
```
omryMen, line 129: os.getenv returns None if doesn't exist.. so you can use it
you're changing a behavior of using the current dir, are you sure we want to do it ?

pazbechor: yes :) 
Check ISSUE BCE-15441

PR #4920, opened 2023-04-18
fix(secrets): timeout in unix to work with signals

omryMen:

## Checklist:

### DIFF
--- a/checkov/secrets/scan_git_history.py
+++ b/checkov/secrets/scan_git_history.py
@@ -157,6 +157,8 @@ class GitHistoryScanner:
         return self.commits_diff
 
     def _run_scan_parallel(self, commits_diff: List[Commit]) -> None:
+        if not commits_diff:
+            return
         results = parallel_runner.run_function(GitHistoryScanner._run_scan_one_bulk, commits_diff)
 
         for result in results:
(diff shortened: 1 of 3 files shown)


maxamel (approved):

checkov/secrets/scan_git_history.py
```diff
@@ -145,13 +147,18 @@ def _get_commits_diff(self, root_folder: str, last_commit_sha: Optional[str] = N
                     curr_diff.add_file(filename=file_path, commit_diff=base_diff_format + file_diff.diff.decode('utf-8'))
                 if not curr_diff.is_empty():
                     self.commits_diff.append(curr_diff)
-            except Exception as e:
-                logging.warning(f"got error while getting commits diff, iteration: {previous_commit_idx}, error: {e}")
+            except TimeoutException:
+                logging.error(f"stopped while getting commits diff, iteration: {previous_commit_idx}")
+                return []
```
marynaKK, line 152: why not return the `return self.commits_diff` ?

omryMen: because we won't use it for anything
also if it's in self, we don't actually need to return it

marynaKK: there is 1 function that uses the assignment of the variable (a workaround for tests, so we do need to return it if we want it.
but also, this is what Taylor wrote in the readme of git history today:
Use `--secrets-history-timeout` to set how long the secrets scan will run on history before stopping and returning results. The default is `12h`.

so I think we want all the results until the timeout.

PR #5021, opened 2023-05-03
feat(terraform): Set TF Modules for_each env var to true

ChanochShayner:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description
Set TF Modules for_each env var to true.

## Checklist:


checkov/terraform/graph_builder/utils.py
```diff
@@ -305,9 +305,6 @@ def get_file_path_to_referred_id_networkx(graph_object: DiGraph) -> dict[str, st
 
 def get_file_path_to_referred_id_igraph(graph_object: igraph.Graph) -> dict[str, str]:
     file_path_to_module_id = {}
-    for v in graph_object.vs:
-        if v[CustomAttributes.BLOCK_TYPE] == BlockType.MODULE:
-            modules = v
```
achiar99, line 310: why?

achiar99 (approved):

checkov/terraform/graph_builder/utils.py
ChanochShayner: Take a look on the following lines, they are overriding the var and doing exactly the same

gruebel: this looks redundant, when I see the line below 😄

checkov/terraform/graph_builder/local_graph.py
```diff
@@ -287,13 +287,13 @@ def _build_edges_for_vertex(self, origin_node_index: int, vertex: TerraformBlock
...
-            ]
-            if len(target_variables) == 1:
-                self._create_edge(target_variables[0], origin_node_index, "default", cross_variable_edges)
+            target_variable = 0
+            for index in self.vertices_block_name_map.get(BlockType.VARIABLE, {}).get(vertex.name, []):
+                if self.get_dirname(self.vertices[index].path) == self.get_dirname(vertex.path):
+                    target_variable = index
+                    break
+            if target_variable:
+                self._create_edge(target_variable, origin_node_index, "default", cross_variable_edges)
```
gruebel, line 296: just double checking, this has no bad side effect, because we would create more often edges, right?

ChanochShayner: I moved it back to a list and checked the length.

gruebel: if we for some reason get more than 1 `target_variables`, we didn't create an edge before and just continued. If the tests are passing, then it is probably not an issue 😄

gruebel (approved):
🚗

ankon:
Since this commit which changes `CHECKOV_ENABLE_MODULES_FOREACH_HANDLING` to `True` by default our checkov tests hang. Will file a separate issue in a minute or two.

PR #5032, opened 2023-05-04
fix(terraform): skipping unknown customer modules

tomerBZ:

## Description

skipping unknown customer modules

## Checklist:

### DIFF
--- a/checkov/common/bridgecrew/integration_features/features/licensing_integration.py
+++ b/checkov/common/bridgecrew/integration_features/features/licensing_integration.py
@@ -48,7 +48,11 @@ class LicensingIntegration(BaseIntegrationFeature):
 
             self.open_source_only = False
             # the API will return True for all modules if they are on resource mode, so we don't actually need the billing plan explicitly here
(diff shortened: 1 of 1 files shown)


checkov/common/bridgecrew/integration_features/features/licensing_integration.py
```diff
@@ -48,7 +48,11 @@ def pre_scan(self) -> None:
 
             self.open_source_only = False
             # the API will return True for all modules if they are on resource mode, so we don't actually need the billing plan explicitly here
-            self.enabled_modules = [CustomerSubscription(m) for m, e in license_details.get(MODULES_KEY).items() if e]
+            self.enabled_modules = [
+                            CustomerSubscription(m)
+                            for m, e in license_details.get(MODULES_KEY, {}).items()
+                            if e
```
bo156, line 54: correct me if I'm wrong, but I assume `m` might be a value outside of the `CustomerSubscription` defined ones, which will cause the exception. So what here prevents it from happening?
Also would really appreciate you adding a test for trying to create a non-existent `CustomerSubscription`

bo156 (changes requested):

PR #5112, opened 2023-05-18
feat(terraform_plan): Expose field changes to python checks

### LINKED ISSUE #5067: [checks][python] How to determine which fields are changing in a resource
While you can use `TF_PLAN_RESOURCE_CHANGE_ACTIONS` to see "did a resource change", I'm wanting to know "what field(s) changed"?

The plan data definitely exposes this, so it should be possible, at the very least. Is this already available and just not explicitly documented? If not, is that on the roadmap to add?

tarfeef101:

## Description

Adds information to python checks about the changed keys/fields in a terraform resource. This allows for checks to be more granular about what they do/do not apply to (for example, an org may require important resource types to be reviewed by special parties when changed, but they may only care about material field changes). This enables that functionality, following the same methodology as `TF_PLAN_RESOURCE_CHANGE_ACTIONS` w/ respect to how the information is exposed.

Fixes #5067

## Checklist:

  - to the extent surrounding code is commented, my personal preference is for more fwiw
  - no, because other "features" like this have not. happy to do so if that's requested, though
  - have also run plans with deletions, creations, and changes through local testing, and it passes all, with correct behaviour
  - no NEW test failures locally were observed
  - N/A


tarfeef101:
(not sure what's going on w/ the tests, line 258 is empty (did this not run on the latest commit?). as for the unit tests, those failed before I made changes)

checkov/terraform/plan_parser.py
```diff
@@ -238,14 +239,24 @@ def _get_resource_changes(template: dict[str, Any]) -> dict[str, dict[str, Any]]
...
+            if not each["change"]["before"]:
+                each["change"]["before"] = {}
+            if not each["change"]["after"]:
+                each["change"]["after"] = {}
+
+            for field in each["change"]["before"]:
+                if each["change"]["before"][field] != each["change"]["after"].get(field):
+                    changes.append(field)
+ 
+            resource_changes_map[each["address"]]["changed_keys"] = changes
```
gruebel, line 259: since this is artificially added please use the field name you created, so it won't likely overlap with a real field.
```suggestion
            resource_changes_map[each["address"]][TF_PLAN_RESOURCE_CHANGE_KEYS] = changes
```

checkov/terraform/plan_parser.py
```diff
@@ -238,14 +239,24 @@ def _get_resource_changes(template: dict[str, Any]) -> dict[str, dict[str, Any]]
...
+            changes = []
+
+            # before + after are None when resources are created/destroyed, so make them safe
+            if not each["change"]["before"]:
+                each["change"]["before"] = {}
+            if not each["change"]["after"]:
+                each["change"]["after"] = {}
+
+            for field in each["change"]["before"]:
+                if each["change"]["before"][field] != each["change"]["after"].get(field):
```
gruebel, line 256:
```suggestion
            for field, value in each["change"]["before"].items():
                if value != each["change"]["after"].get(field):
```

gruebel: also make sure to skip the field names `__startline__` and `__endline__`, there is aconstant `LINE_FIELD_NAMES` which stores them as a set.

checkov/terraform/plan_parser.py
```diff
@@ -238,14 +239,24 @@ def _get_resource_changes(template: dict[str, Any]) -> dict[str, dict[str, Any]]
...
+    if resource_changes and isinstance(resource_changes, list):
+        for each in resource_changes:
+            resource_changes_map[each["address"]] = each
+            changes = []
+
+            # before + after are None when resources are created/destroyed, so make them safe
+            if not each["change"]["before"]:
+                each["change"]["before"] = {}
+            if not each["change"]["after"]:
+                each["change"]["after"] = {}
```
gruebel, line 253: `not` also works, but if it is an empty `dict` then there is no need to override it again
```suggestion
            if each["change"]["before"] is None:
                each["change"]["before"] = {}
            if each["change"]["after"] is None:
                each["change"]["after"] = {}
```

tests/terraform/runner/extra_tf_plan_checks/security_group_rule_protocol_changed.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Any
+
+from checkov.common.models.enums import CheckResult, CheckCategories
+from checkov.terraform.checks.resource.base_resource_check import BaseResourceCheck
+from checkov.terraform.plan_parser import TF_PLAN_RESOURCE_CHANGE_ACTIONS, TF_PLAN_RESOURCE_CHANGE_KEYS
+
+
+class SecurityGroupRuleProtocolChanged(BaseResourceCheck):
```
gruebel, line 10: can you also use this check in a test case to prove you have passing and failing checks?

checkov/terraform/plan_parser.py
```diff
@@ -161,6 +161,7 @@ def _prepare_resource_block(
         changes = resource_changes.get(resource_address)  # type:ignore[arg-type]  # becaus eit can be None
         if changes:
             resource_conf[TF_PLAN_RESOURCE_CHANGE_ACTIONS] = changes.get("change", {}).get("actions") or []
+            resource_conf[TF_PLAN_RESOURCE_CHANGE_KEYS] = changes["changed_keys"]
```
gruebel, line 164:
```suggestion
            resource_conf[TF_PLAN_RESOURCE_CHANGE_KEYS] = changes.get(TF_PLAN_RESOURCE_CHANGE_KEYS) or []
```

gruebel (commented):
nice work, added a couple of comments 💪 and please add a small paragraph in docs about it https://github.com/bridgecrewio/checkov/blob/main/docs/7.Scan%20Examples/Terraform%20Plan%20Scanning.md#deleted-resources

tests/terraform/runner/extra_tf_plan_checks/security_group_rule_protocol_changed.py
tarfeef101: idk, can I? I am not an appdev so I did a lot of guessing/inferencing to arrive at what I just commited :sweat_smile: 

lmk if I did what I intended to

gruebel:
let me check and fix, what's needed to pass CI 🙂

tarfeef101:
> let me check and fix, what's needed to pass CI slightly_smiling_face

thanks :laughing: i think i'm just failing to wrap my head around the import structure for my test, fixed the lint. mypy seems to be unable to correctly apply its logic, the error it's spitting out seems incorrect

gruebel (approved):
thanks for the contribution 🥇 I adjusted the test accordingly, because you put in the wrong area 😉

tarfeef101:
lol so my attempts to test my changes to `checkov/terraform/plan_parser` in `tests/terraform/parser` instead of the `runner` dir (where I originally did it) were misguided after all :upside_down_face: 

thanks for the assist :D

gruebel:
you could have also wrote a parser test, but those are more about the parsing and have nothing to do with created checks. So, your fits better to the runner 🙂

ayajbara (approved):

checkov/terraform/plan_parser.py
```diff
@@ -238,13 +240,24 @@ def _get_resource_changes(template: dict[str, Any]) -> dict[str, dict[str, Any]]
...
     resource_changes_map = {}
-
     resource_changes = template.get("resource_changes")
+
     if resource_changes and isinstance(resource_changes, list):
-        resource_changes_map = {
-            change.get("address", ""): change
-            for change in resource_changes
-        }
+        for each in resource_changes:
```
bo156, line 246: can you rename `each` to something like `resource` or `changed_resource`?
Will be more readable in python code (even though I guess it comes from terraform conventions)

tests/terraform/runner/test_plan_runner.py
```diff
@@ -27,10 +28,10 @@
    {"db_connector": IgraphConnector}
 ])
 class TestRunnerValid(unittest.TestCase):
-
-    def setUp(self) -> None:
-        self.orig_checks = resource_registry.checks
-        self.db_connector = self.db_connector
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.orig_checks = deepcopy(resource_registry.checks)
```
bo156, line 33: why do you need to change this setup and teardown? seems like you are not using it anywhere, unless I missed something

bo156 (changes requested):

tests/terraform/runner/test_plan_runner.py
bo156: I guess that's a question for you @gruebel as you changes this part

gruebel: since this is a unittest testcase, the `setup` and `teardown` methods are invoked automatically before and after each test case. `setUpClass` is only invoked once for the whole file.

bo156 (approved):

PR #5217, opened 2023-06-15
platform(general): Implement SSO Relay State Parameter in Checkov Output Links

SimOnPanw:

## Description

This PR introduces an enhancement to the way Checkov generates output links. Previously, output links were generated and printed to the console without considering any specific state parameters required for Single Sign-On (SSO) procedures.

The code in this PR updates the `_print_to_console` method to incorporate a relay state parameter when generating output links. This is achieved through the implementation of a new method, `get_sso_prismacloud_url`, which calls the Prisma Cloud Platform API to get the necessary relay state from the SAML configuration. This state is then appended to the URL before it's printed to the console.

These changes ensure that the output links generated by Checkov are compatible with SSO procedures that rely on relay state parameters.

Before the merge, Checkov generate the report link without considering any specific relaystate parameters:
image

After the merge, Checkov will use the RelayState Parameter:
image

Please review and let me know if any changes are required.

## Checklist:


checkov/common/runners/runner_registry.py
```diff
@@ -577,16 +579,54 @@ def print_reports(
...
         if output_dest == CONSOLE_OUTPUT:
             del output_formats[output_format]
 
             print(output)
+
             if url:
+                if bc_integration.prisma_api_url is not None and bc_integration.get_auth_token() is not None:
+                    url = self.get_sso_prismacloud_url(report_url=url,
+                                                       prisma_api_url=bc_integration.prisma_api_url,
+                                                       token=bc_integration.get_auth_token())
```
gruebel, line 627: this is not the right place for it, because all related platform interaction is under `common.bridgecrew`. This would probably a good spot to add the function call https://github.com/bridgecrewio/checkov/blob/2df19723d80024544df918bf85cf73b88e5edec7/checkov/common/bridgecrew/platform_integration.py#L576

checkov/common/runners/runner_registry.py
```diff
@@ -577,16 +579,54 @@ def print_reports(
         exit_code = 1 if 1 in exit_codes else 0
         return cast(Literal[0, 1], exit_code)
 
+    # Define the function that will get the relay state from the Prisma Cloud Platform.
+    def get_sso_prismacloud_url(self, report_url: str, prisma_api_url: str, token: str) -> str:
```
gruebel, line 583: please move the method to `checkov/checkov/common/bridgecrew/platform_integration.py`

checkov/common/runners/runner_registry.py
```diff
@@ -577,16 +579,54 @@ def print_reports(
...
+        url_saml_config = f"{prisma_api_url}/saml/config"
+        headers = {'x-redlock-auth': token}
+        try:
+            response = requests.get(url_saml_config, headers=headers, timeout=10)
```
gruebel, line 587: please use `self.http.request(...)` inside the `BcPlatformIntegration` class

checkov/common/runners/runner_registry.py
```diff
@@ -577,16 +579,54 @@ def print_reports(
...
+            response.raise_for_status()  # Raises an HTTPError if the response status is 4xx, 5xx
+            data = response.json()
+        except requests.exceptions.Timeout:
+            print("The request timed out")
+        except requests.exceptions.TooManyRedirects:
+            print("Too many redirects")
+        except requests.exceptions.RequestException as e:
+            print(f"Error while calling Prisma Cloud API: {e}")
+            return report_url
```
gruebel, line 596: instead of print use `logging.info(...)`

gruebel (commented):
thanks for the contribution 💪 I added a couple of comments

checkov/common/runners/runner_registry.py
SimOnPanw: This was integrated into the function get_sso_prismacloud_url moved to platform_integration.py

checkov/common/runners/runner_registry.py
```diff
@@ -579,14 +579,15 @@ def print_reports(
...
     def _print_to_console(self, output_formats: dict[str, str], output_format: str, output: str, url: str | None = None) -> None:
         """Prints the output to console, if needed"""
-
         output_dest = output_formats[output_format]
         if output_dest == CONSOLE_OUTPUT:
             del output_formats[output_format]
 
             print(output)
             if url:
+                url = bc_integration.get_sso_prismacloud_url(report_url=url)
```
gruebel, line 588: as mentioned before, please invoke your method here https://github.com/bridgecrewio/checkov/blob/2df19723d80024544df918bf85cf73b88e5edec7/checkov/common/bridgecrew/platform_integration.py#L576
the URL comes from this method and is used in 2 areas, so both can benefit from it 🙂

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -1041,5 +1042,41 @@ def get_default_headers(self, request_type: str) -> dict[str, Any]:
...
+        headers = merge_dicts(get_auth_header(token),
+                              get_default_get_headers(self.bc_source, self.bc_source_version))
+
+        request = self.http.request("GET", url_saml_config, headers=headers, timeout=10)
+        if request.status == 401:
+            logging.error(f'Received 401 response from Prisma /login endpoint: {request.data.decode("utf8")}')
+            raise BridgecrewAuthError()
+        elif request.status == 403:
+            logging.error('Received 403 (Forbidden) response from Prisma /login endpoint')
+            raise BridgecrewAuthError()
```
gruebel, line 1060: @mikeurbanski1 any thoughts about raising an exception here or should we just log it and return the normal URL?

checkov/common/runners/runner_registry.py
SimOnPanw: @gruebel oh, now I get it. 👍

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -1041,5 +1042,41 @@ def get_default_headers(self, request_type: str) -> dict[str, Any]:
         logging.info(f"Unsupported request {request_type}")
         return {}
 
+    # Define the function that will get the relay state from the Prisma Cloud Platform.
+    def get_sso_prismacloud_url(self, report_url: str) -> str:
+        if not bc_integration.prisma_api_url or report_url is None:
```
gruebel, line 1047:
```suggestion
        if not bc_integration.prisma_api_url or not self.http or not self.bc_source or report_url is None:
```

gruebel: this should hopefully solve the `mypy` issue

gruebel (approved):
nice, looks good from my side, just waiting for @mikeurbanski1 🙂

checkov/common/bridgecrew/platform_integration.py
mikeurbanski1: I like returning the normal URL. Then at least they can go to it after signing in separately.

checkov/common/bridgecrew/platform_integration.py
```diff
@@ -572,7 +573,7 @@ def commit_repository(self, branch: str) -> str | None:
                                                                 get_user_agent_header()
                                                                 ))
                 response = json.loads(request.data.decode("utf8"))
-                url: str = response.get("url", None)
+                url: str = self.get_sso_prismacloud_url(response.get("url", None))
```
mikeurbanski1, line 576: @SimOnPanw what happens if SSO isn't configured?

SimOnPanw: Hello @mikeurbanski1, we will be returning the standard URL that directly points to the platform. To return the URL with RelayState parameter,  you'll have to provide a prisma_api_url and ensure that your Prisma Cloud tenant is configured with both relay_state_param_name and access_saml_url.

Here is an example:
image

checkov/common/bridgecrew/platform_integration.py
SimOnPanw: @mikeurbanski1 @gruebel instead of raising an exception, I could return the normal URL as follow:
```        
request = self.http.request("GET", url_saml_config, headers=headers, timeout=10)  # type:ignore[no-untyped-call]
if request.status >= 300:
  return report_url
```

With the problem that we don't trigger an exception if the BC_API_KEY is not provided. I am not sure this is better. What do you think ?

checkov/common/bridgecrew/platform_integration.py
SimOnPanw: Hello @mikeurbanski1 and @gruebel, any news on this PR ?

mikeurbanski1: But what does `{bc_integration.prisma_api_url}/saml/config` return if SSO is not configured? It looks like we are always going to try to get this, and throw an exception if it doesn't work, so I am worried that we won't fall back.

SimOnPanw: It returns a json with an empty fields for relayStateParamName and redLockAccessSamlUrl, even if a tenant has no SAML configuration.
I checked if there is any value return below in the code.

checkov/common/bridgecrew/platform_integration.py
SimOnPanw: @mikeurbanski1, of course, you are correct. I committed the change.

mikeurbanski1 (approved):

PR #5230, opened 2023-06-19
fix(terraform_plan): add terraform plan vertices to terraform graph if not exist

YaaraVerner:

## Description
added all resources that weren't found in the terraform folder (when deep-analysis enabled) to the terraform graph before scanning it in order to run graph checks over all existing resources in terraform plan files

## Checklist:


achiar99 (approved):
💯

checkov/terraform/deep_analysis_plan_graph_manager.py
```diff
@@ -5,45 +5,70 @@
...
+    def append_vertex_to_terraform_graph(self, tf_plan_vertex, tf_plan_vertex_index, address):
+        new_vertex_idx = len(self.tf_graph.vertices)
+        self.tf_graph.vertices.append(tf_plan_vertex)
+        self._address_to_tf_idx_and_vertex_map[address] = (new_vertex_idx, tf_plan_vertex)
+
+        def __get_tf_vertex_idx_from_tf_plan_vertex(v):
+            v = self._address_to_tf_idx_and_vertex_map.get(v.attributes.get('__address__'))
+            if not v:
+                return None
+            return v[1]
```
gruebel, line 40: I don't think this needs to be a nested method, just move it to the same level as the others.

gruebel (commented):
nice work 🍻

checkov/terraform/plan_runner.py
```diff
@@ -181,11 +181,12 @@ def _get_graph_report(
...
         return self.get_graph_checks_report(root_folder, runner_filter)
 
-    def _create_terraform_graph(self) -> TerraformLocalGraph:
+    def _create_terraform_graph(self, runner_filter) -> TerraformLocalGraph:
         graph_manager = TerraformGraphManager(db_connector=self.db_connector)
         tf_local_graph, _ = graph_manager.build_graph_from_source_directory(
             self.repo_root_for_plan_enrichment,
-            render_variables=True
+            render_variables=True,
+            download_external_modules=runner_filter.download_external_modules
```
maxamel, line 189: I see runner_filter can be defined as None in the parent function, so seems we should check that before accessing this field

checkov/terraform/deep_analysis_plan_graph_manager.py
```diff
@@ -5,45 +5,70 @@
...
+            if origin_index:
+                self.tf_graph.create_edge(origin_index, new_vertex_idx, edge.label)
+
     def enrich_tf_graph_attributes(self) -> None:
-        for address, tf_plan_vertex in self._address_to_tf_plan_vertex_map.items():
-            tf_vertex = self._address_to_tf_vertex_map.get(address)
-            if not tf_vertex:
-                logging.info(f'Cant find this address: {address} in tf graph')
+        for address, tf_plan_idx_and_vertex in self._address_to_tf_plan_idx_and_vertex_map.items():
+            tf_plan_vertex_index, tf_plan_vertex = tf_plan_idx_and_vertex
```
bo156, line 55: Based on the type annotation you actually have `Tuple[Block, int]` which means the unpacking should be the opposite.
It seems like you inserted it in the order that you unpack it though so it works, but the type annotation is wrong (which is also probably one of the reasons your `mypy` check fails)

checkov/terraform/deep_analysis_plan_graph_manager.py
```diff
@@ -5,45 +5,70 @@
...
+            tf_idx_and_vertex = self._address_to_tf_idx_and_vertex_map.get(address)
+            if not tf_idx_and_vertex:
+                logging.info(f'Cant find this address: {address} in tf graph, adding it')
+                self.append_vertex_to_terraform_graph(tf_plan_vertex, tf_plan_vertex_index, address)
```
bo156, line 59: aren't you adding vertices to the `terraform_plan` graph instead of `terraform`?
If you are doing it like you wrote, can you explain me why?

tests/terraform/runner/resources/plan_and_tf_combine_graph_with_missing_resources/source/main.tf
```diff
@@ -0,0 +1,14 @@
+provider "aws" {
```
bo156, line 1: I'm writing it here as I don't have a better place - 
Basically I understand the test because we talked about it, but it's hard to read it if you are new to this bug.
Could you specify which resources are missing from the plan? 
For example, you can call the `aws_s3_bucket` by the name `not-in-tf-plan` instead of `example` (assuming this resource is missing)

bo156 (changes requested):

checkov/terraform/deep_analysis_plan_graph_manager.py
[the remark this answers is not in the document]
YaaraVerner: it's a tuple of two items so yes

checkov/terraform/deep_analysis_plan_graph_manager.py
```diff
@@ -5,45 +5,70 @@
...
-            vertex.attributes[TF_PLAN_RESOURCE_ADDRESS]: vertex
-            for vertex in self.tf_plan_graph.vertices
+        self._address_to_tf_plan_idx_and_vertex_map = {
+            vertex.attributes[TF_PLAN_RESOURCE_ADDRESS]: (i, vertex)
+            for i, vertex in enumerate(self.tf_plan_graph.vertices)
             if vertex.block_type == BlockType.RESOURCE
         }
 
+    def append_vertex_to_terraform_graph(self, tf_plan_vertex, tf_plan_vertex_index, address):
+        new_vertex_idx = len(self.tf_graph.vertices)
```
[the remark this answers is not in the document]
YaaraVerner: we are adding the new vertex to this list so the length will be increased every time

checkov/terraform/deep_analysis_plan_graph_manager.py
YaaraVerner: The terraform graph contains more data on the variables, locals, edges and etc so it's better to run the queries on it. We are adding the missing resources from the terraform_plan graph for the edge case that the terraform graph won't contain it

bo156 (approved):

PR #5272, opened 2023-06-28
fix(terraform): add module prefix to the resource ID of a module

gruebel:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

- added `module.` prefix to the resoruce ID of the module resources, now they are also more aligned to the address of normal resources

## Checklist:

### DIFF
--- a/checkov/terraform/graph_builder/local_graph.py
+++ b/checkov/terraform/graph_builder/local_graph.py
@@ -603,6 +603,11 @@ class TerraformLocalGraph(LocalGraph[TerraformBlock]):
             context_parser = parser_registry.context_parsers[vertex.block_type]
             entity_context_path = context_parser.get_entity_context_path(vertex.config)
             resource_id = '.'.join(entity_context_path) if entity_context_path else vertex.name
+
+            if vertex.block_type == BlockType.MODULE:
+                # prefix modules with 'module' to properly identify them
+                resource_id = f"{BlockType.MODULE}.{resource_id}"
+
             address = f'{address_prefix}{resource_id}'
             vertex.attributes[CustomAttributes.TF_RESOURCE_ADDRESS] = address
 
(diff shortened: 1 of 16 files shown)


lirshindalman (approved):

checkov/terraform/graph_builder/local_graph.py
```diff
@@ -603,6 +603,11 @@ def update_nested_modules_address(self) -> None:
             context_parser = parser_registry.context_parsers[vertex.block_type]
             entity_context_path = context_parser.get_entity_context_path(vertex.config)
             resource_id = '.'.join(entity_context_path) if entity_context_path else vertex.name
+
+            if vertex.block_type == BlockType.MODULE:
+                # prefix modules with 'module' to properly identify them
+                resource_id = f"{BlockType.MODULE}.{resource_id}"
```
bo156, line 609: how come we don't need to do it for any other resource type?

tests/terraform/graph/graph_builder/test_graph_builder.py
```diff
@@ -316,9 +316,9 @@ def test_nested_modules_address_attribute(self):
         graph_manager = TerraformGraphManager(NetworkxConnector())
         local_graph, _ = graph_manager.build_graph_from_source_directory(resources_dir, render_variables=True)
         module_1 = self.get_vertex_by_name_and_type(local_graph, BlockType.MODULE, 'inner_s3_module')
-        assert module_1.attributes.get(CustomAttributes.TF_RESOURCE_ADDRESS) == 'module.s3_module.inner_s3_module'
+        assert module_1.attributes.get(CustomAttributes.TF_RESOURCE_ADDRESS) == 'module.s3_module.module.inner_s3_module'
```
bo156, line 319: looks like a scary change I have to say - are we sure that we handle this correctly now that we have multiple `module`?

checkov/terraform/graph_builder/local_graph.py
gruebel: currently none of the other non-resource blocks has a prefix, which is not so important, because they are identifiable by their block type. I did this change to align the address field to be identical to the prefix for other block types inside a module.

tests/terraform/graph/graph_builder/test_graph_builder.py
gruebel: I assume so as can be seen in the `expected_local_graph.json` file I adjusted. multi module looks good and also resource inside that module didn't change.

PR #5339, opened 2023-07-16
fix(kubernetes): No dups when extracting images

bo156:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

In some cases a k8s configuration might contain images names more than once (for example - using the same image as `initContainer` and `container`). 
This PR makes sure we don't return any duplications while trying to extract resources from the template

## Checklist:


checkov/kubernetes/image_referencer/provider/k8s.py
```diff
@@ -90,6 +54,20 @@ def extract_images_from_containers(containers: Any) -> list[str]:
     return image_names
 
 
+def _extract_images_from_spec(spec: dict[str, Any]) -> list[str]:
+    image_names: list[str] = []
```
gruebel, line 58: how about making this a `set()` then you don't need to transform it to a set and back to list 🙂

bo156: good idea - fixed

checkov/kubernetes/image_referencer/provider/k8s.py
```diff
@@ -19,77 +19,55 @@ def __init__(self, graph_connector: DiGraph) -> None:
...
             if isinstance(container, dict):
                 image = container.get("image")
                 if image and isinstance(image, str):
-                    image_names.append(image)
+                    image_names.add(image)
 
     return image_names
 
 
+def _extract_images_from_spec(spec: dict[str, Any] | None) -> list[str]:
```
omryMen, line 57: Nice !

checkov/kubernetes/image_referencer/provider/k8s.py
```diff
@@ -19,77 +19,55 @@ def __init__(self, graph_connector: DiGraph) -> None:
...
+    image_names: set[str] = set()
+
+    if isinstance(spec, dict):
+        containers = spec.get("containers")
+        image_names.update(extract_images_from_containers(containers=containers))
+
+        containers = spec.get("initContainers")
+        image_names.update(extract_images_from_containers(containers=containers))
```
omryMen, line 65: maybe update only if containers is not None ?

checkov/kubernetes/image_referencer/provider/k8s.py
omryMen: maybe all of the functions here can return sets ?

gruebel: also thought the same, but I think it is a bigger change, because of the typing mismatch.

bo156: Just didn't watch to change api

checkov/kubernetes/image_referencer/provider/k8s.py
bo156: it just adds some more code and makes it look more ugly :/

omryMen (approved):

gruebel (approved):
🏖️

PR #5346, opened 2023-07-17
feat(general): Infrastructure for custom code logger filter

bo156:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Added a potential log adapter that allows removing templates from logs, to allow decreasing log size for example.
First possible use-case of this adapter is to remove resource code from logs, as this might involve large log sizes.
Currently no existing template is added, this PR only creates the infrastructure to use it. 

## Checklist:


checkov/common/resource_code_logger_filter.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, allow_code_logging: bool = True):
+        super().__init__()
+        self.allow_code_logging = allow_code_logging
+
+    def filter(self, record: LogRecord) -> bool:
+        if self.allow_code_logging:
+            return True
+        msg = record.msg
+        for code_template in ResourceCodeFilter.CODE_TEMPLATES:
```
YaaraVerner, line 26: can we check an extra variable that will be sent in those logs? I don't think it will be possible to create a `CODE_TEMPLATES` list that will include all the code log cases

bo156: what do you call an extra variable? basically we can test anything we want on the record.
each record is created from what we send to the relevant function (`info`/`debug` etc).
Basically we can add arguments to the logging calls, but it will be really awkward to do in my opinion

bo156: what do you call an extra variable? basically we can test anything we want on the record.
each record is created from what we send to the relevant function (info/debug etc).
Basically we can add arguments to the logging calls, but it will be really awkward to do in my opinion

checkov/common/util/type_forcers.py
```diff
@@ -44,15 +45,16 @@ def force_float(var: Any) -> float | None:
...
+def convert_str_to_bool(bool_str: bool | str) -> bool:
     if isinstance(bool_str, str):
         bool_str_lower = bool_str.lower()
         if bool_str_lower in ("true", '"true"'):
             return True
         elif bool_str_lower in ("false", '"false"'):
             return False
 
-    return bool_str
+    # If we got here it must be a boolean, mypy doesn't understand it, so we use cast
```
omryMen, line 56: just use else and you won't need it

bo156: Tried it, doesn't work as it waits for a return value as well outside of the else

checkov/common/resource_code_logger_filter.py
gruebel: as said previously this specific implementation wouldn't work, because the logs a re dynamic you can't map them easily, also using a regex can become tricky, since each IaC tool has a different idea on how their resources look like. I think, also what @YaaraVerner said adding the `extra` keyword to all log messages which needs filtering is easier to realize, but it means work to go through all the logs, which we need to do anyway.
ex.
```python
logger.info(f'Failed on resource {resource}', extra={"mask": True})
```

bo156: 2 things here:
1. There is no problem to add this logic to the filter if we decide this is what we want. That's a few lines changed in the `filter` function which can be easily solved.
2. I assumed this is what @YaaraVerner meant, but I just feels it will be extremely tedious to implement. It also means that everybody who use this package should be aware to this feature, which is extremely problematic in my opinion. If we want to do something like that we also need to make sure every `logging.<something>` usage actually suplies this parameter.

checkov/common/util/type_forcers.py
bo156: Fixed

checkov/common/resource_code_logger_filter.py
bo156: added now :)

tests/common/test_resource_code_logger_filter.py
```diff
@@ -0,0 +1,34 @@
...
+from checkov.common.resource_code_logger_filter import add_resource_code_filter_to_logger
+
+
+TEST_CODE_TEMPLATES_TO_REPLACE = "THIS-SHOULD-BE-REPLACED!"
+
+
+def test_code_logger_filter_do_not_log_if_not_allowed(caplog) -> None:
+    with mock.patch("checkov.common.resource_code_logger_filter.ResourceCodeFilter.CODE_TEMPLATES",
+                    [TEST_CODE_TEMPLATES_TO_REPLACE]):
+        logger = logging.getLogger("code logging allowed")
```
omryMen, line 13: the log is not right

bo156: fixed

omryMen (approved):

checkov/common/resource_code_logger_filter.py
```diff
@@ -0,0 +1,52 @@
...
+        self.allow_code_logging = allow_code_logging
+
+    def filter(self, record: LogRecord) -> bool:
+        if self.allow_code_logging:
+            return True
+        if hasattr(record, "mask"):
+            # Allows filtering using `logging.info("<msg>", extra={"mask": True})`
+            mask = record.mask
+            if not isinstance(mask, bool):
+                raise Exception(f"Expected to get `mask` as boolean for logging function, instead got: {mask}")
```
gruebel, line 29:
```suggestion
                raise Exception(f"Expected to get `mask` as boolean for logging function, instead got: {mask} of type {type(mask)}")
```

gruebel (approved):

PR #5362, opened 2023-07-20
feat(terraform): add new function to create module and definitions with tests

maxamel:
This PR introduces a new function in order to create a multi graph hcl module, instead of the previous single graph hcl module.
At the moment it is not called from anywhere and will be used in subsequent PRs.
The old function will not be removed and it will stay and be called depending on env var.

## Checklist:


checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,58 @@ def parse_hcl_module(
...
+            directory=source_dir, out_evaluations_context={},
+            out_parsing_errors=parsing_errors if parsing_errors is not None else {},
+            download_external_modules=download_external_modules,
+            external_modules_download_path=external_modules_download_path, excluded_paths=excluded_paths,
+            vars_files=vars_files, external_modules_content_cache=external_modules_content_cache
+        )
+        tf_definitions = clean_parser_types(tf_definitions)
+        tf_definitions = serialize_definitions(tf_definitions)
+
+        dirs_to_definitions: dict[str, list[dict[TFDefinitionKey, dict[str, Any]]]] = {}
```
bo156, line 355: would split the creation of this dict to a function like `_build_dirs_to_definitions`

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,58 @@ def parse_hcl_module(
...
+                while source_module is not None:
+                    if source_module.nested_tf_module is None:
+                        dir_path = os.path.dirname(source_module.path)
+                        if dir_path in dirs_to_definitions:
+                            dirs_to_definitions[dir_path].append({tf_definition_key: tf_value})
+                        else:
+                            dirs_to_definitions[dir_path] = [{tf_definition_key: tf_value}]
+                    source_module = source_module.nested_tf_module
+
+        modules: list[Tuple[Module, Dict[TFDefinitionKey, Dict[str, Any]]]] = []
```
bo156, line 376: `tuple` and `dict`

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,58 @@ def parse_hcl_module(
...
+        if create_graph:
+            modules = []
```
bo156, line 378: would call it `modules_and_definitions_tuple`, maybe you can create a `NamedTuple` also for it
(I know this is how I did it but you are better than me! ;) )

checkov/terraform/tf_parser.py
```diff
@@ -391,6 +444,29 @@ def parse_hcl_module_from_tf_definitions(
                     logging.warning(e, exc_info=False)
         return module, tf_definitions
 
+    def parse_hcl_module_from_multi_tf_definitions(
+        self,
+        tf_definitions: list[Dict[TFDefinitionKey, Dict[str, Any]]],
```
bo156, line 449: `dict`

tests/terraform/graph/graph_builder/test_local_graph.py
```diff
@@ -108,6 +108,25 @@ def test_set_variables_values_from_modules(self):
...
                 self.assertEqual(var_value, default_val)
 
+    def test_compare_multi_graph_defs(self):
+        resources_dir = os.path.realpath(os.path.join(TEST_DIRNAME,
+                                                      '../resources/variable_rendering/render_module_postgresql'))
+        hcl_config_parser = TFParser()
+        module, defs = hcl_config_parser.parse_hcl_module(resources_dir, source=self.source)
+        modules = hcl_config_parser.parse_multi_graph_hcl_module(resources_dir, source=self.source)
+        for idx, module_to_definitions in enumerate(modules):
+            assert module_to_definitions[0] == module
```
bo156, line 118: why would that be true? is it just one module always for this example?

tests/terraform/graph/graph_builder/test_local_graph.py
```diff
@@ -108,6 +108,25 @@ def test_set_variables_values_from_modules(self):
             else:
                 self.assertEqual(var_value, default_val)
 
+    def test_compare_multi_graph_defs(self):
```
bo156, line 111: good test 👍

tests/terraform/graph/graph_builder/test_local_graph.py
maxamel: Yes we have a single module and then a list of TF definitions, one per file.

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,58 @@ def parse_hcl_module(
...
+        create_graph: bool = True,
+    ) -> list[Tuple[Module, Dict[TFDefinitionKey, Dict[str, Any]]]]:
+        tf_definitions = self.parse_directory(
+            directory=source_dir, out_evaluations_context={},
+            out_parsing_errors=parsing_errors if parsing_errors is not None else {},
+            download_external_modules=download_external_modules,
+            external_modules_download_path=external_modules_download_path, excluded_paths=excluded_paths,
+            vars_files=vars_files, external_modules_content_cache=external_modules_content_cache
+        )
+        tf_definitions = clean_parser_types(tf_definitions)
```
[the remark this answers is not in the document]
maxamel: From what I saw it keeps the keys sorted. The main reason for using it here is that it appears in the old function `parse_hcl_module` :)

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,67 @@ def parse_hcl_module(
 
         return module, tf_definitions
 
+    '''
```
bo156, line 333: the doc should be under the function not above (so line 350 here after the definition).
Otherwise the documentation is good, but I will add that `terraform` also works per directory so that's the proper way to resolve it anyway

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,67 @@ def parse_hcl_module(
...
+
+        modules_and_definitions_tuple: list[tuple[Module, list[dict[TFDefinitionKey, dict[str, Any]]]]] = []
+        if create_graph:
+            for source_path, definitions in dirs_to_definitions.items():
+                module, parsed_tf_definitions = self.parse_hcl_module_from_multi_tf_definitions(definitions, source_path, source)
+                modules_and_definitions_tuple.append((module, parsed_tf_definitions))
+
+        return modules_and_definitions_tuple
+
+    def create_definition_by_dirs(self, tf_definitions: dict[TFDefinitionKey, dict[str, list[dict[str, Any]]]]
```
bo156, line 370: I would add specific unit tests to this function as well, as this is critical logic

checkov/terraform/graph_builder/graph_components/blocks.py
```diff
@@ -81,6 +81,14 @@ def __init__(
             self.for_each_index: Optional[Any] = None
             self.foreach_attrs: list[str] | None = None
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, TerraformBlock):
+            return False
+
+        return self.name == other.name and self.config == other.config and self.path == other.path and \
```
omryMen, line 88: let's add a comment in the init func that if adding var to self you need to update this as well

checkov/terraform/graph_builder/graph_components/module.py
```diff
@@ -44,8 +44,23 @@ def __init__(
         self.enable_nested_modules = strtobool(os.getenv('CHECKOV_ENABLE_NESTED_MODULES', 'True'))
         self.use_new_tf_parser = strtobool(os.getenv('CHECKOV_NEW_TF_PARSER', 'True'))
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Module):
+            return False
```
omryMen, line 49: same as above

tests/terraform/graph/graph_builder/test_local_graph.py
omryMen, line 111: Nice!

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,67 @@ def parse_hcl_module(
...
+        dirs_to_definitions: dict[str, list[dict[TFDefinitionKey, dict[str, Any]]]] = {}
+        for tf_definition_key, tf_value in tf_definitions.items():
+            source_module = tf_definition_key.tf_source_modules
+            if source_module is None:
+                # No module - add new entry to dirs_to_definitions with the path as key
+                dir_path = os.path.dirname(tf_definition_key.file_path)
+                if dir_path in dirs_to_definitions:
+                    dirs_to_definitions[dir_path].append({tf_definition_key: tf_value})
+                else:
+                    dirs_to_definitions[dir_path] = [{tf_definition_key: tf_value}]
```
gruebel, line 381: you could also create a `defaultdict`, then you don't have to this check manually 
```python
dirs_to_definitions = defaultdict(list)
dirs_to_definitions[dir_path].append({tf_definition_key: tf_value})
```

checkov/terraform/tf_parser.py
```diff
@@ -329,6 +330,67 @@ def parse_hcl_module(
...
+            else:
+                # iterate over nested modules while adding directories on the way
+                while source_module is not None:
+                    if source_module.nested_tf_module is None:
+                        dir_path = os.path.dirname(source_module.path)
+                        if dir_path in dirs_to_definitions:
+                            dirs_to_definitions[dir_path].append({tf_definition_key: tf_value})
+                        else:
+                            dirs_to_definitions[dir_path] = [{tf_definition_key: tf_value}]
```
gruebel, line 390: same her, wouldn't be needed

checkov/terraform/tf_parser.py
```diff
@@ -391,6 +453,29 @@ def parse_hcl_module_from_tf_definitions(
...
+        self.add_tfvars_with_source_dir(module, source, source_dir)
+        copy_of_tf_definitions = pickle_deepcopy(tf_definitions)
+        for tf_def in copy_of_tf_definitions:
+            for file_path, blocks in tf_def.items():
+                for block_type in blocks:
+                    try:
+                        module.add_blocks(block_type, blocks[block_type], file_path, source)
+                    except Exception as e:
+                        logging.warning(f'Failed to add block {blocks[block_type]}. Error:')
+                        logging.warning(e, exc_info=False)
```
gruebel, line 476: just thinking about the log level, if maybe `info` is enough, depends on how critical it is to the user.

checkov/terraform/tf_parser.py
```diff
@@ -420,6 +505,14 @@ def add_tfvars(self, module: Module, source: str) -> None:
                 block = [{var_name: {"default": default}}]
                 module.add_blocks(BlockType.TF_VARIABLE, block, path, source)
 
+    def add_tfvars_with_source_dir(self, module: Module, source: str, source_dir: str) -> None:
+        if not self.external_variables_data:
+            return
+        for (var_name, default, path) in self.external_variables_data:
```
gruebel, line 511:
```suggestion
        for var_name, default, path in self.external_variables_data:
```
I don't think you need the brackets

gruebel (commented):
nice work, looking forward to the next steps 🎉

checkov/terraform/tf_parser.py
maxamel: Since it shouldn't happen normally I'm OK with warning, this way we can also search for it in our logs more easily. It's not critical enough to log error in my opinion.
Also, in the original function it's logged as warning as well.

checkov/terraform/tf_parser.py
```diff
@@ -369,25 +370,19 @@ def parse_multi_graph_hcl_module(
 
     def create_definition_by_dirs(self, tf_definitions: dict[TFDefinitionKey, dict[str, list[dict[str, Any]]]]
                                   ) -> dict[str, list[dict[TFDefinitionKey, dict[str, Any]]]]:
-        dirs_to_definitions: dict[str, list[dict[TFDefinitionKey, dict[str, Any]]]] = {}
+        dirs_to_definitions: dict[str, list[dict[TFDefinitionKey, dict[str, Any]]]] = defaultdict(list)
```
bo156, line 373: nice improvement :)

tests/terraform/graph/graph_builder/test_local_graph.py
```diff
@@ -108,6 +109,39 @@ def test_set_variables_values_from_modules(self):
             else:
                 self.assertEqual(var_value, default_val)
 
+    def test_definition_creation_by_dirs(self):
```
bo156, line 112: nice test!
However, I still think we need some more for the logic - maybe add 1 with more than 1 created module.
you can find a good example under `tests/terraform/graph/resources/modules`.
This one even utilizes different modules with a reference to other directories, so I feel it will be a very strong test.
(You don't have to take the whole folder as it might be also too big, maybe just one sub directory like `stacks`)

gruebel (approved):
🏖️

bo156 (approved):

PR #5371, opened 2023-07-23
chore: improve typing in Terraform local_graph

gruebel:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

- also added a couple of missing type hints in TF runner

## Checklist:


checkov/terraform/graph_builder/local_graph.py
```diff
@@ -244,12 +249,15 @@ def _build_edges_for_vertex(self, origin_node_index: int, vertex: TerraformBlock
...
                             if not self.use_new_tf_parser:
                                 source_module_object = referenced_module_object if source_module_object else None
-                            dest_node_index = self._find_vertex_index_relative_to_path(
-                                vertex_reference.block_type, reference_name, referenced_module_path,
-                                vertex.module_dependency,
-                                vertex.module_dependency_num, referenced_module_idx,
-                                source_module_object=source_module_object
-                            )
+                            if referenced_module_path is None:
+                                dest_node_index = -1
```
bo156, line 253: is this correct to use? why not just returning here?

checkov/terraform/graph_builder/local_graph.py
```diff
@@ -478,9 +495,14 @@ def update_vertex_attribute(
         vertex_index: int,
         attribute_key: str,
         attribute_value: Any,
-        change_origin_id: int,
+        change_origin_id: int | None,
         attribute_at_dest: Optional[Union[str, List[str]]],
+        transform_step: bool = False,
```
bo156, line 500: where do you use this var? or was it something like a method inheritance with broken interface?

checkov/terraform/graph_builder/local_graph.py
```diff
@@ -495,13 +517,14 @@ def update_vertices_configs(self) -> None:
...
-    def update_vertex_config(self, vertex: TerraformBlock, changed_attributes: Union[List[str], Dict[str, Any]], dynamic_blocks: bool = False) -> None:
+    @staticmethod
+    def update_vertex_config(vertex: TerraformBlock, changed_attributes: Union[List[str], Dict[str, Any]], dynamic_blocks: bool = False) -> None:
         if not changed_attributes:
             # skip, if there is no change
             return
 
         vertex_name = vertex.name
-        updated_config = pickle_deepcopy(vertex.config)
+        updated_config: Any = pickle_deepcopy(vertex.config)
```
bo156, line 527: maybe `dict[str, Any]` is better here

checkov/terraform/runner.py
```diff
@@ -292,7 +290,7 @@ def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter)
                     report.add_record(record=record)
         return report
 
-    def get_entity_context_and_evaluations(self, entity):
+    def get_entity_context_and_evaluations(self, entity: dict[str, Any]) -> tuple[dict[str, Any] | None, None]:
```
bo156, line 293: the second return type is always `None`? seems we can just drop it no?

bo156 (commented):
Great job tackling this issues :) 💯

checkov/terraform/graph_builder/local_graph.py
gruebel: since it is inside a for loop it can change with the next iteration, that's why I can't just return it.

checkov/terraform/graph_builder/local_graph.py
gruebel: in the parent, I need the signature to be identical.

checkov/terraform/graph_builder/local_graph.py
gruebel: the problem is a few lines later it overridden with a different type. But I can try to make to separate it somehow to be a bit more precise on the type.

checkov/terraform/runner.py
gruebel: yeah, it looks like. Checked current usage, also in downstream projects, and it seems not to be set in a subclass.

bo156 (approved):

ChanochShayner (approved):

PR #5375, opened 2023-07-25
feat(terraform): Use new function to create multi graph instead of single graph

maxamel:
This PR builds on top of https://github.com/bridgecrewio/checkov/pull/5362 to put the new graph splitting function to use depending on env var. Added tests for when the env var will be set.

## Checklist:


tests/terraform/runner/test_runner.py
```diff
@@ -438,6 +444,15 @@ def test_provider_uniqueness(self):
         provider = next(check for check in result.passed_checks if check.resource == "aws.one-line")
         self.assertIsNotNone(provider.file_line_range)
 
+    def test_entire_resources_folder(self):
```
bo156, line 447: really cool solution for the tests!

checkov/common/output/record.py
```diff
@@ -228,7 +228,7 @@ def get_output_id(self, use_bc_ids: bool) -> str:
         return self.bc_check_id if self.bc_check_id and use_bc_ids else self.check_id
 
     def get_unique_string(self) -> str:
-        return f"{self.check_id}.{self.check_result}.{self.file_abs_path}.{self.file_line_range}.{self.resource}"
+        return f"{self.check_id}.{self.file_abs_path}.{self.file_line_range}.{self.resource}"
```
bo156, line 231: why is this changed?

checkov/common/runners/base_runner.py
```diff
@@ -125,16 +126,19 @@ def load_external_checks(self, external_checks_dir: List[str]) -> None:
...
+    def run_graph_checks_results(self, runner_filter: RunnerFilter, report_type: str, graph: Graph | None = None
+                                 ) -> dict[BaseGraphCheck, list[_CheckResult]]:
         checks_results: "dict[BaseGraphCheck, list[_CheckResult]]" = {}
-        if not self.graph_manager or not self.graph_registry:
+        if graph is None and (not self.graph_manager or not self.graph_registry):
             # should not happen
             logging.warning("Graph components were not initialized")
             return checks_results
 
+        if graph is None and self.graph_manager is not None and isinstance(self.graph_manager, GraphManager):
```
bo156, line 137: the check here for `self.graph_manager` is not None can be removed as `isinstance` will be sufficient to catch it as well

checkov/common/runners/base_runner.py
```diff
@@ -125,16 +126,19 @@ def load_external_checks(self, external_checks_dir: List[str]) -> None:
     def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter) -> Report:
         return Report(check_type="not_defined")
 
-    def run_graph_checks_results(self, runner_filter: RunnerFilter, report_type: str) -> dict[BaseGraphCheck, list[_CheckResult]]:
+    def run_graph_checks_results(self, runner_filter: RunnerFilter, report_type: str, graph: Graph | None = None
```
bo156, line 129: why do you need to send the `graph` here as well? (instead of always using `get_reader_endpoint`)

checkov/common/runners/base_runner.py
```diff
@@ -125,16 +126,19 @@ def load_external_checks(self, external_checks_dir: List[str]) -> None:
...
             # should not happen
             logging.warning("Graph components were not initialized")
             return checks_results
 
+        if graph is None and self.graph_manager is not None and isinstance(self.graph_manager, GraphManager):
+            graph = self.graph_manager.get_reader_endpoint()
         for r in itertools.chain(self.external_registries or [], [self.graph_registry]):
-            r.load_checks()
-            registry_results = r.run_checks(self.graph_manager.get_reader_endpoint(), runner_filter, report_type)  # type:ignore[union-attr]
+            r.load_checks()  # type:ignore[union-attr]
```
bo156, line 140: seems like previously the only ignore was one line below. do you need it here as well?

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
+                        download_external_modules=runner_filter.download_external_modules,
+                        external_modules_download_path=runner_filter.external_modules_download_path,
+                        parsing_errors=parsing_errors,
+                        excluded_paths=runner_filter.excluded_paths,
+                        vars_files=runner_filter.var_files,
+                        create_graph=CHECKOV_CREATE_GRAPH,
+                    )
+                    local_graph = []
+                    for graph, definitions in graphs_with_definitions:
+                        for definition in definitions:
```
bo156, line 140: that's to update all definition after variable renedering (and other actions)?

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
+                            self.definitions.update(definition)
```
bo156, line 141: so just to make sure - basically we still have one runner with all of the definitions, but multiple graphs.

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
+                        source_dir=root_folder,
+                        local_graph_class=self.graph_class,
+                        download_external_modules=runner_filter.download_external_modules,
+                        external_modules_download_path=runner_filter.external_modules_download_path,
+                        parsing_errors=parsing_errors,
+                        excluded_paths=runner_filter.excluded_paths,
+                        vars_files=runner_filter.var_files,
+                        create_graph=CHECKOV_CREATE_GRAPH,
+                    )
+                    # Make graph a list to allow single processing method for all cases
```
bo156, line 154: nice!

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
             elif files:
                 files = [os.path.abspath(file) for file in files]
                 root_folder = os.path.split(os.path.commonprefix(files))[0]
                 self.parser.evaluate_variables = False
                 self._parse_files(files, parsing_errors)
 
                 if CHECKOV_CREATE_GRAPH:
-                    local_graph = self.graph_manager.build_graph_from_definitions(self.definitions)
+                    # local_graph needs to be a list to allow supporting multi graph
+                    local_graph = [self.graph_manager.build_graph_from_definitions(self.definitions)]
```
bo156, line 164: shouldn't that part also use logic of build multiple graphs?

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
-                    source_dir=root_folder,
-                    local_graph_class=self.graph_class,
-                    download_external_modules=runner_filter.download_external_modules,
-                    external_modules_download_path=runner_filter.external_modules_download_path,
-                    parsing_errors=parsing_errors,
-                    excluded_paths=runner_filter.excluded_paths,
-                    vars_files=runner_filter.var_files,
-                    create_graph=CHECKOV_CREATE_GRAPH,
-                )
+                tf_split_graph = strtobool(os.getenv('TF_SPLIT_GRAPH', 'False'))
```
bo156, line 126: as this part begins to be a bit big - what do you think about splitting the `if/else` block to a new function called `_generate_graphs_list` (or something like that)?

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
         parsing_errors: dict[str, Exception] = {}
         self.load_external_checks(external_checks_dir)
         local_graph = None
-
+        all_graphs = []
```
bo156, line 120: seems like you only use this variable at the end of this function, so I think it's better to move it there

checkov/terraform/runner.py
```diff
@@ -212,9 +244,9 @@ def get_connected_node(self, entity, root_folder) -> Optional[Dict[str, Any]]:
         connected_node_data['resource_address'] = connected_entity_context.get('address')
         return connected_node_data
 
-    def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter) -> Report:
+    def get_graph_checks_report(self, root_folder: str, runner_filter: RunnerFilter, graph: igraph.Graph | None = None) -> Report:
         report = Report(self.check_type)
-        checks_results = self.run_graph_checks_results(runner_filter, self.check_type)
+        checks_results = self.run_graph_checks_results(runner_filter, self.check_type, graph)
```
bo156, line 249: why do you need to send the graph as well?

checkov/terraform/runner.py
maxamel: Yes it was inside at the beginning but then I got a bunch of errors like: `Local variable referenced before assignment`
The reason is we use this starting at line 195, so I have to declare it out of the inner scope

checkov/common/output/record.py
maxamel: This is the solution for filtering out checks with same resource id. Until we tackle it in a broader way this seems as the simplest and fastest approach that won't potentially break things. It seems we could've used this unique string all along and there was no real need for the check result to be part of this string.

checkov/terraform/runner.py
maxamel: I thought about it but the problem is here we are outputting the graph but also updating the self.definitions. So the definitions update will be like a silent side effect.

checkov/terraform/runner.py
maxamel: Yes multiple graphs and definitions is still a dict as before.

checkov/terraform/runner.py
maxamel: yes the rendering occurs before that.

checkov/common/runners/base_runner.py
maxamel: We want to run on a specific graph, because we have multiple graphs now, not one as before

checkov/terraform/runner.py
maxamel: I think it's connected to this comment: https://github.com/bridgecrewio/checkov/pull/5375#discussion_r1273370241
We need to run on a specific graph now

checkov/terraform/runner.py
```diff
@@ -116,47 +117,73 @@ def run(
...
+                                vertex_id = vertex.id
+                            report.add_resource(f'{vertex.path}:{vertex_id}')
+                    igraph_graph = self.graph_manager.save_graph(graph)
+                    all_graphs.append(igraph_graph)
+                    current_definitions, current_breadcrumbs = convert_graph_vertices_to_tf_definitions(
+                        graph.vertices,
+                        root_folder,
+                    )
+                    self.definitions.update(current_definitions)
+                    self.breadcrumbs.update(current_breadcrumbs)
```
gruebel, line 186: maybe it is time to move this to a separate method?

tests/terraform/runner/test_runner.py
```diff
@@ -438,6 +444,15 @@ def test_provider_uniqueness(self):
...
+        current_dir = os.path.dirname(os.path.realpath(__file__))
+        valid_dir_path = current_dir + "/resources/"
+        runner = Runner(db_connector=self.db_connector())
+        if isinstance(runner.db_connector, IgraphConnector):
+            result = runner.run(root_folder=valid_dir_path, external_checks_dir=None, runner_filter=RunnerFilter())
```
gruebel, line 452: can you please limit it to fixed set of policies, because otherwise you would need to adjust the amount passed/failed checks with each new policy added

checkov/common/output/record.py
bo156: So 2 things here:
1. why does this even solve the uniqueness? In the case we saw togther the check_result was the same, so it seems like this won't be the right thing to solve (perhaps it wasn't the case here but we could have 2 checks with the same results on the same resource, so this won't filter it correctly).
2. I would like some way of knowing that the change we do here is temporary. For example - adding another function with clear documentation about why do we handle/remove duplicates. This way it will be better documented for others as well

checkov/common/runners/base_runner.py
bo156: sound right, hadn't though about it.

checkov/terraform/runner.py
```diff
@@ -161,29 +161,15 @@ def run(
 
                 if CHECKOV_CREATE_GRAPH:
                     # local_graph needs to be a list to allow supporting multi graph
```
bo156, line 163: I think this comment should now move to line 167

checkov/terraform/runner.py
```diff
@@ -216,6 +202,26 @@ def run(
 
         return report
 
+    def update_definitions_and_breadcrumbs(self, all_graphs, local_graph, report, root_folder):
```
bo156, line 205: should probably be private function to make sure it's more clear it's an inner function

tests/terraform/runner/test_runner.py
```diff
@@ -449,10 +449,35 @@ def test_entire_resources_folder(self):
         valid_dir_path = current_dir + "/resources/"
         runner = Runner(db_connector=self.db_connector())
         if isinstance(runner.db_connector, IgraphConnector):
-            result = runner.run(root_folder=valid_dir_path, external_checks_dir=None, runner_filter=RunnerFilter())
-            self.assertEqual(len(result.passed_checks), 701)
-            self.assertEqual(len(result.failed_checks), 896)
-            self.assertEqual(len(result.skipped_checks), 7)
+            result = runner.run(root_folder=valid_dir_path, external_checks_dir=None, runner_filter=RunnerFilter(
+                checks=['CKV_AWS_21', 'CKV_AWS_42', 'CKV_AWS_62', 'CKV_AWS_53', 'CKV_AWS_18', 'CKV_AWS_61', 'CKV_AWS_144',
```
bo156, line 453: why did you change it to include only certain checks? and less overall results?

tests/terraform/runner/test_runner.py
```diff
@@ -449,10 +449,35 @@ def test_entire_resources_folder(self):
...
+        valid_dir_path = current_dir + "/resources"
+        runner = Runner(db_connector=self.db_connector())
+        if isinstance(runner.db_connector, IgraphConnector):
+            res = []
+            for (dir_path, dir_names, file_names) in os.walk(valid_dir_path):
+                for file in file_names:
+                    res.append(os.path.join(dir_path, file))
+            result = runner.run(files=res, root_folder=None, external_checks_dir=None,
+                                runner_filter=RunnerFilter(
+                                    checks=['CKV_AWS_21', 'CKV_AWS_42', 'CKV_AWS_62', 'CKV_AWS_109', 'CKV_AWS_168',
```
bo156, line 472: same here

tests/terraform/runner/test_runner.py
gruebel: otherwise this test needs to be adjusted each time we add new checks

gruebel (approved):
🏆

checkov/common/output/record.py
maxamel: Discussed offline,
After the resource_id eventual fix we won't need to revert this since it's unique enough

tests/terraform/runner/test_runner.py
maxamel: As Anton suggested, to avoid having to update this test for every added policy, I chose the policy IDs with the most results so we have fixed amount of policies, but still have enough cases and results

bo156 (approved):

PR #5412, opened 2023-08-03
feat(secrets): Make non-entropy signatures take precedence over entropy signatures

ChanochShayner:

## Description
Make non-entropy signatures take precedence over entropy signatures.

## Checklist:


checkov/secrets/runner.py
```diff
@@ -72,6 +72,9 @@
     'Twilio API Key': 'CKV_SECRET_18',
     'Hex High Entropy String': 'CKV_SECRET_19'
 }
+
+ENTROPY_CHECK_IDS = ('CKV_SECRET_6', 'CKV_SECRET_19')
```
omryMen, line 76: we also have CKV_SECRET_80 in the platform

checkov/secrets/runner.py
```diff
@@ -231,12 +234,8 @@ def run(
                     logging.info(
                         f"Removing secret due to UUID filtering: {hashlib.sha256(secret.secret_value.encode('utf-8')).hexdigest()}")
                     continue
-                if secret_key in secrets_duplication:
-                    logging.debug(
-                        f'Secret was filtered - secrets_duplication. line_number {secret.line_number}, check_id {check_id}')
-                    continue
-                else:
-                    secrets_duplication[secret_key] = True
+                if secret_key in secret_records.keys() and secret_records[secret_key].check_id in ENTROPY_CHECK_IDS and check_id not in ENTROPY_CHECK_IDS:
```
omryMen, line 237: what if the order will be the other way ?

checkov/secrets/runner.py
```diff
@@ -231,12 +234,8 @@ def run(
...
+                    secret_records.pop(secret_key)
```
bo156, line 238: don't you need to also add it back in the end?

checkov/secrets/runner.py
bo156, line 76: So basically this means that anyone adding a new check which uses entropy has to remember/know that this list exists. Is there a place or something you have to inherit to create a new secret check?
If we have something like that, we can add an API for understanding wether it's an entopy check or not

checkov/secrets/runner.py
ChanochShayner: Nope, we just removed the existing and the new will add in this iteration.

checkov/secrets/runner.py
ChanochShayner: 1. Don't think that we will add a new entropy.
2. There is no place to inherit from.

ChanochShayner: Makes very sense, but no place to inherit from.

omryMen (approved):

lirshindalman (approved):

tests/secrets/custom_and_entropy/main.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_instance" "a" {
+  test_pass: "z2b7k2cQfzc+yjP2K8cjuQ8uoorHBpEvC+XWhU3Z5+IdrPQYwr991Lj73xfZ+RA2GzC0wTedDTvb1C2NX+3Gpw==",
```
gruebel, line 2:
```suggestion
  test_pass = "z2b7k2cQfzc+yjP2K8cjuQ8uoorHBpEvC+XWhU3Z5+IdrPQYwr991Lj73xfZ+RA2GzC0wTedDTvb1C2NX+3Gpw=="
```
this should handle the TF parsing error

Issue #5627, opened 2023-10-06
CKV_AWS_33: error in check results in false negative for terraform


Michael-McClelland:
**Describe the issue**
Check CKV_AWS_33 generates a false negative when the key policy size exceeds a certain threshold for the Terraform check.  The variable assignment [here](https://github.com/bridgecrewio/checkov/blob/8b576e699f93dde4b330347908e8e04562f4e356/checkov/terraform/checks/resource/aws/KMSKeyWildcardPrincipal.py#L19) is the source of the issue.  Once the key policy exceeds a certain size the type of this variable changes to a string from a dictionary.  The following for loop cannot iterate it properly and the check silently succeeds despite a failing statement existing in the policy.

**Examples**
```
resource "aws_kms_key" "thiskeyshouldfailbutdoesnot" {
  description             = "KMS Key Example that should fail check 33"
  deletion_window_in_days = 30

  policy              = <<EOF
{
    "Version": "2012-10-17",
    "Statement": [        
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "*"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
            }
        },
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:root"
            },
            "Action": [
                "kms:*"
            ],
            "Resource": "*"
        },        {
            "Effect": "Allow",
            "Principal": {
[...]

PR #5628, opened 2023-10-06
fix(terraform): change CKV_AWS_33 into CKV2 check

Michael-McClelland:

## Description

~An error occurs in check CKV_AWS_33 when the key policy is large.  The dictionary that the check passes to the for loop becomes a string once a certain line count threshold is reached.  Since no exception is thrown as the loop processes the string the check produces no error and silently and incorrectly passes.~

Converts CKV_AWS_33 into V2 formatted check

~Fixes~ (#5627)

### Fix
~When executed in the new graph check format format a check similar to CKV_AWS_33 produces no error.  Thus the simplest correction is to upgrade the check to the new format.~

Upgrades check to graph format method.

## Checklist:


checkov/terraform/checks/graph_checks/aws/KmsKeyWildcardPrincipal.yaml
```diff
@@ -0,0 +1,28 @@
...
+      resource_types:
+            - "aws_kms_key_policy"
+      attribute: "policy.Statement[?(@.Effect == 'Allow' & @.Principal == '*')]"
+      operator: "jsonpath_not_exists"
+    - cond_type: "attribute"
+      resource_types:
+            - "aws_kms_key_policy"
+      attribute: "policy.Statement[?(@.Effect == 'Allow')].Principal.AWS[*]"
+      operator: "jsonpath_not_equals"
+      value: "*"
```
gruebel, line 28: your policy is handling key to policy connection, but not when the policy is inlined, that's why the actual test fails for all resources.

checkov/terraform/checks/graph_checks/aws/KmsKeyWildcardPrincipal.yaml
```diff
@@ -0,0 +1,28 @@
+---
+metadata:
+  id: "CKV2_AWS_67"
```
gruebel, line 3:
```suggestion
  id: "CKV_AWS_33"
```
no need to create a new ID

checkov/terraform/checks/graph_checks/aws/KmsKeyWildcardPrincipal.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: "Ensure KMS key policy does not contain wildcard (*) principal"
+  category: "GENERAL_SECURITY"
```
gruebel, line 5:
```suggestion
  category: "IAM"
```
this fits best

gruebel (commented):
I think it is a good idea to migrate it over to a YAML based policy to be able to cover the case of key to policy connection 🏅

checkov/terraform/checks/graph_checks/aws/KmsKeyWildcardPrincipal.yaml
Michael-McClelland: Id change made

checkov/terraform/checks/graph_checks/aws/KmsKeyWildcardPrincipal.yaml
Michael-McClelland: category change set

Michael-McClelland:
There is a bug in this I cannot explain.  When evaluating keys that use a data block for the policy unless the resources involved trigger a fail from the check it doesn't report the resource as having been evaluated.  Passes are not reported when the key policy is coming from a data resource, only failures.  

For example this reports a non-compliant resource as intended

```
provider "aws" {
  region = "us-east-1"
}

data "aws_iam_session_context" "current" {
  arn = data.aws_caller_identity.current.arn
}
data "aws_caller_identity" "current" {}
data "aws_organizations_organization" "current" {}
data "aws_partition" "current" {}
data "aws_region" "current" {}

data "aws_iam_policy_document" "keypolicy" {

  statement {
    principals {
      type = "AWS"
      identifiers = [
        "*"
      ]
    }
    effect = "Allow"
    actions = [
      "kms:*"
    ]
    resources = [
      "*",
    ]
  }

}


resource "aws_kms_key" "TestKey" {
  description                        = "testkey"
  deletion_window_in_days            = 7
  key_usage                          = "ENCRYPT_DECRYPT"
  customer_master_key_spec           = "SYMMETRIC_DEFAULT"
  policy                             = data.aws_iam_policy_document.keypolicy.json
  bypass_policy_lockout_safety_check = false
  is_enabled                         = true
  enable_key_rotation                = true
}
```

but this silently passes and is not acknowledged as compliant
```
provider "aws" {
  region = "us-east-1"
}


data "aws_iam_session_context" "current" {
  arn = data.aws_caller_identity.current.arn
}
data "aws_caller_identity" "current" {}
data "aws_organizations_organization" "current" {}
data "aws_partition" "current" {}
data "aws_region" "current" {}

data "aws_iam_policy_document" "keypolicy" {

  statement {
    principals {
      type = "AWS"
      identifiers = [
        "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.id}:root"
      ]
    }
    effect = "Allow"
    actions = [
      "kms:*"
    ]
    resources = [
      "*",
    ]
  }

}


resource "aws_kms_key" "TestKey" {
  description                        = "testkey"
  deletion_window_in_days            = 7
  key_usage                          = "ENCRYPT_DECRYPT"
  customer_master_key_spec           = "SYMMETRIC_DEFAULT"
  policy                             = data.aws_iam_policy_document.keypolicy.json
  bypass_policy_lockout_safety_check = false
  is_enabled                         = true
  enable_key_rotation                = true
}
```

Saarett:
@Michael-McClelland How do you feel about @gruebel 's [comment](https://github.com/bridgecrewio/checkov/issues/5627#issuecomment-1761821012)? I tend to agree with him

PR #5691, opened 2023-10-25
platform(general): Do not crash the run if S3 integration fails during setup, upload, or finalize

mikeurbanski1:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Adds logic to gracefully handle issues that occur during the integrations API calls or S3 uploads to still allow results to be visible locally.

If at any point we encounter a failure with this flow, then we still stop trying subsequent actions and just save that an error occurred. The scan will continue locally, and instead of a details URL, we will show an error message.

Actual auth errors will still stop the run, so this mainly guards against bugs or crashes in the integrations API, which has caused us some issues lately.

This also fixes an issue where a bug-based error response from the integrations POST request results in an infinite loop that blocks checkov.

Note that we intentionally still stop the run if the `runConfiguration` call fails (unless the customer uses the flag to ignore it), because we know from experience that missing this data (enforcement rules, custom policies, etc) gives the user unexpected results that they then raise an issue about anyways.

Also also adds some logic to handle the edge case of when this happens and the customer uses the `--support` flag, so that we can still attempt to save their debug logs somewhere they can access.

Also also also logs some response headers to make finding requests easier from the logs.

## Checklist:

### DIFF
--- a/checkov/common/bridgecrew/platform_integration.py
+++ b/checkov/common/bridgecrew/platform_integration.py
@@ -661,7 +661,7 @@ class BcPlatformIntegration:
                 self.s3_setup_failed = True
             except JSONDecodeError:
                 if request:
-                    logging.warning(f"Response (status: {request.status}) of {self.integrations_api_url}: {request.data.decode('utf8')}")  # nosec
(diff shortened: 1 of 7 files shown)


rotemavni (approved):

gruebel (approved):
🍺

PR #5807, opened 2023-11-28
chore: Expose graph to Python checks

ChanochShayner:

## Description
Expose graph to Python checks + UT as an example.

## Checklist:

### DIFF
--- a/checkov/terraform/runner.py
+++ b/checkov/terraform/runner.py
@@ -396,7 +396,7 @@ class Runner(BaseTerraformRunner[_TerraformDefinitions, _TerraformContext, TFDef
                 entity_evaluations = BaseVariableEvaluation.reduce_entity_evaluations(
                     variables_evaluations, entity_context_path
                 )
-            registry.graph = self.all_graphs[0][0]
+            registry.graph = self._get_graph_for_registry()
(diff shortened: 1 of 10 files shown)


lirshindalman (approved):
nice!

checkov/terraform/runner.py
```diff
@@ -396,6 +396,12 @@ def run_block(
                 entity_evaluations = BaseVariableEvaluation.reduce_entity_evaluations(
                     variables_evaluations, entity_context_path
                 )
+            registry.graph = None
+            if self.all_graphs and isinstance(self.all_graphs, list):
```
bo156, line 400: why do you take only the first graph?

ChanochShayner: fixed

bo156: I think you are still missing some graph objects.
Based on the type - `self.all_graphs: list[tuple[LibraryGraph, Optional[str]]] = []`, we can have multiple tuples of sub-graphs here. What you are doing is only taking the first one you see rather then all of the graphs.
You can watch the creation of `self.all_graphs` under `_update_definitions_and_breadcrumbs`.

Maybe you want to just send all of the graphs instead of breaking? or maybe merge the graphs to 1?

PR #5887, opened 2023-12-20
fix(general): fix multiprocess abilities

AdamDev:

**There are 2 bugs:**
1. SCA and SAST are not working well in Darwin via Fork mode (security issue of Mac).
2. Can't override Windows parallel type to None.

**Solution:**
1. In the case of Darwin, change the parallel type to Threads, the best practice is 'spawn ' but it starts the child process with a fresh memory, and Checkov is not supporting this case yet, the runners still use the parent's memory.
2. Support ability for None parallel type in Windows.

## Checklist:


checkov/kustomize/runner.py
```diff
@@ -674,47 +671,21 @@ def run_kustomize_to_k8s(
...
-                target=self._run_kustomize_parser,
-                args=(
-                    filePath,
-                    shared_kustomize_file_mappings,
-                    self.kustomizeProcessedFolderAndMeta,
-                    self.templateRendererCommand,
-                    self.target_folder_path
-                )
+        shared_kustomize_file_mappings: dict[str, str] = {}
+        for file_path in self.kustomizeProcessedFolderAndMeta:
```
gruebel, line 679: this makes the whole `kustomize` processing sequential, right?

gruebel (commented):
feels like cheating by excluding the `secrets` runner, which failed the last time 😄

checkov/kustomize/runner.py
AdamDev: This is a temporary solution because, there is a central place for managing parallel processes (parallel_runner.py) and this code is creating processes by itself, which cause issues, so until it will also use parallel_runner I prefer to disable it's feature

checkov/common/parallelizer/parallel_runner.py
```diff
@@ -19,46 +20,88 @@
...
             # 'spawn' mode results in a strange error, which needs to be investigated on an actual Windows machine
-            self.type = ParallelizationType.THREAD
+            type = ParallelizationType.THREAD
+        elif self.os == "Darwin":
+            # 'fork' mode is not supported on 'Darwin'
+            type = ParallelizationType.SPAWN
+        else:
+            type = ParallelizationType.FORK
+        if type == ParallelizationType.SPAWN and getattr(sys, 'frozen', False):
+            # if application is running from a frozen executable, spawn mode is not supported
```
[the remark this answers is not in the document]
AdamDev: Yes, I've noticed that it doesn't works with spawn

checkov/common/parallelizer/parallel_runner.py
```diff
@@ -19,46 +20,88 @@
 
 class ParallelRunner:
     def __init__(
-        self, workers_number: int | None = None, parallelization_type: ParallelizationType = ParallelizationType.FORK
+        self, workers_number: int | None = None
```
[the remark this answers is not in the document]
AdamDev: The root issue is that SCA and SAST are not working well in Darwin via Fork mode (security issue of Mac).
There are multiple options to solve this:
1. **use Spawn** (default of Windows) - this is cross-platform solution but it erase the memory of each new process (not like fork the copy the parent process memory) - but to make it works need to go all over the parents memory and pass it to the child processes, it can take lots of time, I do see some attempt to support it:

```
            if parallel_runner.type == ParallelizationType.SPAWN:
                platform_integration_data = bc_integration.generate_instance_data()
```

2. **in case of macOS (Darwin) use threads** - this option is less isolated, and has less performance because of the GIL, but it will solve this issue.

3. **in case of macOS (Darwin) use threads only for SCA and SAST runners** - same as above, but the fork issue can happen in the future for other runners.

In my opinion it should be 2, to support all macOs cases, 
WDYT?

checkov/common/parallelizer/parallel_runner.py
```diff
@@ -33,10 +33,13 @@ def __init__(
             # PYCHARM_HOSTED env variable equals 1 when debugging via jetbrains IDE.
             # To prevent JetBrains IDE from crashing on debug run sequentially
             self.type = ParallelizationType.NONE
-        elif self.os == "Windows":
-            # 'fork' mode is not supported on 'Windows'
-            # 'spawn' mode results in a strange error, which needs to be investigated on an actual Windows machine
-            self.type = ParallelizationType.THREAD
+        elif self.os == "Windows" or self.os == "Darwin":
```
gruebel, line 36: can you also allow to overwrite the behaviour via the `CHECKOV_PARALLELIZATION_TYPE` env var here? Maybe do an initial if condition check for the env var and use it instead of overwriting it for both OS, then users could also set it to none, if needed?

AdamDev: Sure I'll add

gruebel (approved):
🍰

SteveVaknin (approved):

achiar99 (approved):

PR #5892, opened 2023-12-21
feat(sast): add dataflow to checkov report from sast

RabeaZr:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/sast/engines/prisma_engine.py
+++ b/checkov/sast/engines/prisma_engine.py
@@ -311,16 +311,15 @@ class PrismaEngine(SastEngine):
                     file_abs_path = location.path
                     file_path = file_abs_path.split('/')[-1]
                     file_line_range = [location.start.row, location.end.row]
-
+                    split_code_block = [line + '\n' for line in location.code_block.split('\n')]
+                    code_block = get_code_block_from_start(split_code_block, location.start.row)
+                    data_flow = []
                     if match.metadata.taint_mode is not None:
-                        code_block = get_data_flow_code_block(match.metadata.taint_mode.data_flow)
-                    else:
-                        split_code_block = [line + '\n' for line in location.code_block.split('\n')]
-                        code_block = get_code_block_from_start(split_code_block, location.start.row)
+                        data_flow = get_data_flow_code_block(match.metadata.taint_mode.data_flow)
 
                     record = SastRecord(check_id=check_id, check_name=check_name, resource="", evaluations={},
                                         check_class="", check_result=check_result, code_block=code_block,
-                                        file_path=file_path, file_line_range=file_line_range,
+                                        file_path=file_path, file_line_range=file_line_range, data_flow=data_flow,
                                         file_abs_path=file_abs_path, severity=severity, cwe=check_cwe,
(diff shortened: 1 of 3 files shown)


checkov/sast/record.py
```diff
@@ -43,6 +45,7 @@ def __init__(self,
         self.cwe = cwe
         self.owasp = owasp
         self.show_severity = show_severity
+        self.dataflow = data_flow
```
arielkru, line 48:
```suggestion
        self.data_flow = data_flow
```
no?

checkov/sast/record.py
```diff
@@ -14,6 +15,7 @@ def __init__(self,
                  check_name: str,
                  check_result: _CheckResult,
                  code_block: List[Tuple[int, str]],
+                 data_flow: List[Tuple[int, str]],
```
arielkru, line 18: what if its not taint mode? why always add this as an empty list? let's default it to None and not add it to the final json in that case

checkov/sast/record.py
arielkru: I think its better to add a `metadata` that keeps the structure of our report, so its easier to extend or change in the future. So in this case its `metadata.taint_flow` (originaly its taint_mode but I think tait_flow is clearer)

marynaKK:
Please add tests with some test cases to see the addition is working as expected and will not break in the future on future changes :)

marynaKK (approved):

checkov/sast/engines/prisma_engine.py
```diff
@@ -21,6 +21,7 @@
 from checkov.common.sast.consts import SastLanguages
 from checkov.common.sca.reachability.sast_contract.data_fetcher_sast_lib import SastReachabilityDataFetcher
 from checkov.common.typing import _CheckResult
+from checkov.common.typing import _Metadata
```
arielkru, line 24:
```suggestion
from checkov.common.typing import _CheckResult, _Metadata
```

checkov/sast/engines/prisma_engine.py
```diff
@@ -311,16 +312,15 @@ def create_report(self, prisma_report: PrismaReport) -> List[SastReport]:
                     file_abs_path = location.path
                     file_path = file_abs_path.split('/')[-1]
                     file_line_range = [location.start.row, location.end.row]
-
+                    split_code_block = [line + '\n' for line in location.code_block.split('\n')]
+                    code_block = get_code_block_from_start(split_code_block, location.start.row)
+                    metadata = None
                     if match.metadata.taint_mode is not None:
```
arielkru, line 318:
```suggestion
                    metadata = _Metadata(taint_flow=get_data_flow_code_block(match.metadata.taint_mode.data_flow)) if match.metadata.taint_mode else None
```
I think this or extracting the population logic to another function and here just call `metadata = parse_metadata(match.metadata)` are better approaches, rather than declaring the `metadata = None` and then conditionally overriding it (with only one option). At least in python its not needed in this case

arielkru (approved):

PR #6119, opened 2024-03-25
feat(terraform): fix for_each resource handling

maxamel:
This PR fixes the extraction of module from resource of for_each, preventing correct handling of first resource (index 0) and thus missing any violations for this resource.

## Checklist:

### DIFF
--- a/checkov/terraform/modules/module_utils.py
+++ b/checkov/terraform/modules/module_utils.py
@@ -242,7 +242,7 @@ def get_module_name(file_path: TFDefinitionKey) -> str | None:
     if not file_path.tf_source_modules:
         return None
     module_name = file_path.tf_source_modules.name
-    if file_path.tf_source_modules.foreach_idx is not None:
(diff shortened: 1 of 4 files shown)


checkov/terraform/modules/module_utils.py
```diff
@@ -242,7 +242,7 @@ def get_module_name(file_path: TFDefinitionKey) -> str | None:
     if not file_path.tf_source_modules:
         return None
     module_name = file_path.tf_source_modules.name
-    if file_path.tf_source_modules.foreach_idx:
+    if file_path.tf_source_modules.foreach_idx is not None:
```
lirshindalman, line 245: in case of an empty string, we want to get `foreach_or_count = '"' ` or `= ''`?

maxamel: I think none of them, I didn't realize foreach_idx can be string. I changed it so it checks if it is int, otherwise it keeps the previous behavior.

SteveVaknin (approved):

lirshindalman (approved):

PR #6242, opened 2024-05-05
feat(ARM): check - AppServiceDisallowCors 

RachelBorzi:
policy that checks the cors to server


ChanochShayner (commented):
Already implemented here - https://github.com/bridgecrewio/checkov/pull/6244
Can we close this PR?

PR #6389, opened 2024-06-03
feat(terraform): add fix for not contains solver

### LINKED ISSUE #6333: CKV2_AWS_44 is showing false negative
**Describe the issue**
CKV2_AWS_44 (Ensure AWS route table with VPC peering does not contain routes overly permissive to all traffic) is showing false-negative.

The policy is passing even when one of the inline routes in aws route table is overly permissive. 

**Examples**
```
resource "aws_route_table" "private_route_table" {
  vpc_id = aws_vpc.vpc.id

  route {
    cidr_block                            = "10.0.0.0/32"
    vpc_peering_connection_id = var.vpc_peering_connection_id1
  }

  route {
    cidr_block                            = "0.0.0.0/0"
    vpc_peering_connection_id = var.vpc_peering_connection_id1
  }
}
```

 - Checkov Version: 3.2.98

**Additional context**
thank you for the [fix](https://github.com/bridgecrewio/checkov/pull/6301) for a similar (false-positive) [issue](https://github.com/bridgecrewio/checkov/issues/6277). I've tested the fix - it solved the false-positives.

maxamel:
This PR fixes the not_contains solver to take into account lists of values so that all values are checked and not any one of them.

Fixes #6333

## Checklist:

### DIFF
--- a/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
@@ -154,8 +154,7 @@ class BaseAttributeSolver(BaseSolver):
     def _evaluate_attribute_matches(
         self, vertex: dict[str, Any], attribute_matches: list[str], filtered_attribute_matches: list[str]
     ) -> bool | None:
-        from common.checks_infra.solvers import NotContainsAttributeSolver
-        if self.is_jsonpath_check or isinstance(self, NotContainsAttributeSolver):
+        if self.is_jsonpath_check:
             if self.resource_type_pred(vertex, self.resource_types) and all(
                 self._get_operation(vertex=vertex, attribute=attr) for attr in filtered_attribute_matches
             ):
--- a/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
+++ b/checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
@@ -151,13 +151,14 @@ class BaseAttributeSolver(BaseSolver):
         else:
             failed_vertices.append(data)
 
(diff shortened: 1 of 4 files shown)


checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -151,10 +151,13 @@ def _process_node(
...
             failed_vertices.append(data)
 
+    def is_needs_all_condition(self) -> bool:
+        return self.is_jsonpath_check
+
     def _evaluate_attribute_matches(
         self, vertex: dict[str, Any], attribute_matches: list[str], filtered_attribute_matches: list[str]
     ) -> bool | None:
-        if self.is_jsonpath_check:
+        if self.is_needs_all_condition():
```
bo156, line 160: shouldn't we always check everything by default? or every time we use a "list" check for example?

maxamel: No we shouldn't. For example for "contains" we should keep the old behavior where we need just one match. That's why I added this exception for "not_contains" only.

checkov/common/checks_infra/solvers/attribute_solvers/base_attribute_solver.py
```diff
@@ -151,10 +151,13 @@ def _process_node(
         else:
             failed_vertices.append(data)
 
+    def is_needs_all_condition(self) -> bool:
```
bo156, line 154: can you add documentation to when should we override this?
also - I think grammar wise it is better to use `should_check_all_condition`

bo156 (approved):

lirshindalman (approved):

PR #6404, opened 2024-06-05
feat(terraform): add fix for ckv_aws_300

maxamel:
This PR fixes the CKV_AWS_300 check by taking into account the filterings available for the S3 lifecycle directive.

## Checklist:

### DIFF
--- a/tests/terraform/runner/test_plan_runner.py
+++ b/tests/terraform/runner/test_plan_runner.py
@@ -68,6 +68,26 @@ class TestRunnerValid(unittest.TestCase):
         self.assertEqual(report.get_summary()["failed"], 3)
         self.assertEqual(report.get_summary()["passed"], 3)
 
+    def test_tf_plan_filtered_rule(self):
+        if not self.db_connector == RustworkxConnector:
+            return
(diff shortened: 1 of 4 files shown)


lirshindalman (approved):

tests/terraform/runner/test_plan_runner.py
```diff
@@ -68,6 +68,26 @@ def test_runner_two_checks_only(self):
         self.assertEqual(report.get_summary()["failed"], 3)
         self.assertEqual(report.get_summary()["passed"], 3)
 
+    def test_tf_plan_filtered_rule(self):
```
bo156, line 71: can you add a matching test for `terraform` as well (not just `tf_plan`)

maxamel: We have pretty thorough TF tests for this case here: `tests/terraform/checks/resource/aws/test_S3AbortIncompleteUploads.py`
That's why I added the tf_plan part, which is what was missing in our coverage and would've caught this.

bo156 (approved):

PR #6572, opened 2024-07-16
feat(general): Refactor all resource type handling in Checkov

lirshindalman:
The changes in this PR support using 'all' as resources_type and not copying the heard-coded resources list. 
in the start we want to leave the ['all'] in the resources_type list, and in the second step in the resource filtering to check if we have 'all' and the data we get its nit module or variable.


SteveVaknin (approved):
Nice 🏆

checkov/common/checks_infra/checks_parser.py
```diff
@@ -246,7 +247,11 @@ def _parse_raw_check(self, raw_check: Dict[str, Any], resources_types: Optional[
                     or (isinstance(resource_type, str) and resource_type.lower() == "all")
                     or (isinstance(resource_type, list) and resource_type[0].lower() == "all")
             ):
-                check.resource_types = resources_types or []
+                if bool(convert_str_to_bool(os.getenv('CKV_SUPPORT_ALL_RESOURCE_TYPE', False))):
```
bo156, line 250: when adding a new env_var in checkov, you should add it in `checkov/common/util/env_vars_config.py` and use it from there, allows better handling of those variables throughout the code

checkov/common/bridgecrew/integration_features/features/custom_policies_integration.py
```diff
@@ -103,7 +103,8 @@ def _convert_raw_check(policy: dict[str, Any]) -> dict[str, Any]:
             'id': policy['id'],
             'name': policy['title'],
             'category': policy['category'],
-            'frameworks': policy.get('frameworks', [])
+            'frameworks': policy.get('frameworks', []),
+            'scope': {'provider': policy.get('provider', '').lower()}
```
bo156, line 107: just writing here as I have no place else - can you add a description to your PR and explain what has changed?
I can say from my standpoint - this PR without an explanation doesn't make it clear how did we change the behavior

bo156 (approved):

PR #6598, opened 2024-07-24
feat(general): filter resource by provider for all resources types

lirshindalman:
 after we add the support for the resource_type = 'all' without using the resources_list by provider.
Up to now, resources were categorized by provider; meaning only the provider’s resource_type could be found if checked. 
in this PR we add the solution for it. 
We add the `provider` to the `solver` and check if the format of the `resource_type` is suitable for the provider we are examining (for example, if the provider is GCP, we would check if the resource_type contains ‘_google’).


checkov/common/graph/checks_infra/solvers/base_solver.py
```diff
@@ -28,12 +30,34 @@ def _get_operation(self, *args: Any, **kwargs: Any) -> Any:
...
+            resource_type_match_provider = self.resource_match_provider(v.get(resource_type, ''))
+            support_all_resources = bool(resource_type in v and is_all_resources and v.get(resource_type) != 'module' and resource_type_match_provider)
+
+            return not resource_types or support_all_resources
+
+        return not resource_types or (resource_type in v and v[resource_type] in resource_types)
+
+    def resource_match_provider(self, resource_type: str) -> bool:
+        # Based on the resource names in iac frameworks
+        aws_keys = ['aws_', 'AWS::', 'aws-']
```
bo156, line 46: could you extract those lists to consts?

checkov/common/checks_infra/checks_parser.py
```diff
@@ -204,6 +204,7 @@ def parse_raw_check(self, raw_check: Dict[str, Dict[str, Any]], **kwargs: Any) -
         check.guideline = raw_check.get("metadata", {}).get("guideline")
         check.check_path = kwargs.get("check_path", "")
         solver = self.get_check_solver(check)
+        solver.providers = providers
```
bo156, line 207: please add description to your PR

bo156 (approved):

SteveVaknin (approved):
Nice 👌

PR #6603, opened 2024-07-25
feat(terraform): - TF and CFN - Add a policy for ensuring AWS Bedrock Agent is encrypted with a CMK

inbalavital:
## Description

Add TF and CFN policy to check for the use of customer_encryption_key_arn in aws_bedrockagent_agent

## New policies
TF and CFN policy to check for the use of customer_encryption_key_arn in aws_bedrockagent_agent

## Checklist:


tests/cloudformation/checks/resource/aws/test_BedrockAgentEncrypted.py
```diff
@@ -0,0 +1,45 @@
+import os
+import unittest
+
+from checkov.cloudformation.checks.resource.aws.BedrockAgentEncrypted import check
+from checkov.cloudformation.runner import Runner
+from checkov.runner_filter import RunnerFilter
+
+
+class TestBedrockAgentEncrypted(unittest.TestCase):
+
```
ChanochShayner, line 10: Maybe removing this blank line (between the class and the func) will solve the weird cfn-lint issue

inbalavital: It didn't help, the issue is that `cfn-lint` should run only over yml and json files, it fails on every python file I checked locally

.github/workflows/pr-test.yml
```diff
@@ -43,16 +43,25 @@ jobs:
...
+          fi
+      - name: Install cfn-lint
+        if: env.RELEVANT_FILES_CHANGED == 'true'
         run: |
           pip install -U cfn-lint
       - name: Lint Cloudformation templates
-        if: steps.changed-files-specific.outputs.any_changed == 'true'
-        env:
-          ALL_CHANGED_FILES: ${{ steps.changed-files-specific.outputs.all_changed_files }}
+        if: env.RELEVANT_FILES_CHANGED == 'true'
```
maxamel, line 62: Do we need to check this again?
Can this be run in the previous step where we already checked?

.github/workflows/pr-test.yml
```diff
@@ -43,16 +43,25 @@ jobs:
...
           files: tests/cloudformation/checks/resource/aws/**/*
-      - name: Install cfn-lint
+      - name: Filter YAML and JSON files
         if: steps.changed-files-specific.outputs.any_changed == 'true'
+        id: filter-files
+        run: |
+          YAML_JSON_FILES=$(echo ${{ steps.changed-files-specific.outputs.all_changed_files }} | tr ' ' '\n' | grep -E '\.ya?ml$|\.json$' | tr '\n' ' ')
+          if [ -n "$YAML_JSON_FILES" ]; then
+            echo "YAML_JSON_FILES=$YAML_JSON_FILES" >> "$GITHUB_ENV"
+            echo "RELEVANT_FILES_CHANGED=true" >> "$GITHUB_ENV"
```
maxamel, line 53: Isn't this redundant? RELEVANT_FILES_CHANGED
Why we're just not using YAML_JSON_FILES and iterating over it ? I think one env var should be enough here

.github/workflows/pr-test.yml
inbalavital: you're right, the 2 steps can be merged

.github/workflows/pr-test.yml
inbalavital: fixing

maxamel (approved):

PR #6653, opened 2024-08-14
fix(terraform): Don't pass existed resources in non_exists resource checks

inbalavital:

*When using the blocklist feature, we should treat passing resources as UNKNOWN or skipped not Passed to prevent very large outputs and payloads*

## Checklist:


checkov/common/checks_infra/solvers/resource_solvers/base_resource_solver.py
```diff
@@ -49,6 +49,8 @@ def run(
 
         for _, data in graph_connector.nodes():
             result = self.get_operation(resource_type=data.get(CustomAttributes.RESOURCE_TYPE))
+            if result is None:
```
bo156, line 52: seems like we have `unknown_vertices` so it should probably be appended there

checkov/common/checks_infra/solvers/resource_solvers/not_exists_resource_solver.py
```diff
@@ -9,5 +9,6 @@
 class NotExistsResourcerSolver(ExistsResourcerSolver):
     operator = Operators.NOT_EXISTS  # noqa: CCE003  # a static attribute
 
-    def get_operation(self, *args: Any, **kwargs: Any) -> bool:
-        return not super().get_operation(*args, **kwargs)
+    def get_operation(self, *args: Any, **kwargs: Any) -> Optional[bool]:
```
bo156, line 12: why change this?

tests/terraform/graph/checks_infra/resource_solvers/not_exists_solver/test_solver.py
```diff
@@ -8,7 +8,7 @@
 
 
 @parameterized_class([{"graph_framework": "NETWORKX"}, {"graph_framework": "IGRAPH"}])
-class NotExistsSolver(TestBaseSolver):
+class TestNotExistsSolver(TestBaseSolver):
```
bo156, line 11: nice

checkov/common/checks_infra/solvers/resource_solvers/base_resource_solver.py
```diff
@@ -49,6 +49,8 @@ def run(
...
+                continue
             if result:
```
lirshindalman, line 54: why change the logic? `if result is None:` and the preview `if result:` is the same.
you can just not add to the passed_vertices if we have the result (it means it is not the resource in the policy).

inbalavital: because I wanted to have an option of `None` result which indicates that the resource "passed", and in this flow it means it should be skipped.
pls pay attention that `base_resource_solver` `run` and `_process_node` functions handle both `exists_resource_solver` and `not_exists_resource_solver`.
In `ExistsResourcerSolver.get_operation ` we can get either True/False, while in `NotExistsResourcerSolver.get_operation` we can get False/None

checkov/common/checks_infra/solvers/resource_solvers/not_exists_resource_solver.py
inbalavital: because I wanted to have an option of `None` result which indicates that the resource "passed", and in this flow it means it should be skipped.
pls pay attention that `base_resource_solver` `run` and `_process_node` functions handle both `exists_resource_solver` and `not_exists_resource_solver`.
In `ExistsResourcerSolver.get_operation ` we can get either True/False, while in `NotExistsResourcerSolver.get_operation` we can get False/None

checkov/common/checks_infra/solvers/resource_solvers/base_resource_solver.py
inbalavital: After clarifying things with @maxamel, I appended the None results to unknown also here.

lirshindalman (approved):

bo156 (approved):

PR #6677, opened 2024-08-22
feat(terraform): add support for TF cloudsplaining evaluated_keys

maxamel:
This PR adds support for evaluated key calculation in cloudsplaining TF, both for data and resource blocks.

## Checklist:


checkov/terraform/checks/resource/base_cloudsplaining_resource_iam_check.py
```diff
@@ -41,3 +43,24 @@ def convert_to_iam_policy(self, conf: Dict[str, Any]) -> PolicyDocument:
             policy = conf['policy'][0]
 
         return PolicyDocument(policy)
+
+    def cloudsplaining_enrich_evaluated_keys(self, policy: PolicyDocument,
```
bo156, line 47: this seems to be a copy of the one from `data` checks, can we make it in a shared place?
note we have this file - `checkov/terraform/checks/utils/base_cloudsplaining_iam_scanner.py`

itai1357 (approved):

lirshindalman (approved):

PR #7058, opened 2025-03-17
feat(gha): ensure action sources use commit hash

bt-macole:
## Description

Github Actions should pin the actions in workflows to SHA to prevent supply chain attacks such as this:
https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066

Fixes # 7057

## New/Edited policies

CKV_GHA_8

### Description
Ensure GitHub Action sources use a commit hash instead of a branch or tag to help prevent supply chain attacks.

### Fix
replace tag with the commit SHA with an optional comment referencing the version

example:
```yaml
jobs:
  checkov-job:
    runs-on: ubuntu-latest
    name: checkov-action
    steps:
      - name: Checkout with hash and comment on the same line
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```

## Checklist:


checkov/github_actions/checks/job/RevisionHash.py
```diff
@@ -0,0 +1,36 @@
...
+
+class RevisionHash(BaseGithubActionsCheck):
+    def __init__(self) -> None:
+        name = "Ensure GitHub Action sources use a commit hash"
+        id = "CKV_GHA_8"
+        super().__init__(
+            name=name,
+            id=id,
+            block_type=BlockType.ARRAY,
+            supported_entities=('jobs', 'jobs.*.steps[]')
```
andycoellis, line 22: You may also want to add `'runs.steps[]'` to your supported_entities to capture the `uses` that can appear under `runs`. https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runsstepsuses

andycoellis:
Great work getting this up! Much needed 😄

checkov/github_actions/checks/job/RevisionHash.py
bt-macole: thats a good find. however, runs.steps[] is not a currently supported [resource type ](https://github.com/bridgecrewio/checkov/blob/main/checkov/github_actions/graph_builder/graph_components/resource_types.py)for github actions. seems like it should definitely be added though.

PR #7094, opened 2025-04-09
fix(terraform_plan): Ignore start and end code lines for provider resources during plan parsing

### LINKED ISSUE #7093: Terraform plan should not be emitted to results output for provider checks
**Describe the issue**

Tested with checkov 3.2.402

When running provider checks against a Terraform plan, portions of the plan (or the complete plan, if it's all on 1 line) can end up in the JSON results output.

**Example**

For the following simple Terraform:

```hcl
provider "aws" {
    region = "us-east-2"
}
```

Generate a plan.json file:

```
terraform init
terraform plan -out plan.out
terraform show -json plan.out > plan.json
```

Then run the built-in rule that checks for access keys in provider blocks:

```
checkov -f ./plan.json -c CKV_AWS_41 --output-file-path=./results.json -o json
```

The results output contains the entire plan in the `code_block` field since the plan JSON is on 1 line:

```
{
    
    "check_type": "terraform_plan",
    "results": {
        "passed_checks": [
            {
                "check_id": "CKV_AWS_41",
                "bc_check_id": "BC_AWS_SECRETS_5",
                "check_name": "Ensure no hard coded AWS access key and secret key exists in provider",
                "check_result": {
                    "result": "PASSED",
                    "evaluated_keys": []
                },
                "code_block": [
                    [
                        1,
                        "{\"format_version\":\"1.2\",\"terraform_version\":\"1.11.3\",\"planned_values\":{\"root_module\":{}},\"configuration\":{\"provider_config\":{\"aws\":{\"name\":\"aws\",\"full_name\":\"registry.terraform.io/hashicorp/aws\",\"expressions\":{\"region\":{\"constant_value\":\"us-east-2\"}}}},\"root_module\":{}},\"timestamp\":\"2025-04-09T20:30:44Z\",\"applyable\":false,\"complete\":true,\"errored\":false}\n"
                    ]
                ],
                "file_path": "/plan.json",
                "file_abs_path": "/home/alan/Temporary/checkov_issue/plan.json",
                "repo_file_path": "/plan.json",
                "file_line_range": [
                    0,
                    1
                ],
                ...
}
```

This can lead to very large JSON results files if the plan is large and there are several provider rules.

I noticed that when checkov is run against a plan for non-provider resources, checkov doesn't attempt to print a code block. 

**Additional**

I worked on a fix, I'll post the PR shortly.

alanszlosek:

## Description

When running provider checks against a Terraform plan, portions of the plan (or the complete plan, if it's all on 1 line) can end up in the JSON results output. This can leak secrets, and can result in very large JSON results files if the plan is large and multiple provider rules are run.

This PR uses 0 for all start/end lines for provider resources that come from a Terraform plan, which will prevent the plan from being printed.

Fixes #7093 

## Checklist:


tsmithv11 (changes requested):
Thanks for the contribution @alanszlosek! However, this is not a good UX as it would be harder to find and debug findings if we don't provide the line numbers. If you would like to set a maximum length of the block and have it configurable via an environment variable, with -1 as no maximum, then this is more likely to work. Then you can set the length to 1 to avoid having too large of a block in your JSON output.

alanszlosek:
> Thanks for the contribution @alanszlosek! However, this is not a good UX as it would be harder to find and debug findings if we don't provide the line numbers. If you would like to set a maximum length of the block and have it configurable via an environment variable, with -1 as no maximum, then this is more likely to work. Then you can set the length to 1 to avoid having too large of a block in your JSON output.

I agree, that would be poor UX.

Instead of investing the time in this fix, we'll work around the issue by ensuring our plan JSON is pretty-printed before using it with checkov.

Going to close this.

PR #7107, opened 2025-04-15
fix(terraform): Handle explicitly-specified tfvars explicitly

### LINKED ISSUE #4321: Checkov ignores terraform tfvars file specified in `--var-file` flag
**Describe the issue**
Checkov ignores path to tfvars file specified in `--var-file` when file is in different directory than terraform root module.
Deep analysis has been made in other PR discussion: https://github.com/bridgecrewio/checkov/issues/3580#issuecomment-1346950236

**Examples**

main.tf:
```
provider "azurerm" {
    features {}
}

variable "storage" {
    type = object({
        name = string
        tls = string
    })
}

resource "azurerm_storage_account" "this" {
  name = var.storage.name
  min_tls_version = var.storage.tls
}
``` 

terraform.tfvars:
```
storage = {
    name = "checkovTest"
    tls = "TLS1_2"
}
```

Directory structure:
```
.
└── checkov_debug
    ├── main.tf
    └── vars
        └── terraform.tfvars
```

checkov execution (irrevelant output omitted):
```
 $ checkov -d checkov_debug/ -c CKV_AZURE_44 --var-file=./checkov_debug/vars/terraform.tfvars

terraform scan results:

Passed checks: 0, Failed checks: 1, Skipped checks: 0

Check: CKV_AZURE_44: "Ensure Storage Account is using the latest version of TLS encryption"
	FAILED for resource: azurerm_storage_account.this
	File: /main.tf:12-15
	Guide: https://docs.bridgecrew.io/docs/bc_azr_storage_2

		12 | resource "azurerm_storage_account" "this" {
		13 |   name = var.storage.name
		14 |   min_tls_version = var.storage.tls
		15 | }
```

Expected checkov output (irrevelant output omitted):
```
terraform scan results:

Passed checks: 1, Failed checks: 0, Skipped checks: 0

Check: CKV_AZURE_44: "Ensure Storage Account is using the latest version of TLS encryption"
	PASSED for resource: azurerm_storage_account.this
	File: /main.tf:12-15
	Guide: https://docs.bridgecrew.io/docs/bc_azr_storage_2
```

**Version:**
 - 2.2.297

thentenaar:

## Description

* Allow for ``--var-file`` arguments outside of the tf dir to be handled correctly
* Use a dict to enforce tfvars precedence, grouping vars by the directory for which they were defined, reducing the number of tfvar nodes added to the graph for a given directory.
* Create variable -> tfvars edges based on the directory for which the var was defined.
* Have _load_files take a list of str instead of os.DirEntry.
* Changed a usage of printf() to logging.debug()

Fixes #4321

## Checklist:

### DIFF
--- a/checkov/terraform/tf_parser.py
+++ b/checkov/terraform/tf_parser.py
@@ -52,7 +52,7 @@ class TFParser:
         self.external_modules_source_map: Dict[Tuple[str, str], str] = {}
         self.module_address_map: Dict[Tuple[str, str], str] = {}
         self.loaded_files_map: dict[str, dict[str, list[dict[str, Any]]] | None] = {}
-        self.external_variables_data: list[tuple[str, Any, str]] = []
+        self.external_vars: dict[str, dict[str, tuple[Any, str]]] = {}
         self.temp_tf_definition: dict[str, Any] = {}
 
     def _init(self, directory: str,
@@ -164,7 +164,6 @@ class TFParser:
             if not data:
                 continue
             self.out_definitions[TFDefinitionKey(file)] = data
-            self.add_external_vars_from_data(data, file)
 
         force_final_module_load = False
         for i in range(0, 10):
@@ -183,24 +182,24 @@ class TFParser:
 
     def _load_files(
             self,
-            files: list[os.DirEntry[str]],
+            files: list[str],
     ) -> list[tuple[str, dict[str, list[dict[str, Any]]] | None]]:
         def _load_file(
-                file: os.DirEntry[str]
+                file: str
         ) -> tuple[tuple[str, dict[str, list[dict[str, Any]]] | None], dict[str, Exception]]:
             parsing_errors: dict[str, Exception] = {}
             result = load_or_die_quietly(file, parsing_errors)
             for path, e in parsing_errors.items():
                 parsing_errors[path] = e
 
-            return (file.path, result), parsing_errors
+            return (file, result), parsing_errors
 
         files_to_data: list[tuple[str, dict[str, list[dict[str, Any]]] | None]] = []
         files_to_parse = []
         for file in files:
-            data = self.loaded_files_map.get(file.path)
+            data = self.loaded_files_map.get(file)
             if data:
-                files_to_data.append((file.path, data))
+                files_to_data.append((file, data))
             else:
                 files_to_parse.append(file)
 
@@ -489,7 +488,7 @@ class TFParser:
             source_dir=source_dir,
             external_modules_source_map=self.external_modules_source_map,
         )
-        self.add_tfvars_with_source_dir(module, source, source_dir)
+        self.add_tfvars(module, source)
         copy_of_tf_definitions = pickle_deepcopy(tf_definitions)
         for tf_def in copy_of_tf_definitions:
             for file_path, blocks in tf_def.items():
@@ -524,21 +523,13 @@ class TFParser:
         return get_tf_definition_object_from_module_dependency(key, nested_key, module_name)
 
     def add_tfvars(self, module: Module, source: str) -> None:
-        if not self.external_variables_data:
+        if not self.external_vars:
             return
-        for (var_name, default, path) in self.external_variables_data:
-            if ".tfvars" in path:
-                block = [{var_name: {"default": default}}]
-                module.add_blocks(BlockType.TF_VARIABLE, block, path, source)
 
-    def add_tfvars_with_source_dir(self, module: Module, source: str, source_dir: str) -> None:
-        if not self.external_variables_data:
-            return
-        for var_name, default, path in self.external_variables_data:
-            if ".tfvars" in path:
-                if Path(source_dir) in Path(path).parents:
-                    block = [{var_name: {"default": default}}]
-                    module.add_blocks(BlockType.TF_VARIABLE, block, path, source)
+        for load_dir, i in self.external_vars.items():
+            for name, (default, path) in i.items():
+                block = [{name: {'default': default, 'load_dir': load_dir}}]
+                module.add_blocks(BlockType.TF_VARIABLE, block, path, source)
 
     def get_dirname(self, path: TFDefinitionKey) -> str:
         file_path = path.file_path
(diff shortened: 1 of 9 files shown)


checkov/terraform/tf_parser.py
```diff
@@ -526,19 +526,18 @@ def get_new_nested_module_key(
     def add_tfvars(self, module: Module, source: str) -> None:
         if not self.external_variables_data:
             return
-        for (var_name, default, path) in self.external_variables_data:
-            if ".tfvars" in path:
-                block = [{var_name: {"default": default}}]
+        for (var_name, default, path, explicit) in self.external_variables_data:
+            if explicit or ".tfvars" in path:
```
omriyoffe-panw, line 530: Would this not allow for files other then .tfvars to be added here?

checkov/terraform/tf_parser.py
```diff
@@ -526,19 +526,18 @@ def get_new_nested_module_key(
...
     def add_tfvars_with_source_dir(self, module: Module, source: str, source_dir: str) -> None:
         if not self.external_variables_data:
             return
-        for var_name, default, path in self.external_variables_data:
-            if ".tfvars" in path:
-                if Path(source_dir) in Path(path).parents:
-                    block = [{var_name: {"default": default}}]
-                    module.add_blocks(BlockType.TF_VARIABLE, block, path, source)
+        for var_name, default, path, explicit in self.external_variables_data:
+            if explicit or (".tfvars" in path and Path(source_dir) in Path(path).parents):
```
omriyoffe-panw, line 538: Here as well.

checkov/terraform/tf_parser.py
```diff
@@ -605,38 +604,43 @@ def handle_variables(
...
         for var_file, data in sorted(auto_var_files_to_data, key=lambda x: x[0]):
             if data:
-                self.external_variables_data.extend([(k, v, var_file) for k, v in data.items()])
+                self.external_variables_data.extend([(k, v, var_file, False) for k, v in data.items()])
 
-        explicit_var_files_to_data = self._load_files(explicit_var_files)
-        # it's possible that os.scandir returned the var files in a different order than they were specified
         if vars_files:
-            for var_file, data in sorted(explicit_var_files_to_data, key=lambda x: vars_files.index(x[0])):
+            for f in [Path(x).resolve() for x in vars_files]:
```
omriyoffe-panw, line 629: Maybe add a comment here explaining this search over the directory since it's not very clear from the code.

omriyoffe-panw (commented):
Hey @thentenaar,
First of all thank you for your contribution!
I left a few minor comments, however along with them I am not sure I understand your use of the "explicit" variable. If I understand correctly the goal is to allow for .tfvars file in sub-directories of the project using the -var-file flag. I saw you implemented a search on the directory for .tfvars files but I don't see why you would need the "explicit" variable and I can see some use cases where it might allow for files other then .tfvars to be passed in the flag causing errors.
Would love if you could add a short explanation for this implementation.
Thanks!

checkov/terraform/tf_parser.py
thentenaar: Both .tfvars and environment vars are added as 'explicit,' the other half of that expression is for tfvars files scanned in the terraform dir; but I added a check on the extension.

checkov/terraform/tf_parser.py
thentenaar: I added a comment explaining that this gathers the absolute paths of the entries in vars_files.

checkov/terraform/tf_parser.py
```diff
@@ -605,38 +604,48 @@ def handle_variables(
...
-            elif vars_files and file.path in vars_files:
-                explicit_var_files.append(file)
             elif file.name.endswith(".tf") or file.name.endswith('.hcl'):  # TODO: add support for .tf.json
                 tf_files_to_load.append(file)
 
         for key, value in self.env_vars.items():
             if not key.startswith("TF_VAR_"):
                 continue
-            self.external_variables_data.append((key[7:], value, f"env:{key}"))
+            self.external_variables_data.append((key[7:], value, f"env:{key}", True))
```
gruebel, line 613: the variable resolution hierarchy is incorrect and should reflect the official docs https://developer.hashicorp.com/terraform/language/values/variables#variable-definition-precedence 
probably the most accurate approach would be to mark each of them with an enum value (ex. ExternalVariableType.ENV, .TFVARS, etc.) and when a new variable is read check and override an existing one if needed, then you have the up-to-date hierarchy when processing them at the end.

checkov/terraform/tf_parser.py
```diff
@@ -605,38 +604,48 @@ def handle_variables(
...
+            for f in [Path(x).resolve() for x in vars_files]:
+                if str(f) in [x[2] for x in self.external_variables_data]:
+                    continue
+                if not str(f).endswith('.tfvars'):
+                    continue
+                for ent in os.scandir(f.parent):
+                    if not ent.is_file() or Path(ent.path).resolve() != f:
+                        continue
+                    explicit_var_files.append(ent)
+                    break
```
gruebel, line 640: this doesn't make much sense, if you pass a var file via the flag, then you expect to load a single file (or multiple when used more than once) and not read the whole folder.

checkov/terraform/tf_parser.py
```diff
@@ -605,38 +604,48 @@ def handle_variables(
...
-                self.external_variables_data.extend([(k, v, var_file) for k, v in data.items()])
+                self.external_variables_data.extend([(k, v, var_file, False) for k, v in data.items()])
 
-        explicit_var_files_to_data = self._load_files(explicit_var_files)
-        # it's possible that os.scandir returned the var files in a different order than they were specified
+        # Load vars files specified via the '--var-file' option, first
+        # collecting their absolute paths.
         if vars_files:
-            for var_file, data in sorted(explicit_var_files_to_data, key=lambda x: vars_files.index(x[0])):
+            for f in [Path(x).resolve() for x in vars_files]:
```
gruebel, line 631: it is wasteful to create a list and then iterate over it, just iterate directly over it. Also using `Path` and then multiple times creating a string out of it is a bad idea, you can use `os.path.abspath()` instead.

gruebel (commented):
overall the suggested implementation doesn't reflect the actual Terraform behavior and with the added `explicit` marker makes it harder to maintain than it actually should be, because it is not a simple override mechanism but rather a hierarchical resolution order.

checkov/terraform/tf_parser.py
thentenaar: 100%. This is because _load_files expects a list of os.DirEntry (probably because it presumes directory scanning,) though it would be better just to have it take a list of paths, which I've now done.

checkov/terraform/tf_parser.py
thentenaar: In that case, I'd argue that it'd be more proper to use a dictionary and evaluate them in the correct order. If the order is truly incorrect (in which case I'd have to see why the tests that check it pass,) I'd be willing to take a stab at it in a follow-up PR.

lirshindalman (approved):

omriyoffe-panw (approved):

PR #7111, opened 2025-04-17
fix(terraform): Better utilization of managed modules (if enabled)

thentenaar:

## Description

If we have managed modules, the output from terraform should be authoratative. This also allows checkov to scan terraform which only pulls in a module, which pulls in submodules, from the root module.

## Checklist:

### DIFF
--- a/checkov/terraform/module_loading/module_finder.py
+++ b/checkov/terraform/module_loading/module_finder.py
@@ -7,9 +7,9 @@ import re
 from pathlib import Path
 from typing import List, Callable, TYPE_CHECKING
 
+from checkov.common.util.env_vars_config import env_vars_config
 from checkov.common.parallelizer.parallel_runner import parallel_runner
 from checkov.common.util.file_utils import read_file_with_any_encoding
-from checkov.common.util.type_forcers import convert_str_to_bool
 from checkov.terraform.module_loading.registry import module_loader_registry
 
 if TYPE_CHECKING:
@@ -31,6 +31,29 @@ class ModuleDownload:
 
     def __str__(self) -> str:
         return f"{self.source_dir} -> {self.module_link} ({self.version})"
+
+
+def find_tf_managed_modules(path: str) -> List[ModuleDownload]:
+    """
+    Leverage modules.json to better inform discovery. If we have this,
+    there should be no need to walk and gather modules.
+    """
+    modules_found: list[ModuleDownload] = []
+
+    tf_modules_file = Path(path) / '.terraform' / 'modules' / 'modules.json'
+    if not tf_modules_file.exists():
+        return modules_found
+
+    for mod in json.loads(tf_modules_file.read_bytes())['Modules']:
+        if mod['Key']:
+            md = ModuleDownload(path)
+            md.module_name = mod['Key']
+            md.module_link = mod['Dir']
+            md.version = mod['Version'] if 'Version' in mod else 'latest'
+            md.address = f"{mod['Source']}:{md.version}"
+            md.tf_managed = True
+            modules_found.append(md)
+    return modules_found
 
 
 def find_modules(path: str) -> List[ModuleDownload]:
@@ -107,14 +130,13 @@ def load_tf_modules(
     stop_on_failure: bool = False
 ) -> None:
     module_loader_registry.root_dir = path
+    if not modules_to_load and env_vars_config.CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES:
+        modules_to_load = find_tf_managed_modules(path)
     if not modules_to_load:
         modules_to_load = find_modules(path)
 
-    # load terraform managed modules first, before pulling out distinct modules, as address attribute changes
-    replaced_modules = replace_terraform_managed_modules(path=path, found_modules=modules_to_load)
-
     # To avoid duplicate work, we need to get the distinct module sources
-    distinct_modules = list({m.address: m for m in replaced_modules}.values())
+    distinct_modules = list({m.address: m for m in modules_to_load}.values())
 
     downloadable_modules = [
         (module_loader_registry, m)
@@ -124,7 +146,7 @@ def load_tf_modules(
     if run_parallel:
         list(parallel_runner.run_function(_download_module, downloadable_modules))
     else:
-        logging.info(f"Starting download of modules of length {len(replaced_modules)}")
+        logging.info(f"Starting download of modules of length {len(downloadable_modules)}")
         for m in downloadable_modules:
             success = _download_module(*m)
             if not success and stop_on_failure:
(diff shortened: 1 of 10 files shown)


checkov/terraform/module_loading/module_finder.py
```diff
@@ -36,6 +36,22 @@ def __str__(self) -> str:
 def find_modules(path: str) -> List[ModuleDownload]:
     modules_found: list[ModuleDownload] = []
 
+    # Leverage modules.json to better inform discovery. If we have this,
+    # there should be no need to walk and gather modules
+    if convert_str_to_bool(os.getenv('CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES', True)):
```
omriyoffe-panw, line 41: Please use `checkov/common/util/env_vars_config.py` for environment variables.

checkov/terraform/module_loading/module_finder.py
```diff
@@ -153,48 +166,3 @@ def _download_module(ml_registry: ModuleLoaderRegistry, module_download: ModuleD
         return False
 
     return True
-
-
-def replace_terraform_managed_modules(path: str, found_modules: list[ModuleDownload]) -> list[ModuleDownload]:
```
omriyoffe-panw, line 158: Won't the previous flow fail if the env var is set to false now that this function is deleted?

checkov/terraform/module_loading/registry.py
```diff
@@ -42,15 +43,25 @@ def load(
...
         if module_address is None:
             module_address = f'{source}:{source_version}'
         if module_address in self.module_content_cache:
             logging.debug(f'Used the cache for module {module_address}')
             return self.module_content_cache[module_address]
-        else:
-            logging.debug(f'Cache miss for {module_address}')
 
+        # If we have tf managed modules, we likely have whatever :latest is in the cache
+        if convert_str_to_bool(os.getenv('CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES', True)):
```
omriyoffe-panw, line 53: here as well use `checkov/common/util/env_vars_config.py`

omriyoffe-panw (commented):
Hey @thentenaar,
Thanks for you contribution again. Please make sure that you use the env var correctly and wrap the entire new flow with it to ensure that if not explicitly set to true, the flow will remain as it is today since this is based on an experimental feature.

checkov/terraform/module_loading/module_finder.py
thentenaar: I've added that one to env_vars_config. The flow otherwise remains unchanged.

checkov/terraform/module_loading/module_finder.py
thentenaar: This function is basically a no-op if that env var doesn't evaluate to True.

checkov/terraform/module_loading/registry.py
thentenaar: Thanks for the comments, will update this branch (and the other) shortly

checkov/terraform/module_loading/module_finder.py
omriyoffe-panw: You can see an example of how you can use it here:
https://github.com/bridgecrewio/checkov/blob/e6041558ece3e5dee556b6bf935d960a2e91168d/checkov/terraform/graph_builder/foreach/foreach_entity_handler.py#L74

checkov/terraform/module_loading/module_finder.py
```diff
@@ -36,6 +37,22 @@ def __str__(self) -> str:
...
+            for mod in json.loads(tf_modules_file.read_bytes())['Modules']:
+                if len(mod['Key']):
+                    md = ModuleDownload(path)
+                    md.module_name = mod['Key']
+                    md.module_link = mod['Dir']
+                    md.version = mod['Version'] if 'Version' in mod else 'latest'
+                    md.address = f"{mod['Source']}:{md.version}"
+                    md.tf_managed = True
+                    modules_found.append(md)
+            return modules_found
```
gruebel, line 54: this changes the original behavior of being supplemental to authoritative as mentioned. it makes the implementation simpler but loses the original flexibility.

checkov/terraform/module_loading/module_finder.py
```diff
@@ -36,6 +37,22 @@ def __str__(self) -> str:
 def find_modules(path: str) -> List[ModuleDownload]:
     modules_found: list[ModuleDownload] = []
 
+    # Leverage modules.json to better inform discovery. If we have this,
+    # there should be no need to walk and gather modules
+    if env_vars_config.CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES:
+        tf_modules_file = Path(path) / '.terraform' / 'modules' / 'modules.json'
+        if tf_modules_file.exists():
+            for mod in json.loads(tf_modules_file.read_bytes())['Modules']:
+                if len(mod['Key']):
```
gruebel, line 46:
```suggestion
                if mod['Key']:
```
it is faster to check for truthiness, which means it is not empty.

checkov/terraform/module_loading/registry.py
```diff
@@ -42,15 +43,25 @@ def load(
...
+        if env_vars_config.CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES:
+            if source_version == 'latest':
+                vers = sorted([k.rsplit(':', 1)[1] for k in self.module_content_cache.keys() if k.startswith(f'{source}:')])
+                if vers:
+                    logging.debug(f'Used the cache for module {module_address}')
+                    return self.module_content_cache[f'{source}:{vers[-1]}']
+                vers = sorted([k.rsplit(':', 1)[1] for k in self.module_content_cache.keys() if k.startswith(f'registry.terraform.io/{source}:')])
+                if vers:
+                    logging.debug(f'Used the cache for module (from tf registry) {module_address}')
+                    return self.module_content_cache[f'registry.terraform.io/{source}:{vers[-1]}']
```
gruebel, line 62: sorting doesn't give any benefit and just costs time. also iterating two times over the same collection seems rather wasteful and can be done by a simple loop and checking for either the normal case or the registry prefix. accessing the module cache with the concatenated key adds an additional lookup (plus the string concatenation), you can can iterate over the `.items()` and have the value right next to the key.

thentenaar: Even better, just keep a separate lookaside.

checkov/terraform/module_loading/registry.py
```diff
@@ -42,15 +44,25 @@ def load(
...
-
         if module_address is None:
             module_address = f'{source}:{source_version}'
         if module_address in self.module_content_cache:
             logging.debug(f'Used the cache for module {module_address}')
             return self.module_content_cache[module_address]
-        else:
-            logging.debug(f'Cache miss for {module_address}')
 
+        # If we have tf managed modules, we likely have whatever :latest is in the cache
```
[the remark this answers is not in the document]
thentenaar: Right, "latest" from terraform's perspective when it last updated the modules, but not necessarily the latest version in the registry at the time when checkov runs.

checkov/terraform/module_loading/registry.py
```diff
@@ -42,15 +44,25 @@ def load(
...
+        if env_vars_config.CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES:
+            if source_version == 'latest':
+                reg = f'registry.terraform.io/{source}'
```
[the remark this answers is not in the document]
gruebel: what I meant is, just move it a few lines lower, where it is actually used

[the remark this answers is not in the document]
thentenaar: Done

checkov/terraform/module_loading/module_finder.py
```diff
@@ -36,6 +36,22 @@ def __str__(self) -> str:
 def find_modules(path: str) -> List[ModuleDownload]:
     modules_found: list[ModuleDownload] = []
 
+    # Leverage modules.json to better inform discovery. If we have this,
+    # there should be no need to walk and gather modules
+    if env_vars_config.CHECKOV_EXPERIMENTAL_TERRAFORM_MANAGED_MODULES:
```
lirshindalman: Please consider extracting this logic into a separate function for better readability

thentenaar: Done

omriyoffe-panw (approved):

PR #7159, opened 2025-05-15
fix(terraform): dont move clone to internal dir

lirshindalman:

since we're already using inner_module when building dest_dir, there's no need to add it again after cloning.
like in this example - git::github.com/terraform-aws-modules/security-group/aws//modules/http-80
 as this 'git::**https://**github.com/terraform-aws-modules/terraform-aws-security-group'


omriyoffe-panw (approved):

omriyoffe-panw:
@tsmithv11 Added unit tests for the function

omriyoffe-panw:
> @lirshindalman can you add a test?

@tsmithv11 Added unit tests

checkov/common/goget/github/get_git.py
```diff
@@ -69,9 +69,6 @@ def do_get(self) -> str:
         clone_dir = self.temp_dir + "/clone/" if self.create_clone_and_res_dirs else self.temp_dir
         self._clone(git_url, clone_dir)
 
-        if internal_dir:
```
bo156, line 72: Maybe we should also add a test which specifically shows the generated url is without `git::https://github` but only `git::github`?

Are there any other options when we do want to add them? Or was it entirely a bug?

omriyoffe-panw: According to what I understood from @lirshindalman, in the case of git external modules, we would do a duplicate parsing of the url in the base class and in the git loader class making the resulted URL invalid. This is the removal of the parsing in the git loader derived class.

bo156 (approved):

PR #7219, opened 2025-06-22
feat(kubernetes): include hidden folders in scan

mayblo:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:


checkov/kubernetes/graph_manager.py
```diff
@@ -25,9 +25,10 @@ def build_graph_from_source_directory(
         parsing_errors: Optional[dict[str, Exception]] = None,
         download_external_modules: Optional[bool] = False,
         excluded_paths: Optional[list[str]] = None,
+        include_hidden: Optional[list[str]] = None,
```
bo156: this function has the same signature across all graph managers, adding this parameter here breaks it.
assuming you don't want to add it across all frameworks, I can suggest controlling this behaviour with an env var rather than a parameter:
1. Add a matching env var under the file `checkov/common/util/env_vars_config.py`, with default False.
2. In the specific function you wanted to use it, check for the value (using import) and define behavior accordingly

checkov/kubernetes/kubernetes_utils.py
```diff
@@ -29,18 +29,19 @@
...
-        filter_ignored_paths(root, f_names, excluded_paths)
+        filter_ignored_paths(root, d_names, excluded_paths, include_hidden)
+        filter_ignored_paths(root, f_names, excluded_paths, include_hidden)
 
         for file in f_names:
             file_ending = os.path.splitext(file)[1]
             if file_ending in K8_POSSIBLE_ENDINGS:
                 full_path = os.path.join(root, file)
-                if "/." not in full_path and file not in EXCLUDED_FILE_NAMES:
+                if ("/." not in full_path or any(inc in full_path for inc in include_hidden)) and file not in EXCLUDED_FILE_NAMES:
```
bo156: please add a test for this behavior

checkov/kubernetes/kubernetes_utils.py
```diff
@@ -109,11 +110,12 @@ def create_definitions(
     runner_filter = runner_filter or RunnerFilter()
     definitions: dict[str, list[dict[str, Any]]] = {}
     definitions_raw: dict[str, list[tuple[int, str]]] = {}
+    runner_filter.include_hidden = runner_filter.include_hidden or []
```
bo156: following my first comment, this is not needed

bo156 (changes requested):

checkov/kubernetes/graph_manager.py
mayblo: Regarding the breaking - I've tested a checkov run on both k8s and terraform and it worked. Still do you recommend the change?

tests/kubernetes/graph/test_kubernetes_utils.py
```diff
@@ -23,3 +27,24 @@ def test_build_resource_id_from_empty_labels(self) -> None:
...
+    def test_get_folder_definitions_ignore_hidden(self) -> None:
+        test_root_dir = Path(TEST_DIRNAME) / RELATIVE_PATH
+        definitions, _ = get_folder_definitions(root_folder=test_root_dir, excluded_paths=[])
+        file_list = list(definitions.keys())
+        
+        self.assertNotIn(FILE_UNDER_HIDDEN_FOLDER, file_list)
+        self.assertIn(FILE_NOT_HIDDEN, file_list)
+    
+    def test_get_folder_definitions_do_not_ignore_hidden(self) -> None:
+        os.environ["CKV_IGNORE_HIDDEN_DIRECTORIES"] = "False"
```
bo156: this will set it for the rest of the process, you can use `monkeypatch` instead to set it just for the test

tests/kubernetes/graph/test_kubernetes_utils.py
```diff
@@ -23,3 +27,24 @@ def test_build_resource_id_from_empty_labels(self) -> None:
...
+        importlib.reload(k8s_utils)
```
bo156: that's not common practice, please use mocks for the test using `monkeypatch`

tests/kubernetes/graph/resources/definitions/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
bo156: is this file here on purpose? seems unrelated to k8s tests

checkov/kubernetes/kubernetes_utils.py
```diff
@@ -26,6 +26,7 @@
 PARENT_RESOURCE_KEY_NAME = "_parent_resource"
 PARENT_RESOURCE_ID_KEY_NAME = "_parent_resource_id"
 FILTERED_RESOURCES_FOR_EDGE_BUILDERS = ["NetworkPolicy"]
+IGNORE_HIDDEN_DIRECTORY_ENV = strtobool(os.getenv("CKV_IGNORE_HIDDEN_DIRECTORIES", "True"))
```
bo156: this line should be under the `env_var_config.py` rather than here (under the `common` dir with the rest of the env vars)

bo156 (changes requested):

OfekShimko (approved):

PR #7230, opened 2025-07-08
fix(general): Proxy ignored when pulling image data

jumiles-pa:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

checkov/common/images/image_referencer.py:_fetch_image_results_async() uses the Aiohttp library which documents that the environment variables will not be used unless trust_env is set to True. This breaks scans when and proxy is in use through environment variables which is documented in Checkov to be supported.

Fixes # (issue)
The fix is to add trust_env=True to the two spots that instantiate ClientSession objects. To test this please search for the fail logs in the function referred to in description when scans should be blocked by a proxy for the image result and license calls.

Please note that this will read all environment variables for these specific calls so may have unrelated impact. Other portions of the tool properly use proxy settings and the error that lead to this discovery was finding a different number of findings than others without a proxy. 

https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support

## Checklist:

### DIFF
--- a/checkov/common/images/image_referencer.py
+++ b/checkov/common/images/image_referencer.py
@@ -185,11 +185,22 @@ class ImageReferencerMixin(Generic[_Definitions]):
         This is an async implementation of `_fetch_image_results`. The only change is we're getting a session
         as an input, and the asyncio behavior is managed in the calling method.
         """
-        async with aiohttp.ClientSession() as session:
+        async with aiohttp.ClientSession(trust_env=True) as session:
             results: list[dict[str, Any]] = await asyncio.gather(*[
                 image_scanner.get_scan_results_from_cache_async(session, f"image:{i}")
                 for i in image_names_to_query
             ])
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
+        try:
+            logging.debug(session)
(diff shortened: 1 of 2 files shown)


jumiles-pa:
My intention is to remove extra logging once I'm able to test this more thoroughly.

checkov/common/images/image_referencer.py
```diff
@@ -185,11 +185,22 @@ async def _fetch_image_results_async(image_names_to_query: list[str]) -> list[di
...
         as an input, and the asyncio behavior is managed in the calling method.
         """
-        async with aiohttp.ClientSession() as session:
+        async with aiohttp.ClientSession(trust_env=True) as session:
             results: list[dict[str, Any]] = await asyncio.gather(*[
                 image_scanner.get_scan_results_from_cache_async(session, f"image:{i}")
                 for i in image_names_to_query
             ])
+        logger = logging.getLogger()
+        logger.setLevel(logging.DEBUG)
```
AdamDev, line 194: Hi @jumiles-pa
Why are we hardcoding the log level to DEBUG?
The log level should be injected from an external configuration.

PR #7257, opened 2025-07-22
chore(terraform): TerraformLocalGraph - add caching for get_module_vertices_mapping

emaohi:
avoid nested loop

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/terraform/graph_builder/local_graph.py
+++ b/checkov/terraform/graph_builder/local_graph.py
@@ -287,14 +287,14 @@ class TerraformLocalGraph(LocalGraph[TerraformBlock]):
 
         # Match vertices using the lookup
         for vertex in self.vertices:
-            smo = vertex.source_module_object
-            if not smo or not smo.name:
+            source_module_object = vertex.source_module_object
+            if not source_module_object or not source_module_object.name:
                 continue
             composed_key = (
-                smo.name,
-                smo.path,
-                smo.nested_tf_module,
-                smo.foreach_idx,
+                source_module_object.name,
+                source_module_object.path,
+                source_module_object.nested_tf_module,
+                source_module_object.foreach_idx,
             )
             module_vertice_idx = module_lookup.get(composed_key)
             if module_vertice_idx is not None:


checkov/terraform/graph_builder/local_graph.py
```diff
@@ -274,20 +274,31 @@ def get_module_vertices_mapping(self) -> None:
         For each vertex, if it's originated in a module import, add to the vertex the index of the
         matching module vertex as 'source_module'
         """
+        module_lookup = {}
```
bo156, line 277: can you explain what is the meaning of the change you made?

checkov/terraform/graph_builder/local_graph.py
```diff
@@ -274,20 +274,31 @@ def get_module_vertices_mapping(self) -> None:
...
+                module_vertex.path,
+                module_vertex.source_module_object,
+                module_vertex.for_each_index,
+            )
+            module_lookup[composed_key] = module_idx
+
+        # Match vertices using the lookup
         for vertex in self.vertices:
-            if not vertex.source_module_object:
+            smo = vertex.source_module_object
```
bo156: I don't think `smo` is readable enough, you can keep it as `source_module_object`

checkov/terraform/graph_builder/local_graph.py
emaohi: yea this is the old code:
```
 for vertex in self.vertices:
            if not vertex.source_module_object:
                continue
            for idx in self.vertices_by_block_type[BlockType.MODULE]:
                if vertex.source_module_object.name != self.vertices[idx].name:
                    continue
                if vertex.source_module_object.path != self.vertices[idx].path:
                    continue
                if vertex.source_module_object.nested_tf_module != self.vertices[idx].source_module_object:
                    continue
                if vertex.source_module_object.foreach_idx != self.vertices[idx].for_each_index:
                    continue
                vertex.source_module.add(idx)
                break
        return
```
for each vertice in the graph we're iterating thorugh all modules vertices. in the modified code, i've looped once on all module vertices, saving their composed data as key and its index as value. then i loop just once for all vertices, and lookup at this lookup map for the corresponding module (by looking at the composing key). this makes it from N^2 worst-case to N

yuvalmich (approved):

PR #7260, opened 2025-07-24
fix(dockerfile): Use proxy env vars in aiohttp client requests

inbalavital:
Moved all aiohttp requests to an existed wrapper.
Extended the wrapper to support different methods, get url as a URL type as well as str, and changed payload to optional (for GET requests).
In the wrapper, added usage of proxy env vars in order to avoid a bug a customer faced where its configured proxy env vars weren't in use in certain requests (the ones that were made using aiohttp lib).

### DIFF
--- a/checkov/common/util/http_utils.py
+++ b/checkov/common/util/http_utils.py
@@ -12,8 +12,6 @@ import aiohttp
 import asyncio
 from typing import Any, TYPE_CHECKING, cast, Optional, overload
 
-from aiohttp import ClientResponse
-
 from checkov.common.util import env_vars_config
 from urllib3.response import HTTPResponse
 from urllib3.util import parse_url
@@ -29,6 +27,7 @@ if TYPE_CHECKING:
     from checkov.common.bridgecrew.bc_source import SourceType
     from requests import Response
     from aiohttp.typedefs import StrOrURL
+    from aiohttp import ClientResponse
 
 # https://requests.readthedocs.io/en/latest/user/advanced/#timeouts
 REQUEST_CONNECT_TIMEOUT = force_float(os.getenv("CHECKOV_REQUEST_CONNECT_TIMEOUT")) or 3.1


inbalavital:
> Did you decide not to wrap it in an env var?

@omriyoffe-panw yes, because I added a standard proxy support, and not used the entire env vars configurations.
We have this kind of support in other libraries as well without env var wrapping.

checkov/common/util/http_utils.py
```diff
@@ -204,36 +206,54 @@ def request_wrapper(
...
         headers: dict[str, Any],
-        payload: dict[str, Any]
-) -> int:
+        payload: dict[str, Any] | None = None,
+) -> tuple[int, Any]:
     request_max_tries = int(os.getenv('REQUEST_MAX_TRIES', 3))
     sleep_between_request_tries = float(os.getenv('SLEEP_BETWEEN_REQUEST_TRIES', 1))
 
+    # 1. Read proxy URL (may include user:pass for authentication)
+    proxy_url = os.environ.get("HTTPS_PROXY", os.environ.get("HTTP_PROXY"))
```
bo156: those values can be 2 different fields, you don't have to use the same proxy for both http and https

checkov/common/util/http_utils.py
```diff
@@ -204,36 +206,54 @@ def request_wrapper(
...
+    proxy_auth = None
+    if proxy_url:
+        parsed_proxy_url = urlparse(proxy_url)
+        if parsed_proxy_url.username and parsed_proxy_url.password:
+            proxy_auth = aiohttp.BasicAuth(login=parsed_proxy_url.username, password=parsed_proxy_url.password)
+    # 2. Read path to custom certificate bundle
+    ca_bundle_path = os.getenv('BC_CA_BUNDLE')
```
bo156: all env vars shoul be under `checkov/common/util/env_vars_config.py`, please use the values from there (also for http/s proxy above)

checkov/common/util/http_utils.py
```diff
@@ -204,36 +206,54 @@ def request_wrapper(
...
+                    return 0, response_json
                 elif i != request_max_tries - 1:
                     await asyncio.sleep(sleep_between_request_tries * (i + 1))
                     continue
                 else:
                     logging.error(f"[http_utils](aiohttp_client_session_wrapper) - Failed to send report to "
                                   f"url \'{url}\'")
                     logging.error(f"Status code: {response.status}, Reason: {response.reason}, Content: {content}")
-                    return 1
+                    return 1, None
```
bo156: why do you return both return code and json? isn't it better to just return the response itself, and the relvent caller will decide what to do with it?
If you do want to do it like that, let's create an object for it to be more readable (even a `namedTuple` can be good)

checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
```diff
@@ -147,7 +147,8 @@ async def get_scan_results_from_cache_async(image_id: str) -> Dict[str, Any]:
             headers = bc_integration.get_default_headers("GET")
             logging.debug(f"Invoking API {url}")
 
-            _, response_json = await aiohttp_client_session_wrapper("GET", URL(url, encoded=True), headers=headers)
+            response = await aiohttp_client_session_wrapper("GET", URL(url, encoded=True), headers=headers)
+            response_json = await response.json()
```
bo156, line 151: do you need another await if you already did it in the previous line?

checkov/common/sca/output.py
```diff
@@ -698,9 +698,10 @@ async def get_license_statuses_async(packages: list[dict[str, Any]], image_name:
     if not requests_input:
         return {'image_name': image_name, 'licenses': []}
     try:
-        _, response_json = await aiohttp_client_session_wrapper("POST", url,
-                                                                headers=bc_integration.get_default_headers("POST"),
-                                                                payload={"packages": requests_input})
+        response = await aiohttp_client_session_wrapper("POST", url,
+                                                        headers=bc_integration.get_default_headers("POST"),
+                                                        payload={"packages": requests_input})
+        response_json = await response.json()
```
bo156, line 704: same

bo156 (approved):

checkov/common/bridgecrew/vulnerability_scanning/image_scanner.py
inbalavital: First await: Waits for the HTTP request to complete and returns the ClientResponse object
Second await: The response.json() method is itself an async method that reads the response body from the network stream and parses it as JSON

yuvalmich (approved):

PR #7264, opened 2025-07-28
Fix logic

inbalavital:


checkov/common/checks_infra/solvers/complex_solvers/and_solver.py
```diff
@@ -16,12 +16,26 @@ def __init__(self, solvers: List[BaseSolver], resource_types: List[str]) -> None
...
             operation = solver.get_operation(vertex)
             if operation is None:
                 has_unrendered_attribute = True
+            elif operation == "not found":
+                continue
             elif not operation:
                 return False
-        return None if has_unrendered_attribute else True
+        if not found:
+            return "not found"
```
bo156: this should be extracted to a const

checkov/common/checks_infra/solvers/complex_solvers/and_solver.py
```diff
@@ -16,12 +16,26 @@ def __init__(self, solvers: List[BaseSolver], resource_types: List[str]) -> None
     def _get_operation(self, *args: Any, **kwargs: Any) -> Any:
         return reduce(and_, args)
 
-    def get_operation(self, vertex: Dict[str, Any]) -> Optional[bool]:
+    def get_operation(self, vertex: Dict[str, Any]) -> Union[Optional[bool], str]:
```
bo156, line 19: I don't like that the return type changed like that, specifically if we want to return `not found` we should use `Literal["not found"]` to be more specific.
In addition - is it really needed? Can't we just return the value which will not change the check?
For example in `and-solver` we can return `True` and in `or-solver` return `False`, to make sure this is simpler

inbalavital: this was my last resort, it's essential because otherwise it affects checks results, which is not what we want.

PR #7272, opened 2025-08-11
fix(terraform): Handled git external module loading with sub-directory but without protocol

bo156:

## Description
Handled git external module loading with sub-directory but without protocol, specifically when using `git::git` rather than `git::https://...` (or any other protocol instead of https).
This makes sure we correctly parse the sub-directory inside the source url.

## Checklist:

### DIFF
--- a/checkov/common/util/http_utils.py
+++ b/checkov/common/util/http_utils.py
@@ -12,7 +12,6 @@ import aiohttp
 import asyncio
 from typing import Any, TYPE_CHECKING, cast, Optional, overload
 
-from checkov.common.util import env_vars_config
 from urllib3.response import HTTPResponse
 from urllib3.util import parse_url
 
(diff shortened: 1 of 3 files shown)


checkov/common/util/http_utils.py
```diff
@@ -213,6 +212,8 @@ async def aiohttp_client_session_wrapper(
         headers: dict[str, Any],
         payload: dict[str, Any] | None = None,
 ) -> ClientResponse:
+    from checkov.common.util import env_vars_config
```
bo156, line 215: this import moved due to an import loop of `env_var_config` and `http_utils`

checkov/terraform/module_loading/loaders/git_loader.py
```diff
@@ -105,8 +105,12 @@ def _parse_module_source(self, module_params: ModuleParams) -> ModuleSource:
             version = "HEAD"
 
         if len(module_source_components) < 3:
-            root_module = module_source_components[-1]
-            inner_module = ""
+            if len(module_source_components) == 2 and "git::git" in module_source_components[0]:
```
yuvalmich, line 108: I think it would be helpful if we had a comment here (and in other places in this method) which would explain which use case it solves.

i.e. in this case if I got it right it would be

`// git::git@github.com:test-inner-module/out-module//inner-module?ref=main
`

yuvalmich (approved):

checkov/terraform/module_loading/loaders/git_loader.py
bo156: Added :)

PR #7319, opened 2025-10-05
fix(terraform_plan): Don't add values to empty list values in after_unknown

bo156:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes a bug in which empty lists in after_unknown section were inserted a new value, now the value stays an empty list.

## Checklist:

### DIFF
--- a/checkov/terraform/plan_parser.py
+++ b/checkov/terraform/plan_parser.py
@@ -263,9 +263,21 @@ def _handle_complex_after_unknown(k: str, resource_conf: dict[str, Any], v: Any)
             continue
         if inner_key not in resource_conf[k]:
             if isinstance(resource_conf[k][0], dict):
-                resource_conf[k][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
+                if _validate_after_unknown_list_not_empty(inner_key, k, resource_conf):
+                    resource_conf[k][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
             elif isinstance(resource_conf[k][0], list) and isinstance(resource_conf[k][0][0], dict):
-                resource_conf[k][0][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
+                if _validate_after_unknown_list_not_empty(inner_key, k, resource_conf):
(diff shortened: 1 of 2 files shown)


checkov/terraform/plan_parser.py
```diff
@@ -263,9 +263,21 @@ def _handle_complex_after_unknown(k: str, resource_conf: dict[str, Any], v: Any)
             continue
         if inner_key not in resource_conf[k]:
             if isinstance(resource_conf[k][0], dict):
-                resource_conf[k][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
+                if _validate_after_unknown_list_not_empty(inner_key, k, resource_conf):
+                    resource_conf[k][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
             elif isinstance(resource_conf[k][0], list) and isinstance(resource_conf[k][0][0], dict):
-                resource_conf[k][0][0][inner_key] = _clean_simple_type_list([TRUE_AFTER_UNKNOWN])
+                if _validate_after_unknown_list_not_empty(inner_key, k, resource_conf):
```
maxamel: Should we really invoke the same condition in both cases?
I would expect the second condition to be invoked with k[0], since the second one is operating at a deeper level.

tests/terraform/parser/test_plan_parser.py
```diff
@@ -146,6 +146,23 @@ def test_handle_complex_after_unknown(self):
...
 
+    def test_handle_complex_after_unknown_with_empty_list(self):
+        resource = {"network_configuration": [
+            {
+                "endpoint_configuration": [
+                ]
+            }
+        ]}
+        key: str = 'network_configuration'
+        value: list = [
```
maxamel: Can we have more than one item in this list?

checkov/terraform/plan_parser.py
bo156: right, fixing

tests/terraform/parser/test_plan_parser.py
bo156: updated the function to handle case where there are more than 1 by going in for loop

Issue #3910, opened 2022-11-21
CKV_K8S_21 - ability to pass namespace for Helm managed templates


Constantin07:
**Describe the issue**
CKV_K8S_21
The check does report that the `default` namespace is used when in fact Helm install uses a custom namespace when deploying to k8s.

**Examples**
```
Check: CKV_K8S_21: "The default namespace should not be used"
	FAILED for resource: ServiceAccount.default.release-name-hello-kubernetes
	File: /hello-kubernetes/templates/sa.yaml:3-13
	Guide: https://docs.bridgecrew.io/docs/bc_k8s_20

		3  | apiVersion: v1
		4  | kind: ServiceAccount
		5  | metadata:
		6  |   name: release-name-hello-kubernetes
		7  |   namespace: default
		8  |   labels:
		9  |     app.kubernetes.io/name: hello-kubernetes
		10 |     helm.sh/chart: hello-kubernetes-1.0.24
		11 |     app.kubernetes.io/instance: release-name
		12 |     app.kubernetes.io/managed-by: Helm
		13 |     app.kubernetes.io/version: "1.4"
```

As per Helm best practices here https://github.com/helm/helm/issues/5465#issuecomment-473942223 it's not recommended to hardcode the namespace as helm uses the parameter `-namespace` at installation time which namespace to deploy to.
So the below template for ServiceAccount, from helm perspective is completely valid.  
```yaml
cat sa.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ include "hello-kubernetes.fullname" . }}
  labels: {{- include "hello-kubernetes.labels" . | nindent 4 }}
```

**Version (please complete the following information):**
 - version: 2.2.80 

**Additional context**
Even if I add namespace to `metadata`, it still reports as FAILED.
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ include "hello-kubernetes.fullname" . }}
  namespace: {{ .Release.Namespace }}
  labels: {{- include "hello-kubernetes.labels" . | nindent 4 }}
```

I'm running checkov in container in Jenkins like described here https://www.checkov.io/4.Integrations/Jenkins.html.
```groovy
    stage('Static code analysis') {
      agent {
        docker {
          image 'bridgecrew/checkov:2.2.80'
          args "--entrypoint=''"
        }
      }
      steps {
        sh "checkov -d ${env.DIRECTORY} --framework helm"
      }
    }


```

[...]

PR #7355, opened 2025-10-23
fix(helm): Check HELM_NAMESPACE env var in CKV_K8S_21 

inbalavital:
Description
HELM_NAMESPACE env var should be checked in CKV_K8S_21
https://github.com/bridgecrewio/checkov/issues/3910

Fixes # (issue)
3910

### DIFF
--- a/tests/kubernetes/checks/test_DefaultNamespace.py
+++ b/tests/kubernetes/checks/test_DefaultNamespace.py
@@ -1,5 +1,6 @@
 import os
 import unittest
+from unittest import mock
 
 from checkov.kubernetes.checks.resource.k8s.DefaultNamespace import check
 from checkov.kubernetes.runner import Runner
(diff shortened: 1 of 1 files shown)


maxamel:
Can you add a test for this? I think you can use an existing test if we have one for this check.

omriyoffe-panw (approved):

tests/kubernetes/checks/test_DefaultNamespace.py
```diff
@@ -21,6 +21,19 @@ def test_summary(self):
         self.assertEqual(summary['skipped'], 0)
         self.assertEqual(summary['parsing_errors'], 0)
 
+    def test_summary_with_env_var(self):
+        runner = Runner()
+        current_dir = os.path.dirname(os.path.realpath(__file__))
+        os.environ['HELM_NAMESPACE'] = 'non-default'
```
maxamel: usually we don't set os.environ in tests since it can effect subsequent tests.
Can you use mock.patch or a similar mechanism to limit the env var setting only to this test?

PR #7369, opened 2025-11-17
fix(terraform): Fixed variable rendering of complex variables to avoid changing type

bo156:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Addresses issue in terraform variable rendering of complex types like list/dict. 
The issue was that we wrapped most variables in a `'` to make it pass `asteval` (as the syntax `var.name` is not valid python syntax for variable), which could result in "changing" the variable type from list to str for example when evaluating it.

Note - a more substantial solution would be renaming the var names to valid python format, and passing them to `asteval` when evaluating the dict( similar to `SAFE_EVAL_DICT`). This solution should be implemented in the future.

## Checklist:

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -139,6 +139,7 @@ jobs:
       - uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e  # v4
       - uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78  # v3
         with:
+          version: "v3.19.1"  # the tests break starting v4 as checkov cannot support it, needs to be investigated
           token: ${{ secrets.GITHUB_TOKEN }}
       - uses: imranismail/setup-kustomize@a76db1c6419124d51470b1e388c4b29476f495f1  # v2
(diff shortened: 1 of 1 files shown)


lirshindalman (approved):

checkov/terraform/graph_builder/variable_rendering/evaluate_terraform.py
```diff
@@ -155,7 +162,7 @@ def _find_new_value_for_interpolation(origin_str: str, str_to_replace: str, new_
             return new_value
 
         # Second part - checking if escaped is valid
-        escaped_new_value = f"'{new_value}'"
+        escaped_new_value = f'"{new_value}"'
```
maxamel, line 165: Is this change needed? Seems like only the order of " and ' changed

bo156: It is actually a core part of the change - when we parse terraform using `hcl2` the generated string uses `'`.
This means that some changes are causing bad escaping of strings if we use the same character. Without this change the added test will not pass.

PR #7378, opened 2025-11-22
fix(kubernetes): Fix CKV_K8S_21

Saarett:
When the name or kind of the metadata are missing, the policy would crash the run


maxamel (approved):

checkov/kubernetes/checks/resource/k8s/DefaultNamespace.py
```diff
@@ -36,15 +36,15 @@ def __init__(self) -> None:
     def scan_spec_conf(self, conf: dict[str, Any]) -> CheckResult:
         metadata = conf.get("metadata")
         if metadata:
-            if "namespace" in metadata and metadata["namespace"] != "default":
+            if metadata.get("namespace") != "default":
```
inbalavital: this one should stay as it was, since metadata.get("namespace") might equal None and is different from "default", but the check wants to make sure there is a namespace that is different than "default", and empty should fail (if other conditions don't apply)

Issue #7407, opened 2025-12-23
Recommendation from SCA


sergiomarotco:
**Describe the issue**
Is it possible to update base image and components? My security team says that using Checkov is not secure.
SCA says:

<img width="1619" height="476" alt="Image" src="https://github.com/user-attachments/assets/9ef7a4fc-a1a8-4383-9edd-a57e2f0aeea8" />

<img width="1639" height="696" alt="Image" src="https://github.com/user-attachments/assets/cc4045c5-eeff-4f60-b79b-385f5e81a75d" />
.... and more

<img width="1613" height="353" alt="Image" src="https://github.com/user-attachments/assets/5f76805c-0735-415a-b321-968e50aa124b" />

**Version (please complete the following information):**
 - Checkov Version: bridgecrew/checkov:3.2.495 - Docker image

**Additional context**

1. SBOM: [bom_bridgecrew_checkov_f7b4b05e_cyclonedx_v1_6.json](https://github.com/user-attachments/files/24306656/bom_bridgecrew_checkov_f7b4b05e_cyclonedx_v1_6.json)
2. Report: [report_bridgecrew_checkov_f7b4b05e.pdf](https://github.com/user-attachments/files/24306752/report_bridgecrew_checkov_f7b4b05e.pdf)

PR #7418, opened 2026-01-15
chore: Update Docker base image and components

omryMen:

## Description

Changes applied to both root Dockerfile and kubernetes/Dockerfile. All changes are backward-compatible and non-breaking. Addresses security team concerns about outdated components.

- Upgrade base image from python:3.11-slim to python:3.12-slim
- Add explicit pip, setuptools, and wheel upgrade before installing dependencies
- Maintain all existing security features (apt-get upgrade, checksum verification)
- Add --no-cache-dir flags to pip install commands for reduced image size

Fixes # ([issue](https://github.com/bridgecrewio/checkov/issues/7407))

## Checklist:


kubernetes/Dockerfile
```diff
@@ -1,27 +1,53 @@
...
+RUN groupadd -g 12000 -r checkov \
+ && useradd -u 12000 --no-log-init -r -g checkov checkov
+
+# Create necessary directories
+RUN mkdir /data && mkdir /app && mkdir /home/checkov \
+ && chown checkov:checkov /data /app /home/checkov
+
+# Install yq v4.49.2 (latest stable, built with Go 1.25.4 - still has CVE-2025-61729, needs Go 1.25.5+)
+RUN YQ_VERSION="v4.49.2" \
+ && curl -fL -o /usr/bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64" \
```
omriyoffe-panw: This curl would download a binary file but the grep in line 41 is for the .tar.gz file. Is it intentional?

kubernetes/Dockerfile
```diff
@@ -1,27 +1,53 @@
...
+ && curl -fL -o yq.sha256 "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/checksums" \
+ && grep "yq_linux_amd64.tar.gz" yq.sha256 || true \
```
omriyoffe-panw: Is this supposed to verify the checksum of the binary file? If so than the grep should be for "yq_linux_amd64" instead of the .tar.gz file (as I said on the previous comment) and also this command only checks if the text string exists in the checksum file. It does not calculate the hash of the downloaded binary.
Additionally the || true ensures the build continues even if the grep fails, essentially making this step do nothing.

Suggestion:
`grep "yq_linux_amd64 " yq.sha256 | awk '{print $2 " /usr/bin/yq"}' | sha256sum --check
`

Dockerfile
```diff
@@ -1,34 +1,49 @@
...
+RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
+    && pip install --no-cache-dir setuptools==79.0.0 urllib3==2.6.3
+
+# Install Helm v3.17.2 (built with Go 1.25.5+ to fix CVE-2024-25621, CVE-2025-22869, CVE-2025-22868, and Go stdlib CVEs)
+RUN HELM_VERSION="v3.17.2" \
+    && curl -fsSL "https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz" -o helm.tar.gz \
+    && curl -fsSL "https://get.helm.sh/helm-${HELM_VERSION}-linux-amd64.tar.gz.sha256sum" -o helm.sha256 \
+    && cat helm.sha256 \
+    && sha256sum helm.tar.gz \
+    && grep $(sha256sum helm.tar.gz | awk '{print $1}') helm.sha256 || echo "Checksum verification skipped" \
```
omriyoffe-panw: Why do we allow to skip the helm checksum verification and not the k8s one? If we allow skipping let's make it uniform also for k8s and yq with the added log of the "Checksum verification skipped"

Dockerfile
```diff
@@ -1,34 +1,49 @@
...
+    && tar -zxf helm.tar.gz \
+    && mv linux-amd64/helm /usr/local/bin/helm \
+    && chmod +x /usr/local/bin/helm \
+    && rm -rf helm.tar.gz helm.sha256 linux-amd64
+
+# Install Kustomize v5.7.1 (built with Go 1.25.5+ to fix Go stdlib CVEs: CVE-2025-47907, CVE-2025-58183, CVE-2025-61729)
+RUN KUSTOMIZE_VERSION="v5.7.1" \
+    && curl -fsSL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2F${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz" -o kustomize.tar.gz \
```
omriyoffe-panw: Same as the previous comment, let's make the checksum verification uniform if we can.

Dockerfile
```diff
@@ -1,34 +1,49 @@
...
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        git \
+        curl \
+        openssh-client \
+    && rm -rf /var/lib/apt/lists/*
 
+# Upgrade pip and install Python dependencies (fixes CVE-2026-21441, CVE-2025-66471, CVE-2025-66418, GHSA-58pv-8j8x-9vj2)
+RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
+    && pip install --no-cache-dir setuptools==79.0.0 urllib3==2.6.3
```
omriyoffe-panw, line 17: The main Pipfile for checkov pins setuptools = "==78.1.1", but the Dockerfiles are manually installing setuptools==79.0.0. We might want to update that as well if there is a CVE there.

omriyoffe-panw (approved):
Looks Good!

SteveVaknin (approved):

PR #7436, opened 2026-02-11
fix(terraform): return inner module path when dest_dir already exists on Linux

guyg-palo:
## Description

When multiple Terraform modules share the same base source but reference different inner modules (using `//` syntax), the `RegistryLoader._load_module` method incorrectly returned the base `dest_dir` when it already existed on disk.

This caused only the **first set of inner modules** to be scanned, while subsequent inner modules from the same base source were silently skipped because the scanner received the base directory instead of the specific inner module directory.

### Root Cause

In `checkov/terraform/module_loading/loaders/registry_loader.py`, the `_load_module` method has an early return on line 77-78:

```python
if os.path.exists(module_params.dest_dir):
    return ModuleContent(dir=module_params.dest_dir)
```

This does not account for `inner_module`. When the `dest_dir` already exists (because a previous module with the same base source was downloaded), it returns the base directory instead of the inner module subdirectory.

### Why it only happens on Linux

- **Windows**: Uses MD5 hashes of the **full source** (including inner module path) for directory names, creating unique directories for each module. The bug is never triggered.
- **Linux**: Uses the source path directly. After `_process_inner_registry_module` splits by `//`, modules sharing the same base source share the same `dest_dir`. When the second module is loaded, `os.path.exists(dest_dir)` is `True`, triggering the buggy early return.

### Example

Given two modules with sources:
- `registry.example.com/org/repo//modules/asg`
- `registry.example.com/org/repo//modules/ssm_document`

Both resolve to the same `dest_dir` after `_process_inner_registry_module` splits by `//`. When the second module is loaded, `dest_dir` already exists, and the code was returning `ModuleContent(dir=dest_dir)` instead of `ModuleContent(dir=os.path.join(dest_dir, inner_module))`.

### Fix

Added a check for `inner_module` before returning the directory path:

```python
if os.path.exists(module_params.dest_dir):
    if module_params.inner_module:
        return ModuleContent(dir=os.path.join(module_params.dest_dir, module_params.inner_module))
    return ModuleContent(dir=module_params.dest_dir)
```

This mirrors the existing logic on lines 121-122 (in the download path) that already correctly handles `inner_module`.


maxamel:
Is this covered by any of the existing tests? If not, can one be added?

guyg-palo:
> Is this covered by any of the existing tests? If not, can one be added?

Thanks @maxamel.

Two tests were added to `tests/terraform/module_loading/loaders/test_registry_loader.py`:

1. `test_load_module_returns_inner_module_path_when_dest_dir_exists` - Simulates the exact bug scenario: two modules sharing the same base source (`repo//modules/asg` and `repo//modules/ssm_document`) with `dest_dir` already existing on disk. Asserts each call to `_load_module` returns the correct inner module subdirectory path, and that the two paths are different.

2. `test_load_module_returns_dest_dir_when_no_inner_module` - Regression test confirming modules without `inner_module` still return `dest_dir` as before.

All 19 tests pass (17 existing + 2 new)

tests/terraform/module_loading/loaders/test_registry_loader.py
```diff
@@ -162,3 +162,90 @@ def test_is_matching_loader_git_sources(source_url):
...
+    """
+    # given
+    loader = RegistryLoader()
+    base_dir = tmp_path / "base_module"
+    inner_module_a = base_dir / "modules" / "asg"
+    inner_module_b = base_dir / "modules" / "ssm_document"
+
+    # Simulate a previously downloaded module: create the base dir and both inner module dirs
+    inner_module_a.mkdir(parents=True)
+    inner_module_b.mkdir(parents=True)
```
maxamel, line 185: Are those cleaned up anywhere? I don't see cleanup in the test, you can verify if after the test the created files are still on the filesystem

tests/terraform/module_loading/loaders/test_registry_loader.py
```diff
@@ -162,3 +162,90 @@ def test_is_matching_loader_git_sources(source_url):
...
+
+def test_load_module_returns_dest_dir_when_no_inner_module(tmp_path):
+    """
+    When dest_dir already exists and there is no inner_module,
+    _load_module should return dest_dir as before (no regression).
+    """
+    # given
+    loader = RegistryLoader()
+    base_dir = tmp_path / "base_module"
+    base_dir.mkdir(parents=True)
```
maxamel, line 234: same as https://github.com/bridgecrewio/checkov/pull/7436/changes#r2837377449

tests/terraform/module_loading/loaders/test_registry_loader.py
guyg-palo: The directories created by `inner_module_a.mkdir()` and `inner_module_b.mkdir()` on lines 184-185 are created **inside** the `tmp_path` directory, which is a built-in pytest fixture.

**How `tmp_path` cleanup works:**
- `tmp_path` creates a unique temporary directory per test under the system temp dir
- Directories persist after the test session ends (this is by design - useful for debugging failed tests)
- Pytest automatically cleans up old sessions, keeping only the **last 3 session directories**
- On the next `pytest` run, older temp directories are automatically removed

@maxamel let me know if you think we need a specific handling here.

tests/terraform/module_loading/loaders/test_registry_loader.py
guyg-palo: same reply :)
https://github.com/bridgecrewio/checkov/pull/7436#discussion_r2837409583

maxamel (approved):

PR #7471, opened 2026-03-17
fix(terraform): resolve module output edges by identity, not path alone

### LINKED ISSUE #7470: Variable resolution fails when two module instances share the same source and a third module references one's output
## Description

When two module calls share the same source, and a third module (different source) references one of their outputs, Checkov fails to resolve variables in the third module. Explicitly-passed argument values remain as unresolved `var.*` references, causing false positives on any check that inspects those attributes.

The bug is **ordering-dependent**: it only triggers when the sibling module's name sorts alphabetically **before** the producer's name (e.g., module `"a"` before module `"b"`).

**Checkov version**: 3.2.508

### Directory layout

```
.
├── main.tf
└── modules/
    ├── mod_a/
    │   └── main.tf
    └── mod_c/
        └── main.tf
```

### `main.tf`

```hcl
module "b" {
  source = "./modules/mod_a"
}

module "a" {
  source = "./modules/mod_a"
  result = module.b.result
}

module "c" {
  source = "./modules/mod_c"

  default_action = "Deny"

  ref = module.b.result.some_attr
}
```

### `modules/mod_a/main.tf`

```hcl
variable "result" {
  default = null
}

output "result" {
  value = var.result
}
```

### `modules/mod_c/main.tf`

```hcl
variable "default_action" {
  type    = string
  default = "Deny"
}

variable "ref" {
  default = null
}

resource "azurerm_storage_account" "sa" {
  network_rules {
    default_action = var.default_action
  }
}
```

### Run

```bash
checkov --directory . --check CKV_AZURE_35 --framework terraform
```

## Expected result

```
Passed checks: 1, Failed checks: 0
```

`default_action = "Deny"` is explicitly passed to module `"c"`. The variable should resolve to `"Deny"`.

## Actual result

```
Passed checks: 0, Failed checks: 1

Check: CKV_AZURE_35: "Ensure default network access rule for Storage Accounts is set to deny"
        FAILED for resource: module.c.azurerm_storage_account.sa
        File: /modules/mod_c/main.tf
        Calling File: /main.tf
```

The resource config shows `default_action = var.default_action` — the variable reference was never resolved.

## Root cause

In `TerraformLocalGraph._should_add_edge()`, the second condition uses a loose path-only comparison (`self.get_abspath(vertex.source_module_object.path) == self.get_abspath(module_node.path)`) to decide which output vertex to connect an edge to. When two modules share the same source, their outputs live at the same path, so the edge connects to whichever output is encountered first (alphabetically by module name) rather than the correct module instance.

The fix is to replace the path comparison with `get_vertex_as_tf_module(module_node)`, which compares the full module identity (name + path + nested modules) instead of just the path.

## Elimination matrix

Each row changes one thing from the failing reproduction:

| Change                                                      | Result   |
| ----------------------------------------------------------- | -------- |
| Remove module `"a"` (the same-source sibling)               | **PASS** |
| Remove `ref` argument from module `"c"`                     | **PASS** |
| Replace `module.b.result.some_attr` with a literal string   | **PASS** |
| Rename `"a"` → `"z"` (consumer now sorts after producer)    | **PASS** |

pszypowicz:
## Summary

- Fix `_should_add_edge()` to match output vertices by full module identity (`get_vertex_as_tf_module()`) instead of path only, preventing wrong edge connections when two module calls share the same source
- Add regression test that verifies variable resolution works correctly in this scenario

## Details

When two module calls share the same source (e.g., modules `"a"` and `"b"` both using `./mod_a`), and a third module `"c"` references one of their outputs, the old path-only check in `_should_add_edge()` could connect the edge to the wrong module instance. This corrupted variable resolution for module `"c"`, leaving `var.*` references unresolved and causing false positives (e.g., CKV_AZURE_35).

The fix is a one-line change replacing the loose path comparison with `get_vertex_as_tf_module(module_node)`, which already exists in the codebase and compares the full module identity (name, path, nested modules).

Fixes #7470

## Test plan

- [x] New test `test_module_output_edge_with_same_source_siblings` passes with the fix
- [x] Same test fails without the fix (`'Deny' != 'var.default_action'`)
- [x] All existing graph builder tests pass (49/49)
- [x] All variable rendering tests pass (205/205)


pszypowicz:
**Why this works:**

The bug is in `_should_add_edge()` at `local_graph.py:845`. The method decides which output vertex to connect an edge to when resolving `module.<name>.<output>` references.

The old second condition:
```python
self.get_abspath(vertex.source_module_object.path) == self.get_abspath(module_node.path)
```
Only compared file paths. When modules `"a"` and `"b"` both use source `./mod_a`, their output vertices live at the same path — so the first match wins (alphabetically), connecting to the wrong module instance.

The fix:
```python
vertex.source_module_object == get_vertex_as_tf_module(module_node)
```
`get_vertex_as_tf_module()` (already used elsewhere in this file, line 416) constructs a `TFModule` from the vertex, which includes the module **name**, path, and nested module chain. This distinguishes module `"a"` from module `"b"` even when they share the same source.

PR #7519, opened 2026-04-17
fix(terraform): improve performance in nested modules by reducing allocations

### LINKED ISSUE #7524: Performance regressions for deeply-nested terraform modules
At various points in time we've observed serious performance regressions while scanning our terraform projects, notably those that have deeply nested modules and lots of use of count/for_each. I have notes from my teammate who investigated, saying:
> 3.2.34 took runs from 15->30m, then 3.2.119 took them to 4h+

At the time that teammate developed a patch that avoids some object allocation inside inner loops, and that brought performance back to acceptable levels. We've been carrying on applying that patch when we install checkov, but it recently broke and rather than patch the patch, we thought we'd try to upstream the fix. 

I already created a [PR](https://github.com/bridgecrewio/checkov/pull/7519) but am filing an issue to ensure that the issue gets the proper visibility.

patrickvinograd:

## Description

Improves the performance of terraform scans in states with deeply nested modules:
* Moves the allocation of a static list of source types out of the inner loop of `S3AllowsAnyPrincipal.py`
* Uses a list instead of an array in the vertex-walking code in `module_handler.py` to avoid that list growing very large, also within an inner loop.

Motivation: We have several terraform projects that are highly modularized with deeply nested modules. We observed step performance regressions in 3.2.34 and 3.2.119 and after my teammate did some local profiling, have been patching checkov locally with these fixes, but would like to upstream them. With these fixes we see a 6-7x performance improvement when scanning such terraform projects.

I did not see any obvious performance tests that I could add to but functionally these changes seem equivalent to existing behavior and low-risk given existing test coverage. 

Fixes: #7524

## Checklist:


maxamel:
Thanks @patrickvinograd , can you check the failing workflows please?

patrickvinograd:
I was out last week, but taking a look at the failing tests now.

patrickvinograd:
In the course of looking at the failing tests, I discovered that our patch is *not* functionally equivalent - in fact it is aborting out of the module-foreach graph traversal and swallowing an exception, because TerraformBlock is not hashable, so can't be added to a set. When looking in the logs I see "Failed to process foreach handling, error: unhashable type: 'TerraformBlock'" 

So effectively by not building any of the module foreach graph edges, we were doing the equivalent of running with `CHECKOV_ENABLE_MODULES_FOREACH_HANDLING=false` and that's where all the performance improvements were coming from. 

The change in S3AllowsAnyPrincipal.py is likewise not significant with respect to performance, although it's a little nicer to read.

patrickvinograd:
Closing this PR due to the above.

PR #7529, opened 2026-04-26
fix(terraform): fix wrong windows path

yuvalmich:

# fix: preserve subdirectory paths in graph check results on Windows

## Description

Fixes a bug where CKV2_* (graph) checks on Windows would lose intermediate subdirectory components from `file_path`, reporting `\NESTED-TEST.tf` instead of `\Nested\NESTED-TEST.tf`.

## Root Cause

A previous workaround for Windows cross-drive `os.path.relpath` failures (which raises `ValueError` when paths are on different drives) unconditionally replaced `root_folder` with the file's parent directory via `os.path.split(full_file_path)[0]`. This caused `os.path.relpath` to compute paths relative to the wrong base, stripping all intermediate directories.

## Fix

Introduced a `safe_relpath()` utility that wraps `os.path.relpath()` in a try/except, falling back to `os.path.basename()` **only** when a `ValueError` is raised (the actual cross-drive case). Removed the blanket `platform.system() == "Windows"` root folder overrides from:

- `checkov/terraform/base_runner.py` — `get_graph_checks_report()` and `_get_connected_node_data()`
- `checkov/common/runners/object_runner.py` — `add_python_check_results()` and `add_graph_check_results()`
- `checkov/terraform/plan_runner.py` — `_get_file_path()`

## Before / After

| Check Type | Before (Windows) | After (Windows) |
|---|---|---|
| CKV_AZURE_59 (Python) | `\Nested\NESTED-TEST.tf` ✅ | `\Nested\NESTED-TEST.tf` ✅ |
| CKV2_AZURE_47 (Graph) | `\NESTED-TEST.tf` ❌ | `\Nested\NESTED-TEST.tf` ✅ |

## Checklist:

### DIFF
--- a/checkov/common/util/file_utils.py
+++ b/checkov/common/util/file_utils.py
@@ -23,7 +23,7 @@ def convert_to_unix_path(path: str) -> str:
     return path.replace('\\', '/')
 
 
-def safe_relpath(file_path: str, root_folder: str | None) -> str:
+def safe_relpath(file_path: str, root_folder: str | Path | None) -> str:
     """Compute os.path.relpath(file_path, root_folder) safely on Windows.
 
     On Windows ``os.path.relpath`` raises ``ValueError`` when *file_path* and
@@ -38,6 +38,10 @@ def safe_relpath(file_path: str, root_folder: str | None) -> str:
         return os.path.relpath(file_path, root_folder)
     except ValueError:
         # Cross-drive on Windows – best-effort: return the basename only
+        logger.warning(
+            f"Could not compute relative path for '{file_path}' from root '{root_folder}' "
+            f"(cross-drive path on Windows); falling back to basename."
+        )
         return os.path.basename(file_path)
 
 
--- a/checkov/common/util/file_utils.py
+++ b/checkov/common/util/file_utils.py
@@ -23,6 +23,24 @@ def convert_to_unix_path(path: str) -> str:
     return path.replace('\\', '/')
 
 
+def safe_relpath(file_path: str, root_folder: str | None) -> str:
+    """Compute os.path.relpath(file_path, root_folder) safely on Windows.
+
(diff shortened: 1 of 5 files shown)


checkov/common/util/file_utils.py
```diff
@@ -23,6 +23,24 @@ def convert_to_unix_path(path: str) -> str:
...
+    caller still gets a usable (albeit shortened) relative path instead of
+    crashing.
+
+    On non-Windows platforms this is a thin wrapper around ``os.path.relpath``.
+    """
+    try:
+        return os.path.relpath(file_path, root_folder)
+    except ValueError:
+        # Cross-drive on Windows – best-effort: return the basename only
+        return os.path.basename(file_path)
```
omriyoffe-panw, line 45: shouldn't it be `root_folder = os.path.split(file_path)[0]` like the original solution used?

omriyoffe-panw: Also maybe add a log when we fallback to the original flow so it would be easy to spot if errors would arise here again

yuvalmich: **In regards to the first comment:**

os.path.basename(file_path) — returns only the filename (last component), e.g. "main.tf". It strips the entire directory path.
os.path.split(file_path)[0] — returns the directory part of the path (everything except the filename), e.g. "/home/user/project". It's the opposite of basename.
Neither is what the fallback in [safe_relpath()](vscode-webview://0aqjet3cc3b0qhlc46a5tr38v4l5k6nsssaj4i1ji1oga3drics4/checkov/common/util/file_utils.py:26) is doing — the fallback os.path.basename(file_path) returns just the filename, which is a reasonable best-effort when a cross-drive relative path can't be computed on Windows.

**And the second comment:**
Added a warning log for the fallback case

omriyoffe-panw (approved):

inbalavital (approved):
let's add a test for this case

PR #7540, opened 2026-05-10
fix(secrets): report all multiline regex matches per file, not just first occurrence - FIX

eshmayovitz:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:

### DIFF
--- a/checkov/secrets/plugins/custom_regex_detector.py
+++ b/checkov/secrets/plugins/custom_regex_detector.py
@@ -168,25 +168,45 @@ class CustomRegexDetector(RegexBasedDetector):
                 multiline_regex = self.pattern_by_prerun_compiled.get(regex.pattern)
                 if multiline_regex is None:
                     continue
-                multiline_matches = multiline_regex.findall(file_content)
-                for mm in multiline_matches:
-                    mm = self._extract_real_regex_match(mm)
-                    if isinstance(mm, tuple):
-                        mm = mm[0]
-                    line_num = find_line_number(file_content, mm, line_number)
-                    quoted_mm = f"'{mm}'"
-                    ps = PotentialSecret(
-                        type=regex_data["Name"],
-                        filename=filename,
-                        secret=quoted_mm,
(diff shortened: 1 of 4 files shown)


maxamel (approved):

checkov/secrets/plugins/custom_regex_detector.py
```diff
@@ -168,25 +168,45 @@ def _find_potential_secret(
...
+                            line_number=line_num,
+                            is_verified=is_verified,
+                            is_added=is_added,
+                            is_removed=is_removed,
+                            is_multiline=True,
+                        )
+                        ps.check_id = regex_data["Check_ID"]
+                        output.add(ps)
+                    except Exception:
+                        logging.warning('Failed to process multiline match.')
```
yuvalmich: you swallow the exception here, which will make it hard to debug.

Can you add the exception details here? (I would also consider log error instead of log warning)

eshmayovitz: I omitted the exception details intentionally to avoid leaking secrets into logs. 
Exception messages may contain the matched secret value, so a static message is safer. 
Also, warning fits better than error here since this is recoverable - we skip the match and continue scanning.

yuvalmich: So maybe let's change it to something more informative like:

```
except Exception as e:
    logging.warning(
        'Failed to process multiline match for check %s in %s at offset %d: %s',
        regex_data.get("Check_ID", "unknown"),
        filename,
        regex_match.start(),
        type(e).__name__,
    )
    continue
```

This way if something breaks in production, the log would show something like:

WARNING - Failed to process multiline match for check CKV_SECRET_123 in /app/config.yaml at offset 4821: AttributeError

Instead of just:

WARNING - Failed to process multiline match.

yuvalmich (approved):

PR #7548, opened 2026-05-19
fix(terraform): CKV_AWS_304 correctly evaluates cron() schedule expressions instead of unconditionally passing

tejgokani:
## Description

`CKV_AWS_304` ("Ensure Secrets Manager secrets should be rotated within 90 days") 
previously returned `PASSED` for **all** `cron()` schedule expressions due to an 
unimplemented TODO — meaning a secret rotating once a year (e.g., `cron(0 12 1 1 ? *)`) 
was silently reported as compliant.

This PR implements `_check_cron_expression()` which parses the AWS EventBridge cron 
month field and calculates the maximum possible gap between rotations. If that gap can 
exceed 90 days, the check returns `FAILED`. The method fails safe — malformed or 
unparseable expressions return `FAILED` rather than `PASSED`.

Also uncomments the `fail_scheduled_cron` test fixture that was left commented out 
with `# Will handle later`, and updates the test to include it in expected failures.

Fixes # (issue)

### Description

`aws_secretsmanager_secret_rotation` resources that use a `schedule_expression` with 
`cron()` syntax could configure rotation intervals far exceeding 90 days (e.g., once 
per year, once per quarter) and still pass the check. This violates the intent of 
CKV_AWS_304 and the CIS AWS Foundations Benchmark requirement for regular secret rotation.

AWS EventBridge cron docs: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-cron-expressions.html

### Fix

Use `automatically_after_days` set to 90 or fewer, or use a `schedule_expression` 
with either a `rate()` expression within 90 days, or a `cron()` expression whose 
month field guarantees rotation at least every 90 days (e.g., monthly `*` or 
every-2-months `*/2`).

## Checklist:

### DIFF
--- a/checkov/terraform/checks/resource/aws/SecretManagerSecret90days.py
+++ b/checkov/terraform/checks/resource/aws/SecretManagerSecret90days.py
@@ -29,6 +29,55 @@ class SecretManagerSecret90days(BaseResourceCheck):
                 return value <= 129600  # 90 days * 24 hours * 60 minutes
         return False
 
+    def _check_cron_expression(self, expression: str) -> bool:
+        """
+        Parse AWS EventBridge cron format: cron(minutes hours day-of-month month day-of-week year)
+        Returns True (PASS) only if the rotation is guaranteed to run at least every 90 days.
+        Fails safe — if parsing fails or is ambiguous, returns False (FAIL).
+        """
+        inner = expression[5:-1]  # strip cron( and )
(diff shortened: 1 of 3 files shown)


tejgokani:
Hey @maxamel @Saarett @AdamDev, sorry for the ping. Could somebody have a look at this?

PR #7556, opened 2026-05-26
feat(general): verify ECDSA-P256 signatures on external custom checks before loading

mayblo:

## Description

Adds optional cryptographic verification of every Python custom check loaded
via `--external-checks-dir` / `--external-checks-git`.

By default, Checkov loads every `.py` file under `--external-checks-dir`
(or `--external-checks-git`) and executes it. In environments where the
custom-checks tree must be tamper-evident, Checkov can now verify a
cryptographic signature on each file before loading it.

When the operator passes one or more PEM-encoded ECDSA P-256 public keys
(new `--external-checks-public-key` flag, or `CKV_EXTERNAL_CHECKS_PUBLIC_KEY`
env var), Checkov refuses to load any `.py` file lacking a valid
`# checkov-digest: <hex>` trailer signed by one of the configured keys, and
exits before any scan runs.

Strictly opt-in and fully backward compatible. When no key is configured,
the loader executes exactly the same `spec_from_file_location → exec_module`
code path as before this MR.

## Highlights

* **Crypto primitives pinned**: ECDSA P-256 + SHA-256, strict DER encoding,
  `r,s ∈ [1, n-1]` range check. No low-S enforcement (required for AWS-KMS /
  OpenSSL interop; safe — no cache is keyed by signature bytes). Hash
  algorithm and DER encoding are both pinned: JWS / raw `r∥s` signatures
  are rejected as `BAD_SIGNATURE`; SHA-384 / SHA-512 / pre-hashed signing
  produce `UNKNOWN_KEY`.
* **Byte-exact wire format**: trailer prefix is exactly `# checkov-digest: `
  (one ASCII space), lowercase hex `[126, 144]` chars, file ends with
  exactly one `\n`. No CRLF, no BOM. A file ending with two trailer lines
  produces a dedicated `DOUBLE_TRAILER` result.
* **Exit semantics**: verification failure exits via `Checkov.exit_run()`
  so the documented `--no-fail-on-crash` contract ("Return exit code 0
  instead of 2") is honoured for all failure types, including verification
  failures. The stderr diagnostic and the
  `./checkov-verification-failures.log` file are written regardless of the
  flag — only the exit code differs.
* **Defence in depth**: the loader independently re-checks every walked
  file against the realpath-normalised allowlist registered at verify
  time; any `.py` file that appears on disk between verification and load
  is refused and surfaces as `SignatureVerificationError`.
* **Multi-key rotation supported** (pass the flag multiple times).
* **`sast_custom_policies` carve-out**: the platform-supplied directory is
  appended *after* the verification block; it reaches Checkov through a
  separate TLS-authenticated channel and is out of scope for trailer
  signing.

## Scope

Verification covers `.py` files only. YAML / JSON graph-check definitions
in the same directory are loaded without signature verification — they
don't execute as code and gating them is intentionally out of scope for
this MR. Binary loadable files (`.pyc`, `.pyi`, `.so`, `.pyd`) are
hard-rejected when found in a verified directory so operators get a clear
error rather than a silent skip.

## Dependency

Adds [`ecdsa>=0.19.0,<1.0.0`](https://pypi.org/project/ecdsa/) to the core
install — a small, audited, pure-Python ECDSA implementation. No native
compilation, no transitive dependencies.

## Files

**New package** `checkov/common/external_checks/verification/`:
`__init__.py`, `errors.py`, `keys.py`, `trailer_format.py`, `verifier.py`,
`enforce.py`, `sources_registry.py`, `verified_loader.py`.

**Modified**:
* `base_check_registry.py` — verified-load path integrated; legacy
  unverified branch is byte-identical to today.
* `main.py` — chokepoint in `get_external_checks_dir`,
  verification-failure reporting helper.
* `ext_argument_parser.py` — new `--external-checks-public-key` flag with
  `CKV_EXTERNAL_CHECKS_PUBLIC_KEY` env var.
* `setup.py` / `Pipfile` — add `ecdsa` dependency.

**Docs**: `docs/3.Custom Policies/Python Custom Policies.md` — new
"Signing custom checks (optional)" section with the portable OpenSSL
recipe (`od -An -tx1 | tr -d ' \n'`), key-rotation example, and explicit
statement that signing covers `.py` files only.

## Tests

**149 tests** across 7 files under `tests/common/external_checks/verification/`:

* `test_verifier.py` — happy path, missing/unknown-key/partial signing
  rejection, multi-key rotation, ignored non-Python files, binary-file
  rejection, error message format.
* `test_regression.py` — `--no-fail-on-crash` behaviour, sticky-flag
  absence pinning, realpath/symlink consistency, post-verification disk
  drift, and other prior-review regressions.
* `test_chokepoint.py` / `test_chokepoint_git.py` — chokepoint
  integration including the `sast_custom_policies` carve-out and
  `--external-checks-git` path.
* `test_sources_registry.py` — process-global registry lifecycle.
* `test_verified_loader.py` — in-memory loader + meta-path finder
  behaviour for cross-directory transitive imports.
* `test_signer_interop.py` — cross-tool interop with Python
  `cryptography`, OpenSSL CLI (subprocess), KMS-style high-S signatures,
  JWS / raw `r∥s` rejection, and hash-algorithm pinning.

Expected: **`149 passed, 1 skipped`** (the skip is a probabilistic
short-DER-signature test).

## Checklist:


Pipfile
Saarett, line 1: Please update the lock file as well

MaryArmaly (approved):

checkov/main.py
```diff
@@ -756,12 +767,65 @@ def commit_repository(self) -> str | None:
...
+            extra = len(bullets) - _VERIFICATION_FAILURE_INLINE_LIMIT
+            inline_detail = (
+                "\n".join(visible)
+                + f"\n  ... and {extra} more "
+                f"(see ./{_VERIFICATION_FAILURE_LOG_FILENAME} for the full list)"
+            )
+        else:
+            inline_detail = "\n".join(bullets)
+
+        print(
```
AdamDev: why not using logging lib like all other loggers?

checkov/main.py
```diff
@@ -756,12 +767,65 @@ def commit_repository(self) -> str | None:
...
+            "External checks signature verification failed; "
+            "refusing to run the scan. Offending files:\n"
+            f"{inline_detail}",
+            file=sys.stderr,
+        )
+        log_path = f"./{_VERIFICATION_FAILURE_LOG_FILENAME}"
+        try:
+            with open(log_path, "w") as f:
```
AdamDev: Why use special log file and not printing into the main log file? 
Product requirement? my concern is that it won't upload on dev status.

checkov/main.py
```diff
@@ -756,12 +767,65 @@ def commit_repository(self) -> str | None:
...
+                f.write(
+                    "External checks signature verification failed; "
+                    "refusing to run the scan. Offending files:\n"
+                )
+                f.write("\n".join(bullets) + "\n")
+        except OSError as exc:
+            # Read-only cwd is the usual cause; tell the operator
+            # explicitly so they don't chase a non-existent file.
+            print(
```
AdamDev: logging instead of print

checkov/common/external_checks/verification/enforce.py
```diff
@@ -0,0 +1,164 @@
...
+from .errors import SignatureVerificationError
+from .keys import VerificationKey
+from .verifier import VerificationResult, verify_file
+
+
+logger = logging.getLogger(__name__)
+
+
+LOADABLE_SUFFIXES = (".py",)
+_BINARY_LOADABLE_SUFFIXES = (".pyc", ".pyi", ".so", ".pyd")
```
AdamDev: Do we really need to support binaries files?

checkov/common/external_checks/verification/trailer_format.py
```diff
@@ -0,0 +1,81 @@
...
+    return body, last_nl, body[last_nl + 1:]
+
+
+def parse_trailer(file_bytes: bytes) -> "tuple[bytes, bytes] | None":
+    split = _split_last_line(file_bytes)
+    if split is None:
+        return None
+    _body, last_nl, trailer_line = split
+
+    if not trailer_line.startswith(TRAILER_PREFIX):
```
AdamDev, line 30: does it remove spaces from start?

AdamDev (approved):

PR #7557, opened 2026-05-26
fix(dockerfile): support custom-prefix Dockerfile names like AdtDockerfile

### LINKED ISSUE #7541: Enhance Pattern DOCKERFILE_MASK in dockerfile.py
**Description**
I am facing the problem that Checkov is silently ignoring some of our Dockerfiles though they are present and valid. Some digging into Checkov's code brought up this pattern found here: https://github.com/bridgecrewio/checkov/blob/da500c3412faf9fb1a007bc0b0fd739cca737491/checkov/common/util/dockerfile.py#L3

Now we have some files not named "Dockerfile" or "dockerfile" but instead something like "AdtDockerfile" which will not be matched by the pattern.

Now the question I'd like to ask is if there is any special reason why Checkov only accepts these base dockerfile names? Are there any plans on enhancing this pattern? If not, would you accept a PR on this issue?

**Version:**
 - 3.2.562

Thanks! Bernd

AlexanderSanin:
## Summary

- `DOCKERFILE_MASK` in `checkov/common/util/dockerfile.py` only matched `Dockerfile`/`dockerfile` at the start of the filename or after a dot separator (e.g. `dev.Dockerfile`).
- Files with a CamelCase or word prefix *without* a dot separator — such as `AdtDockerfile` or `MyServiceDockerfile` — were silently skipped, leaving those images unscanned.
- The regex is updated to add a second alternative that allows any non-dot prefix followed by an uppercase `D`, so `AdtDockerfile` and similar variants are now recognised.
- Lowercase-only prefix variants (e.g. `ddockerfile`) remain invalid to avoid ambiguity, preserving all existing test expectations.

Closes #7541

## Test plan

- [ ] All previously-valid names still pass `is_dockerfile()`: `Dockerfile`, `dockerfile`, `Dockerfile.prod`, `dev.Dockerfile`, `team1.product.dockerfile`, etc.
- [ ] All previously-invalid names still fail: `.dockerfile`, `ddockerfile`, `docker-file`, `dockerfile1`, `Dockerfile.env.dockerignore`, etc.
- [ ] New valid names now pass: `AdtDockerfile`, `MyServiceDockerfile`, `AdtDockerfile.prod`
- [ ] Run `pytest tests/dockerfile/test_utils.py::test_is_dockerfile`

### DIFF
--- a/checkov/common/util/dockerfile.py
+++ b/checkov/common/util/dockerfile.py
@@ -1,6 +1,6 @@
 import re
 
-DOCKERFILE_MASK = re.compile(r"^(?:.+\.)?[Dd]ockerfile(?:\..+)?$(?<!\.[Dd]ockerignore)")
+DOCKERFILE_MASK = re.compile(r"^(?:(?:.+\.)?[Dd]|.+D)ockerfile(?:\..+)?$(?<!\.[Dd]ockerignore)")
 
 
 def is_dockerfile(file: str) -> bool:
--- a/tests/dockerfile/test_utils.py
+++ b/tests/dockerfile/test_utils.py
@@ -12,6 +12,9 @@ VALID_DOCKERFILE_NAMES = [
     "Dockerfile.Product1",
     "dev.Dockerfile",
     "team1.product.dockerfile",
+    "AdtDockerfile",
+    "MyServiceDockerfile",
+    "AdtDockerfile.prod",
 ]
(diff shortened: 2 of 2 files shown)


AlexanderSanin:
Hey @yuvalmich @talazuri @omriyoffe-panw. Could you, please, have a look at this?

PR #7560, opened 2026-05-30
fix(terraform): CKV_AZURE_215 - fix false positive on azurerm_api_management_backend for unresolvable url references

### LINKED ISSUE #7412: CKV_AZURE_215: False positive on azurerm_api_management_backend - protocol parameter doesn't control transport security
**Describe the issue**
CKV_AZURE_215 triggers on azurerm_api_management_backend resources with two fundamental problems:
1. The protocol parameter doesn't control transport security - it specifies API type (rest vs soap)
2. The url parameter can typically be an inferred/variable value, making static validation impossible

**Example**
```
resource "azurerm_api_management_backend" "backend" {
  for_each = local.api_configs
  
  name                = "${each.value.name}-backend"
  resource_group_name = var.resource_group_name
  api_management_name = var.api_management_name
  
  protocol = "http"  # This means REST API, NOT the transport protocol
  url      = each.value.backend_url  # Variable value - cannot be statically validated
}
```
The protocol parameter specifies API type (i.e. "http" = rest api, "soap" = soap api) and not transport security and are the only two valid values as per the [Azure provider documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/api_management_backend#protocol). The Azure provider does not accept "https" as a valid value.

Transport security is controlled by the url parameter, but in real-world scenarios, backend URLs are passed as variables, not hardcoded so static analysis cannot validate variable values at scan time. This makes the check impractical for actual Terraform usage patterns which triggers a false positive:
```
# Variable definition (validated elsewhere, in tests, runtime, etc.)
url (i.e. backend_url in above example) = "https://api.example.com"

# Checkov cannot see the actual value, only sees:
url = each.value.backend_url  # Triggers CKV_AZURE_215
```

**Expected Behavior**
CKV_AZURE_215 should not apply to azurerm_api_management_backend because:
- The protocol parameter doesn't control what the check is testing
- The url parameter is typically inferred and cannot be statically validated
- Transport security is enforced at runtime and through other validation mechanisms (tests, variable validation, etc.)

**Version:**
 - latest Checkov Version

JoshMcLeanTx:
## Description
Fixes #7412

CKV_AZURE_215 produces false positives on azurerm_api_management_backend
when the url parameter is set to a variable reference (var., local., each.,
module., data.) that cannot be statically resolved at scan time. Previously
the check returned FAILED for any URL not containing 'https', including
unresolvable variable references.

### Description
When a backend url is defined as a variable reference, Checkov cannot
determine the actual value at scan time. Returning FAILED in this case
is a false positive since the URL may well be HTTPS at runtime.

### Fix
Configure the url to a hardcoded https:// value, or ensure the variable
resolves to an https:// URL via a variables.tf default value.

## Checklist:


JoshMcLeanTx:
Anything needed from my side to get this reviewed? Happy to make any changes.

PR #7564, opened 2026-06-02
fix(CKV2_AWS_19): allow EIPs attached to Network Load Balancers to pass

### LINKED ISSUE #7532: CKV2_AWS_19 still fails when "domain = vpc"
**Describe the issue**
I have an Elastic IP is allocated to a network load balancer and received CKV2_AWS_19, which states that `vpc` should be set to `true`. However this setting is no longer available. A fix was attempted in #5378 but it doesn't seem to cover this case, if it worked for other cases.

**Examples**
```
resource "aws_eip" "foo" {
  domain               = "vpc"
  network_border_group = "us-east-1"
  public_ipv4_pool     = "amazon"
  tags = {
    Name = "foo"
  }
}
```

 - Checkov Version 3.2.525
 - AWS provider 6.37.0

**Additional context**

AlexanderSanin:
## Summary

- CKV2_AWS_19 was producing a false positive for `aws_eip` resources with `domain = "vpc"` when the EIP is used in the `subnet_mapping.allocation_id` of an `aws_lb` (Network Load Balancer).
- The check already validates VPC membership correctly (via `vpc = true` or `domain = "vpc"`), but the connection check only allowed `aws_instance`, `aws_nat_gateway`, `aws_transfer_server`, and `aws_eip_association` — not NLBs.
- Add `aws_lb` and `aws_alb` as valid connected resource types in the YAML definition so EIPs legitimately assigned to NLBs are no longer flagged.

Closes #7532

## Test plan

- [ ] Added `aws_eip.ok_eip_nlb` test resource in `tests/terraform/graph/checks/resources/EIPAllocatedToVPCAttachedEC2/main.tf` — an EIP with `domain = "vpc"` referenced by an `aws_lb` via `subnet_mapping.allocation_id`
- [ ] Updated `expected.yaml` to include `aws_eip.ok_eip_nlb` in the `pass` list
- [ ] Existing test `test_EIPAllocatedToVPCAttachedEC2` passes locally with the new test case

### DIFF
--- a/checkov/terraform/checks/graph_checks/aws/EIPAllocatedToVPCAttachedEC2.yaml
+++ b/checkov/terraform/checks/graph_checks/aws/EIPAllocatedToVPCAttachedEC2.yaml
@@ -54,6 +54,13 @@ definition:
             - aws_transfer_server
           operator: exists
           cond_type: connection
+        - resource_types:
+            - aws_eip
+          connected_resource_types:
+            - aws_lb
+            - aws_alb
+          operator: exists
+          cond_type: connection
         - cond_type: attribute
           resource_types:
           - aws_eip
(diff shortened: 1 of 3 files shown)


AlexanderSanin:
Hey @yuvalmich @omriyoffe-panw @maxamel. Could you, please, have a look at this?

PR #7575, opened 2026-06-09
fix(terraform): pass CKV_GCP_123 when remove_default_node_pool is true

### LINKED ISSUE #7406: CKV_GCP_123 triggers even if remove_default_node_pool is set
**Describe the issue**

As noted in the [GKE Terraform provider documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster) it is impossible to create a GKE cluster without creating the default node pool, so as a workaround the `remove_default_node_pool` flag can be used to immediately delete it after the cluster is created.

The check for `CKV_GCP_123` is simply checking that the nested `node_pool` block does not exist to configure the default node pool, however this has some issues:
1. In some cases, you may _have to_ define characteristics for the default node pool due to org policies etc or the cluster creation will fail, even if you intend to immediately delete the default node pool anyway
2. The lack of that nested configuration block does _not_ imply that a default node pool isn't created

The second problem is mostly a question of what exactly the check is trying to protect against - I would argue that it really should trigger when `remove_default_node_pool != true` to align closer to what the recommendation actually is saying.

**Examples**

```terraform
resource "google_container_cluster" "example" {
  initial_node_count       = 1
  remove_default_node_pool = true

  node_pool {
    node_config {
      service_account = google_service_account.default.email
    }
  }
} 
```

 - 3.2.461

**Additional context**
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#remove_default_node_pool-1

AlexanderSanin:
## Summary

- `CKV_GCP_123` checks that `google_container_cluster` resources do not inline a `node_pool` block, because node pools should be managed as separate `google_container_node_pool` resources.
- However, the Terraform `google` provider requires at least one node pool to exist when a cluster is created. The canonical workaround is to set `remove_default_node_pool = true`, which destroys the inline pool immediately after cluster creation.
- The old check fired on these clusters even though the inline `node_pool` is only a temporary scaffold — causing a false positive for a valid, recommended pattern.

**Fix:** override `get_excluded_key()` / `check_excluded_condition()` from `BaseResourceNegativeValueCheck` so the check returns `PASSED` whenever `remove_default_node_pool = true`, regardless of any inline `node_pool` blocks.

Closes #7406

## Test plan

- [ ] New Terraform fixture `google_container_cluster.pass_remove_default` added to `example_GKEDontUseNodePools/main.tf` — resource has both `remove_default_node_pool = true` and an inline `node_pool` block.
- [ ] `TestGKEDontUseNodePools.test` updated to expect `google_container_cluster.pass_remove_default` in the passing set (2 passes, 1 fail).
- [ ] `python -m pytest tests/terraform/checks/resource/gcp/test_GKEDontUseNodePools.py -v` passes locally.


AlexanderSanin:
Hey @maxamel @Saarett @AdamDev. Could you, please, have a look at this?

maxamel (approved):

AlexanderSanin:
Hey @Saarett @AdamDev. Could you, please, review also? we'll need 2 approvals

omriyoffe-panw (approved):

PR #7580, opened 2026-06-15
fix(kubernetes): CKV_K8S_40 should pass when hostUsers is false

### LINKED ISSUE #7527: CKV_K8S_40 (and others) should not trigger if .spec.hostUsers is false
**Describe the issue**
User Namespaces are enabled by default: https://kubernetes.io/blog/2025/04/25/userns-enabled-by-default/

> This means containers can run as root and be mapped to a non-root user on the host. Inside the container the process will think it is running as root (and therefore tools like apt, yum, etc. work fine), while in reality the process doesn't have privileges on the host.
> Furthermore, as users on each pod will be mapped to different non-overlapping users in the host, it is limited what they can do to other pods too.
> Capabilities granted to a pod are also limited to the pod user namespace and mostly invalid out of it, some are even completely void. Here are two examples:
>
>    CAP_SYS_MODULE does not have any effect if granted to a pod using user namespaces, the pod isn't able to load kernel modules.
>    CAP_SYS_ADMIN is limited to the pod's user namespace and invalid outside of it.

For sure `CKV_K8S_40` is completely invalid if `.spec.hostUsers` is set.  I would argue `CKV_K8S_23` should also check that field.  More research would need to be done if all capabilities need to be dropped.

**Examples**

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: isolated-workload
spec:
  hostUsers: false
  containers:
  - name: app
    image: fedora:42
    securityContext:
      runAsUser: 0
```

This should pass `CKV_K8S_40` and, I would argue, `CKV_K8S_23`.

 - Checkov Version: 3.2.92 but I spot-checked the code

**Additional context**
https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/

UTKARSH698:
## Description

`CKV_K8S_40` ("Containers should run as a high UID to avoid host conflict") enforces `runAsUser >= 10000` so that, if a container process runs as root, its UID will not collide with a privileged UID on the host.

When a pod sets `spec.hostUsers: false` it runs in a **private user namespace** ([enabled by default since Kubernetes 1.33](https://kubernetes.io/blog/2025/04/25/userns-enabled-by-default/)). Container UIDs are then remapped to unprivileged, non-overlapping host UIDs, so the host-collision risk this check guards against can no longer occur. In that case the high-UID requirement is moot and the check should `PASS`.

This PR makes `RootContainersHighUID.scan_spec_conf` return `PASSED` early when `spec.hostUsers is False`.

Fixes #7527

## Behaviour

- `hostUsers: false` -> `PASSED` (private user namespace; UID collision impossible)
- `hostUsers: true` or unset -> unchanged (high-UID check still applies)

The guard uses an explicit `is False` check, so only the namespaced case is exempted; an unset or `true` value keeps the original behaviour.

## Tests

Extended `tests/kubernetes/checks/test_RootContainersHighUID.py` example manifests:
- added a `hostUsers: false` pod with no `runAsUser` -> expected `PASSED` (proves the fix)
- added an explicit `hostUsers: true` pod with no `runAsUser` -> expected `FAILED` (guards against over-broadening the exemption)

Expected summary counts updated to `passed=4`, `failed=7`. Test passes locally.

### DIFF
--- a/checkov/kubernetes/checks/resource/k8s/RootContainersHighUID.py
+++ b/checkov/kubernetes/checks/resource/k8s/RootContainersHighUID.py
@@ -21,6 +21,12 @@ class RootContainersHighUID(BaseK8sRootContainerCheck):
 
         # Collect results
         if spec and isinstance(spec, dict):
+            # With a private user namespace (hostUsers: false) container UIDs are
+            # remapped to unprivileged host UIDs, so the high-UID host-collision
+            # concern this check guards against no longer applies.
+            if spec.get("hostUsers") is False:
+                return CheckResult.PASSED
+
             results: dict[str, Any] = {"pod": {}, "container": []}
             results["pod"]["runAsUser"] = self.check_runAsUser(spec, 10000)
 
--- a/tests/kubernetes/checks/example_RootContainersHighUID/rootContainersHighUIDFAILED.yaml
+++ b/tests/kubernetes/checks/example_RootContainersHighUID/rootContainersHighUIDFAILED.yaml
@@ -88,3 +88,16 @@ spec:
   - name: main2
     image: alpine
     command: ["/bin/sleep", "999999"]
+---
+# hostUsers: true keeps the host user namespace, so the high-UID check still
+# applies and a missing runAsUser must FAIL (guards the hostUsers exemption)
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod5
+spec:
+  hostUsers: true
+  containers:
+  - name: main
+    image: alpine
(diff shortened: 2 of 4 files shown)


UTKARSH698:
Friendly ping — this has been open ~3 weeks, CI is green and it's mergeable. It fixes a false positive where `CKV_K8S_40` fails even when `hostUsers: false` is set. @maxamel could you take a look? Thanks!

maxamel (approved):

talazuri (approved):

PR #7581, opened 2026-06-15
fix(terraform): prevent crash in S3AllowsAnyPrincipal with unparsed v…

### LINKED ISSUE #7571: CKV_AWS_70 crashes on mixed concat([dict], var.list) Statement shape — PR #7097 only covers all-strings case
**Describe the issue**

`CKV_AWS_70` (S3AllowsAnyPrincipal) crashes with `TypeError: string indices must be integers, not 'str'` on an `aws_s3_bucket_policy` whose `Statement` is built via `concat([static_dict], var.unresolved_list)`.

Reference: #7090 was closed via #7097 which added an `all('Effect' not in statement ...)` early-return. That guard handles the all-strings case (every statement is unresolved, none contains `"Effect"` as a substring). It does **not** handle the mixed shape below.

---

**Reproduction**

```hcl
variable "additional_policy_statements" {
  type    = list(any)
  default = []
}

resource "aws_s3_bucket_policy" "this" {
  bucket = aws_s3_bucket.this.id
  policy = jsonencode({
    Version = "2012-10-17"
    Statement = concat(
      [
        {
          Sid       = "DenyInsecureTransport"
          Effect    = "Deny"
          Principal = "*"
          Action    = "s3:*"
          Resource  = [aws_s3_bucket.this.arn, "${aws_s3_bucket.this.arn}/*"]
          Condition = { Bool = { "aws:SecureTransport" = "false" } }
        },
      ],
      var.additional_policy_statements,
    )
  })
}
```

```
TypeError: string indices must be integers, not 'str'
  File "checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py", line 90
    if 'Effect' not in statement or statement['Effect'] == 'Deny' or 'Principal' not in statement:
                                    ~~~~~~~~~^^^^^^^^^^
```

---

**Why #7097's guard doesn't fire**

1. Static element is a real dict with an `"Effect"` key → `'Effect' not in dict` returns `False` → `all(...)` returns `False` → early-return skipped.
2. Unresolved var element is serialized by Terraform as a string that contains the substring `"Effect"` (e.g. `"{Effect = ..., Principal = '*', ...}"`).
3. `'Effect' not in '...Effect...'` returns `False` (substring check). The OR doesn't short-circuit. `statement['Effect']` on a string → crash.

---

**Impact**

The crash silently downgrades the SARIF report to zero findings (rule errors don't become findings), so consumers' gate verdict comes back as `allow` even though the rule never executed. High-impact failure mode because the verdict looks legitimate.

---

**Suggested fix**

Defense-in-depth `isinstance` guard inside the for-loop, on top of #7097's all-strings pre-check.

```diff
--- a/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
+++ b/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
@@ -84,6 +84,10 @@ class S3AllowsAnyPrincipal(BaseResourceCheck):
             if all('Effect' not in statement for statement in statements):
                 return CheckResult.UNKNOWN
             for statement in statements:
+                if not isinstance(statement, dict):
+                    # Unresolved variable element — Terraform serializes such
+                    # items as strings that may contain the substring "Effect".
+                    # Cannot introspect; skip cleanly.
+                    continue
                 if 'Effect' not in statement or statement['Effect'] == 'Deny' or 'Principal' not in statement:
                     continue
                 principal = statement['Principal']
```

**Suggested test fixture** — append to `tests/terraform/checks/resource/aws/example_S3AllowsAnyPrincipal/main.tf`:

```hcl
# Mixed concat: static-dict + unresolved-var list. Should return UNKNOWN
# (rule cannot introspect the unresolved element). Regression coverage
# for the case beyond the all-strings shape fixed in #7097.
variable "extra_statements_pass" {
  type    = list(any)
  default = []
}

resource "aws_s3_bucket_policy" "concat_mixed_pass" {
  bucket = "example"
  policy = jsonencode({
    Version = "2012-10-17"
    Statement = concat(
      [
        {
          Sid       = "DenyInsecureTransport"
          Effect    = "Deny"
          Principal = "*"
          Action    = "s3:*"
          Resource  = "*"
          Condition = { Bool = { "aws:SecureTransport" = "false" } }
        },
      ],
      var.extra_statements_pass,
    )
  })
}
```

And in `tests/terraform/checks/resource/aws/test_S3AllowsAnyPrincipal.py`, add `"aws_s3_bucket_policy.concat_mixed_pass"` to the expected-pass set (or expected-unknown set, depending on how the test class assertions are structured — both behaviorally equivalent for the gate: no SARIF finding emitted, no crash).

Without the patch, this test fixture reproduces the `TypeError`. With the patch, the scan completes and the resource is reported as UNKNOWN.

---

**Version**

- checkov: 3.2.530 (also confirmed in `main` HEAD at the time of this report)
- python: 3.13
- terraform: 1.15.3 (CI runner; local repro on 1.15.5; project constraint `>= 1.10, < 2.0`)

marcoHosseini:

## Description

This PR fixes a parser crash in the `CKV_AWS_70` (`S3AllowsAnyPrincipal`) check when it encounters an unparsed Terraform variable within a `concat` statement for an S3 Bucket Policy.

A `TypeError` was previously thrown when the parser expected a dictionary but received a string (the unparsed variable placeholder). I added an `isinstance` guard inside the loop to safely skip non-dictionary elements, along with a corresponding Terraform test fixture to prevent regression.

Fixes #7571

## Checklist:

### DIFF
--- a/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
+++ b/checkov/terraform/checks/resource/aws/S3AllowsAnyPrincipal.py
@@ -87,6 +87,8 @@ class S3AllowsAnyPrincipal(BaseResourceCheck):
             if all('Effect' not in statement for statement in statements):
                 return CheckResult.UNKNOWN
             for statement in statements:
+                if not isinstance(statement, dict):
+                    continue
                 if 'Effect' not in statement or statement['Effect'] == 'Deny' or 'Principal' not in statement:
                     continue
                 principal = statement['Principal']
--- a/tests/terraform/checks/resource/aws/example_S3AllowsAnyPrincipal/main.tf
+++ b/tests/terraform/checks/resource/aws/example_S3AllowsAnyPrincipal/main.tf
@@ -518,3 +518,27 @@ resource "aws_s3_bucket_policy" "logs" {
     )
   })
 }
+# Test for Issue 7571: Unparsed variables in concat should not crash the parser
+variable "additional_policy_statements" {
+  default = []
+}
+
+resource "aws_s3_bucket_policy" "mixed_concat_crash" {
+  bucket = "bucket"
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = concat(
+      [
+        {
+          Sid       = "DenyInsecureTransport"
+          Effect    = "Deny"
+          Principal = "*"
+          Action    = "s3:*"
+          Resource  = ["arn:aws:s3:::bucket/*"]
+        }
+      ],
+      var.additional_policy_statements
+    )
+  })
+}
--- a/tests/terraform/checks/resource/aws/test_S3AllowsAnyPrincipal.py
+++ b/tests/terraform/checks/resource/aws/test_S3AllowsAnyPrincipal.py
@@ -28,6 +28,7 @@ class TestS3AllowsAnyPrincipal(unittest.TestCase):
             "aws_s3_bucket.pass_w_condition5",
             "aws_s3_bucket.pass_w_condition6",
             "aws_s3_bucket_policy.vpce_account_condition",
+            "aws_s3_bucket_policy.mixed_concat_crash",
         }
         failing_resources = {
             "aws_s3_bucket.fail",


maxamel:
@marcoHosseini tests are failing

marcoHosseini:
> @marcoHosseini tests are failing

Rebased onto the latest main and re-triggered CI.

I investigated the previously failing `test__parse_files` cases — they pass 
consistently locally across all parallelization types (fork/spawn/thread/none):
- macOS / Python 3.12: 5x repeated runs, all green
- Linux / Python 3.12 (fork, matching the CI runner): 3x runs, all green

This looks like flakiness in the parallel runner rather than a code issue — 
the change only touches the S3AllowsAnyPrincipal check, not the Terraform 
parser those tests exercise. Could you approve the workflow run when you get 
a chance, @maxamel ? Happy to address anything else.

yuvalmich (approved):

maxamel (approved):

PR #7585, opened 2026-06-23
fix(kubernetes): scope NetworkPolicyEdgeBuilder connections to same namespace

### LINKED ISSUE #7474: CKV2_K8S_6: NetworkPolicyEdgeBuilder ignores namespace scoping — single NetworkPolicy with empty podSelector passes all pods across all namespaces
## Describe the issue

The NetworkPolicyEdgeBuilder used by CKV2_K8S_6 does not consider namespace when connecting a NetworkPolicy to Pod resources. A NetworkPolicy with `podSelector: {}` in one namespace causes all pods in all namespaces to pass the check, even though Kubernetes NetworkPolicies are namespace-scoped.

This leads to:
- False positives (PASS) in repos that happen to have one NetworkPolicy with an empty podSelector anywhere
- False negatives (FAIL) in identical repos that lack that single NetworkPolicy

## Steps to reproduce

Given two directories with identical Deployments across multiple namespaces:

Repo A — has a single NetworkPolicy in namespace namespaceA:
```yaml
kind: NetworkPolicy
metadata:
  name: allow-policy
  namespace: namespaceA
spec:
  podSelector: {}
  ingress: [...]
  egress: [...]
```

Repo B — has no NetworkPolicy at all.

Both repos have the same Deployment in namespace namespaceB:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: namespaceB
  namespace: namespaceB
```

Running `checkov -d . --check CKV2_K8S_6`:
- Repo A: all pods PASS (including namespaceB in namespace namespaceB)
- Repo B: all pods FAIL

## Expected behavior

The namespaceB Deployment in namespace namespaceB should FAIL in both repos, because neither repo has a NetworkPolicy in the namespaceB namespace. Kubernetes NetworkPolicies are namespace-scoped — a NetworkPolicy in namespaceA does not protect pods in namespaceB.

## Root cause

In [NetworkPolicyEdgeBuilder.find_connections()](https://github.com/bridgecrewio/checkov/blob/main/checkov/kubernetes/graph_builder/graph_components/
edge_builders/NetworkPolicyEdgeBuilder.py), when podSelector is empty (no matchLabels), the code connects the NetworkPolicy to every Pod in the graph without checking if they share the same namespace:

```python
# the network policy has a podSelector property with no labels and should apply for all pods
else:
    connections.append(potential_pod_index)
```

The fix should add a namespace comparison, e.g.:

```python
else:
    if pod.attributes.get("namespace") == network_policy.attributes.get("namespace"):
        connections.append(potential_pod_index)
```

This same namespace check should likely also be applied to the matchLabels branch above it.

## Version

- checkov: 3.2.510
- Python: 3.14

AlexanderSanin:
## Summary

Fixes false-positive PASS results in `CKV2_K8S_6` (Require all pods to have a NetworkPolicy) caused by `NetworkPolicyEdgeBuilder` ignoring namespace scoping.

Kubernetes NetworkPolicies are namespace-scoped: a policy in `namespace-a` does not protect pods in `namespace-b`. Before this fix, the edge builder connected a `NetworkPolicy` with an empty `podSelector: {}` to **every Pod in the entire graph**, regardless of namespace. This meant a single wildcard NetworkPolicy in any namespace caused pods in all other namespaces to incorrectly pass the check.

**Root cause**: In `NetworkPolicyEdgeBuilder.find_connections()`, both the empty-podSelector wildcard branch and the `matchLabels` branch appended pod connections without comparing namespaces.

**Fix**: Extract the namespace from the NetworkPolicy and Pod vertex attributes (falling back to `"default"` when unset, matching the Kubernetes API default), and skip the connection when namespaces differ.

Closes #7474

## Test plan

- [x] Existing test `test_LabelSelectorEdgeBuilder_on_templates_with_network_policy` continues to pass (all resources in the same `default` namespace — behaviour unchanged)
- [x] New test `test_NetworkPolicyEdgeBuilder_respects_namespace_scoping` verifies that a NetworkPolicy in `namespace-a` with an empty `podSelector` and a label-based selector only connects to pods in `namespace-a`, not pods in `namespace-b`
- [x] All 15 tests in `tests/kubernetes/graph/test_local_graph.py` pass

### DIFF
--- a/checkov/kubernetes/graph_builder/graph_components/edge_builders/NetworkPolicyEdgeBuilder.py
+++ b/checkov/kubernetes/graph_builder/graph_components/edge_builders/NetworkPolicyEdgeBuilder.py
@@ -29,12 +29,19 @@ class NetworkPolicyEdgeBuilder(K8SEdgeBuilder):
         """
 
         connections: list[int] = []
+        network_policy = vertex
+        network_policy_namespace = (network_policy.attributes.get("metadata") or {}).get("namespace") or "default"
+
         for potential_pod_index, potential_vertex in enumerate(vertices):
             if potential_vertex.id == vertex.id or potential_vertex.attributes.get("kind") != "Pod":
                 continue
 
-            network_policy = vertex
             pod = potential_vertex
+            pod_namespace = (pod.attributes.get("metadata") or {}).get("namespace") or "default"
+
+            # NetworkPolicies are namespace-scoped and only apply to pods in the same namespace
+            if pod_namespace != network_policy_namespace:
+                continue
 
             pod_spec = network_policy.attributes.get("spec", {})
             if pod_spec is None:
@@ -54,7 +61,7 @@ class NetworkPolicyEdgeBuilder(K8SEdgeBuilder):
                 shared_labels = [k for k in match_labels if k in pod_labels and match_labels[k] == pod_labels[k]]
                 if len(shared_labels) == len(match_labels):
                     connections.append(potential_pod_index)
-            # the network policy has a podSelector property with no labels and should apply for all pods
+            # the network policy has a podSelector property with no labels and should apply for all pods in the namespace
             else:
                 connections.append(potential_pod_index)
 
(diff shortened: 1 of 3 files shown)


AlexanderSanin:
Hey @maxamel @omriyoffe-panw. Could you, please, have a look at this?

PR #7602, opened 2026-07-07
fix(terraform): allow aws_lb as valid EIP attachment target in CKV2_AWS_19

### LINKED ISSUE #7532: CKV2_AWS_19 still fails when "domain = vpc"

NarahariRaghava:

## Description

Fixes #7532

`CKV2_AWS_19` checks that EIPs allocated to a VPC are attached to a valid resource. The check already accepts both `vpc = true` and `domain = "vpc"` as valid VPC allocation attributes, but the allowed attachment targets only included `aws_instance`, `aws_nat_gateway`, and `aws_transfer_server`.

Attaching an EIP to a Network Load Balancer via `subnet_mapping.allocation_id` is a standard AWS pattern for giving an NLB a static IP. This configuration was incorrectly flagged as a false positive. Adding `aws_lb` to the connected resource types fixes it.

## Checklist:

- [x] I have added tests that prove my fix is effective and works

### DIFF
--- a/checkov/terraform/checks/graph_checks/aws/EIPAllocatedToVPCAttachedEC2.yaml
+++ b/checkov/terraform/checks/graph_checks/aws/EIPAllocatedToVPCAttachedEC2.yaml
@@ -54,6 +54,12 @@ definition:
             - aws_transfer_server
           operator: exists
           cond_type: connection
+        - resource_types:
+            - aws_eip
+          connected_resource_types:
+            - aws_lb
+          operator: exists
+          cond_type: connection
         - cond_type: attribute
           resource_types:
           - aws_eip
--- a/tests/terraform/graph/checks/resources/EIPAllocatedToVPCAttachedEC2/expected.yaml
+++ b/tests/terraform/graph/checks/resources/EIPAllocatedToVPCAttachedEC2/expected.yaml
@@ -5,5 +5,6 @@ pass:
   - "aws_eip.eip_ok_transer_server"
   - "aws_eip.ok_eip_domain"
   - "aws_eip.ok_eip_domain_assoc"
+  - "aws_eip.ok_eip_nlb"
 fail:
   - "aws_eip.not_ok_eip"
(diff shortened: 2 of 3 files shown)


checkov/terraform/checks/resource/azure/AppServiceDotnetFrameworkVersion.py
```diff
@@ -23,6 +23,8 @@ def scan_resource_conf(self, conf) -> CheckResult:
             site_config = conf.get('site_config')[0]
             if site_config.get('dotnet_framework_version') and isinstance(site_config.get('dotnet_framework_version'), list):
                 version = site_config.get('dotnet_framework_version')[0]
+                if not isinstance(version, str):
+                    return CheckResult.UNKNOWN
```
maxamel: Is this related to the CKV2_AWS_19?

NarahariRaghava: Apologies for the confusion! I was working on two separate fixes simultaneously, this one for CKV2_AWS_19 (allowing aws_lb as a valid EIP attachment target) and another for CKV_AZURE_80 (guarding against a dict dotnet version crash). The Azure change accidentally got included in this branch when I pushed. It has nothing to do with CKV2_AWS_19. I've force-pushed to remove it, the branch now only contains the EIP/NLB attachment fix. The Azure fix lives in its own separate PR (#7597).

PR #7610, opened 2026-07-14
fix(terraform): support GitHub immutable OIDC subject IDs in gh_repo_regex

Dashtid:

## Description

GitHub is rolling out **immutable OIDC subject claims**, where the `repo:` segment embeds numeric owner and repository IDs so the identity survives renames — `repo:octo-org@123456/octo-repo@456789:ref:refs/heads/main` instead of the classic `repo:octo-org/octo-repo:...`. Per the [GitHub changelog](https://github.blog/changelog/2026-04-23-immutable-subject-claims-for-github-actions-oidc-tokens/), this format becomes the default for **newly created and renamed repositories on 2026-07-15**.

The shared helper `gh_repo_regex` in `checkov/common/util/oidc_utils.py` validates the `owner/repo` segment of a `repo:` subject claim, but its character classes contain no `@`, so it rejects the immutable owner segment:

```python
# before
gh_repo_regex = re.compile(r"(\$\{)?[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)*(\})?/[^/]+")
# after
gh_repo_regex = re.compile(r"(\$\{)?[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)*(@[0-9]+)?(\})?/[^/]+")
```

`re.match` on `octo-org@123456/octo-repo@456789` stops at the first `@` and returns `None`, so a correctly-pinned immutable subject is treated as an invalid repo format and the check **fails**. The single added token `(@[0-9]+)?` accepts an optional numeric ID on the owner side; the repo side (`[^/]+`) already accepts `@456789`.

This helper is the single validation point reached — on the `split[0] == "repo"` branch — by all four GitHub-OIDC checks, so one line fixes every consumer:

| Check | Consumer |
|---|---|
| `CKV_AZURE_249` | `azuread_application_federated_identity_credential` |
| `CKV_AWS_358` | `data.aws_iam_policy_document` (HCL + terraform_plan) |
| `CKV_AWS_393` | `aws_iam_role.assume_role_policy` |
| `CKV_GCP_125` | `google_iam_workload_identity_pool_provider` |

Without the fix, every new or renamed repository authenticating via GitHub Actions OIDC would trigger a false positive across AWS, Azure, and GCP trust policies from 2026-07-15 onward.

The change is inert on existing inputs: exact `owner/repo`, org-only `owner/*`, and `${var}`-interpolated subjects all still match, and `owner*` (missing `/`) still correctly fails.

### Tests

Added a `pass_immutable` fixture with subject `repo:octo-org@123456/octo-repo@456789:ref:refs/heads/main` to each of the four suites (HCL for all four, plus a GitHub-OIDC data source in `tfplan.json` to exercise the `CKV_AWS_358` plan path). Reverting the regex alone, with the fixtures in place, fails all five new/updated assertions — so they are genuine regression guards, not just added coverage.

References:
- GitHub changelog — immutable subject claims for GitHub Actions OIDC tokens: https://github.blog/changelog/2026-04-23-immutable-subject-claims-for-github-actions-oidc-tokens/
- GitHub OIDC security hardening / subject claim format: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect

## Checklist:

---

**Open question (intentionally out of scope here):** `gh_repo_regex` also matches the org-wide `repo:myOrg/*` value, so an org-wide subject currently PASSES `CKV_AWS_358` / `CKV_AWS_393`. The inline comment in `GithubActionsOIDCTrustPolicyOnRole.py` marks that as deliberate (`"repo:myOrg/*" -> PASS`), so I've left it untouched. Is the org-wide-wildcard pass intended long-term — any repository the org can create is allowed to assume the role — or would tightening it to require a concrete repository segment be worth a separate issue?

### DIFF
--- a/checkov/common/util/oidc_utils.py
+++ b/checkov/common/util/oidc_utils.py
@@ -5,10 +5,11 @@ This file provides utility functions for handling OIDC-related operations, parti
 Constants:
     gh_repo_regex (re.Pattern): A regular expression pattern that matches GitHub repository paths.
         # Matches patterns like: "owner/repo", "${var}/repo", "org.name/repo"
+        # Also matches GitHub immutable OIDC subjects: "owner@123456/repo@456789"
         # Allows for variable substitution syntax ${} and organization names with dots
 
     gh_abusable_claims (list): A list of GitHub OIDC claims that could potentially be abused in security contexts.
 """
-gh_repo_regex = re.compile(r"(\$\{)?[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)*(\})?/[^/]+")
+gh_repo_regex = re.compile(r"(\$\{)?[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)*(@[0-9]+)?(\})?/[^/]+")
 gh_abusable_claims = ["workflow", "environment", "ref", "context", "head_ref", "base_ref"]
 gh_sub_condition = re.compile(r"^token\.actions\.githubusercontent\.com(?:/[a-zA-Z0-9_-]+)?:sub$")
--- a/tests/terraform/checks/data/aws/example_GithubActionsOIDCTrustPolicy/main.tf
+++ b/tests/terraform/checks/data/aws/example_GithubActionsOIDCTrustPolicy/main.tf
@@ -262,3 +262,30 @@ data "aws_iam_policy_document" "pass-gh-org" {
     }
   }
 }
+
+# pass_immutable - GitHub immutable OIDC subject (owner@id/repo@id numeric suffixes)
+data "aws_iam_policy_document" "pass_immutable" {
+  version = "2012-10-17"
+
+  statement {
+    effect = "Allow"
+    action = [
+      "sts:AssumeRoleWithWebIdentity"
+    ]
+    principals {
+      identifiers = ["arn:aws:iam::123456123456:oidc-provider/token.actions.githubusercontent.com"]
+      type        = "Federated"
+    }
+    condition {
+      test     = "StringEquals"
+      values   = ["repo:octo-org@123456/octo-repo@456789:ref:refs/heads/main"]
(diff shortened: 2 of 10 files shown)


Dashtid:
This has been open since 14 July with `start-security-scan` still sitting in `waiting`, so no CI has actually run on it yet — as a first-time external contributor the workflows need a maintainer to approve them before anything executes. Flagging that rather than the review itself, since it's the one thing I can't unblock from my side.

Everything verifiable locally is green: flake8 and mypy clean on `oidc_utils.py`, and the four affected suites pass (reverting just the regex with the fixtures in place fails them, so they're genuine regression guards rather than added coverage). The branch is rebased on current main and the title matches the enforced `fix(terraform):` format.

Is there anything needed from my side to get the workflows approved — or would you prefer this split differently before it runs?

Dashtid:
The red X on this PR is not a test result, and I think it is worth saying so explicitly because it makes the PR look failed to anyone skimming the list.

All four workflows report `conclusion: failure`, and none of them ran anything. Three of them — `PR Test`, `CodeQL` and `PR Title` — have no job records at all (`actions/runs/<id>/jobs` returns `total_count: 0`). The fourth, `security`, has two: `start-security-scan`, which sat pending from 2026-07-30T08:44Z until it failed at 2026-08-29T08:44Z, and the `security` job itself, skipped because that gate had failed. Thirty days to the second is the expiry on the "approve and run workflows" gate for fork PRs. So nothing in this branch has been compiled, linted or tested; the failure is the approval clock running out, not the patch.

There is a second gate underneath it: `security` runs on `pull_request_target` and declares `environment: scan-security`, so it needs a deployment-environment approval on top of the workflow approval.

The branch is still `MERGEABLE`. Happy to rebase if that helps, or to close it if the check is no longer wanted — is a workflow approval all this is waiting on?

PR #7615, opened 2026-07-19
fix(kustomize): surface non-zero build failures as parsing errors

### LINKED ISSUE #7599: Kustomize build failures are silently ignored, resulting in no Kubernetes checks being executed
## Describe the issue

When Checkov invokes Kustomize and the build fails, Checkov silently continues instead of reporting an error.

As a result:

- No Kubernetes manifests are generated.
- No Kubernetes checks are executed.
- Checkov exits without clearly indicating that the Kustomize build failed.
- Users may incorrectly assume that all checks passed, while in reality nothing was scanned.

This behavior is particularly misleading in CI pipelines, where a failing Kustomize build should cause the scan to fail rather than produce an apparently successful run with no Kubernetes results.

I would expect Checkov to:

- report that the Kustomize build failed,
- display the stderr output from Kustomize,
- return a non-zero exit code,
- not continue scanning as if nothing happened.

---

## Examples

Example directory:

```
kustomization.yaml
deployment.yaml
```

Suppose `kustomization.yaml` references a resource that does not exist:

```yaml
resources:
  - deployment.yaml
  - missing-resource.yaml
```

Running:

```sh
checkov -d . --framework kustomize

       _               _
   ___| |__   ___  ___| | _______   __
  / __| '_ \ / _ \/ __| |/ / _ \ \ / /
 | (__| | | |  __/ (__|   < (_) \ V /
  \___|_| |_|\___|\___|_|\_\___/ \_/

By Prisma Cloud | version: 3.3.6 

```

Current behavior:

- Checkov invokes Kustomize.
- Kustomize fails.
- Checkov silently skips scanning the generated manifests.
- No Kubernetes checks are reported.
- The user receives little or no indication that the scan never actually occurred.

Expected behavior:

```
Error: Kustomize build failed.

<stderr from kustomize>

Scanning aborted.
```

The process should exit with a non-zero exit code so that CI pipelines correctly fail.

---

## Exception Trace

Running with debug logging:

```sh
LOG_LEVL=DEBUG checkov -d . --framework kustomize           

       _               _
   ___| |__   ___  ___| | _______   __
  / __| '_ \ / _ \/ __| |/ / _ \ \ / /
 | (__| | | |  __/ (__|   < (_) \ V /
  \___|_| |_|\___|\___|_|\_\___/ \_/

By Prisma Cloud | version: 3.3.6 


```

Debug output shows that Kustomize failed, but Checkov continues execution instead of treating it as an error.

```
kustomize build .
Error: accumulating resources: accumulation err='accumulating resources from 'deploment.yaml': evalsymlink failure on '/tmp/checkov-bug/deploment.yaml' : lstat /tmp/checkov-bug/deploment.yaml: no such file or directory': must build at directory: not a valid directory: evalsymlink failure on '/tmp/checkov-bug/deploment.yaml' : lstat /tmp/checkov-bug/deploment.yaml: no such file or directory

```

---

## Desktop (please complete the following information)

- OS: Ubuntu 24.04
- Checkov Version: 3.3.6

---

## Additional context

This issue can be difficult to notice because the output may simply contain no Kubernetes results, leading users to believe that their manifests were successfully scanned.

This is especially problematic in combination with the current Kustomize executable auto-detection (`kubectl kustomize` vs `kustomize`), where users may already be troubleshooting different Kustomize implementations. If the selected implementation fails and Checkov silently ignores the failure, diagnosing the root cause becomes significantly more difficult.

In general, a failure to generate manifests should be treated as a fatal error rather than allowing the scan to continue with no resources to evaluate.

alexchen-sys:
## Description

When `kustomize`/`kubectl kustomize` exits non-zero, Checkov previously discarded stderr and continued as if the build had produced an empty successful scan. That made CI look green while no Kubernetes checks ran.

This change:

- treats a non-zero build exit as a failure
- logs an ERROR with the build command, exit code, and stderr
- records the failed directory and attaches it to `report.parsing_errors`
- returns the failed path from parallel parser workers so parent processes still see the failure

## Related Issue

Fixes #7599

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/bridgecrewio/checkov/blob/main/CONTRIBUTING.md)
- [x] I have added tests that prove my fix works
- [x] Local: `python -m pytest tests/kustomize/test_runner.py -q -o addopts=''` → 14 passed, 7 skipped

## Notes

`report.parsing_errors` is the existing Checkov surface for parse/build failures. Process exit code still follows `CKV_PARSE_ERROR_FAIL` (same as other frameworks). Operators who want hard CI fail on parse errors should set that env var.

## Tests

```
python -m pytest tests/kustomize/test_runner.py -q -o addopts=''
# 14 passed, 7 skipped
```

### DIFF
--- a/checkov/kustomize/runner.py
+++ b/checkov/kustomize/runner.py
@@ -432,6 +432,8 @@ class Runner(BaseRunner[_KubernetesDefinitions, _KubernetesContext, "KubernetesG
         self.kustomizeFileMappings: "dict[str, str]" = {}
         self.templateRendererCommand: str | None = None
         self.target_folder_path = ''
+        # Paths where kustomize/kubectl build failed; surfaced as report.parsing_errors (#7599)
+        self.kustomize_build_errors: "list[str]" = []
 
         self.checkov_allow_kustomize_file_edits = convert_str_to_bool(os.getenv("CHECKOV_ALLOW_KUSTOMIZE_FILE_EDITS",
                                                                                 False))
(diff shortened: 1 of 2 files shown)


ltshb:
@l46983284-cpu thanks for the fix, I hop this will get reviewed and merge soon

PR #7618, opened 2026-07-21
Validate unknown check IDs passed to --check

### LINKED ISSUE #7614: Unknown check IDs passed to --check are silently ignored
Codex:

> ## Describe the issue
>
> Checkov silently ignores unknown policy IDs supplied through `--check`. If at least one requested policy exists and passes, Checkov exits successfully even though another explicitly requested policy never ran.
>
> This makes a misspelled, renamed, deleted, or unregistered external policy indistinguishable from a successful scan in CI.
>
> ## Version
>
> `checkov 3.3.7` with Python 3.12.
>
> ## Minimal reproduction
>
> Create a Dockerfile whose known policy passes:
>
> ```dockerfile
> FROM alpine:3.20
> HEALTHCHECK CMD true
> ```
>
> Run one valid check and one nonexistent check:
>
> ```bash
> checkov \
>   --framework dockerfile \
>   --file Dockerfile \
>   --check CKV_DOCKER_2,CKV_DOCKER_DOES_NOT_EXIST \
>   --quiet
>
> echo "$?"
> ```
>
> Checkov reports the valid check as passed, says nothing about `CKV_DOCKER_DOES_NOT_EXIST`, and exits `0`.
>
> I also reproduced this with Python checks loaded through `--external-checks-dir`: one valid external ID plus one nonexistent external ID returned `0`.
>
> ## Expected behavior
>
> When the user explicitly supplies an exact policy ID through `--check`, Checkov should either:
>
> 1. exit nonzero when any requested ID is unavailable; or
> 2. provide a strict option that validates every requested ID before scanning.
>
> At minimum, the CLI should emit a clear warning naming every unknown requested ID.
>
> ## Why this matters
>
> Teams use `--check` as an enforcement allowlist. External policies can fail to register or be renamed while the remaining policies still run and keep CI green. Consumers currently need to parse structured output themselves to verify that every requested policy executed.

haneul-24:
Fixes #7614
I added validation for check IDs passed with --check.
If a check ID is not found in the registry, checkov now shows an error instead of ignoring it.
I also added tests for valid, invalid, wildcard, and BC ID cases.

### DIFF
--- a/checkov/main.py
+++ b/checkov/main.py
@@ -350,7 +350,11 @@ class Checkov:
                 report_sast_imports=bool(convert_str_to_bool(os.getenv('CKV_ENABLE_UPLOAD_SAST_IMPORTS', False))),
                 report_sast_reachability=bool(convert_str_to_bool(os.getenv('CKV_ENABLE_UPLOAD_SAST_REACHABILITY', False)))
             )
-
+            
+            invalid_checks = runner_filter.validate_checks() 
+            if invalid_checks:
+                self.parser.error(f"Invalid Checks: {', '.join(invalid_checks)}" )
+             
             source_env_val = os.getenv('BC_SOURCE', 'cli')
             source = source_type if source_type else get_source_type(source_env_val)
             if source == SourceTypes[BCSourceType.DISABLED]:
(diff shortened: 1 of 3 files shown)


haneul-24:
Hi @gruebel @schosterbarak, please review this PR and share your feedback.

haneul-24:
Hi @gruebel @schosterbarak, just following up on this PR, please take a look and share your feedback.

PR #7624, opened 2026-07-25
Fix baseline suppression for same resources in different directories

haneul-24:
## Description

fixes: #7486

This fixes an issue where baseline matching only checked the resource and check ID. Because of that, resources with the same name in different directories could be treated as the same finding and get suppressed.
I added the file path to the comparison so only the matching file is suppressed.
I also added a regression test to cover this case.

### Testing
- Reproduced the issue using the provided example.
- Verified that the issue occurs without the fix.
- Verified that only the matching file is suppressed after the fix.
- Ran the new regression test and existing baseline tests successfully.

### DIFF
--- a/checkov/common/output/baseline.py
+++ b/checkov/common/output/baseline.py
@@ -85,10 +85,12 @@ class Baseline:
     def _is_check_in_baseline(self, check: Record) -> bool:
         failed_check_id = check.check_id
         failed_check_resource = check.resource
+        failed_check_path = check.file_path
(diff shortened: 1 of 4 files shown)


haneul-24:
Hi @maxamel @GillSami @Saarett, please review this PR and share your feedback. Thanks!

haneul-24:
Hi @maxamel @GillSami @Saarett, just following up on this PR. please take a look and share your feedback.

PR #7627, opened 2026-07-27
fix(terraform): CKV_AZURE_249 should flag pull_request OIDC subjects

Dashtid:

## Description

`CKV_AZURE_249` currently passes a federated identity credential whose subject is `repo:<org>/<repo>:pull_request`.

That subject matches tokens minted by **`pull_request`-triggered workflow runs** — runs that execute a *proposed* change rather than the protected default branch. GitHub's security-hardening guidance treats the `pull_request` subject as a distinct, broader trust surface for exactly this reason: the code that runs during a pull request has not necessarily been reviewed or merged, and on public repositories it can originate from a fork.

Since an Azure FIC subject is an exact string match, a credential written this way trusts every pull-request run of that repository. The check inspects the subject closely enough to reject wildcards and abusable leading claims, so this seemed like a gap in the same family rather than an intentional allowance.

### Change

One additional comparison in `scan_resource_conf`, after the existing repo-format check:

```python
# Fourth check -> pull_request event
if (
    split_condition[0] == "repo"
    and len(split_condition) > 2
    and split_condition[2] == self.GH_PULL_REQUEST_EVENT
):
    return CheckResult.FAILED
```

It matches only the **event segment** (index 2) of a `repo:` subject, so it is deliberately narrow:

| Subject | Before | After |
|---|---|---|
| `repo:myOrg/myRepo:pull_request` | PASSED | **FAILED** |
| `repo:myOrg/myRepo:environment:pull_request` | PASSED | PASSED (environment *named* `pull_request`) |
| `repo:myOrg/pull_request:ref:refs/heads/x` | PASSED | PASSED (repo *named* `pull_request`) |
| `repo:myOrg/myRepo:ref:refs/heads/main` | PASSED | PASSED |
| `repo:myOrg/myRepo:environment:Production` | PASSED | PASSED |
| `repo:*`, `*`, `workflow:...` | FAILED | FAILED |

No other check or shared helper is touched, so `CKV_AWS_358` / `CKV_AWS_393` / `CKV_GCP_125` are unaffected (their suites pass unchanged).

### Tests

Two fixtures: `fail_pull_request` (the flagged subject) and `pass_environment_named_pull_request`, which pins the precision boundary so a later broadened match would fail the suite.

References:
- GitHub — security hardening with OpenID Connect (subject claim / `pull_request`): https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect
- Microsoft Entra — workload identity federation considerations (FIC subject is an exact match): https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-considerations

## Checklist:

---

Two things I noticed while working in this file and left alone, in case they are of interest:

1. `validate_subject_claim` appears to be unused — `scan_resource_conf` re-implements the same sequence inline, and I could not find a caller. I mirrored the new condition into both so they do not drift further, but would removing it (or having `scan_resource_conf` call it) be preferable?

2. The check registers only `azuread_application_federated_identity_credential`. The AzureAD provider added `azuread_application_flexible_federated_identity_credential` in 3.7, where matching moves to a `claims_matching_expression` and the `matches` operator honours `*` — so `claims['sub'] matches 'repo:org/*'` there really does admit the whole org, unlike a classic FIC subject where `*` is an inert literal. Is flexible FIC out of scope by design, or would a follow-up covering that resource be welcome?

---

Note on overlap: #7610 (immutable OIDC subject IDs) adds a fixture to the same two Azure test files, so whichever of the two lands first will leave the other with a small fixture-level conflict. The changes are independent — that one widens `gh_repo_regex`, this one adds an event-segment comparison — and I am happy to rebase whichever comes second.

### DIFF
--- a/checkov/terraform/checks/resource/azure/GithubActionsOIDCTrustPolicy.py
+++ b/checkov/terraform/checks/resource/azure/GithubActionsOIDCTrustPolicy.py
@@ -6,6 +6,12 @@ from checkov.common.util.oidc_utils import gh_abusable_claims, gh_repo_regex
 
 
 class AzureGithubActionsOIDCTrustPolicy(BaseResourceCheck):
+    # Subjects of the form "repo:<org>/<repo>:pull_request" trust tokens minted by
+    # pull_request-triggered workflows, which run proposed changes rather than the
+    # protected default branch. Only the event segment (index 2) is matched, so an
+    # environment or ref that happens to be named "pull_request" is unaffected.
+    GH_PULL_REQUEST_EVENT = "pull_request"
+
     def __init__(self) -> None:
         name = "Ensure Azure GitHub Actions OIDC trust policy is configured securely"
         id = "CKV_AZURE_249"
@@ -39,6 +45,10 @@ class AzureGithubActionsOIDCTrustPolicy(BaseResourceCheck):
             if not gh_repo_regex.match(claim_parts[1]):
                 return False
 
+            # Reject pull_request-triggered subjects
+            if len(claim_parts) > 2 and claim_parts[2] == self.GH_PULL_REQUEST_EVENT:
+                return False
+
         return True
 
     def scan_resource_conf(self, conf: Dict[str, List[Any]]) -> CheckResult:
@@ -65,6 +75,14 @@ class AzureGithubActionsOIDCTrustPolicy(BaseResourceCheck):
 
             # Third check -> repo format
             if split_condition[0] == "repo" and not gh_repo_regex.match(split_condition[1]):
+                return CheckResult.FAILED
+
+            # Fourth check -> pull_request event
+            if (
+                split_condition[0] == "repo"
+                and len(split_condition) > 2
+                and split_condition[2] == self.GH_PULL_REQUEST_EVENT
+            ):
                 return CheckResult.FAILED
 
             return CheckResult.PASSED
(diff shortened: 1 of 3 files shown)


Dashtid:
`start-security-scan` on this PR is now showing as failed rather than pending. The job is the one-line `echo` step in `.github/workflows/security.yml` whose only gate is `environment: scan-security`, and run 30527901430 sat awaiting that approval from 30 July until it hit Actions' 720-hour ceiling on 29 August. Nothing executed — the red X is the approval window expiring, not a test result.

It doesn't appear specific to this branch: the same `720h0m1s` signature is on the other open fork PRs opened before late July (#7615, #7619, #7625 and #7628 among them), while fork PRs opened in the last few weeks still show it pending. The net effect is that external PRs turn from pending to failing at the 30-day mark without anything having run.

The branch is unchanged since it was opened and still merges cleanly. Happy to push a rebase if re-queuing is the easier path — or would approving the existing run in place be preferable, given a fresh run would start the same 30-day clock over?

Dashtid:
The red X on this PR is not a test result, and I think it is worth saying so explicitly because it makes the PR look failed to anyone skimming the list.

All four workflows report `conclusion: failure`, and none of them ran anything. Three of them — `PR Test`, `CodeQL` and `PR Title` — have no job records at all (`actions/runs/<id>/jobs` returns `total_count: 0`). The fourth, `security`, has two: `start-security-scan`, which sat pending from 2026-07-30T08:44Z until it failed at 2026-08-29T08:44Z, and the `security` job itself, skipped because that gate had failed. Thirty days to the second is the expiry on the "approve and run workflows" gate for fork PRs. So nothing in this branch has been compiled, linted or tested; the failure is the approval clock running out, not the patch.

There is a second gate underneath it: `security` runs on `pull_request_target` and declares `environment: scan-security`, so it needs a deployment-environment approval on top of the workflow approval.

The branch is still `MERGEABLE`. Happy to rebase if that helps, or to close it if the check is no longer wanted — is a workflow approval all this is waiting on?

PR #7638, opened 2026-08-07
fix(terraform): handle empty key vault ip rules

### LINKED ISSUE #7384: CKV_AZURE_189 should account for empty `ip_rules`
**Describe the issue**
When passing an empty `ip_rules` to an `azurerm_key_vault` this results in a Key-Vault which is not publicly available. However, CKV_AZURE_189 complains that it is publicly reachable. 

**Examples**
This example should be approved, since it results in a non-reachable key-vault.
```
resource "azurerm_key_vault" "main" {
  name                          = "test"
  public_network_access_enabled = true
  purge_protection_enabled = true

  network_acls {
    bypass         =  "None"
    default_action =  "Deny"
    ip_rules = []
  }
}
```

3.2.460

Mukeshkr-19:

## Description

Fixes #7384.

`CKV_AZURE_189` incorrectly fails Azure Key Vault configurations where public network access is enabled but `network_acls` uses `default_action = "Deny"` with an explicitly empty `ip_rules` list.

Terraform can represent an empty `ip_rules` value as either a nested empty list or a normalized empty list depending on the processing path. The check previously treated the empty value as falsy and fell through to `FAILED`.

This change:
- recognizes an explicitly empty `ip_rules` list after normalization
- requires `default_action = "Deny"` for the new pass condition
- preserves existing behavior for missing, unresolved, and non-empty values
- adds regression coverage for both raw and normalized empty-list representations

## Checklist:

### DIFF
--- a/checkov/terraform/checks/resource/azure/KeyVaultDisablesPublicNetworkAccess.py
+++ b/checkov/terraform/checks/resource/azure/KeyVaultDisablesPublicNetworkAccess.py
@@ -38,6 +38,14 @@ class KeyVaultDisablesPublicNetworkAccess(BaseResourceValueCheck):
                         ip_rules = ip_rules[0] if ip_rules and isinstance(ip_rules, list) else ip_rules
                         if ip_rules:
                             return CheckResult.PASSED
+                        # An explicitly empty or rendered-empty ip_rules list (e.g. [[]] or [])
+                        # combined with an explicitly resolved default_action = "Deny" also blocks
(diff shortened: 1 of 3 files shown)


Mukeshkr-19:
Hi maintainers, it looks like the fork-triggered workflows expired after waiting for approval, so the downstream security job was skipped. Would someone be able to approve and re-run the workflows when convenient? The focused and related tests pass locally. Thank you!

PR #7650, opened 2026-08-18
fix(sca): correct Windows path handling in image referencer

omriyoffe-panw:

## Description

Image scan findings produced on Windows report paths that don't match the recorded file path. Two causes:

**1. A separator is dropped.** The dockerfile runner prepends `/` to a native relative path, giving a mixed-separator string. Stripping the root with `path.replace(Path(path).anchor, "", 1)` then resolves `anchor` to a bare `\`, and since the string starts with `/` that isn't a prefix — so `str.replace` removes the first backslash *anywhere* in the string:

```
/sub\client\Dockerfile.base   ->   //subclient\Dockerfile.base
```

**2. Reported paths keep `\` separators.** `dockerFilePath` and `relatedResourceId` are matched against the recorded file path, which always uses `/`.

## Changes

- `str_utils.py`: add `get_rootless_path()`, a positionally-aware replacement for the `anchor` idiom, and an `IS_WINDOWS` constant. The helper strips only a genuine leading drive/UNC/root and leaves separators as-is, so CLI output keeps its current form.
- `image_referencer.py`, `sca_image/runner.py` (x2): use the helper **only when `IS_WINDOWS`**; POSIX keeps the existing `anchor` expression. The helper uses `PureWindowsPath`, which would read a leading `//` — a valid POSIX root — as a UNC share and drop two segments, so it must not run off-Windows.
- `docker_image_scanning.py`: apply `align_path()` to `dockerFilePath` and `relatedResourceId`. These are computed independently of the CLI report path, so console output is unaffected. No guard needed: `align_path` only replaces backslashes, which POSIX paths don't contain.

`sca_package_2/runner.py` is the only other place using this construct, and it is **not** affected: it calls `str(Path(...))` first, which normalizes separators, so the anchor is always a real prefix. Confirmed by fuzzing 88 path shapes.

## Tests

- `test_str_utils.py`: 9-case matrix (Windows `/`-prefixed, drive-absolute, UNC, root-relative, relative; POSIX absolute/relative; nested; single file), plus a case documenting why the helper is Windows-only — `//srv/share/repo/Dockerfile` yields `repo/Dockerfile` under Windows semantics but must stay `srv/share/repo/Dockerfile` on POSIX.
- `test_base_image_referencer.py`: patches `IS_WINDOWS` and `Path` so the Windows condition reproduces on any host — runs in CI on Linux.
- `test_docker_image_scanning.py`: Windows alignment, POSIX unchanged, `None` `relatedResourceId`.

907 passed, 1 skipped across `tests/common`, `tests/dockerfile`, `tests/sca_image`, `tests/sca_package_2`. No existing test needed changing. `flake8` clean, `mypy` unchanged from baseline.

POSIX output was compared against `main` across 10 path shapes (including `//`, `///`, drive-like and spaced paths) through the real code path: byte-identical.

Verified on Windows against a fixture with `sub\client\Dockerfile.base`:

```
before   //subclient\Dockerfile.base (node:24.15.0 lines:1-1 (...))
after    /sub\client\Dockerfile.base (node:24.15.0 lines:1-1 (...))
```

The CLI path keeps native separators; the reported `dockerFilePath` is `/sub/client/Dockerfile.base`.

## Checklist:


yuvalmich (approved):
LGTM

PR #7658, opened 2026-08-24
feat(terraform): add CKV_AWS_394 for unconstrained aws_availability_zones data source

### LINKED ISSUE #7657: Add check: flag aws_availability_zones data source that does not pin zone identity
**Describe the issue**

Checkov currently has no check that flags an unconstrained Terraform `aws_availability_zones` data source (there is no existing check ID for this; the string `aws_availability_zones` appears only in test fixtures today). When a data source is declared with no filtering, e.g. `data "aws_availability_zones" "available" {}`, it returns *every* AZ in the region. This is a silent, environment-dependent expansion: config that worked for years suddenly changes behavior when AWS adds a new AZ to a region, with no code change on the user's side.

The logic for the addition: fail an `aws_availability_zones` data block unless it constrains *which* zones are returned — i.e. it pins zone identity. The recognized identity-pinning arguments are `exclude_names`, `exclude_zone_ids`, or an identity-based `filter` (on `zone-name` / `zone-id`). Arguments that constrain zone *attributes* rather than identity — `state`, `all_availability_zones`, or a `filter` on `opt-in-status` — do NOT prevent a newly-added zone from appearing and therefore do NOT satisfy the check on their own. This is the same class of risk as the existing `WhoAMI` check (`CKV_AWS_386`), which fails an under-constrained `aws_ami` data source, so there is clear precedent for scanning `data` blocks this way. The check would target the root cause (the unconstrained data source) rather than the downstream `count`/`for_each` usage.

**Examples**

Terraform:

The check passes only when the block pins zone *identity* (`exclude_names`, `exclude_zone_ids`, or an identity-based `filter`). Attribute-only constraints like `state` still return a non-deterministic set of zones and FAIL.

```hcl
# FAIL - no constraining argument at all; silently expands when a new AZ is added to the region
data "aws_availability_zones" "unfiltered" {}

# FAIL - `state` constrains zone health, not identity; a newly-added healthy AZ
# still appears, so the returned set is non-deterministic across regions/time
data "aws_availability_zones" "state_only" {
  state = "available"
}

# PASS - constrained via exclude_names (pins identity)
data "aws_availability_zones" "excluded_by_name" {
  exclude_names = ["us-east-1e"]
}

# PASS - constrained via exclude_zone_ids (pins identity)
data "aws_availability_zones" "excluded_by_id" {
  exclude_zone_ids = ["use1-az3"]
}

# PASS - constrained via an identity-based filter (zone-name)
data "aws_availability_zones" "filtered_by_name" {
  filter {
    name   = "zone-name"
    values = ["us-east-1a", "us-east-1b", "us-east-1c"]
  }
}
```

Expected outcomes:
- `data.aws_availability_zones.unfiltered` -> FAILED
- `data.aws_availability_zones.state_only` -> FAILED
- `data.aws_availability_zones.excluded_by_name` -> PASSED
- `data.aws_availability_zones.excluded_by_id` -> PASSED
- `data.aws_availability_zones.filtered_by_name` -> PASSED

- Checkov Version 3.3.13

**Additional context**

This targets the `data` block itself, which is the correct root-cause fix; it does not attempt to catch the downstream logic bug where resources fan out over `length(data.aws_availability_zones.available.names)`.

A downstream guard on `length(data.aws_availability_zones.available.names)` (or a `slice`/`precondition`/conditional `count`) can legitimately mitigate the risk. Checkov cannot reliably prove such guards across arbitrary expression forms, and a length guard is not even a complete mitigation (zones can be renamed/reordered). Therefore this check intentionally evaluates only the data block itself; users with a verified downstream mitigation should suppress it per-block via `# checkov:skip=CKV_AWS_XXX: <reason>`. A future graph-based variant could grant partial credit when the data source is connected to a resource whose `count`/`for_each` is length-constrained, but that is explicitly out of scope for the initial check.

The pass/fail line is drawn on an objective property — does the argument pin zone *identity* — not on a subjective judgment of how "secure" a filter value is. `exclude_names`, `exclude_zone_ids`, and an identity-based `filter` (`zone-name` / `zone-id`) constrain *which* zones are returned, producing a deterministic set that does not grow when AWS adds an AZ. `state`, `all_availability_zones`, and a `filter` on `opt-in-status` constrain zone *attributes*, so the returned set can still expand — these do not satisfy the check on their own. This categorical distinction (identity vs. attribute) is deliberately kept separate from grading filter *contents*: the check does not, for example, judge whether a given `exclude_names` list is complete. That mirrors `WhoAMI` (`CKV_AWS_386`), which flags an unset owner + wildcard name but does not grade how "secure" a given owner or pattern is.

Implementation note on `filter` blocks: a `filter` only pins identity when its `name` is `zone-name` or `zone-id`. Distinguishing an identity-based `filter` from an attribute-based one (`opt-in-status`, `zone-type`, `region-name`) by inspecting the filter's `name` is a reasonable v2 refinement; the simplest defensible initial implementation is to require at least one of `exclude_names` / `exclude_zone_ids`, treating bare `state` and non-identity `filter` blocks as FAIL.

madhavvishnubhatta:

## Description

Adds a new check, CKV_AWS_394, that flags a Terraform `aws_availability_zones` data source when it does not constrain *which* zones are returned. An unconstrained data source (e.g. `data "aws_availability_zones" "available" {}`) returns every AZ in the region, so configuration that relied on a fixed AZ count silently changes behavior when AWS adds a new AZ to the region — with no code change on the user's side.

This is the same class of risk as the existing WhoAMI check (CKV_AWS_386), which flags an under-constrained `aws_ami` data source, so there is precedent for scanning `data` blocks this way. The check targets the root cause (the unconstrained data source) rather than downstream `count`/`for_each` usage.

Fixes #7657 

### Description

The check fails an `aws_availability_zones` data block unless it pins zone *identity*. Identity-pinning arguments are:

- `exclude_names`
- `exclude_zone_ids`
- a `filter` on `zone-name` or `zone-id`

Arguments that constrain zone *attributes* rather than identity do NOT prevent a newly-added zone from appearing and therefore do NOT satisfy the check on their own: `state`, `all_availability_zones`, a `filter` on `opt-in-status`, and grouping filters such as `group-name` / `group-long-name` (a new AZ joins the existing group and still appears).

Docs:
- Terraform data source: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/availability_zones
- AWS DescribeAvailabilityZones filters: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeAvailabilityZones.html

### Fix

Constrain the data source so its result set is deterministic, e.g.:

    data "aws_availability_zones" "available" {
      exclude_names = ["us-east-1e"]
    }

or filter on zone identity:

    data "aws_availability_zones" "available" {
      filter {
        name   = "zone-name"
        values = ["us-east-1a", "us-east-1b", "us-east-1c"]
      }
    }

Users with a verified downstream mitigation (e.g. a length-guarded
`count`/`for_each`) can suppress per-block:

    # checkov:skip=CKV_AWS_394: <reason>

## Checklist:


Saarett:
Hello @madhavvishnubhatta , thank you for the contribution. Please pull latest dev and push it so we can re-run the tests.

madhavvishnubhatta:
> Hello @madhavvishnubhatta , thank you for the contribution. Please pull latest dev and push it so we can re-run the tests.

@Saarett Thanks for taking a look. I have done it now. Please re-run the tests and let me know if anything else is needed.

OfekShimko (approved):

maxamel:
Thanks @madhavvishnubhatta for the PR. Can you explain why you are relying on exclusion list for pinning, and not only on filter? Exclusion list won't pin zones, it will only determine what zones are not included.

madhavvishnubhatta:
> Thanks @madhavvishnubhatta for the PR. Can you explain why you are relying on exclusion list for pinning, and not only on filter? Exclusion list won't pin zones, it will only determine what zones are not included.

Fair question. You are right, exclusion doesn't pin AZs. I guess I was thinking that it shows they they at least thought about making sure only certain AZs are pulled. But I think flagging even exclusion as a potential issue does seem to make sense.

I am open to make the change. Let me know and I will do it.

maxamel:
> > Thanks @madhavvishnubhatta for the PR. Can you explain why you are relying on exclusion list for pinning, and not only on filter? Exclusion list won't pin zones, it will only determine what zones are not included.
> 
> Fair question. You are right, exclusion doesn't pin AZs. I guess I was thinking that it shows they they at least thought about making sure only certain AZs are pulled. But I think flagging even exclusion as a potential issue does seem to make sense.
> 
> I am open to make the change. Let me know and I will do it.

Sure please do.

madhavvishnubhatta:
Updated the code to treat exclude_names / exclude_zone_ids as FAIL for the aws_availability_zones check

maxamel (approved):

PR #7665, opened 2026-08-31
fix(terraform): CKV_AWS_358 & CKV_AWS_393 - inspect every value of a multi-value OIDC sub condition

Dashtid:

## Description

IAM evaluates the values of a single condition key with a logical OR — "each context key must resolve to true for at least one key value for the desired `Allow` or `Deny` effect to be invoked" ([IAM User Guide, single-valued context key policy examples](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition_examples-single-valued-context-keys.html)). In a GitHub Actions OIDC trust policy the loosest value in the list therefore sets the trust boundary, and a tight value beside it constrains nothing:

```hcl
condition {
  test     = "StringLike"
  variable = "token.actions.githubusercontent.com:sub"
  values   = ["repo:myOrg/myRepo:ref:refs/heads/MyBranch", "*"]
}
```

Both AWS GitHub-OIDC checks stop after one value of that list.

**`CKV_AWS_393`** (`GithubActionsOIDCTrustPolicyOnRole.py`) is the clear case. `_evaluate_sub_conditions` returns the verdict of the first value it classifies — including `PASSED` — so every later value is unreachable. `assume_role_policy` is `jsonencode`'d and reaches the check as one opaque string, so the author's JSON order survives intact. Measured on `main` @ `d8aec9db` through the Terraform `Runner`:

| `token.actions.githubusercontent.com:sub` values | verdict |
|---|---|
| `["repo:myOrg/myRepo:ref:refs/heads/MyBranch", "*"]` | **PASSED** |
| `["repo:myOrg/myRepo:ref:refs/heads/MyBranch", "workflow:github-actions:*"]` | **PASSED** |
| `["*"]` | FAILED |
| `["workflow:github-actions:*"]` | FAILED |

A bare wildcard in any non-first position is invisible.

**`CKV_AWS_358`** reads `condition_value[0]` and breaks, but its exposure is narrower and worth stating up front rather than leaving for a reviewer to rediscover: `clean_parser_types_lst` ([`checkov/terraform/modules/module_utils.py:224`](https://github.com/bridgecrewio/checkov/blob/master/checkov/terraform/modules/module_utils.py#L224), reached from `tf_parser.parse_file`) sorts string values lexicographically before any check runs. That incidentally drags `"*"` and `"environment:*"` to index 0, so those are caught today. What is missed is any unsafe value that sorts **after** every safe one — of the six `gh_abusable_claims` only `workflow` sorts after `repo:`, plus a misused `repo:<owner-with-no-slash>` whose owner sorts high:

| `values` (HCL) | verdict |
|---|---|
| `["repo:myOrg/myRepo:ref:refs/heads/MyBranch", "workflow:github-actions:*"]` | **PASSED** |
| `["repo:myOrg/myRepo:ref:refs/heads/MyBranch", "repo:zzz-org*"]` | **PASSED** |
| each of those second values alone | FAILED |

Narrower than `CKV_AWS_393`, still a miss. **The sort is deliberately left untouched** — it is load-bearing across the parser and is not the bug.

AWS's create-time validation does not close this either: IAM verifies only that the `sub` condition key "is present and that its value is not solely a wildcard character (\* and ?) or null" ([IAM User Guide, configuring a role for GitHub OIDC identity provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html)) — a presence-and-shape check that says nothing about the remaining values of a list. A static check is the layer that can catch the poisoned one.

### Fix

Classify **every** string value of a `sub` condition: `FAILED` on the first unsafe value, `PASSED` only once all values of that condition are safe.

Deliberately unchanged:

- single-value semantics;
- the intentional `repo:org/*` pass (#7221), including the `pass-fm-customer` regression fixture;
- the existing "first `sub` condition decides" behaviour when several condition blocks or operators each constrain `:sub` — IAM ANDs those, so an unrelated loose block does not make a tightly-pinned policy unsafe. A parity fixture covering that shape returns the same verdicts as unfixed `main`.

### Tests

There is no multi-value coverage upstream today. `fail-multivalue-wildcard`, `fail-multivalue-abusable` and `pass-multivalue-pinned` are added to both suites, with labels mirrored across the data and resource variants. Reverting the two check files alone, with the fixtures in place, fails the new assertions — so they are regression guards, not just added coverage.

## Checklist:

---

**Sequencing:** this is independent of #7610 and #7627, but it touches the same two AWS test suites as #7610. The fixture additions are append-only, so if #7610 lands first I will rebase this — no need to hold either on the other's account.

It is also the sharper half of the open question left in #7610's body. That one asked whether the org-wide `repo:myOrg/*` pass is intended long-term; this is the case where the answer does not depend on that judgement at all, since a list can pair a fully-pinned value with one that is unambiguously unsafe. Is there a reason the value loop was scoped to a single element, or would failing closed on any unsafe value in the list be the behaviour you would want here?


Dashtid:
A note on the checks here, since the state is easy to misread: `start-security-scan` (the `environment: scan-security` gate in `security.yml`) has been `waiting` since 2026-08-31, and `PR Test`, `CodeQL` and `PR Title` are parked at `action_required` behind the fork-PR approval, so nothing in this branch has been compiled or tested yet. At the 720-hour ceiling, around 2026-09-30, the waiting job will flip to a red `failure` with zero jobs -- the same thing that happened on #7610 and #7627 on 2026-08-29. The branch is 9 commits behind `main` with no conflicts. Is a workflow approval all this is waiting on?

PR #7672, opened 2026-09-06
fix(cloudformation): don't crash on unparseable FromPort/ToPort

### LINKED ISSUE #7670: CKV_AZURE_27 false positive for azurerm_mssql_server_security_alert_policy with AzureRM provider v5.3.0
**Describe the issue**

`CKV_AZURE_27` reports a failure for an `azurerm_mssql_server_security_alert_policy` resource even though:

- `state` is set to `"Enabled"`
- `email_account_admins_enabled` resolves to `true`

The problem started after upgrading the AzureRM provider to version `5.3.0`.

The same result is reproducible with the latest available Checkov version. (3.3.16)

### Check being reported

```text
CKV_AZURE_27 - Ensure that 'Email service and co-administrators' is 'Enabled' for MSSQL servers
```
**Examples**

The resource is declared in a module called `mssql_server`

```hcl
resource "azurerm_mssql_server_security_alert_policy" "mssql_alert_policy_replica" {

  resource_group_name = "example-resource-group"
  server_name         = azurerm_mssql_server.sql_server.name
  state               = "Enabled"
  email_addresses     = var.owner_emails
  email_account_admins_enabled = var.admin_alerts
  retention_days = 7
}

```
The module is called with var.admin_alerts set to `true` like this

```hcl

module "mssql_server" {
  source = "../.modules/mssql-server"

  ...........
  sql_security_alerts_recipients = var.owner_emails
  admin_alerts                   = true
  alerts_enabled                 = var.alerts_enabled
  .......
}

```
Expected behavior
CKV_AZURE_27 should pass because the MSSQL server security alert policy is enabled and email notifications to subscription administrators are explicitly enabled.

Actual behavior
Checkov reports CKV_AZURE_27 as failed for the resource.

Steps to reproduce
1. Configure an azurerm_mssql_server_security_alert_policy inside a Terraform module.
2. Assign a module input variable to email_account_admins_enabled.
3. Pass the literal boolean value true to that module input.
4. Use AzureRM provider version 5.3.0.
5. Run Checkov against the Terraform configuration.
6. Observe that CKV_AZURE_27 fails.

 - Checkov Version [e.g. 22]: 3.3.16

**Additional context**

**Minimal reproduction**

main.tf

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "5.3.0"
    }
  }
}

module "mssql_server" {
  source = "./modules/mssql-server"

  resource_group_name = "example-resource-group"
  server_name         = "example-sql-server"
  admin_alerts        = true
}
```
modules/mssql-server/variables.tf
```hcl
variable "resource_group_name" {
  type = string
}

variable "server_name" {
  type = string
}

variable "admin_alerts" {
  type = bool
}
```
modules/mssql-server/main.tf
```hcl
resource "azurerm_mssql_server_security_alert_policy" "example" {
  resource_group_name = var.resource_group_name
  server_name         = var.server_name

  state                        = "Enabled"
  email_account_admins_enabled = var.admin_alerts
  retention_days               = 7
}
```

This was working fine before the AzureRM provider version update.

karlhillx:
Closes #7670.

`AbsSecurityGroupUnrestrictedIngress.range()` raised `ValueError` on an empty `FromPort`/`ToPort`, which escaped `scan_resource` and left the resource with no result. `range()` now returns `None` when either bound is unparseable; the caller reports `UNKNOWN`, matching the file's existing precedent for unusable input. Valid integer and string port ranges are unchanged.

Adds a regression fixture (`SecurityGroupUnresolvedPort-UNKNOWN.yaml`) and two tests covering the new and existing behavior. CHANGELOG entry under `[Unreleased]`.


karlhillx:
Updated to address both Copilot findings:

- `bool | None` → `Optional[bool]` (drops the 3.10+ syntax / future-import dependency)
- Short-circuit mask: skipped rules no longer pre-empt a definitive `FAILED` verdict on a sibling. Two new tests lock the new semantics: `test_unparseable_rule_does_not_mask_failed_sibling` and `test_all_rules_unparseable_still_returns_unknown`. 286 tests pass in the cloudformation suite.

PR #7687, opened 2026-09-16
chore(general): upgrade parallelism and drain

GillSami:

[//]: # "
    # PR Title
    #    
    Allowed prefixs:
    #
    ex.
"

## Description

Fixes # (issue)

### Description

### Fix

## Checklist:


Saarett (approved):

checkov/secrets/runner.py
```diff
@@ -93,6 +93,42 @@
...
+#
+# Default THREAD mode is GIL-bound for the mostly-Python per-line work, so it
+# uses roughly one core. CHECKOV_SECRETS_PARALLEL_MODE=spawn runs the scan on a
+# dedicated spawn pool (separate processes and GILs) for better multi-core use.
+# spawn is safe on all platforms (it is the default context on macOS/Windows).
+#
+# Spawned workers start with blank memory, so each rebuilds the detect-secrets
+# settings and disables the same filters as the parent via _init_secrets_worker;
+# otherwise a worker would run unconfigured and miss findings.
+SECRETS_PARALLEL_MODE = os.getenv("CHECKOV_SECRETS_PARALLEL_MODE", "thread").lower()
```
AdamDev, line 107: Why does the default parallel mode is thread? 
it decided by the os by default, for example on Linux the normal default is FORK.

Fork also has separate GIL, like spawn

checkov/common/parallelizer/parallel_runner.py
```diff
@@ -121,11 +150,55 @@ def func_wrapper(original_func: Callable[[Any], _T], items_group: List[Any], con
...
+            if group_len > 0
+        }
+
+        while remaining:
+            # wait() blocks until at least one pipe has data. It is generic over
+            # waitable objects, but we only ever pass Connection objects.
+            for conn in cast(List["Connection"], mp_wait(list(remaining.keys()))):
+                try:
+                    value = conn.recv()
+                except EOFError:
```
AdamDev, line 178: **Permanent Deadlock**

This EOFError block will never actually catch a killed worker process because of how the processes were spawned above (lines 126–129). 

When parent_conn, child_conn = multiprocessing.Pipe(duplex=False) is created in _run_function_multiprocess_fork, the parent process keeps child_conn (the write end) open in its own file descriptor table. In Unix, an OS pipe does not deliver EOF (POLLHUP / EOFError) as long as any process still holds a write descriptor open.

If a worker process is killed (SIGKILL, OOM, segfault), the parent itself still holds child_conn open, so mp_wait() never wakes up with EOF and will hang indefinitely.

To make this recovery logic actually work, close child_conn in the parent loop immediately after process.start():
process.start()
child_conn.close()  # Parent doesn't write; close so only the child holds the write end

checkov/secrets/runner.py
```diff
@@ -524,14 +560,35 @@ def _prioritise_secrets(secret_records: Dict[str, SecretsRecord], secret_key: st
...
         ]
-        results = parallel_runner.run_function(func=Runner._safe_scan, items=items)
+
+        if SECRETS_PARALLEL_MODE == "spawn" and plugins_used is not None:
+            # Dedicated spawn runner for the secrets scan only (plugins_used is
+            # required so workers can rebuild their detect-secrets config).
+            # ParallelRunner downgrades spawn->thread on macOS/Windows in
+            # __init__; re-assert the type to opt this scan into real
+            # multi-process parallelism on every platform.
+            spawn_runner = ParallelRunner(parallelization_type=ParallelizationType.SPAWN)
```
AdamDev, line 586: I don't think we should force SPAWN by overriding ParallelRunner.type.

ParallelRunner intentionally falls back to THREAD on macOS and Windows, and bypassing that behavior introduces several risks:

1. **Encapsulation**: Directly overriding .type bypasses the safeguards and platform-specific decisions implemented by ParallelRunner.
2. **Frozen/PyInstaller compatibility**: SPAWN may break packaged Checkov binaries because multiprocessing.freeze_support() is not currently configured. In the worst case, child processes can re-execute Checkov and recursively create more processes.
3. **Performance**: SPAWN has significant startup overhead because every worker starts a fresh Python interpreter and initializes/imports the required modules. This is especially expensive for small scans.
4. **Resource usage**: Creating a full spawn pool for only a few files wastes CPU and memory on idle workers.
5. **Platform behavior**: Forcing SPAWN removes the OS-specific behavior that ParallelRunner currently provides and may introduce issues on platforms where this mode has not been validated.

AdamDev (approved):

